PR #741, opened 2020-09-10
[MIRROR] The BEAUTIFUL Supplypod Renaissance Part 5: An Ode To Admins Everywhere (total UI overhaul)

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/52611
--------------------

Ekd03EV
<sup>ignore the tiny bit of overlap w/ the top buttons in compact mode that has since been fixed</sup>
## About The Pull Request
This one was a while in the making. 

- Presets! Save and load all your pod settings. They persist across rounds and servers!
- Added ability to filter what gets taken by reverse pods. This includes mobs, anchored/unanchored objs, wallmounted objs, under-turf objs (pipes, cables, etc.), and even turfs and walls, which required some unique code to get working right.
- Added a view for the bay you're launching items from, the pod, and the location you're dropping the pod off at
- Added Quick-access buttons to game panel and buildmode
- Tooltips have been made MUCH more concise overall. No more giant boxes of dumb self-explanatory info
- Updated the supplypod selector icon and added an icon that shows your dropoff location. Both of these are only visible to the holder's client.
- The new UI is like half the size of the old one, and it can be made even more compact with a press of a button, for those supplypod launches on the go!
- Added a quick way to tell the projectile type and magnitude at a glance (via tooltip)

And on the TGUI side:
- Adds a way to generate UUID's (All credit to stylemistake)
- Adds a way to allow knobs to be manually set over their limit
image
- Tooltips now support multiline
- Tooltips can now be placed on the top-left and top-right of elements

## Why It's Good For The Game
good UI good

## Changelog
:cl:
admin: The supplypod menu no longer sucks
/:cl:

image


Azarak:
travis failing due to wrong tests

PR #795, opened 2020-09-14
[Non modular] Small booze changes to make it more sensible

Azarak:
## About The Pull Request
Now you get drunk 60% slower, but your drunkiness decays around 300% slower. Allowing people to get smashed after a proper amount of alcohol, not just one glass, but also making so the alcohol persists through a long duration of time.
This is the change we've had before and before

## Why It's Good For The Game
more RP friendly drunkiness

## Changelog
:cl:
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -505,7 +505,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 		handle_hallucinations()
 
 	if(drunkenness)
-		drunkenness = max(drunkenness - (drunkenness * 0.04) - 0.01, 0)
+		drunkenness = max(drunkenness - (drunkenness * 0.01) - 0.01, 0) //SKYRAT EDIT - booze - changed the 0.04 decay to 0.01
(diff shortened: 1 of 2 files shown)


Gandalf2k15:
Perhaps an upstream change?

Azarak:
> 
> 
> Perhaps an upstream change?

TG rounds end up being quite short most often than not, not sure if this is something they'd like

code/modules/mob/living/carbon/life.dm
```diff
@@ -505,7 +505,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 		handle_hallucinations()
 
 	if(drunkenness)
-		drunkenness = max(drunkenness - (drunkenness * 0.04) - 0.01, 0)
```
Gandalf2k15, line 508: Remember to have original

Gandalf2k15 (changes requested):

code/modules/mob/living/carbon/life.dm
Azarak: done!

PR #841, opened 2020-09-17
Adds different random title screens! (I'm bored)

Gandalf2k15:
Just adds some more title screens.


Gandalf2k15:
Temporary change until we get the OG ones.

PR #901, opened 2020-09-20
Removes random placement [MODULAR]

Gandalf2k15:
## About The Pull Request

When dropping an item it no longer randomly pixels on the ground.

## Why It's Good For The Game

OCD.

## Changelog
:cl:
tweak: Items that are dropped now centre on the ground.
/:cl:


Azarak:
Not sure about first randomizing the position of the item, and then resetting it. I myself like the random offset on dropped items, makes piles of things look way more natural than just stacked on top of each other.

SpaceVampire:
I think random placement is a good feature. Having everything stack perfectly looks worse, and is less useful.

PR #982, opened 2020-09-25
Horns are rendered after ears for mammals

Azarak:
## Changelog
:cl:
/:cl:


Azarak:
Actually, this one is tricky, some will look worse, some will look better

PR #1063, opened 2020-09-29
[MIRROR] Fixes visual bugs with mothpeople markings

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/54051
--------------------
##  About The Pull Request

Uses a species trait to apply six mutable appearances in `handle_body()` so that as mothpeople are dismembered/augmented their biological markings will no longer be shown.
The marking icons are based off of their preferences, and so mixing them is not an option.

## Why It's Good For The Game

image

This is how things are on live. Markings are shown in places that have no limb and over limbs that are augmented.

image
With these changes you can see how the markings look in game in various stages of dismemberment/augmentation.

I've already addressed the wings having limb-shaped holes in https://github.com/tgstation/tgstation/pull/53969 so that bug isn't on live, and this pr does not regress that either. 

## Changelog
:cl:
fix: The markings on mothpeople will not show over missing or augmented limbs any more.
/:cl:


Azarak:
blame cobalt for the CI

PR #1118, opened 2020-10-02
[NON-MODULAR] Lowers event frequency, disables some problematic events

Azarak:
## Changelog
:cl:
/:cl:

### DIFF
--- a/code/modules/events/swarmer.dm
+++ b/code/modules/events/swarmer.dm
@@ -2,7 +2,8 @@
 	name = "Spawn Swarmer Beacon"
 	typepath = /datum/round_event/spawn_swarmer
 	weight = 8
-	max_occurrences = 1 //Only once okay fam
+	//max_occurrences = 1 //Only once okay fam //ORIGINAL
+	max_occurrences = 0 //SKYRAT EDIT CHANGE - EVENTS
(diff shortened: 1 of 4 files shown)


code/modules/events/swarmer.dm
```diff
@@ -2,7 +2,8 @@
 	name = "Spawn Swarmer Beacon"
 	typepath = /datum/round_event/spawn_swarmer
 	weight = 8
-	max_occurrences = 1 //Only once okay fam
+	//max_occurrences = 1 //Only once okay fam //ORIGINAL
```
Gandalf2k15, line 5: what's wrong with swarmers?

Azarak: Ever had a round versus them? It's completely atrocious, they keep breaking things, stealing mats, locking people, completely derailing the entire round as they snowball. It can't be a random occurance.

Gandalf2k15: I think we need to decide on policy level if we're going to allow the game to run the round for us or the admins, because I'm not sure which will do a better job.

Azarak: I think swarmers should be disabled regardless

PR #1233, opened 2020-10-09
[MIRROR] Mobility refactor: no more update_mobility()

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/54183
--------------------
This is a pretty big change all around. The gist of it is that it moves the `mobility_flags` into traits or variables that can track the sources, and to which we can append code to react to the events, be it via signals or via on_event-like procs.

For example, `MOBILITY_STAND` could mean, depending on context, that the mob is either already standing or that it may be able to stand, and thus is lying down.

There was a lot of snowflakery and redefinitions on top of redefinitions, so this is bound to create bugs I'm willing to fix as I learn them.

The end-goal is for every living mob to use the same mobility system, for the traits to mean the same among them, and for no place to just mass-change settings without a way to trace it, such as with `mobility_flags = NONE` and `mobility_flags = ALL`

* Fixes #54173
* Closes #54174
* Fixes #54175
* Fixes AIs being able to strip nearby people. They've lost their hands usage.
* Fixes #54270
* Fixes #54271


Gandalf2k15:
updated robot to use update_resting(), also fixed a bug with pulling and offsets.

PR #1324, opened 2020-10-15
Makes yells be carried through areas as an echo

Azarak:
## About The Pull Request
This is a port from our old base, originally done by Bhjin i believe. I tweaked the radius to not be halved, as it doesn't really impact the echo too much but solves some fringe cases where it would feel like it doesn't work, ie. the yell range being just near the next area, but not quite reaching it.

## Changelog
:cl:
/:cl:

### DIFF
--- a/code/modules/mob/living/living_say.dm
+++ b/code/modules/mob/living/living_say.dm
@@ -303,7 +303,7 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 				//SKYRAT EDIT ADDITION BEGIN - YELLING ECHOES
 				if(yellareas)
 					var/area/A = get_area(M)
-					if(istype(A) && A.ambientsounds != SPACE && A in yellareas)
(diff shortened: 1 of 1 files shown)


ForrestWick:
I would not be sad in the slightest if this wasn't ported tbh. All this ever did on Citadel was encourage people to just use !! at the end of messages to make them carry through walls, and I have no doubt it'd happen here too once people began to find out

Azarak:
> 
> 
> I would not be sad in the slightest if this wasn't ported tbh. All this ever did on Citadel was encourage people to just use !! at the end of messages to make them carry through walls, and I have no doubt it'd happen here too once people began to find out

You are "yelling" after all. I'd prefer if someone used this instead of comming a scream of help, hearing yells of help or anything like that is quite immersive and cool too, or parts of a loud quarrel.

Gandalf2k15 (approved):

PR #1356, opened 2020-10-18
[MODULAR] [READY FOR TESTING] Glowshroom-spacevines-blob-PVE-encounter MOLD

Azarak:
## About The Pull Request
100% PVE
Spreads much like glowshrooms and space vines, has a main core like a blob. During the spreading it builds structures, currently there's 2. One of them will release something harmful on proximity, another will spawn enemies. The blob also releases harmful stuff depending on the type, on cooldown.

The resin grows and bashes through doors and firelocks, when it meets a space tile it walls it off. It isn't too hard of an encounter and it doesn't spread too fast. 

There's 4 types right now, for each type people should take be taking different tactics and pre-cautions.

Currently not tied to any event, just admin spawn for the purposes of tests and stuff.

When the core dies, everything slowly decays away.

dreamseeker_2020-10-18_16-19-42
dreamseeker_2020-10-18_16-20-28
dreammaker_2020-10-18_15-28-17

## Why It's Good For The Game
a PVE event that shouldn't be too easy, nor tedious, isn't an "all hands on deck or station dies" kinda deal either. I figure it's quite healthy to get some interesting PVE content so people can get their unga out on the mold. No under-handed, station-ending tactics like the PVP blob either.

## Changelog
:cl:
add: Added THE MOLD
/:cl:


Erokusan:
This is a fantastic idea and I love it, I would like to ask however, is the announcement for it arriving on the station going to be like the blobs announcement? Because just calling it a lvl 5 biohazard (Because it would be really close to one, like this is maybe biohazard lvl 6) will certainly call it out as a blob even though it isn't, only because it looks very similar

Owai-Seek:
I think the intention is to replace blob with this, as people who know what they're doing can easily round end the station, and ones who don't get slapped very quickly. This can be balanced if it's too weak or strong.

Jeweledbug:
I don't think it's a good idea. It's an NRP antagonist that won't fufill the "unga" since if it's just a sec problem assuming it's not all hands on deck that means that they'll be the sole ones to deal with it and let that emotion/attitude out. Either than or the first few people with welders like a vine, and that hasn't really kept the attitude in check either.

It doesn't fit in with the RP first attitude of the server, the focus of creating more roleplay than the most basic of LRP server, all it'll do is make yet another vines issue where it'll need constant tweaking, maintenance and more for the sake of an event that's better suited to ERIS than Skyrat.

Azarak:
> 
> 
> This is a fantastic idea and I love it, I would like to ask however, is the announcement for it arriving on the station going to be like the blobs announcement? Because just calling it a lvl 5 biohazard (Because it would be really close to one, like this is maybe biohazard lvl 6) will certainly call it out as a blob even though it isn't, only because it looks very similar

No announcement yet, and since this doesn't spread particularly quickly, nor is it an immediate threat, perhaps it shouldnt have one, or have a very delayed one I guess.

Azarak:
> 
> 
> I don't think it's a good idea. It's an NRP antagonist that won't fufill the "unga" since if it's just a sec problem assuming it's not all hands on deck that means that they'll be the sole ones to deal with it and let that emotion/attitude out. Either than or the first few people with welders like a vine, and that hasn't really kept the attitude in check either.
> 
> It doesn't fit in with the RP first attitude of the server, the focus of creating more roleplay than the most basic of LRP server, all it'll do is make yet another vines issue where it'll need constant tweaking, maintenance and more for the sake of an event that's better suited to ERIS than Skyrat.

I was designing this encounter with the flaws of vines and blob in mind, trying to avoid them and make the best of it. There will be plenty to do for all sorts of personnel, depending on the type. Engineering may need to fix air alarms, or help in firefighting against the fire variant. Medical will require to patch up wounded people, and maybe treat a fungal disease contracted from it. Science may be asked to research something. Miners could be pressed to mine missing materials to print whatever's nessecary. And since it's designed to be kinda slow, but punishing if you fight it improperly (it doesn't snowball, it just needs tactics), it leaves room for RP aswell as doing other tasks that arent just weldering the mold

YakumoChen:
Is blob removed? It's one of the most fun antags we have to fight, and the whole station coming together is good.

Zenitheevee:
So, question:
Are the blob textures just fillers, or is this going to replace blob as a whole?

Jeweledbug:
>it leaves room for RP aswell as doing other tasks that arent just weldering the mold

Not really. Here's my reasoning as to why. 

Like a blob, or vines, what this does upon being found is it pauses everything around it. Let's say you're security in the middle of a tense situation, you've finally hit the jackpot, you're roleplaying with antags, talking, communicating, etc. Then, this pops up, and like others it NEEDS to be dealt with. Maybe it won't destroy the whole station but it'll cause an immense amount of loss and possibly many lives. So rather than roleplay sec has to drop everything, they have to go out and kill this thing.

Same goes for any kind of RP currently going on. It all ceases to be and is instead replaced with the need to take down the threat. It halts RP, forcing it to disappear in return for.... busy work. For the engineer walking with mats, the sec officer shooting his gun, the doctor applying a patch. Things that make those departments amongst the least RP driven of any of the departments. 

Sec needs to kill it NOW.

Doctors need to heal the patients NOW

Engineers need to fix the power/station/etc NOW.

There is no or at best, minimal roleplaying. Instead if favors busy work and diminished existing roleplay especially for departments that could stand to have it instead of busy work.

Owai-Seek:
Outside of balance concerns because I think it'll be easy enough to taper a PvE encounter if it's too hard.

- The heartbeat from the core will proc before the sound of the first beat has ended, making it stack continually.
- The first test I saw it, I was a bit annoyed with all of the simplemob bodies piling up, as it's hard to clean even if the mold disappears. (I think some of them are non-butcherable?) I don't know if this has changed since the first test of it.
- The first 3 spawns or so that the mold spits out are player controlled.
- ~~Rot mold (Or maybe Glowing?) can apparently give people Fungal TB.~~
- Rot mold was -bugged- and was not meant to be spitting out Fungal TB, but you're addressing the issue already. <3

YakumoChen:
Easily the worst event I've ever had to deal with in a round. It takes all the worst parts of Blob and Space Vines and combines them into one abomination that is incredibly tedious, annoying, and destructive to the round's rhythm without actually being fun to fight.

It's not as dangerous as a blob to warrant an all hands on deck response, nor does it share with blob the actual PVP action that makes blob enjoyable to fight and an actual threat. Like blob it spews out a bunch of simple mobs which are very obnoxious to deal with and make it even harder. Like blob you can just march up to the core and kill it though, which makes it easy in a way to stop. It's like space vines in that it's very tedious to fully get rid of and isn't really annouced and isn't super destructive to the station or atmos. Having everyone trying to bash down the fire mold because welders obviously don't work is shitty and unfun when it's spewing out fire zombies and making atmos uninhabitable because there's fire everywhere.

 In my round in particular engineering refused to fight it because the sentient zombies started crying about having a family and the crew sympathized with a fucking mold like a bunch of grognards (if blob did this it would probably win every time), essentially ruining the round when it spread all over maint until someone finally just walked up to the core and bashed it with a fire extinguisher.

Gandalf2k15 (approved):

Gandalf2k15:
We'll FM this, run more tests, and update the module as time goes on. Fed up of this shitty TM conflicts, and I want to work on this actively.

PR #1451, opened 2020-10-24
[WIP] [non-modular] Showers dont run out of water

### LINKED ISSUE #1389: Showers have finite water aAAAAAAAaaaaAAaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
AAAAAAAAAAAAAAAAaaaaaaaaaaaaaaaaaaaaaAAAAAaAAaaaAAaaaAAAAAAAAAAAAAAaaaaaaaaaaaaaaaaaaaa this ruins the shower rp, give us unlimited water back.

Azarak:
## About The Pull Request
fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/1389

## Changelog
:cl:
/:cl:

### DIFF
--- a/code/game/objects/structures/shower.dm
+++ b/code/game/objects/structures/shower.dm
@@ -127,7 +127,7 @@
 
 /obj/machinery/shower/process()
 	if(on && reagents.total_volume >= 5)
-		reagents.remove_any(5)
+		//reagents.remove_any(5) //SKYRAT EDIT REMOVAL - QOL
 		wash_atom(loc)
 		for(var/am in loc)
 			var/atom/movable/movable_content = am


Owai-Seek:
The hero we need. As a sidenote, it's the same for sinks, but I (guess) that's not as big an issue. Seems silly TG would do it at all.

Azarak:
> 
> 
> The hero we need. As a sidenote, it's the same for sinks, but I (guess) that's not as big an issue. Seems silly TG would do it at all.

realism, you wouldnt want someone to turn on all the showers and just leave them be with no care for water, now they'd run out and you'd have someone to blame, IC conflict. The more real a game feels the better the roleplaying experience is

Gandalf2k15:
Eh I don't agree with this one. It takes us away from roleplay.

Azarak:
> 
> 
> Eh I don't agree with this one. It takes us away from roleplay.

not worth it to go with realism on here, people like shower scenes, let them enjoy them

Gandalf2k15:
This was a TG change, they weren't much about realism and these showers last long enough for a scene. Fuck me we don't need to be bending over backwards to cater towards ERP.

Azarak:
> 
> 
> This was a TG change, they weren't much about realism and these showers last long enough for a scene. Fuck me we don't need to be bending over backwards to cater towards ERP.

How is this bending backwards for ERP? It's a very small change, really only noticable for those who do want to enjoy it

Gandalf2k15:
Please also remember this will cause showers to produce infinite chemicals, the reason for tg doing this was to prevent abuse as you can change the contents.

Hemlock32:
Don't showers use plumbing with the new system? Doesn't that mean you could fuck with the plumbing and make infinite chemical showers?

Owai-Seek:
Are showers pre-piped to get more water? How long does 200 units last?

Worst case scenario, we could make a subclass of shower specifically for the shower rooms/dorms that -only- produce water.

Azarak:
Oh it does use plumbing?

Azarak:
Okay this changes the things

MLGTASTICa:
Maybe do a creative fix and add a reagent synthethizer + chemister tank hooked up to showers ? instead of just deleting another way of creating RP by shower saboutage?

Azarak:
> 
> 
> Maybe do a creative fix and add a reagent synthethizer + chemister tank hooked up to showers ? instead of just deleting another way of creating RP by shower saboutage?

Good idea, but requires map maintenance which we currently dont do.
Inb4 sometime in the future we'll have disposal pipes, atmos pipes and hydraulic pipes all over the map

YakumoChen:
Surely there's a subtype of shower that has the old functionality? 
If not then just do a partial revert of the code.

Azarak:
i'll mess with this later

PR #1458, opened 2020-10-25
[MIRROR] Interview System / Soft Panic Bunker

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/54465
--------------------

## About The Pull Request

Ports and improves my interview system that has been previously used in the summer ball and toolbox tournament events. Allows for a 'softer' panic bunker, wherein players who fall below the required living time limit can still join the server and be restricted to filling out a questionnaire. Upon completing the questionnaire, the player may be allowed into the server by an administrator. If the application is approved, they get a notification that they will be reconnected and upon reconnecting will have all verbs as they usually would. If the application is denied the user is put on a cooldown after which they may submit a new questionnaire.

Players who are being interviewed (herein interviewees) have no verbs other than those required for the stat panel to function, as well as a verb to pull up the interview panel. Interviews do not persist through restarts, and the ability to join that is granted by an accepted interview is only valid for the duration of that round. 

Open interviews are listed under a new 'interviews' tab for admins, which is VERY similar to the existing tickets tab.

Below is what a player who is flagged as an interviewee will see when they join the server. They can do nothing but respond to the questionnaire or leave.
image

This is what an administrator sees after an interview is submitted, they will also see a corresponding message within their chatbox, and an age-old BWOINK when an interview is submitted.
image

The interviews tab, which is similar to the tickets menu. You can open the interview manager panel to view all active (including non-submitted) interviews, queued (submitted) interviews, and closed interviews.

image

### FAQ:
**What happens if someone submits an interview when no admins are on?** It's treated like adminhelps are, the message gets sent to TGS to be dispatched off to configured end-points (like Discord or IRC), and the user is notified that their interview was handled this way.

**Can you configure the questions?** Yes, in ``config/`` there is now a ``interviews.txt`` file in which the welcome message and the individual questions can be set and modified.

**Can this be turned on and off during a round?** Yes, it can be toggled like the panic bunker. It requires the panic bunker to be raised in order to function.

**Can interviewees have further questions asked to them?** Yes, if you admin-pm them, which is possible using regular means or a conveniently placed button on the interview UI, they will be able to respond to the message.

### Technical details

To use the interview system you must have the panic bunker enabled, this is an additional setting for the panic bunker. It can be set through the ``PANIC_BUNKER_INTERVIEW`` setting in ``config.txt``, or alternatively enabled in-game as prompted during the panic bunker toggling process. It also can be toggled on its own using a verb added for this purpose, ``Toggle PB Interviews`` found under the server tab. These new actions are included in the logging for the panic bunker. I have also added a reporting stat to the world topic  status keyword, which now reports if the interview system is on using the keyword ``interviews``.

As mentioned above, for server operators, configure the questions and welcome message in ``config/interviews.txt``.

Note to maintainers and those with big brains I had to add a call to ``init_verbs`` on the stat panel window being ready because seemingly a race condition exists wherein the ``add_verb`` of the 'view my interview' verb doesn't cause a refresh of the tabs (and therefore doesn't show the 'Interview' tab) when running in dream daemon but running it directly from visual studio code properly shows the tab. Adding a ``init_verbs`` call directly after adding the verb didn't seem to help.

### A note for downstreams

If you don't use the HTML stat panel (which may not be a bad thing) then you will have to do some conversion from the HTML stat panel stuff used here to the old style stat panels. It's pretty trivial, but just be aware of that. You can see how I used to use the old stat panels in my PR from the summer ball, [here](https://github.com/CitrusGender/tgstation/pull/113), which should be helpful.

## Why It's Good For The Game

This allows for a softer version of the panic bunker which impedes the flow of malicious players while allowing genuine players a chance to enter a round to gain enough time to not be affected by the panic bunker's restrictions.

## Changelog
:cl: bobbahbrown
add: Added the interview system, a 'soft' panic bunker which lets players who would normally be blocked from joining be interviewed by admins to be selectively allowed to play.
/:cl:


bobbah:
upset?

PR #1507, opened 2020-10-28
Added SQL logging to atoms

MissFox0810:
## About The Pull Request
Modifies Atoms.dm to add SQL logging for the message handler.
This allows SQL logging for all actions (Say,PDA,Attack,Admin,dsay etc) to allow for future work on an admin interface for log handling.

I only want this test murged for a round, both to see if it causes any performance impact on the server and to then collect the table and work on the web interface.

The system also has a flag "sql_LOG" that can be changed to disable the SQL logging without removing the code.
Could also be added so its a dynamic config option that can be enable/disabled mid-round in case of issues

Requires adding a table with the following structure to the database.
```sql
CREATE TABLE `game_log` (
  `ID` int(11) NOT NULL,
  `datetime` datetime NOT NULL,
  `round_id` int(11) NOT NULL,
  `ckey` varchar(60) COLLATE utf8_bin NOT NULL,
  `loc` varchar(60) COLLATE utf8_bin NOT NULL,
  `type` varchar(10) COLLATE utf8_bin NOT NULL,
  `message` varchar(1000) COLLATE utf8_bin NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
````

## Why It's Good For The Game
It's been a major request from nearly all the staff that the logging be improved. currently to search over past round you have to open each rounds file for each of the actions you want (If you wanted to check what "MrBadAtRp" was doing in a round, you have to open game.log, pda.log and attack.log then search and filter out everything you wanted)
With this PR it adds the ability to in future make a web interface were you can search all the logs from all the rounds at once.
Allowing you to see inline attack logs from multiple people at once, or see everyone talking in one area of the station.
Would also allow augmenting tickets and bans with log entries, so making the PC's job easier as well.

Here is some examples that can be done with SQL Querys to the log.
Example 1: Showing all actions in area
image

Example 2: Getting logs from two players at once, inline so you can much more easily see who attacked first.
image

## Changelog
:cl:
add: Added SQL Logging to Atoms.dm
Update: SQL Schema and comments on atoms.dm
/:cl:

### DIFF
--- a/SQL/tgstation_schema.sql
+++ b/SQL/tgstation_schema.sql
@@ -10,6 +10,21 @@
 /*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
 
 --
+-- Table Structure for table 'game_log'
+--
+-- Skyrat edit to enable SQL logging of all messages.
+
+CREATE TABLE `game_log` (
+  `ID` int(11) NOT NULL,
+  `datetime` datetime NOT NULL,
+  `round_id` int(11) NOT NULL,
+  `ckey` varchar(32) COLLATE utf8_bin NOT NULL,
+  `loc` varchar(60) COLLATE utf8_bin NOT NULL,
+  `type` varchar(10) COLLATE utf8_bin NOT NULL,
+  `message` varchar(1000) COLLATE utf8_bin NOT NULL
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+--
 -- Table structure for table `admin`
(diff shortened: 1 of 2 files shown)


MissFox0810:
Noticed that the varchar limit for "ckey" in the table needs changing from 32 to 60. to allow for long ckeys plus long playernames.

MissFox0810:
Updated the SQL Schema to add in the new table with correct varchar lengths. and added skyrat comments to code.

PR #1536, opened 2020-11-01
Light tweaks [SEMI-MODULAR]

Gandalf2k15:
## About The Pull Request

All lights now have a delay for turning on, just like IRL lights. They have a nice cool sound as well!

All light switches come with a new sound.

All areas(provided they have a lightswitch) have a 50% chance to start with lights off. Find. That. Light. Switch.

Example: https://streamable.com/98gwg7

## Why It's Good For The Game

Honestly it's underrated how much this makes things feel more... substantial. 

## Changelog
:cl:
add: Lights now have a new sound for turning on.
add: Light switches now actually switch.
tweak: Lights take some time to warm up and turn on.
tweak: All areas now have a 50% chance to start with lights off(provided the area has a lightswitch!)
/:cl:


code/modules/power/lighting.dm
```diff
@@ -368,31 +368,35 @@
...
-			if(rigged)
-				if(status == LIGHT_OK && trigger)
-					explode()
-			else if( prob( min(60, (switchcount^2)*0.01) ) )
-				if(trigger)
-					burn_out()
-			else
-				use_power = ACTIVE_POWER_USE
-				set_light(BR, PO, CO)
+		spawn(rand(10,30))	//SKYRAT EDIT ADDITION - AESTHETICS
```
Azarak: Use a timer or sleep() instead here perhaps?

Gandalf2k15: This should prevent any edge cases with spamming lights, as they will respect if the switch is OFF or ON, with the check you see, so that lights that are starting up, can still run their code if someone is spamming it. If we used sleep, people could spam the switch and then break lights as the whole code is sleeping. And timers, well, it prevents timer subsystem lag.

modular_skyrat/modules/aesthetics/lights/code/lighting.dm
```diff
@@ -1,3 +1,40 @@
 /obj/machinery/light
 	icon = 'modular_skyrat/modules/aesthetics/lights/icons/lighting.dmi'
 	overlayicon = 'modular_skyrat/modules/aesthetics/lights/icons/lighting_overlay.dmi'
+	var/maploaded = FALSE //So we don't have a lot of stress on startup.
+	var/turning_on = FALSE //More stress stuff.
+
+/obj/machinery/light/proc/turn_on(trigger)
+	turning_on = FALSE
+	if(!on || QDELETED(src))
+		return
```
Azarak:
```suggestion
	if(QDELETED(src) || !on)
		return
	turning_on = FALSE
```

modular_skyrat/modules/aesthetics/lights/code/lighting.dm
```diff
@@ -1,3 +1,40 @@
 /obj/machinery/light
 	icon = 'modular_skyrat/modules/aesthetics/lights/icons/lighting.dmi'
 	overlayicon = 'modular_skyrat/modules/aesthetics/lights/icons/lighting_overlay.dmi'
+	var/maploaded = FALSE //So we don't have a lot of stress on startup.
+	var/turning_on = FALSE //More stress stuff.
+
+/obj/machinery/light/proc/turn_on(trigger)
+	if(QDELETED(src) || !on)
+		return
+	turning_on = FALSE
```
Azarak:
```suggestion
	if(QDELETED(src))
		return
	turning_on = FALSE
	if(!on)
		return
```

PR #1554, opened 2020-11-03
[MIRROR] [READY] Swarmer QOL and Balance Update

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/54340
--------------------
## About The Pull Request

This update seeks to iron out some of the problems revolving around swarmers and the rework which occurred in June.  The following changes were made:

- Swarmers now teleport hulks away with their beam akin to simple mobs.  Before this change, hulks could not be stopped by swarmers outside of spamming walls, but that could never save the beacon.
- Swarmer drones now automatically follow the main swarmer and will stop to chase targets.  The middle-click mechanic was tedious at best.  Rule of thumb for how it works now, when you move, they move towards you.
- Ventcrawling has been removed.  Unfortunately due to the balance of ventcrawling in its current state, having a swarmer eat the ORM before you could even know they spawned just fails to be fun.  Will be watching to see if further buffs will be needed to compensate for this change.
- Swarmers now start with 20 resources upon being spawned.  This lets them set up some basic defenses or build a drone from the start.
- Swarmer event chance has been brought in line with other midround antagonists, making them slightly more common.
- Added swarmers to Dynamic midround spawns.  Its a personal thing of mine that every antag eventually get added to the system, and that includes swarmers.

## Why It's Good For The Game

Fixes some gripes about swarmers which persisted through the rework, and tackles the swarmer-eat-ORM-instantly issue without removing the possibility entirely.  Also gives some nice QOL changes in the form of drones auto-following you and having hulks no longer be unstoppable, along with 20 resources at spawn to quickly get something going.

## Changelog
:cl:
add: Dynamic Mode can now spawn Swarmers
balance: Swarmers have received a software update, changing some of their capabilities.  Nanotrasen reports a slight increase in strange objects in the neighboring sector.
/:cl:


Gandalf2k15:
I think we can re-enable swarmed.

Azarak:
> 
> 
> I think we can re-enable swarmed.

yeah they may be kinda better now without ventcrawling, cant do all the underhanded stuff so easily

Azarak:
I'll lower their weights though

PR #1850, opened 2020-11-26
Heads can no longer become most antags

Azarak:
## Changelog
:cl:
balance: heads can no longer roll antagonist roles
/:cl:


Azarak:
self speedmerging because no other maintainers up and im about to update the server and this is kinda important

PR #1853, opened 2020-11-26
Numbers are now accepted in name sanitization

### LINKED ISSUE #1845: Numerical Characters Don’t Save for IPC / Synthetic Lizardperson Names!
## Round ID:
45

## Reproduction:

This issue is rather singular, and affects the character names of character slots belonging to the synthetic lizardperson and I.P.C. species - numbers entered into the text box do not save, and those produced by the random name generator are purged from the character’s nameslot after reconnecting.

Recreation is quite simple! Create a new character of a synthetic lizardperson or I.P.C. species, and attempt to type a name with numbers in it (e.g. “ARMA-199”). Note that the name will save as “ARMA-“, and this lack of numbers will carry over into the next shift. Using randomly generated names will allow you to use numbers in said name until you restart the client or attempt to further edit the character slot’s name, at which point even the randomly generated name will suffer the same deletion of numerical characters. (The character slot name itself will not change unless saved after having been affected by the glitch, at which point it too will lose any numbers already purged from the nameslot.) Note that this glitch applies to every nameslot barring Cyborg and AI; I suspect it’s caused because the game doesn’t consider the synthetic lizardperson or I.P.C. species to be silicon and therefore prevents them from having non-roman characters in its name; this seems very out of place and I’m fairly certain this was not always the case, especially considering the list of randomly generated names does not account for it.

Azarak:
## About The Pull Request
Couple of people's characters have numbers in their names

fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/1845

## Changelog
:cl:
fix: name sanitization now accepts numbers
/:cl:

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -147,7 +147,8 @@
   * * strict - return null immidiately instead of filtering out
   * * allow_numbers - allows numbers and common special characters - used for silicon/other weird things names
   */
-/proc/reject_bad_name(t_in, allow_numbers = FALSE, max_length = MAX_NAME_LEN, ascii_only = TRUE, strict = FALSE)
+///proc/reject_bad_name(t_in, allow_numbers = FALSE, max_length = MAX_NAME_LEN, ascii_only = TRUE, strict = FALSE) //ORIGINAL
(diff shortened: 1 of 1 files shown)


Afya212:
Should link to #1845, looks good though.

Azarak:
> 
> 
> Should link to #1845, looks good though.

good catch

PR #1885, opened 2020-11-28
Disables ninja event

Azarak:
## About The Pull Request
Wow dynamic doesnt have a max occurences thing

## Why It's Good For The Game
Request from Xyel

## Changelog
:cl:
del: Ninja event has been disabled
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
+++ b/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm
@@ -644,6 +644,7 @@
 	requirements = list(101,101,101,80,60,50,30,20,10,10)
 	repeatable = TRUE
 	var/list/spawn_locs = list()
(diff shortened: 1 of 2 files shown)


Gandalf2k15:
Let's just remove the ability to spawn more events, instead of removing this all together

KathrinBailey:
I'm almost certain we can still make Space Ninja through PP transform and TP, just has to be done manually. So we just have to admin dchat and ask who wants it, open notepad.exe, jot down names and just roll a dice.
_Plus, that has the added benefit of not letting the shitter types take it._

hugmeorelse:
This is for dynamic. We need to disable this for event. We do not use dynamic.

Change max_occurrences variable here: https://github.com/Skyrat-SS13/Skyrat-tg/blob/a77acf11a48a2104d959633f00356bf635a7c86e/code/modules/events/space_ninja.dm

PR #1908, opened 2020-11-28
Adds logging to the SSD indicator.

Azarak:
## About The Pull Request
It would be useful in a ticket we've had today, where it was hard to determine if it actually showed up due to a long timeout

## Changelog
:cl:
/:cl:


KathrinBailey (approved):
Heckin' needs to happen ASAP so I can easily bonk people who mess with SSD players.

PR #1939, opened 2020-11-29
Brig timers now calculate from real time, not time dilation time. You can set brig timers to 60 minutes

Azarak:
## About The Pull Request
Title. I know realtime has inaccuracies, i wanna TM this and see if it works well anyway, as this will be useful

## Why It's Good For The Game
Saves possibly hours of time from dirty criminals, lost due to timespace continuum errors

## Changelog
:cl:
tweak: You can now set brig timers to 30 minutes
balance: Brig timers are now calculated from real time, not applying TD in effect
/:cl:


Gandalf2k15:
Hmm, could use timeofday instead here?

Azarak:
> 
> 
> Hmm, could use timeofday instead here?

Why?

Gandalf2k15 (approved):

Gandalf2k15:
> > Hmm, could use timeofday instead here?
> 
> Why?

Timeofday is more relevant as it only includes the time and not the date since 2000, its less computing power for byond to deal with.

Azarak:
> 
> 
> > > Hmm, could use timeofday instead here?
> > 
> > 
> > Why?
> 
> Timeofday is more relevant as it only includes the time and not the date since 2000, its less computing power for byond to deal with.

What if the day turns into the next day?

KathrinBailey:
I hate to be that guy but I feel like TG would want this x_x

Azarak:
I'll need to make some system keeping track of real time better. As relying on the byond stuff is hard. I'll do it later

Gandalf2k15:
This is old, and has been TM'd for long enough. Hasn't made any issues so far. No point in clogging TM's up with it. Full merge.

PR #1942, opened 2020-11-30
[Semi-Modular] Re-adds DNR and DNC traits, prevents clone pod from cloning synthetics

PatchyDragon:
## About The Pull Request

Adds back the Do Not Revive and Do Not Clone traits as they functioned in Skyrat prior to the rebase. Subjects with either of these traits will be unclonable, and the cloning machine will give a unique message for this failure. 

The code from the old base also used the line checking for DNR/DNC to check for synthetic species, which were previously not accounted for in code. I have added back the error message that should be given when attempting to clone a synthetic.

Oddly, the old DNR trait did not seem to prevent resuscitation via defib, but my goal was to recreate it exactly as it was and to let the dev team make any rebalance changes themselves. Thus, this little inconsistency remains, but after braindeath the subject will (in theory) be unrevivable by most means that call revive()

The old DNR trait made a change to carbon.dm, and I do not think this functionality could be achieved in a modular manner, so when adding that back that I have now also highlighted the change for future merge clarity.

## Why It's Good For The Game

The DNR and DNC traits added variety to character creation, and allowed for players to give themselves drawbacks with (seemingly) no upside for added challenge. Others have asked for it back, and I myself would like it back, so I've done some archaeological work and unearthed it again.

## Changelog
:cl:
add: Adds Do Not Revive and Do Not Clone traits
fix: Prevents synthetic races from being cloned 
/:cl:


PatchyDragon:
Huh. One of the linters failed. Unsure why.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -881,6 +881,10 @@
 	. = ..()
 	if(!getorgan(/obj/item/organ/brain) && (!mind || !mind.has_antag_datum(/datum/antagonist/changeling)))
 		return FALSE
+//SKYRAT EDIT ADDITION BEGIN - TRAIT_DNR
+	if(HAS_TRAIT(src, TRAIT_DNR))
+		return 0
```
Gandalf2k15:
```suggestion
		return FALSE
```

Gandalf2k15 (changes requested):
Looking good, commit the one change and we can go ahead.

Azarak:
Could you make them more flavorful? Throw in some medical terms and stuff, while still keeping it ambigous?

modular_skyrat/modules/cloning/code/game/machinery/computer/cloning.dm
```diff
@@ -558,6 +558,10 @@
 		scantemp = "<font class='bad'>Mental interface failure.</font>"
 		playsound(src, 'sound/machines/terminal_prompt_deny.ogg', 50, FALSE)
 		return
+	if(HAS_TRAIT(mob_occupant, TRAIT_DNC) || HAS_TRAIT(mob_occupant, TRAIT_DNR) || (ROBOTIC_DNA_ORGANS in dna.species?.species_traits))
```
Azarak:
```suggestion
	if(HAS_TRAIT(mob_occupant, TRAIT_DNC) || HAS_TRAIT(mob_occupant, TRAIT_DNR) || mob_occupant.mob_biotypes & MOB_ROBOTIC)
```

PatchyDragon:
> Could you make them more flavorful? Throw in some medical terms and stuff, while still keeping it ambigous?

The problem is DNC/DNR could be chosen due to medical complications, or could be religious or ethics based choice (sorta like how people on Baycode don't take neural laces or whatever), and I can't think of a way where committing to either in the text wouldn't rule out the other to some degree.

PatchyDragon:
Sidenote: I have a new love-hate with how github does suggestions as commits. On the one hand, this is great since I don't have to reopen the code locally. On the other _why did it not apply the batched commits I asked it to the first time?_

Had to go back and do a reset and force push it just to get the thing to accept your suggestions again...

Azarak:
> 
> 
> > Could you make them more flavorful? Throw in some medical terms and stuff, while still keeping it ambigous?
> 
> The problem is DNC/DNR could be chosen due to medical complications, or could be religious or ethics based choice (sorta like how people on Baycode don't take neural laces or whatever), and I can't think of a way where committing to either in the text wouldn't rule out the other to some degree.

Religious beliefs wouldnt stop someone from being cloned, but yeah I do see your point. When we get around to having neural laces then it could be a thing. (Which could not be that far away, since augmentation menu would support it)

PatchyDragon:
> > > Could you make them more flavorful? Throw in some medical terms and stuff, while still keeping it ambigous?
> > 
> > 
> > The problem is DNC/DNR could be chosen due to medical complications, or could be religious or ethics based choice (sorta like how people on Baycode don't take neural laces or whatever), and I can't think of a way where committing to either in the text wouldn't rule out the other to some degree.
> 
> Religious beliefs wouldnt stop someone from being cloned, but yeah I do see your point. When we get around to having neural laces then it could be a thing. (Which could not be that far away, since augmentation menu would support it)

If they RP'd having DNC/DNR tied to their record by request that'd prevent a cloner from doing it

Azarak (approved):

Gandalf2k15 (approved):

PR #1973, opened 2020-11-30
Teleporter changes [NON-MODUALR]

Gandalf2k15:
## About The Pull Request

Let's not go easy on dumb people that use an uncalibrated teleporter;
You'll lose 2 limbs, and you'll also receive a a lot of radiation.

## Why It's Good For The Game

Muh character.

## Changelog
:cl:
balance: Teleproters no longer change species, but are FAR more deadly when used uncalibrated.
/:cl:

### DIFF
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -95,7 +95,7 @@
 							qdel(BP) //WHEEEEE LOST TO SPACE N' TIME
 							dismembered++
 					//SKYRAT EDIT CHANGE END
-					//human.apply_effect((rand(120 - accuracy * 40, 180 - accuracy * 60)), EFFECT_IRRADIATE, 0) - SKYRAT EDIT REMOVAL
+					human.apply_effect((rand(120 - accuracy * 40, 180 - accuracy * 60)), EFFECT_IRRADIATE, 0)
 			calibrated = FALSE
 	return
 
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -88,16 +88,14 @@
 					var/dismembered = 0
 					to_chat(human, "<span class='danger'>Your limbs lose molecular cohesion as you teleport!</span>")
 					for(var/obj/item/bodypart/BP in human.bodyparts)
-						if(dismembered <= 2) //Remove two body parts.
+						if(dismembered <= 2) //Remove 3 body parts.
 							if(BP.name == "chest" || BP.name == "head")
 								continue
 							BP.dismember()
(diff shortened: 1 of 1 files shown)


tf-4:
Ehhhh I think three seems like a lot. It was never an issue before, I'm not sure why it would become one now.

Azarak:
I think 2 is great, 3 may be a bit much. With 3 they could be close to death due to bleeding fast? And 2 would give them enough time to suffer. Plus they'll be suffering from radiation poisoning

ShadeAware:
> 
> 
> I think 2 is great, 3 may be a bit much. With 3 they could be close to death due to bleeding fast? And 2 would give them enough time to suffer. Plus they'll be suffering from radiation poisoning

As far as I know, TG does not have bleeding stump wounds, so losing a limb is essentially losing all of the damage connected to that limb and nothing else. I could be wrong though.

Gandalf2k15:
Changed it to two limbs, and upped the rad damage.

Azarak:
> 
> 
> > I think 2 is great, 3 may be a bit much. With 3 they could be close to death due to bleeding fast? And 2 would give them enough time to suffer. Plus they'll be suffering from radiation poisoning
> 
> As far as I know, TG does not have bleeding stump wounds, so losing a limb is essentially losing all of the damage connected to that limb and nothing else. I could be wrong though.

this will need to be fixed in another PR 👀

code/game/machinery/teleporter.dm
```diff
@@ -77,13 +77,25 @@
...
+					var/dismembered = 0
+					to_chat(human, "<span class='danger'>Your limbs lose molecular cohesion as you teleport!</span>")
+					for(var/obj/item/bodypart/BP in human.bodyparts)
+						if(dismembered <= 1) //Remove 3 body parts.
+							if(BP.name == "chest" || BP.name == "head")
+								continue
+							BP.dismember()
+							qdel(BP) //WHEEEEE LOST TO SPACE N' TIME
+							dismembered++
+					//SKYRAT EDIT CHANGE END
```
Azarak:
```suggestion
						if(dismembered < 2) //Remove 2 body parts.
							if(BP.body_zone == BODY_ZONE_CHEST || BP.body_zone== BODY_ZONE_HEAD)
								continue
							BP.dismember()
							qdel(BP) //WHEEEEE LOST TO SPACE N' TIME
							dismembered++
						else
							break
					//SKYRAT EDIT CHANGE END
```

Owai-Seek:
Doesn't this mean that the teleporter can just, behead you?

*Edit*

Nevermind, had to read more thoroughly.

PR #1991, opened 2020-12-01
Fixes all issues with slime mutations, allows species changes to retain all mutant parts if able

Azarak:
## Changelog
:cl:
fix: Fixes all issues regarding slime transformations
/:cl:


Azarak:
Why is linter being a cunt
firefox_2020-12-01_13-25-12
sublime_text_2020-12-01_13-25-20
it's literally a typed

PR #1997, opened 2020-12-01
Makes lung damage from cigarettes only 50% as effective

Azarak:
## Why It's Good For The Game
Having lung damage from cigs is one thing, but people get lung failure is another, and its not that uncommon, so that's kinda bad. Still is enough damage to like. have doctors warn you not to smoke though, I guess?

## Changelog
:cl:
balance: Cigs will hurt your lungs much less
/:cl:


Gandalf2k15:
Woooo we promoting smoking ;-;

Azarak:
The damage is quite big

PR #2000, opened 2020-12-01
Makes two mood messages first person

tf-4:
## About The Pull Request

Title

## Why It's Good For The Game

Consistency

## Changelog
:cl:
fix: Makes mood messages consistently first person
/:cl:


KathrinBailey:
This is another change that needs to happen upstream, non-modular edits like this just makes the project less viable in the long term. We want as few non-modular edits as possible, so if it's something that would be accepted on TG, it should be PRed there first.
If TG deny it, poke me and I'll reopen.

PR #2033, opened 2020-12-03
Nerfs R window melee resistance

Azarak:
## About The Pull Request
It's overtuned and it was very apparent when a window could stop a horde of monsters when an event was ran.
It still has the increased integrity it's got from the previous buff

## Changelog
:cl:
balance: Reinforced windows now have 50% melee resistance from 80%
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -380,7 +380,8 @@
 	icon_state = "rwindow"
 	reinf = TRUE
 	heat_resistance = 1600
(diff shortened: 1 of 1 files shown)


Owai-Seek:
Does this still make it so you need to meet a certain threshold of damage? I know you can break the windows with spears, but can't break it with weaker (IE: 10 force) weapons.

From what it looks like, it deflects anything that doesn't do at least 11 damage.

PR #2035, opened 2020-12-04
[Ready] [Not Modular] Expansive Smuggling

Zenitheevee:
## About The Pull Request

Changes and expands the Smugglers Satchels at round spawn to have 3 items (As opposed to 2) And to pull from a different pool of loot.

Changes:
>Removed Dope necklace + Tacticool turtleneck. Smugglers have valued these as not worth smuggling. (Also removes posters)
>Added a variety of new items, Including Liberty caps, Syndicate toolboxes, Syndicate soap, Chest rigs, and doctorable ID's, And those ever-rare Telecrystals in single packs
>Stuffed another item into satchels
>Makes sure that no item repeats. No buyer wants to sample the same thing twice.

## Why It's Good For The Game

Makes placing and finding smuggler satchels better for syndicate members

## Changelog
:cl:
tweak: Smugglers have decided to stuff caches a bit more.
tweak: Smugglers have decided to venture into more lucrative caches.
code: Removes super call from with_tools variant of smuggler satchels.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -322,18 +322,28 @@
 	STR.set_holdable(null, list(/obj/item/storage/backpack/satchel/flat)) //muh recursive backpacks)
 
 /obj/item/storage/backpack/satchel/flat/PopulateContents()
-	var/datum/supply_pack/costumes_toys/randomised/contraband/C = new
+	//SKYRAT EDIT CHANGE BEGIN
+	/*
+	/datum/supply_pack/costumes_toys/randomised/contraband/C = new
 	for(var/i in 1 to 2)
 		var/ctype = pick(C.contains)
 		new ctype(src)
 
 	qdel(C)
+	*/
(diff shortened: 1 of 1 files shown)


hugmeorelse:
Smuggling is good. I think adding in more options for smuggling would be interesting as well, especially when it comes to hiding them on yourselves.

Zenitheevee:
After a lot of forced-spawning for weight calculations, I'm happy to say that I'm more than pleased, Personally, with the weighting of the items.

The rarest item, A toolbelt with a set of Syndicate tools including a wrench that does a whole wopping 1 more damage than other wrenches only spawned twice out of 261 total items spawned at a 0.5% weighting. Single TC crystals spawned around 5 times at 1%, And everything else fell within parameters that within normal spawning conditions I feel would be Rare, But not horribly broken for one to have.

Also, Small sidenote that... MAY be good for this, Is that it DOES prevent 2 bottles of "Suspicious pills" from spawning, Which means you'll never get 3 pillbottles. While i'm not sure _why_ this happens, i surmise its because while they're clearly labed as aresnep, LSD and Happy pillbottles in code, In function they're all "Suspicious pill bottles" And such wont pull from them.

However, I've noted that this behavior does not work with Cannabis leaves.

code/game/objects/items/storage/backpack.dm
```diff
@@ -322,18 +322,28 @@
...
+		new contraband_type(src)
+
+	//SKYRAT EDIT CHANGE END
 
 /obj/item/storage/backpack/satchel/flat/with_tools/PopulateContents()
 	new /obj/item/stack/tile/plasteel(src)
 	new /obj/item/crowbar(src)
 
-	..()
+//	..() SKYRAT EDIT REMOVAL
```
Gandalf2k15, line 346: Should this not be a parent callback? . = ..()?

Zenitheevee: Seeing as there are some items within the list that could turn this into a net profit, I think it would be inadvisable for the with_tools variant to call back and pull from the item set.

Ontop of that, Functionally speaking it does not require it, and re-instating it could easily turn a 1TC item into a loot-box scenario "Buy now" item, yet i'd prefer it stay as a 1TC item and so i cut out the callback.

Gandalf2k15: Got it

Gandalf2k15:
Seems fine to me. I'll wait on azaeaks input

PR #2072, opened 2020-12-07
Removes clown car from uplink

Azarak:
## Why It's Good For The Game
It was previously removed for a good reason. The whole nature of the item is way too clowny 

## Changelog
:cl:

del: Removes clown car from uplink
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1818,6 +1818,8 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	cost = 10
 	restricted_roles = list("Research Director", "Scientist", "Roboticist")
 
+//SKYRAT EDIT REMOVAL BEGIN
+/*
 /datum/uplink_item/role_restricted/clowncar
 	name = "Clown Car"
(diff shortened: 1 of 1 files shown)


Owai-Seek:
I think it's an alright item when people use it for gimmicks on lowpop (taxi-service and whatnot), but the chemicals it can spew out are apparently -very- deadly. (From what it looks like, it just spits out random stuff, which could be completely harmless, or kill someone in seconds..)

Having to ask for some items isn't an entirely bad thing, it establishes trust between the admin granting it, and the person asking for it, that there won't be -too- many shenanigans happening.

hugmeorelse:
> I think it's an alright item when people use it for gimmicks on lowpop (taxi-service and whatnot), but the chemicals it can spew out are apparently -very- deadly. (From what it looks like, it just spits out random stuff, which could be completely harmless, or kill someone in seconds..)
> 
> Having to ask for some items isn't an entirely bad thing, it establishes trust between the admin granting it, and the person asking for it, that there won't be -too- many shenanigans happening.

It can be requested by a good clown. Otherwise, unless we make it less easy to grief with, it will be banbait. We do wish to have this return at some point.

PR #2087, opened 2020-12-08
Table flipping! [Modular]

Gandalf2k15:
Flip!

## About The Pull Request

This adds the ability to flip tables, the integrity is equal to that of the table you flipped. There are a few exceptions here, like reinforced tables, they can't be flipped.

To flip a table ctrl+shift+click it. It will do_after the max_integrity/4. This prevents spam and provides some balance.

image

Currently, it will block all objects coming from that dir.

## Why It's Good For The Game

More combat stuff!

## Changelog
:cl:
add: You can now flip tables by ctrl+shift+clicking them!
/:cl:


Owai-Seek:
Is it alt clicking or ctrl clicking that displays everything on a tile?

It might be necessary to check someone's intent so people don't accidently flip stuff when searching through piles.

RatFromTheJungle:
looks great, but altclick is an inconvenient keybind, it'd suck to be trying to get something out of a pile with altclick, just to chimp out against the table instead

NotRanged:
>Currently, it will block all objects coming from that dir.

Make this an armor piercing value check, perhaps? Wood and thin sheet metal aren't exactly well known for stopping bullets. Concealment, not cover. Perhaps there's a way to reduce the accuracy of bullets that pass through it.

Gandalf2k15:
Updated to be ctrl shift click

modular_skyrat/modules/tableflip/code/flipped_table.dm
```diff
@@ -0,0 +1,79 @@
...
+	if(!istype(user) || !user.can_interact_with(src))
+		return
+	if(can_flip)
+		user.visible_message("<span class='danger'>[user] starts flipping [src]!</span>", "<span class='notice'>You start flipping over the [src]!</span>")
+		if(do_after(user, max_integrity/4))
+			var/obj/structure/flippedtable/T = new flipped_table_type(src.loc)
+			T.name = "flipped [src.name]"
+			T.desc = "[src.desc] It is flipped!"
+			T.icon_state = src.base_icon_state
+			T.dir = user.dir
```
Azarak, line 60:
```suggestion
			T.dir = user.dir
			if(T.dir == NORTH)
				T.layer = BELOW_MOB_LAYER
```

modular_skyrat/modules/tableflip/code/flipped_table.dm
```diff
@@ -0,0 +1,79 @@
...
+	opacity = FALSE
+	var/table_type = /obj/structure/table
+
+/obj/structure/flippedtable/CanAllowThrough(atom/movable/mover, turf/target)
+	. = ..()
+	var/attempted_dir = get_dir(loc, target)
+	if(table_type == /obj/structure/table/glass) //Glass table, jolly ranchers pass
+		if(istype(mover) && (mover.pass_flags & PASSGLASS))
+			return TRUE
+	if(istype(mover, /obj/projectile))
```
Azarak, line 18: Allow them to pass if they're facing the direction in which it is flipped? It's how you'd use tables for cover and shooting

Azarak:
Skeptical about making it a new type, but otherwise its fine

There's also matters of icon smoothing and adjacent tables, but it'd be much more work

Also an idea for future: Allowing tables to be dragged very slowly, plus depending on adjacency to other tables. This may make setting up cover much more organic and less anemic like - deconstruct table, construct in another place, flip

modular_skyrat/modules/tableflip/code/flipped_table.dm
Azarak:
```suggestion
			var/new_dir = get_dir(user, T)
			T.dir = new_dir
			if(new_dir == NORTH)
				T.layer = BELOW_MOB_LAYER
```

modular_skyrat/modules/tableflip/code/flipped_table.dm
```diff
@@ -0,0 +1,79 @@
...
+		return FALSE
```
Azarak:
```suggestion
	if(istype(mover, /obj/projectile))
		var/obj/projectile/P = mover
		//Lets through bullets shot from behind the cover of the table
		if(P.trajectory && angle2dir_cardinal(P.trajectory.angle) == dir)
			return TRUE
		return FALSE
```

Azarak:
Also update changelog for the updated control

Azarak (approved):
would prefer it as a functionality of /table but still alright

PR #151, opened 2020-08-01
Ports cloning [WIP] [DNM]

Gandalf2k15:
## About The Pull Request
Does what it says on the tin. - Adds cloning back in.

## TODO:
Make the whole cloning process a little cooler(failed clones etc)
Add defective clone
Add the old space ruin back to spawning
Add boards to cargo
Add back the cloning to medical
Testing

## Why It's Good For The Game
https://github.com/tgstation/tgstation/pull/48668

## Changelog
:cl:
add: Ported cloning
/:cl:

PR #2138, opened 2020-12-11
Lowers stasis bed surgery efficiency by 5%

tf-4:
## About The Pull Request

Title

## Why It's Good For The Game

Nobody uses the damn operating tables, stasis beds are too strong. This just leads to 4 brain surgeries in one big open room in medical which is pretty LRP. Stasis beds need a little more tuning than just this, but it's a start.

## Changelog
:cl:
balance: lowers stasis bed efficiency by 5%
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_helpers.dm
+++ b/code/modules/surgery/surgery_helpers.dm
@@ -3,7 +3,8 @@
 	if(locate(/obj/structure/table/optable, T))
 		return 1
 	else if(locate(/obj/machinery/stasis, T))
-		return 0.9
+		//return 0.9 //ORIGINAL
+		return 0.8 //SKYRAT EDIT CHANGE - drops success chance on stasis beds by 10%. Use a damn operating table
 	else if(locate(/obj/structure/table, T))
 		return 0.8
 	else if(locate(/obj/structure/bed, T))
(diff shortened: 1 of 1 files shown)


Owai-Seek:
Surgery tables are for optional surgeries, stasis beds are so people don't actually die.

Making the surgery success chance lower doesn't promote any RP, it'll just make it so it takes longer for people to get fixed, and taken out of medbay.

davidchan13:
Seems like a remap is in order if we want this to be more RP friendly, people are still going to use these because they are accessible for surgery (and rightly should be.)

Trauma Ward/Treatment Center is basically the emergency room for Medical, the exam rooms are effectively worthless mechanically and the only RP that happens in them is ERP. If you're trying to promote RP then cosmetic details like privacy curtains would make more sense than just nerfing an item being used as intended (for life saving procedures) until the alternative is used isn't accomplishing it.

chesse20:
bruh just bring back cloning so we can keep medbay for rp and just use cyro pods to actually heal and cloning to actually revive rest is just for rp

atakiya:
> 
> 
> bruh just bring back cloning so we can keep medbay for rp and just use cyro pods to actually heal and cloning to actually revive rest is just for rp

Cloning's been in for a month now... #151

Gandalf2k15:
Hello, I don't think we require this. If you can think of a more intuitive way to change medbay, do so.

PR #2141, opened 2020-12-11
Let's turn the amount of retained messages in the chat up. [DNM]

### LINKED ISSUE #1852: Chat history too short.
The current chat system is unable to hold a full rounds worth of stuff. Maybe add an option allowing people to choose how much stuff is saved on their end, along with an option to delete the current logs, like the old pre-rebase server? It would also be nice if you could export logs that are just the logs of the filters you were using in question, instead of the whole thing, but that is less important than the above two.

Useroth:
While lower values would work fine on most gameplay-focused servers, on a roleplay-focused one it would be nice to be able to scroll up a bunch further up.


Useroth:
Also, while the idea is noble, let's actually discuss the merits, advantages and disadvantages before someone presses the button to merge it.

atakiya:
> 
> 
> Also, while the idea is noble, let's actually discuss the merits, advantages and disadvantages before someone presses the button to merge it.

Yeah, raising it that drastically _could_ bring issues for certain clients, especially those with lower system memory.

I don't like how it is currently either though, something akin to modern IM clients where it loads more messages from a DB would be nice, but that is a topic for another day.

I would say, test merging this first for a day or so would be a good way to see if major issues crop up.

Zenitheevee:
so, First thing first worth asking: 
How will this effect other clients using the same system? As TG chat currently stands, all clients retain X chat messages Even after turning off their client. If they log into a different server, They still retain some messages, and i think this would be an important question to delve into.

The advantages are obviously that it can give people much longer times to pull evidence for reports. 

The Disadvantages... well, Memory usage is the first one that comes to mind. And at really long times it can be a pain to search through.

Useroth:
> How will this effect other clients using the same system?

@Zenitheevee 
While I haven't tested, I imagine that when you join other server, with lower message limits there, it would trim your logs to that limit.

atakiya:
> 
> 
> > How will this effect other clients using the same system?
> 
> @Zenitheevee
> While I haven't tested, I imagine that when you join other server, with lower message limits there, it would trim your logs to that limit.

The load function loads *all* messages. The limit is defined in the save function.
See below

```js
const saveChatToStorage = async store => {
  const state = selectChat(store.getState());
  const fromIndex = Math.max(0,
    chatRenderer.messages.length - MAX_PERSISTED_MESSAGES); /* current amount of messages - max saved amount */
  const messages = chatRenderer.messages
    .slice(fromIndex) /* cut it at the above calculated position and return everything after */
    .map(message => serializeMessage(message));
  storage.set('chat-state', state);
  storage.set('chat-messages', messages);
};
```
And the load function
```js
const loadChatFromStorage = async store => {
  const [state, messages] = await Promise.all([
    storage.get('chat-state'),
    storage.get('chat-messages'), /* literally just get everything in the appropiate storage, no limit */
  ]);
  // Discard incompatible versions
  if (state && state.version <= 4) {
    store.dispatch(loadChat());
    return;
  }
  if (messages) {
    const batch = [
      ...messages,
      createMessage({
        type: 'internal/reconnected',
      }),
    ];
    chatRenderer.processBatch(batch, {
      prepend: true,
    });
  }
  store.dispatch(loadChat(state));
};
```

However, any time `rebuildChat()` runs, it also prunes the chat to the **_current_ server's limit**, so the "apply now" in the settings, or just natural prune over time will apply the current server's length, obviously.

Seris02:
How about making this an option in the TGUI chat? As in let them type in how many messages to keep?

atakiya:
> 
> 
> How about making this an option in the TGUI chat? As in let them type in how many messages to keep?

Not impossible, and I would agree (having options is always better than not) - **HOWEVER** most players wont know what setting is good performance wise **with the current implementation**.
We're not talking just "lag" here - we're talking issues like full client crashes / loss of logs.
Personally I think we will need to just improve tgchat first to handle this in a different way. Else we're not getting far, even with having an option to a hardcoded limit

atakiya:
Also here's my experience with the last testmerge for information's sake:
- more soft crashes were observed (non-client crashes)
- client took a lot longer to fully initialize, most likely due to the higher amount of messages having to be loaded back into buffer

Gandalf2k15:
I think there are smarter ways to go about this tbf, this is the least favourable one imo. Let's close this for now and maybe come up with a better solution, if necessary, at a later date.

PR #2166, opened 2020-12-13
[MIRROR] [READY] Creates Datumized AI and applies it to monkeys

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/55238
--------------------
:cl: Qustinnus
 refactor: Monkey AI is now contained in a datum, allowing for it to be applied to other mobs.
/:cl:

This PR introduces an AI system that is a mix between finite state machines and behavior trees, but in a different structure to run better on Byond.

The main structure works like this:
/datum/ai_controller - Datum that is the core behavior controller for the mob, it contains the planning of actions and important data on the ai itself

list/blackboard - An assoc list on the ai_controller with string keys, and dynamic values. This is used to contain data the AI knows about such as if its aggressive, who its target is, etcetera. This allows us to have varying data on a type per type basis while allowing them to be compatible with ai_behaviors

/datum/ai_behavior - A datum which contains an encapsulated task an AI can perform. Although in the best cases these are generic, they can be specific (Such as monkey equipping and monkey disposaling). By utilizing the blackboards on the controllers, these can remain purely as singletons.

The creation of behavior is handled by the ai_controller, which runs through a list of checks as you would do with normal AI, and then decides which behaviors to run. Aslong as there's behaviors running, it doesn't create new plans. Behaviors themselves should handle being cancelled or completing. this lowers the amount of CPU time we use on deciding what the best course of action is, but can make designing them slightly more difficult. I think this is a design pattern we should be able to get into but if it is an issue please let me know!

All in all, memory wise this system is also very lightweight in theory, there is one ai_controller on each atom with AI, and every behavior datum is only instanced once, with them acting as behavior containers and using the controller as data storage.

The main benefits of this system are as follows:

- New AI systems can be built with modularity in mind; allowing for re-usable behaviors
- AI can be made to work on multiple types. E.g. in this PR; the monkey AI will work on humans too (And on any other living mob capable of grabbing/holding things)
- The mobs themselves no longer hold tons of AI behavior, giving us the future possibility of making simplemobs simple again.

Actual use-cases for this in the future could be things such as "Monkeys turned into humans in genetics will have their AI still".

Again, this PR is heavily in production still, but I really wanted to get some eyes on this because it's a big system that I had a lot of trouble with code design-wise. (Also yes this edits balance right now but the idea is that by the time I'm further ahead the monkeys act mostly the same and there is no additional functionality)

vids:
AI on a monkey: https://streamable.com/v378qb (High attack speed atm because I havnt introduced a cooldown system)
AI on human mob: https://streamable.com/dt7hdh (This isn't actually in the game besides admin-spawn, but it shows the capabilities of the system)

todo:

- [x] Clean-up some of the uglier code
- [x] Do more bug-testing with all the types of behaviors monkeys have
- [x] Re-add monkey recruiting
- [x] Add attack cooldowns
- [x] Create readme for AI
- [x] Improve documentation


RatFromTheJungle:
ok, pretty based ill allow it

PR #2175, opened 2020-12-14
Makes command immune to FAMILIES

GuppyLaxx:
Heads of Command cannot spawn as family leaders or join families. This PR also hardcodes Security to be unable to join them at all, even if they lose their mindshield.
:cl: GuppyLaxx
add: Added immunity from Families to Heads of Staff (And sec)
/:cl:

### DIFF
--- a/code/game/gamemodes/gang/gang_things.dm
+++ b/code/game/gamemodes/gang/gang_things.dm
@@ -12,9 +12,9 @@
 
 /obj/item/gang_induction_package/attack_self(mob/living/user)
 	..()
-	if(HAS_TRAIT(user, TRAIT_MINDSHIELD))
+	/*if(HAS_TRAIT(user, TRAIT_MINDSHIELD))
 		to_chat(user, "You attended a seminar on not signing up for a gang and are not interested.")
-		return
+		return*/
 	if(user.mind.has_antag_datum(/datum/antagonist/ert/families))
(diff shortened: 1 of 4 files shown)


Owai-Seek:
Not really a fan of this, it's completely voluntary to join a family so there's no reason a head of staff or security shouldn't be able to if they have reason for it.

Spawning as a family member though, I'm fine if heads or security don't -spawn- as one.

RatFromTheJungle:
heads not starting as one, is a great idea
non security based heads should still be able to join, but security should probably be barred from it
Mostly because antag security are legally, and for the most of the crew who don't want slapped with a powergaming note, mechanically untouchable, and security have everything to gain from joining a gang in a purely OOC sense, because now they can /practically/ do anything.

SpookyTheFox:
Security, and particularly Blue Sheilds, have been abusing families. Good PR.
On a related note, Obsession seems to be picking a lot of security and captains as of late.

Gandalf2k15:
Removing the mindshield should allow conversion to prevent needless killing.

ReynTheLord:
One problem is that one of the gangs objectives, that being the jack bros, requires getting a member of the gang to be captain, or vica versa. If the ability for command to join gangs is removed... well... It'll turn into coup city if the gangs ACTUALLY do gang objectives.

Gandalf2k15:
As you have completed the changes requested, I'm happy to merge this, No point in keeping it TM'd.

PR #1913, opened 2020-11-28
[Not Modular] Defibrillates bonus ore from the ORM. 

ShadeAware:
## About The Pull Request
Better upgrades now equates to more resources, as it was on the oldbase.

## Why It's Good For The Game
Doing literally anything right now is hell and we constantly run out of all materials. It wasn't fun in the old base when we tried it, its not going to be fun now.

## Changelog
:cl:
balance: Upgrades in the ORM actually make the output of the ORM better again.
/:cl:

PR #2232, opened 2020-12-18
[Not Modular] Tries to revive ORM upgrades again with changes

ShadeAware:
## About The Pull Request
ORM upgrades actually do shit again, but not as much as they used to.

## Why It's Good For The Game
Finite materials are fine, but running out of every material every 5 minutes isn't fun, engaging, or actually challenging. Its tedium and makes everything just miserable to deal with.

## Changelog
:cl:
balance: ORM upgrades do thing again.
/:cl:

### DIFF
--- a/code/modules/mining/machine_redemption.dm
+++ b/code/modules/mining/machine_redemption.dm
@@ -35,6 +35,18 @@
 	materials = null
 	return ..()
 
+//Skyrat change begin
+/obj/machinery/mineral/ore_redemption/RefreshParts()
+	var/point_upgrade_temp = 1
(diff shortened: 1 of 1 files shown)


H0gberg:
good PR! +1

atakiya:
Sorry but from objectively checking the situation over weeks, there is no need for this PR.
Materials are supposed to _sometimes_ run out - e.g. IC caused issues - just like pointed out in #1913
Keep in mind, if more materials are needed, theres more ways to get them than just mining as well.
Mining points are also more than enough.

If you can find something else that the parts upgrade can do without touching the above two points, feel free to commit/comment.

PR #2316, opened 2020-12-25
Removes Claps Once spam

GuppyLaxx:
Removes (hopefully) claps once spam by adding a SPECIAL thing that has a chance to happen when you press the claps once button

## Why It's Good For The Game
I hate this emote with a passion and it's annoying constantly seeing borgs and people running around spamming their key bound to this emote

:cl:
add: Added a cool thing that happens if you spam claps once
/:cl:

https://user-images.githubusercontent.com/25938722/103117707-1344e480-4642-11eb-8426-2bed281038fb.mp4


ShadeAware:
This is quite a shoddy method of going about this - a better method would be to have it detect when spam is actually occuring, like how other emote stopgaps work. This just adds a flat 1% chance to punish you whether you're actually spamming it or not.

Azarak:
Very bad method to do it, and also you earrape people

KasparoVy:
Thunderclap.

Owai-Seek:
I don't think anyone who does this will be disincentivized from doing it by a frying pang bong. If anything it'd make them do it more.

GuppyLaxx:
Not really close to earrapey levels, and it's not exactly just "le loud sound" it causes arm damage and prevents the claps once sound from happening for a minute. If you want me to check that someone's spamming it I'm gonna have to add a whole new variable that is processed like what spin does, but spin's variable is used with other things (dizzyness)
Aside from some damage that even a medibot can heal in a quick moment, it doesn't deter people from using it as normal, but spamming it is stopped for a minute each time it happens, and the damage stacks on quickly if you run around spamming the button

atakiya:
+1 for the hilarity of it, but else I agree with the others.
It won't stop the spam. The one minute timeout on that probability is like russian roulette. Some will just be able to spam anyways.

We do plan to tackle this problem in a different and more global way for most emotes, *clap1 isn't the only one being spammed, see #2320 which will also be a temporary (but helpful) bandaid for the time being.

GuppyLaxx:
The deterrent is more so the fact it happens at all. I can easily spam spin, flip, etc. But the fact I can quite literally throw up or fall down and smack myself on the floor through pure rng is enough to just make go "eh, I'll use it once" or even not at all

H0gberg:
this makes me want to clap more

PR #2379, opened 2020-12-30
[MIRROR] Refactors monkeys into a species

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/55614
--------------------
:cl: Qustinnus, Fikou for sprites
refactor: Monkeys are now a humanoid species rather than a subtype of carbon.
/:cl:

This removes living/carbon/monkey and turns it into a species datum.

I might kill off attack_paw in this PR too, but because there's a lot more edge-cases involved with that I might do that in a later PR and make sure this works first.

This'll definitely cause bugs, but it is hard to avoid with mobcode PRs of this size.  PR should be testmerged for a good bit before this is merged.

also this fixes the healthbar below monkey thing.

fixes #52711


hugmeorelse:
monkee

PR #2438, opened 2021-01-02
Lowers down and disables some events

Azarak:
## About The Pull Request
on the request of Xyel

## Changelog
:cl:
tweak: Appendicitis is now less likely to occur
tweak: Spacevines are now less likely to occur
tweak: Brand Intelligence event now cant occur
tweak: Spider Infestation is now less likely to occur
tweak: Random heart attack is now less likely to occur
/:cl:

### DIFF
--- a/code/modules/events/blob.dm
+++ b/code/modules/events/blob.dm
@@ -1,7 +1,8 @@
 /datum/round_event_control/blob
 	name = "Blob"
 	typepath = /datum/round_event/ghost_role/blob
-	weight = 10
+	//weight = 10 //ORIGINAL
+	weight = 5 //SKYRAT EDIT CHANGE
 	max_occurrences = 1
 
 	min_players = 20
--- a/code/modules/events/blob.dm
+++ b/code/modules/events/blob.dm
@@ -1,8 +1,7 @@
 /datum/round_event_control/blob
 	name = "Blob"
 	typepath = /datum/round_event/ghost_role/blob
-	//weight = 10 //ORIGINAL
-	weight = 5 //SKYRAT EDIT CHANGE
+	weight = 10
 	max_occurrences = 1
 
 	min_players = 20
--- a/code/modules/events/brand_intelligence.dm
+++ b/code/modules/events/brand_intelligence.dm
@@ -4,7 +4,8 @@
 	weight = 5
 
 	min_players = 15
-	max_occurrences = 1
+	//max_occurrences = 1 //ORIGINAL
+	max_occurrences = 0 //SKYRAT EDIT CHANGE
 
 /datum/round_event/brand_intelligence
 	announceWhen	= 21
--- a/code/modules/events/heart_attack.dm
+++ b/code/modules/events/heart_attack.dm
@@ -1,7 +1,8 @@
 /datum/round_event_control/heart_attack
 	name = "Random Heart Attack"
 	typepath = /datum/round_event/heart_attack
-	weight = 20
+	//weight = 20 //ORIGINAL
+	weight = 10 //SKYRAT EDIT CHANGE
 	max_occurrences = 2
 	min_players = 40 // To avoid shafting lowpop
 
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -1,7 +1,8 @@
 /datum/round_event_control/spacevine
 	name = "Spacevine"
 	typepath = /datum/round_event/spacevine
-	weight = 15
+	//weight = 15 //ORIGINAL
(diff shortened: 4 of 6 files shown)


code/modules/events/blob.dm
```diff
@@ -1,7 +1,8 @@
 /datum/round_event_control/blob
 	name = "Blob"
 	typepath = /datum/round_event/ghost_role/blob
-	weight = 10
+	//weight = 10 //ORIGINAL
+	weight = 5 //SKYRAT EDIT CHANGE
```
NotRanged: Blob is already so rare, I barely see it once a day :(

EvilCrimsonWitch: Aw, I like blob. That's a shame.

Azarak: I dont mind having it be either, convince Xyel as he wants to have it changed

Azarak: No longer reduced

SpaceVampire:
Why is Brand Intelligence being removed? I never heard anyone complain about it. I just wish we could see the reason for these charges in general. I understand Spider Infestation, because we actually had a vote on that one.

Azarak:
> 
> 
> Why is Brand Intelligence being removed? I never heard anyone complain about it. I just wish we could see the reason for these charges in general. I understand Spider Infestation, because we actually had a vote on that one.

@hugmeorelse

YakumoChen:
> Why is Brand Intelligence being removed? I never heard anyone complain about it. I just wish we could see the reason for these charges in general. I understand Spider Infestation, because we actually had a vote on that one.

It removes every vending machine onboard, severely reducing access to station resources, and makes a ton of holes in the hull, and then the vending machines tend to kill people on top of it. A successful brand intelligence is usually a guaranteed shuttle call.

Oscari-pog:
Pointless to decrease appendicitis and random heart attack, happens once per shift and the person is usually treated as soon as they notice something wrong.

Spacevines are usually killed by a assistant with a welder - no reason to decrease them. The only reason for this I can think of is that one round where ultravines were bussed in.

Didnt we already decrease spiders? Spiders are awful at their current state so I cant complain about that one.

KathrinBailey (approved):

hugmeorelse:
> Pointless to decrease appendicitis and random heart attack, happens once per shift and the person is usually treated as soon as they notice something wrong.
> 
> Spacevines are usually killed by a assistant with a welder - no reason to decrease them. The only reason for this I can think of is that one round where ultravines were bussed in.
> 
> Didnt we already decrease spiders? Spiders are awful at their current state so I cant complain about that one.

There is an opportunity cost with each event. Each event type that occurs blocks possibly more interesting events from happening. There's an upper bound that limits the number of events that can possibly happen per round, and it cannot go any higher unless the round gets extended. Would you prefer something more interesting to happen or appendicitis or heart attacks that get fixed right away? You can have multiples of them per round, and I have seen it happen quite often. Furthermore, events like these are highly disruptive to an individual and can hardly be called interesting or fun for them. They may produce some entertainment for a doctor, but that's about it if they can call it that. Either ways, only a grand total of two people are affected. It's not really much of an event.

As for spacevine, I never see them do anything of note. It's just there which is perfectly fine, but they can happen multiple times per round as well. This is fine, but it's also dull. People don't want anything as crazy as spiders, but they also think the station is too chill at times too.

Secondly, weights are all relative. Lowering the weights of one event increases the weights of others. Removing one event does the same thing. Some of these are adjustments compensating for that or to to increase the weights of other events without having to touch them.

Overall, balancing events is difficult. The only way we can gauge them is by making changes and seeing how it impacts the rhythm of a round.

hugmeorelse:
> Why is Brand Intelligence being removed? I never heard anyone complain about it. I just wish we could see the reason for these charges in general. I understand Spider Infestation, because we actually had a vote on that one.



> > Why is Brand Intelligence being removed? I never heard anyone complain about it. I just wish we could see the reason for these charges in general. I understand Spider Infestation, because we actually had a vote on that one.
> 
> It removes every vending machine onboard, severely reducing access to station resources, and makes a ton of holes in the hull, and then the vending machines tend to kill people on top of it. A successful brand intelligence is usually a guaranteed shuttle call.

Yakumo hit the nail on the head there. You can also be in mid-RP discussion near the vending machine when it suddenly blows up and maims both you and your friends. These were disabled in the oldbase for a reason. Furthermore, I want to start clamping down on items that take away from the atmosphere of the server. We're already looking to remove meat sheets, pizza sheets, and other items that are overly silly. These things will take time, but we need to start somewhere.

PR #2468, opened 2021-01-03
no_dresscode variable but this time it doesn't override important slots

Seris02:
## About The Pull Request

Makes it so the no_dresscode variable is a thing again, but this time improves it so the loadout doesn't override things like ID, PDA, headset and backpack.
Also makes the proc equip_to_appropriate_slot take a slot blacklist variable.

## Why It's Good For The Game

Because it's annoying having to equip the stuff you want equipped every shift start, and also annoying to see grey outfits laying around everywhere.

## Changelog
:cl:
tweak: enables the no_dresscode variable once again but better this time
/:cl:


atakiya:
Would also be appreciated if this could be modified to somehow not do the equip noise on the lobby screen.
As of the previous PR, anyone that would join the round with a loadout overwriting the suit, would play the zipper noise in the lobby.

Azarak:
What's the 'initial' variable for here?

Seris02:
Ah, initial is used because in the equip proc used for equipping jobs, it uses the same variable, which prevents the equipping sound from being played.

code/modules/mob/mob.dm
```diff
@@ -382,9 +383,12 @@
 			ITEM_SLOT_LPOCKET, ITEM_SLOT_RPOCKET,\
 			ITEM_SLOT_DEX_STORAGE\
 		)
-
+	//SKYRAT EDIT CHANGE BEGIN - CUSTOMIZATION
+	if (blacklist)
```
Azarak:
```suggestion
	//SKYRAT EDIT CHANGE BEGIN - CUSTOMIZATION
	/*
	for(var/slot in slot_priority)
		if(equip_to_slot_if_possible(W, slot, FALSE, TRUE, TRUE, FALSE, FALSE)) //qdel_on_fail = FALSE; disable_warning = TRUE; redraw_mob = TRUE;
	*/
	if (blacklist)
```

Azarak: Otherwise I wont know what's the original code

atakiya:
I know I'm late with this question but uh, does this still have the option to *not* do it?
From what I understood the previous PR had that option as a preference, and imo its kinda nice to have options.. 👉👈

Seris02:
No it doesn't, as that would really be unreasonable, there's a grey jumpsuit and black shoes in the loadout, which is all this currently overwrites. Or you could just not pick a suit and shoes. Weh.
Also the PR you're speaking of was closed by me because Azarak told me that it was bad and gave me reasons which I agreed with.

atakiya:
> 
> 
> No it doesn't, as that would really be unreasonable, there's a grey jumpsuit and black shoes in the loadout, which is all this currently overwrites. Or you could just not pick a suit and shoes. Weh.

Ah so it does not override job-specific loadouts like engineering jumpsuits or similar - Only assistant?

Seris02:
Yep, only assistant currently, because no_dresscode isn't enabled for the other jobs.

Gandalf2k15 (approved):
Attempt number 2. All engine are go.

PR #2479, opened 2021-01-03
Improve mold heartbeat sound & tiny cleanup

atakiya:
## About The Pull Request
Adds two new heartbeat sounds, a 3-second loop and a single beat, also denoised them and removed reverb as the mold heartbeat already uses reverb with the included audio DSP system.

As addition, I also went ahead and fixed the master files readme a bit (mostly changing backslashes to forward slashes for cross-OS compatiblity) and removed an unused soundfile.

## Why It's Good For The Game
The heartbeat sound now wont keep playing for up to 51 seconds while maybe having already moved away, or changed direction from where it could come from, in short, making the sound much more responsive.

## Changelog
:cl: Avunia Takiya
tweak: Changed mold cores' heartbeat sound to use a 3-second loop instead of 51, denoised it
soundadd: Added two new heartbeat loops, heart_beat_once.ogg and heart_beat_loop3.ogg
sounddel: Removed unused horror_scream.ogg under modular master files
/:cl:


Low3:
Finaly, having this loud, droning sound after for an entire minute just for barely glancing mold is so annoying for ghost only gang.

Azarak (approved):

PR #2421, opened 2021-01-01
Automatically flag Severe+ intensity ambitions for admin review

NotRanged:
## About The Pull Request

Submitted ambitions of Severe intensity or higher are automatically flagged for review and notify admins now.
Editing your ambitions 'un-submits' them now too, so admins will be notified again if high intensity ambitions are changed and submitted.

## Why It's Good For The Game
Yes

## Changelog
:cl:
admin: Submitting traitor ambitions of intensity 'Severe' or higher automatically flags them for admin review.
/:cl:

PR #2619, opened 2021-01-10
Fixes Duplicate TC from Ambitions Module 

MissFox0810:
## About The Pull Request
Fixes a bug with the Ambitions module.
If a Antag as submitted and approved ambitions. then they drop there PDA, Then change objectives. give_antag is called again giving them another pda and bunch of TC.

Added check to that function to check if they have already been given the antag gear, and if so. just do nothing.
I think this works, but Needs checking by Azarak

## Changelog
:cl: Miss Fox
fix: Fixes to Ambitions module causing duplicate TC
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/datum_traitor.dm
+++ b/code/modules/antagonists/traitor/datum_traitor.dm
@@ -261,9 +261,9 @@
 	killer.add_malf_picker()
 
 /datum/antagonist/traitor/proc/equip(silent = FALSE)
-	if(traitor_kind == TRAITOR_HUMAN && !/var/has_been_antag_geared)
+	if(traitor_kind == TRAITOR_HUMAN && !has_been_antag_geared)
 		owner.equip_traitor(employer, silent, src)
-		/var/has_been_antag_geared = TRUE
+		has_been_antag_geared = TRUE
 
 //TODO Collate
 /datum/antagonist/traitor/roundend_report()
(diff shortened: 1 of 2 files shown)


code/modules/antagonists/traitor/datum_traitor.dm
```diff
@@ -259,8 +261,9 @@
 	killer.add_malf_picker()
 
 /datum/antagonist/traitor/proc/equip(silent = FALSE)
-	if(traitor_kind == TRAITOR_HUMAN)
+	if(traitor_kind == TRAITOR_HUMAN && !/var/has_been_antag_geared)
```
NotRanged: should just be !has_been_antag_geared

code/modules/antagonists/traitor/datum_traitor.dm
```diff
@@ -259,8 +261,9 @@
...
 		owner.equip_traitor(employer, silent, src)
+		/var/has_been_antag_geared = TRUE
```
NotRanged: same here

Azarak:
ambitions_add() shouldn't ever be called twice by a single antag datum

Azarak:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/2421
is at fault as ambitions are designed to only be submitted once

MissFox0810:
~~It does. We tested
Go antag. Submit your ambitions, (You get your pda), approve them.. Then drop your pda. and change your ambitions. you get another pda~~

MissFox0810:
Ooooh

PR #2661, opened 2021-01-12
Defibrillates Split-Personality

SpaceVampire:
## About The Pull Request

Returns Split-Personality to the list of possible traumas.

## Why It's Good For The Game

I agree that it wasn't very fun on the old base, but its been updated here so that you can actually talk with your other half, which I think could lead a lot of good RP. I have seen people who wished they could have it before for that reason, but at the moment there isn't any way at all to spawn it in the code. We already removed the random brain trauma event, so there is isn't a random chance you are going to get it during ERP or anything, given that the list of stuff which can cause it is already pretty low. I just think that its cool enough to bring back, at least in some form. Given that this is the most simple way to do it, I am starting here. I really just want to know what other people have to think about it returning as well.

## Changelog
:cl:
add: Re-added split-personality
/:cl:

### DIFF
--- a/code/datums/brain_damage/split_personality.dm
+++ b/code/datums/brain_damage/split_personality.dm
@@ -1,5 +1,3 @@
-//SKYRAT EDIT REMOVAL BEGIN
-/*
 #define OWNER 0
 #define STRANGER 1
 
@@ -237,5 +235,3 @@
 
 #undef OWNER
 #undef STRANGER
-*/
-// SKYRAT EDIT REMOVAL END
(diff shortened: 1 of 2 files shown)


hugmeorelse:
Is there any possible way we can have it be like PAIs where one can pair up with a friend or even a stranger voluntarily? It sounds like it could be a fun gimmick that way. They can either ask ghosts to be their split personality, select a friend, or maybe choose from a list of potential candidates?

I'll be happy to merge this if we can do it like that. 

Also, have you ever played on a codebase with NIFs? It's possible to setup a framework so we could also use it for NIFs as well.

SpaceVampire:
@hugmeorelse 
I think that is a really cool idea, but I am not competent enough to code it. 
This currently just returns the code to a state where it can be used for some stuff.
Given that I knew some people who wanted to RP having a split.
As I thought with the updates it had been given, it was sad that it couldn't be accessed at all.

Gandalf2k15:
Howdiddlydo, this will need to be held to a vote! Let's get one on the go.

Gandalf2k15:
Vote completed, we will be merging this.

PR #2674, opened 2021-01-13
Fixes a couple game breaking bugs to do with the slimeperson improvements PR

Seris02:
## About The Pull Request

One bug was because I got lax about testing the last two commits and the other was because I can't test everything all the time, ack.

Fixes transparent bodyparts on non-slimes and slightly transparent bodyparts on non-slimes after switching species from a slime to a non-slime.

Also this should probably be speedmerged.

## Why It's Good For The Game

Bug fixing.

## Changelog
:cl:
fix: fixes transparent and slightly transparent bodyparts on non-slimes.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species.dm
@@ -345,9 +345,6 @@ GLOBAL_LIST_EMPTY(customizable_races)
 		var/obj/item/thing = C.get_item_by_slot(slot_id)
 		if(thing && (!thing.species_exception || !is_type_in_list(src,thing.species_exception)))
 			C.dropItemToGround(thing)
-	for (var/B in C.bodyparts)
-		var/obj/item/bodypart/BP = B
-		BP.alpha = specific_alpha
 
 	if(C.hud_used)
 		C.hud_used.update_locked_slots()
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/customization/modules/mob/living/carbon/human/species.dm
```diff
@@ -345,6 +345,10 @@ GLOBAL_LIST_EMPTY(customizable_races)
 		var/obj/item/thing = C.get_item_by_slot(slot_id)
 		if(thing && (!thing.species_exception || !is_type_in_list(src,thing.species_exception)))
 			C.dropItemToGround(thing)
+	for (var/B in C.bodyparts)
+		var/obj/item/bodypart/BP = B
+		BP.alpha = specific_alpha
```
Azarak: Is this nessecary?

Azarak: Specific alpha isnt handled already in update_limb()?

Seris02:
For the others who want to know why it's still there in on_species_gain instead of in update_limb like Az asked, it's because it won't work in update_limb for some reason. No fuckin clue why. It really should, but it doesn't, and it does when it's in on_species_gain.

Azarak (approved):

PR #2681, opened 2021-01-13
[MANUAL MIRROR] Use portable Python for map merge hooks, other tools / Add .dmm merge driver

atakiya:
Original PR: https://github.com/tgstation/tgstation/pull/55658
--------------------
Mappers and spriters on Windows who wish to use our tools will no longer have to manually install Python or deal with pygit2 binary compatibility issues. A bootstrap script will download and unpack a private copy of a pinned version of Python, as well as update PyPI dependencies when needed.

The overall usage and functionality of the Map Merger is not changed.

Functional changes:

- Editing a hook no longer requires re-running `Install.bat`, only adding or removing a hook or merge driver.
- HitboxExpander now has `Hitbox Expander.bat`, onto which `.dmi` files can be drag-and-dropped.
- The path updater has its own folder with `Update Paths.bat`, onto which `.txt` files can be drag-and-dropped.
- The icon merger has its own folder with `Resolve Icon Conflicts.bat`. If the merge driver is not installed, or was not run by a GUI, this script can be run to resolve icon conflicts on an in-progress merge.
- The tgui bundle merge driver is removed in anticipation of #55373

Implementation details:

- All `.bat` files which launch Python scripts now use the portable Python.
- The hook and merge driver installation script is now implemented in Python.
- The `dmi.py` tests are now in their own file.
- HitboxExpander's `requirements.txt` has been folded into the main `requirements.txt`.
- `requirements-install.bat` is removed as this is now handled by the bootstrap script.
- The `tools/` directory is now added to the Python module path instead of `tools/mapmerge2/`.

Linux users:

- There is no portable Python distribution for Linux, so a system Python is required. PyPI packages are installed to a virtualenv.
- As before, run `tools/hooks/install.sh` to install hooks.
- `tools/hooks/python.sh` has been replaced with `tools/bootstrap/python`.

--------------------
Original PR: https://github.com/tgstation/tgstation/pull/55699
--------------------
This is kind of a prototype. It only fully handles a few situations, and doesn't produce particularly easy-to-read conflict markers when it fails. I hope that it can be useful at least some of the time, can be improved over time, and that the lessons learned can influence a future interactive GUI conflict resolver (integrated into StrongDMM?). In the worst case, one can fall back to the tried and true "manually re-do one side's changes" strategy.

Automatic use:

- In `tools/hooks/`, run `Install.bat`
- Everything is automatic from now on

Manual use, for Git GUIs that don't run merge drivers:

- Before committing, in `tools/mapmerge2/`, double-click `Run Before Committing.bat`
- While a merge is in progress, in `tools/mapmerge2/`, double-click `Resolve Map Conflicts.bat`

This PR also removes the error-prone "Prepare Maps.bat" / "mapmerge.bat" workflow, which has no way to recover from forgetting to use "Prepare Maps" at exactly the right time. First-time contributors who opened a PR without map merging can be instructed to run `I Forgot To Map Merge.bat`, which will create a new commit properly converting the maps.


Azarak:
Im not too good with any external tools, if this is good feel free to merge.
Thanks for mirroring the stuff that the bot missed!

Azarak (approved):

PR #2702, opened 2021-01-15
Teleporter now has safety features, unless emagged.

BurgerLUA:
## About The Pull Request

Using a teleporter will shut off the teleporter and clear the teleporter location, on top of uncalibrating it as a safety precaution.
Emagging a teleporter hub will revert it to this previous behavior.
Delimb chance is now 100% when you enter it while emagged and uncalibrated.

## Why It's Good For The Game

On the Skyrat-Citadel code, stepping into an uncalibrated teleporter would no longer uncalibrate it. The captain was free to set it to some location, like the Hotel, and wouldn't require any user input. This was acceptable.

On the current rebase, it now uncalibrates after every time you use it. On top of that, if you enter a teleporter while it's uncalibrated, it has a chance to delimb you unless upgraded. These two features combined is dumb as hell and punishes players who are used to the previous rebase as well as new players who are unfamiliar with how a teleporter works. It bandaid fixes powergamers using it to swarm a location, which barely happens on Skyrat. The change I added is significantly better and makes more sense realisticly.

Implementation in this PR is a sane middle ground that prevents swarms of people teleporting in at once while removing the RNG punishment for doing so, while giving antags the option to punish via an emag charge.

## Changelog
:cl: BurgerBB
add: Teleporter Hubs now shut off and clear their target after each use as a safety measure. Teleporter Hubs can now be emagged to overload safety features, reverting to previous teleporter Hub behaviors.
tweak: Teleporter uncalibration delimb chance is now always 100%.
/:cl:

### DIFF
--- a/code/game/machinery/teleporter.dm
+++ b/code/game/machinery/teleporter.dm
@@ -74,7 +74,7 @@
 	if (ismovable(M))
 		if(do_teleport(M, com.target, channel = TELEPORT_CHANNEL_BLUESPACE))
 			use_power(5000)
-			if(!calibrated && prob(30 - ((accuracy) * 10))) //oh dear a problem
+			if(!calibrated) //oh dear a problem SKYRAT CHANGE: 100% chance to delimb if uncalibrated.
 				if(ishuman(M))//don't remove people from the round randomly you jerks
 					var/mob/living/carbon/human/human = M
 					/* - SKRYAT EDIT CHANGE ORIGINAL
@@ -104,7 +104,27 @@
 					human.apply_effect((rand(480 + rad_mod - accuracy * 40, 880 + rad_mod - accuracy * 60)), EFFECT_IRRADIATE, 0)
 					//SKYRAT EDIT CHANGE END
 			calibrated = FALSE
+			//SKYRAT CHANGE START, remove target, unless emagged.
+			if(!(obj_flags & EMAGGED)) //If we're emagged, don't uncalibrate.
+				if(power_station)
+					power_station.engaged = FALSE
+					if(power_station.teleporter_console)
+						power_station.teleporter_console.target = null
+						power_station.teleporter_console.update_icon()
+					power_station.update_icon()
+				update_icon()
+			//END OF SKYRAT CHANGE
+
 	return
+
+//SKYRAT CHANGE, emagged teleporter.
+/obj/machinery/teleport/hub/emag_act(mob/user)
+	if(obj_flags & EMAGGED)
+		return
+	obj_flags |= EMAGGED
+	playsound(src, "sparks", 75, TRUE, SILENCED_SOUND_EXTRARANGE)
+	to_chat(user, "<span class='notice'>You use the cryptographic sequencer on [src] to short the calibration safeties!</span>")
+//Skyrat change end.
 
 /obj/machinery/teleport/hub/update_icon_state()
 	if(panel_open)


CameronWoof:
Won't this make it impossible to tell where the last person went with the teleporter? That seems like good information to have access to.

BurgerLUA:
> Won't this make it impossible to tell where the last person went with the teleporter? That seems like good information to have access to.

Yes it would. The antag players said that they don't want people following them when they use a teleporter so it just resets.

CameronWoof:
> > Won't this make it impossible to tell where the last person went with the teleporter? That seems like good information to have access to.
> 
> Yes it would. The antag players said that they don't want people following them when they use a teleporter so it just resets.

Oh, is that what antagonist players said? They don't like being followed when they use a teleporter?

Okay.

Weblure:
I think this is a good solution until we figure out how the community wants the teleporter to function overall. As for antags, I think having it shut off after entering is already a decent advantage to them, but I'm not entirely against it resetting the location as well.

One question though, will this affect hand teleporters in any way?

BurgerLUA:
> I think this is a good solution until we figure out how the community wants the teleporter to function overall. As for antags, I think having it shut off after entering is already a decent advantage to them, but I'm not entirely against it resetting the location as well.
> 
> One question though, will this affect hand teleporters in any way?

It shouldn't no. This is for teleporter hubs only.

hugmeorelse:
We can at least test this and see how it goes, but I'm not sure if I'm a fan of antags being able to disappear off of the face of known existence like that by entering a gateway. It seems a bit silly that they wouldn't have like a stored database of where people teleported off to, but I guess a lot of things are silly. However, for balancing purposes, security should have a way to maybe figure it out. Maybe, it could be tied to detective?

BurgerLUA:
> We can at least test this and see how it goes, but I'm not sure if I'm a fan of antags being able to disappear off of the face of known existence like that by entering a gateway. It seems a bit silly that they wouldn't have like a stored database of where people teleported off to, but I guess a lot of things are silly. However, for balancing purposes, security should have a way to maybe figure it out. Maybe, it could be tied to detective?

It's easy to patch out. I just added it because when I brought it up in main chat, people complained that it was too hard to run away from security via the teleporter.

hugmeorelse:
> > We can at least test this and see how it goes, but I'm not sure if I'm a fan of antags being able to disappear off of the face of known existence like that by entering a gateway. It seems a bit silly that they wouldn't have like a stored database of where people teleported off to, but I guess a lot of things are silly. However, for balancing purposes, security should have a way to maybe figure it out. Maybe, it could be tied to detective?
> 
> It's easy to patch out. I just added it because when I brought it up in main chat, people complained that it was too hard to run away from security via the teleporter.

I think we could reach a compromise on that and seek a middle ground. I can see the reasoning from both sides on this, and I think it requires some changes instead of going outright one way or another.

SpookyTheFox:
One thing comes to mind, the hand teleporters. Also the dragnet. They follow where the location of where main station one is set, and unsetting them delists that option in the hand tele, thus breaking that functionality if someone else uses the main teleporter. On top of that, yakety-sax antags are terrible, and I don't support unsetting the teleporter after every jump for that reason all on it's own. Remember that even wizard teleport names it's destination aloud. Having the teleporter name it's last destination is critical to discovering that they are working with cybersun or cheezing the AI or whatever the case may be.

atakiya:
> On the current rebase, it now uncalibrates after every time you use it. 

Only partly true, if upgraded with better parts, it can teleport more without uncalibrating.
T4 parts actually make it always calibrated after inital calibration iirc.

Edit: https://github.com/Skyrat-SS13/Skyrat-tg/blob/35ef1e2dd52b186e8408ba0fbf27f90d209839b0/code/game/machinery/teleporter.dm#L28-L32
https://github.com/Skyrat-SS13/Skyrat-tg/blob/35ef1e2dd52b186e8408ba0fbf27f90d209839b0/code/game/machinery/teleporter.dm#L77
`accuracy` would be `4` with a T4 matterbin.
`30 - (4 * 10) = -10 == False`

Zenitheevee:
> 
> 
> > > Won't this make it impossible to tell where the last person went with the teleporter? That seems like good information to have access to.
> > 
> > 
> > Yes it would. The antag players said that they don't want people following them when they use a teleporter so it just resets.
> 
> Oh, is that what antagonist players said? They don't like being followed when they use a teleporter?
> 
> Okay.

To be honest i dont see why this **wouldn't** already have been put in with the system, Both to prevent people piling into a teleporter to raid cybersun, And to.. Well, Actually make the teleporter a more useful tool to use for antags?

Like, At *Best* you'll have to use the crew monitor to make sure they're still on station/check where they are
And at Worst, You can probably lever someone to code in a teleport report that pops up on the computer everytime its used.

Gandalf2k15:
As far as I'm concerned the way we have implemented is the best way to handle teleporters. The teleporter is not a tool that should be abused. Doing this removes the nerf of it, and enables antags to have a chance if they get to it and use it, so the entire security team cant just rush em. It takes time, planning. That's the whole point. The flypeople thing was removed because it was basically round removing, and this implemented to ensure it was still balanced. The chances of you losing a limb are very low, and generally, if people learn to calibrate the teleporter then they'll be fine.

PR #2727, opened 2021-01-15
[MIRROR] Infinite Gun Culling: Gun Kit Edition

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56133
--------------------

## About The Pull Request

Every prototype firearm in the game now utilizes crafting to construct it, usually sacrificing a standard version of that firearm in order to produce the new one. The guns are made using gun part kits you print from the security protolathe (or buy, in the case of hellfire lasers).

**The weapons that utilize Energy Guns in their crafting include:**
- Advanced Energy Guns (duh)
- Temperature Guns
- Particle Acceleration Rifles

**The weapons that utilize Laser Guns in their crafting include:**
- X-ray Laser Guns (duh again)
- Hellfire Laser Guns (this is a given)
- Decloner
- Ion Carbine

**Some weapons are outliers and use different parts**
- Particle Acceleration Rifles use a flux anomaly and a gravity anomaly as part of their construction.
- Tesla Revolvers use flux anomalies as part of their construction and need no other gun. You build the gun around the anomaly!
- Decloners require 40 units of mutagen (their old protolathe recipe) and 30 units of baldium (ayy lmao).
- Energy Crossbows require a Proto-Kinetic Accelerator and 15 units of radium to construct. 

In addition:

Hellfire Laser Guns now come as gun part kits and are significantly cheaper as goodies.

Energy Crossbows (the big ones) are bulky. They are called large for a reason.

Every weapon that was printed from the lathe now actually spawns with a firing pin when constructed.

## Why It's Good For The Game

@ Timberpoes I've somewhat disappointed you in NOT making them use anomalies but...

I didn't think it entirely fair to lock x-rays behind specifically the limited station supply of anomalies, but I did think they should be more limited. Now, they're limited by station supply and cargo credits. All the more reason for security to contribute to the station funds.

**Real Talk**
Somewhere along the lines we decided that printing out Saber SMGs from the science protolathe was a bad idea, but yet here we are with a ton of some quite powerful weapons freely printable from our current lathes. Then we moved our guns to the security protolathe and the results were pretty much roughly equivalent in my opinion, a gun printer is a gun printer whether it's in science or security, and a bunch of easily produced guns is easily produced guns for the whole crew. If we want to limit weapon access,  we can't have these weapons stay in the state they're in. Not if we want cargo to represent an important part in gun distribution.

On the other side of this coin, I am fairly sure I heard people calling for the removal of beam rifles all the way back when I started playing the game in mid 2018. They rendered blob a cakewalk, and you had plenty of firing pins to just print off half a dozen and decimate any blob threat on the station regardless of where they are or how well they're doing. This change doesn't stop them from being good against blobs, but it should help justify giving beam rifles a broader niche than JUST blob killing, since I'm of the opinion they were deliberately handicapped explicitly because of their mass producibility. It is out of scope for this PR, but I hope this will give some justification for making the beam rifle a weapon worth using in a separate PR.

## Changelog
:cl:
add: Guns are no longer directly printable from the protolathe.
add: Gun kits to craft the various prototype firearms once printable from the security protolathe.
balance: Large energy crossbows now are bulky, like most other printed weapons.
balance: Crafted weapons now come with a firing pin.
/:cl:


dawsonkeyes:
seems like a good change except for some rather strange things (flux anomaly for a tesla revolver wat?) but eeeh

Sir-Lagsalot:
Honestly? Not the worst. Worth a test, at least. TGcode is already super gun-heavy, changing things up could be good. I don't really think requiring anomaly cores for things is ever a great idea, though.

Hemlock32:
> Honestly? Not the worst. Worth a test, at least. TGcode is already super gun-heavy, changing things up could be good. I don't really think requiring anomaly cores for things is ever a great idea, though.

TG code is the least gun-filled codebase I've played on...

PR #2830, opened 2021-01-22
Balances BZ and Miasma effects.

BurgerLUA:
## About The Pull Request

BZ now requires at least 0.1 moles instead of 0.01 to trigger BZ effects.
You now gain a druggy effect if you inhale too much BZ (1 mole).
Miasma now requires at least 0.25 moles instead of 0 moles to trigger a chance to trigger a disease per breath.

## Why It's Good For The Game

Similar tweaks existed on old skyrat and were absolutely needed. They're now needed even more considering Ice Box atmos or lavaland atmos has a chance to contain either BZ or miasma.

## Changelog
:cl: BurgerBB
balance: BZ now requires at least 0.1 moles instead of 0.01 to trigger BZ effects.
balance: You now gain a druggy effect if you inhale too much BZ (1 mole).
balance: Miasma now requires at least 0.25 moles instead of 0 moles to trigger a chance to trigger a disease.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/lungs.dm
+++ b/code/modules/surgery/organs/lungs.dm
@@ -286,11 +286,12 @@
 		var/bz_pp = breath.get_breath_partial_pressure(breath_gases[/datum/gas/bz][MOLES])
 		if(bz_pp > BZ_trip_balls_min)
 			H.hallucination += 10
+			H.druggy += 5 //SKYRAT CHANGE.
 			H.reagents.add_reagent(/datum/reagent/bz_metabolites,5)
 			if(prob(33))
 				H.adjustOrganLoss(ORGAN_SLOT_BRAIN, 3, 150)
 
-		else if(bz_pp > 0.01)
+		else if(bz_pp > 0.1) //SKYRAT CHANGE.
 			H.hallucination += 5
 			H.reagents.add_reagent(/datum/reagent/bz_metabolites,1)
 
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
So basically, if there's BZ in the air of the Icemoon, everyone's going to get druggy because some of it will inedvitably end up in Distro? I mean, to be fair, I don't know how much 1 mole of BZ in the air is like in percentage of the total air, but... I'm honestly just wondering what kind of effect it'll have on the game, hopefully a positive one.

BurgerLUA:
> Does this apply to lings as well? A few days ago, I spent much of a round with no chemicals because there was a very slight amount in the air all over the station. Not a good experience, and I hope this PR helps to address stray BZ doing that.

Yes. It applies to lings as well.

BurgerLUA:
> So basically, if there's BZ in the air of the Icemoon, everyone's going to get druggy because some of it will inedvitably end up in Distro? I mean, to be fair, I don't know how much 1 mole of BZ in the air is like in percentage of the total air, but... I'm honestly just wondering what kind of effect it'll have on the game, hopefully a positive one.

No it won't. Druggy effects only occur if there is more than 1 mole of BZ. It's actually a nerf to BZ as it effectively notifies you that you will be receiving brain damage, as breathing in more than 1 mole of BZ has a 33% chance of causing brain damage.

Gandalf2k15:
Please fill out the skyrat edit comments properly.

Gandalf2k15:
Due to the stagnant position of this PR and the fact the minor atmos changes were updated and fixed, I feel like this is no longer required.

PR #2834, opened 2021-01-22
[NON-MODULAR] Adds a crafting recipe for the double emergency oxygen tank

desertmote:
## About The Pull Request

A simple crafting recipe addition. Two engi tanks and sticky tape makes a double engi tank.

## Why It's Good For The Game

The double emergency tank is nice, it gives about 40 minutes of air. To my knowledge it is only attainable through maintenance spawn as of now.

## Changelog
:cl:
add: Adds a crafting recipe for the double emergency oxygen tank.
/:cl:

### DIFF
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -1030,6 +1030,22 @@
 	result = /obj/item/pickaxe/improvised
 	category = CAT_MISC
 
+// Skryat addition start
+/datum/crafting_recipe/doubletank
+	name = "Double emergency oxygen tank"
+	reqs = list(
+		/obj/item/tank/internals/emergency_oxygen/engi = 2,
+		/obj/item/stack/sticky_tape = 1,
+	)
+	result = /obj/item/tank/internals/emergency_oxygen/double/empty
+	category = CAT_MISC
+// Skyrat addition end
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
2 air is two times the good il allow it

code/datums/components/crafting/recipes.dm
```diff
@@ -1030,6 +1030,17 @@
 	result = /obj/item/pickaxe/improvised
 	category = CAT_MISC
 
+// Skryat addition start
```
Gandalf2k15, line 1033: Perhaps move this to a modular folder, no need for it to be here

desertmote: will look into this

Low3:
What about makeshift double emergency tank (non engie one). I think it will be a nice addition. Also adding crafting recepies is not great because bloat. Can you make it so when you have duct tape in offhand and click one tank with another it consumes few stacks of tape, deletes tanks and puts in hand double one. Also what about tanks contents? Do they get merged or overwritten? For such cases adding check that they both are empty would be nice. And ofcource result will be empty doubletank.

Gandalf2k15 (approved):
Honestly seems good to me.

PR #2870, opened 2021-01-25
Powersinks can no longer be purchased via Traitor Uplink

BurgerLUA:
## About The Pull Request

Removes powersinks from the uplink.

## Why It's Good For The Game

Powersinks are one of those items that just completely halt a round and are never used to further a goal or an objective that is actually meaningful. When they are purchased, they are consistently placed in meta locations or their own custom locations, usually in locations that are absolutely absurd and require EVA equipment not typically issued to crew. Even with a population of 80, there are only a few people willing and able to actually locate a powersink and even fewer able to actually search in locations it's usually in.

With the extremely low price of 10TC (Traitors are given 30TC in total!), it absolutely does too much to damage a round and arguably does more damage to the round of the 18TC Syndicate Bomb, which actually has a highly likelihood of being discovered due to the ticking sound effect. and affects a smaller area of the station. The seriously low price can actually allow a total of 3 purchases per traitor, assuming that it isn't on sale, which can unironically be enough to halt the round for a significant period of time.

I think if an antagonist wishes to use this item, they should contact admins instead and do a TC trade like some people do when they ask for special antag types like wizard, or ask for special items. In moderation, the power sink is fine but I think people have abused this uplink item way too much to remain an actual uplink item. The rebase was a chance to prove everyone wrong when it was added back in rotation but I don't think anyone was proved wrong.

You can see the PR in oldbase here: https://github.com/Skyrat-SS13/Skyrat13/pull/1777

## Changelog
:cl: BurgerBB
del:  Removes power sinks from the uplink.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1510,6 +1510,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/sbeacondrop
 	cost = 10
 
+/* SKYRAT CHANGE, REMOVAL
 /datum/uplink_item/device_tools/powersink
 	name = "Power Sink"
 	desc = "When screwed to wiring attached to a power grid and activated, this large device lights up and places excessive \
@@ -1517,6 +1518,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			traditional bags and boxes. Caution: Will explode if the powernet contains sufficient amounts of energy."
 	item = /obj/item/powersink
 	cost = 10
+/*
 
 /datum/uplink_item/device_tools/rad_laser
 	name = "Radioactive Microlaser"
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1518,7 +1518,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			traditional bags and boxes. Caution: Will explode if the powernet contains sufficient amounts of energy."
 	item = /obj/item/powersink
 	cost = 10
-/*
+*/
 
 /datum/uplink_item/device_tools/rad_laser
 	name = "Radioactive Microlaser"


MLGTASTICa:
We have staff for  a  reason , to enforce RP standards, if someone abuses powersinks to just RP-lessly saboutage you shoud ahelp, also powersinks are fairly easy to find for a competent engineer. The powersink can also be very valuable for  some antagonist gimmicks,and  forcing them to TC  trade is just going to have admins refuse them.
We  had the same issue on oldbase with antags wordlessly using syndicate bombs, and staff action adressed it  ,so i don't see why  you would go for full removal.
Worthy to mention the oldbase PR removed them because the lighting subsystem broke and not because of gameplay issues.

BurgerLUA:
I forgot to mention that the existence of powersinks is worse in the rebase because of the "minecraft" wiring addition, as well as the fact that most of the wiring on stations is circular, so you can't just isolate half of the station.

Arctaisia:
I agree with this, no one's shown the responsibility in using it, it could be used for so much more, like cutting the wires to security, setting one up on security's grid, and using a crowbar for access in when it all happens, but people just use it to grief. We're an RP server, and although some people won't act like it, we still are at least trying to be. These would be much more useful if departments had their own substations on servers like Chompstation, but they don't, so they're not as useful, especially with redstone-style wires.

TL:DR, If people wanna abuse their toys, they should expect them to be taken away.

SpookyTheFox:
The powersink should have a feature that causes it to explode after taking in a certain amount of energy. That would normally time limit them, even if they are not found. Rather than jumping to removal, I'd like to try reducing the amount of charge it absorbs before blowing. TG maps and wiring is pretty harsh right now though.

Zenitheevee:
I mean
It is **Nice** to have an option to essentially depower the station to try and do something...
But
If no ones using it properly, It.. Kinda is a problem.
And while there *Are* Alternative methods to depowering the station... A lot of them are instantly negated if engineers upgrade the SMES's or check power-flow. 

But, To stand by MLGTastic's side, We have staff for a reason. And while it will probably immediately turn into >Ahelp to check if its getting utilized properly
Its less annoying than essentially hard-gating people away from using it by forcing it behind a TC Trade, Especially considering there are a LOT of rounds where Staff may not be totally active to do through with the trade.

Gandalf2k15:
We need to stop removing things like this, if we continue this way, we'll have nothing to actually use. So we need this to be enforced by admins more so players change their mindset.

BurgerLUA:
> We need to stop removing things like this, if we continue this way, we'll have nothing to actually use. So we need this to be enforced by admins more so players change their mindset.

There is currently nothing against the rules about how you place powersinks or how many you can buy. Consistently people have placed powersinks and other round disrupting items since the rebase and as far as I know, no administrative action has ever been taken based on how powergamy the placement is. I've been told by staff it's an IC issue when ahelped or talked about after the round.

And this won't lead to a slippery slope of removals. The only other item that is next to the level of disruption is the EMP bomb, but at least that has some sort of counter that doesn't take 20-30 minutes.

KathrinBailey:
Closing in favour of #2897.
If the nerf turns out to be a soft removal, I'll either reopen this or close the nerf.

PR #2897, opened 2021-01-26
Nerfs Powersinks

BurgerLUA:
image

Gottem.

But no this PR actually adjusts the power sink drain rate to be the same as a supercharged supermatter on an average shift (300,000) and will detonate in 20 seconds if it can leech that power at a consistent rate.

## Changelog
:cl: BurgerBB
balance: Reduced the drain of powersinks to be roughly an SM output. Powersinks should explode now. Powersinks are now 16 TC.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -18,9 +18,9 @@
 	throw_speed = 1
 	throw_range = 2
 	custom_materials = list(/datum/material/iron=750)
-	var/drain_rate = 2000000	// amount of power to drain per tick
+	var/drain_rate = 300000	// amount of power to drain per tick //SKYRAT CHANGE
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 6e8		// maximum power that can be drained before exploding
+	var/max_power = 3e7		// maximum power that can be drained before exploding //SKYRAT CHANGE
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = FALSE // stop spam, only warn the admins once that we are about to boom
 
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1516,7 +1516,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			load on the grid, causing a station-wide blackout. The sink is large and cannot be stored in most \
 			traditional bags and boxes. Caution: Will explode if the powernet contains sufficient amounts of energy."
 	item = /obj/item/powersink
-	cost = 10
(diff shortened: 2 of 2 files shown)


SpookyTheFox:
You're on the right track. 20 is too short though, for a 16TC item. The solars can do 360000 by themselves, and some SMs do multiple megawatts. You might want a minimum active time of like 10 minutes, it should indeed cripple the station by draining all APCs and whatnot for a long enough period of time to incentivize people to go look for and remove it. And a pretty hefty explosion if they fail to find it. It's not hard, it still even makes the sound, right?

Actually, that's something I didn't notice in the hell round with two power sinks. I knew where they were because I was a sneaky ling with x-ray vision, but I didn't hear the buzzing sound they made on the old base.

SpaceVampire:
I'm not sure why you would nerf them into the ground AND buff the TC cost within the same balance patch. I feel like you should really leave the the TC the same for the moment, if you are already adjusting it to be far worse. Given you know, the whole point is that they should last a while, so this is hardly going to do anything as the power will just refill itself the moment that its gone.

MLGTASTICa:
Why not just make the powersink more obvious ? make it do tesla zaps ? i donț think nerfing it to the ground is a solution.
Also making them suck the power of a low-yield SM setup makes them useless.

LocalizedDownpour:
I think one or the other is far more prudent. Being a Drain to the station, but only having 16 would mean you would hive to tie it around your gimmick for the round and make it the center piece of your purchase unless it was discounted. Or leave it at 10TC but making the max it can drain lower...meaning it would explode earlier, so if someone DOES buy three, they would never be able to ruin the round for a long amount of time.

Nerfing it into the floor like this pretty much ensures it will never be bought.

KathrinBailey:
I'll have a look and see how this performs in-game and how it compares to a syndicate bomb.
I honestly don't feel like this is at all necessary, but I'll check and see for myself.

YakumoChen:
Isn't the point of a power sink to.... depower the station? You don't buy it to have a bomb, there's already bombs in the uplink that do that function.

Vizoplez:
Firstly. 20s is way too short for the explosion it creates. You can currently buy a roundstart powersink beacon, rush HoP office maint, plant it, 20s later it explodes with a force larger than our maxcap, 4,8,16,32; allowing you to get AA near instantly.

Secondly. The explosion on these is far too large, bypassing our maxcap. Just... why?
This would make it a lower cost syndicate HE bomb for essentially half the price and a fourth of the original timer, and with no beeping warning.

Gandalf2k15:
This actually makes powersinks more dangerous in the long run, let's not buff things. Powersinks are generally speaking something that gets used when someone has a legitimate reason to do so. We don't need to nerf more shit for antags, especially with the security update.

PR #2917, opened 2021-01-27
[ASSOPS] Removes The Orbital Mech Pad From DS-1

unit0016:
## About The Pull Request
Why was this even in robotics to begin with. Round 606 I tested it out and was able to send a DS-1 death ripley to the station. Probably works the other way around too, and no, I don't want to find out for myself.

## Why It's Good For The Game
as funny as a ghostrole being able to send mail-order ripleys to cargo and robotics is, this is obviously unintentional and could be exploited to hell and back.

## Changelog
Omitted, again


Gandalf2k15:
Rome wasn't built in a day, but you have laid a brick.

PR #2942, opened 2021-01-29
[READY] Adds a pool to Icebox, but MULTI-Z!

unit0016:
## About The Pull Request
Kind of a silly PR to demonstrate multi-z mapping and re-add the ~~in~~famous BoxStation pool. It's. HUGE. I had to move some stuff around for this to fit in properly, but it feels complete to me.
image

## Why It's Good For The Game
big pool make splish splash. Also good to relax in if that's your thing, plus a bunch of personal lockers and a feasible way to exit hydroponics from the lower floor.

## Changelog
:cl:
add: BoxStation has it's pool back! Splish.
/:cl:


Gandalf2k15:
Remember pool controllers and such? Or are those not in?

unit0016:
yeah, they're not in. All that's in right now is the turf itself.

Gandalf2k15:
Seems good to me, we might need to add more detail to the pool, but otherwise it's fine.

PR #2952, opened 2021-01-30
The roundend screen will now not show ckeys.

uomo91:
## About The Pull Request

This makes it so that roundend screen only shows IC names.

## Why It's Good For The Game

Metacliques are bad and people trying to start new statics should feel free to relatively anonymously be able to do so.
I considered trying to make ckey display based on a user config option but it seemed a bit too onerous for the scope of this PR.

Full disclosure is that while I've tried this with many different antags and silicon configurations and stuff, that I'm not sure it covers every possibility. This will definitely need a testmerge.

### DIFF
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -458,7 +458,10 @@
 	for (var/i in GLOB.ai_list)
 		var/mob/living/silicon/ai/aiPlayer = i
 		if(aiPlayer.mind)
-			parts += "<b>[aiPlayer.name]</b> (Played by: <b>[aiPlayer.mind.key]</b>)'s laws [aiPlayer.stat != DEAD ? "at the end of the round" : "when it was <span class='redtext'>deactivated</span>"] were:"
+			//SKYRAT EDIT CHANGE BEGIN - ckey display on roundend screen
+			//parts += "<b>[aiPlayer.name]</b> (Played by: <b>[aiPlayer.mind.key]</b>)'s laws [aiPlayer.stat != DEAD ? "at the end of the round" : "when it was <span class='redtext'>deactivated</span>"] were:" - SKYRAT EDIT - ORIGINAL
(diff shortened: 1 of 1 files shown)


SpaceVampire:
We had a feature like this on the old-base, as a config: https://github.com/Skyrat-SS13/Skyrat13/pull/2198
I feel like that would be a better implementation, if you could port it.

uomo91:
> We had a feature like this on the old-base, as a config: [Skyrat-SS13/Skyrat13#2198](https://github.com/Skyrat-SS13/Skyrat13/pull/2198)
> I feel like that would be a better implementation, if you could port it.

Thanks, I'll look into it.

KathrinBailey (approved):
If you test it locally and just show a quick screenshot just in case anything hecked happens I'll merge. 👀 
You can turn your ghost into a borg, ghost from that borg, turn your ghost into a human and antag it, end the round and see if it's working as intended.

uomo91:
> If you test it locally and just show a quick screenshot just in case anything hecked happens I'll merge. 👀
> You can turn your ghost into a borg, ghost from that borg, turn your ghost into a human and antag it, end the round and see if it's working as intended.

Untitled

I reversed the order of how I did things, because if you drag a ghost into a borg and ghost again you seem to get stuck with the borg name. Verified that this doesn't have anything to do with this PR though, happens the same on current master, just with the addition of the ckey being shown.

SpaceVampire:
@KathrinBailey 
But wouldn't the way we did it before be a better way of handling it?
I would rather have it as a feature you could turn on if you wanted it, instead of something blanket applied.

hugmeorelse:
Isn't this already handled based on preferences though? You can disable this if you don't want your CKEY to show. 

**Edit:** Was it removed? Can't seem to find it in preferences anymore. It may have been on Citadel. In that case, it may be a good idea to port that over.

PR #2954, opened 2021-01-30
[SEC OVERHAUL] Airlocks / ID Card / HUD Tweaks

unit0016:
## About The Pull Request
Something always seemed.. unsettling about the colors used. They felt like they blended together to make this murky green when they were just blue and grey next to each other, something I didn't notice until I went in to fix this.
This PR changes that, making the colors on airlocks, ID cards, and the HUD more vibrant so they pop out and don't look like swamp water (Though I'd personally prefer if security IDs were black, too).
It also reorganizes card.dmi so all the icons slot into their modular designated sections, and makes the headsec's hud icon in line with the other heads. Also gives the blueshield the CORRECT sec colors.
Screenshot:
image

## Why It's Good For The Game
Purdy sprite good.

## Changelog
Omitted


CameronWoof:
We need *less* different shades of blue, not *more*. The uniforms and lockers already don't match the airlocks and floors. Please don't make the situation worse.

PR #2961, opened 2021-01-30
[TM Only] Fixed security slots.

Useroth:
Quick patch for a bug which causes the configs to work incorrectly, because it says "set to 0 to disable" but the config entry forces it to the minimal value of 1.


Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/2995

PR #3002, opened 2021-02-01
Fix Shapeshifter module node

SarmentiCampbell:
## About The Pull Request
The Cyborg Shapeshifter was apparently hidden, as the node won't show up even after the illegal tech was unlocked

## Why It's Good For The Game
Really?

## Changelog
:cl:
fix: fixed the cyborg shapeshifting node.
/:cl:


xXPawnStarrXx:
Yes! It's a fun module and It's not damaging at all.

PR #3020, opened 2021-02-02
makes the pubby monastery ten times the price

Elli-Skala:
## About The Pull Request

makes the pubby monastery shuttle ten times the price (to 500000) 
## Why It's Good For The Game

the pubby shuttle is a non emag shuttle that does more damage the meteor shuttle, keeping it in the nature of the first pr it can still be bought without a emag but you need to work had making a unfunny amount of cash to get it
## Changelog
:cl:
balance: rebalanced the price of the monastery shuttle 
/:cl:

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -232,7 +232,7 @@
 	name = "Grand Corporate Monastery"
 	description = "Originally built for a public station, this grand edifice to religion, due to budget cuts, is now available as an escape shuttle for the right... donation. Due to its large size and callous owners, this shuttle may cause collateral damage."
 	admin_notes = "WARNING: This shuttle WILL destroy a fourth of the station, likely picking up a lot of objects with it."
-	credit_cost = CARGO_CRATE_VALUE * 250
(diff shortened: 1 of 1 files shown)


dawsonkeyes:
this is 5 times that of the hyperfractal gigashuttle, a far more reasonably expensive shuttle
i think simply making it an emag shuttle would be better

SpaceVampire:
Looking at the PR for it, it seems like they already plan to put it behind an emag soon anyway.
I agree that if we were going to make some change to it, just locking it behind one if the better way to handle it.
Not that I honestly think it is an issue in the first place.

Cheshify:
I mean, it causes more damage than the meteor, but it's also safer, larger, and more comfortable. I feel like the suggested idea of locking behind an emag or some other prerequisite is safer, and it'll ACTUALLY get seen, because 500,000 credits probably only happens when an atmos god makes zauker.

YakumoChen:
cargo ideally shouldn't be making even half that on a typical round.

Gandalf2k15:
Admins have the power to override this remember, and it shouldn't really be seen too much. I don't really see it being a problem, it's only been ordered once. TG will probably update it anyway at some point.

PR #3029, opened 2021-02-02
[MIRROR] Cleans up blob code, removes a lot of hardcoded stuff, many new defines

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56306
--------------------

## About The Pull Request
Cleaned up, commented on, and (hopefully) improves a lot of blob code.
Put pretty much everything balance-related (except for the strains themselves) into a define file, for easier viewing, changing, and balancing.
Added a bunch of new functionality that new strains (some of which I plan to add in the future) could use: from increased expansion range to more spores per factory.

~~Small balance changes contained within:
**Relocate core: 30 point cost (was 80)**
Relocating the core is only something you would ever really do when you're seriously fucked. You're /never/ in a situation where you can afford 80 points to relocate your core, as opposed to doing literally anything else. 30 is a more reasonable number for this, and in most situations you're likely better off doing something else with 30 points still.
**Blob core can sustain 1 spore**
Something I thought would be neat: a single spore to keep the lonely core company during the early times.~~

## Why It's Good For The Game
Code improvement, commenting improvements, easier to balance, more strain stuff to work with in the future.

## Changelog
:cl:
refactor: Changed a lot of blob code backend around, making it easier to mess with balance and work with in the future.
/:cl:


NotRanged:
@KathrinBailey

PR #3031, opened 2021-02-02
Chameleon projectors now have a tell

CameronWoof:
## About The Pull Request

Currently, chameleon projectors are a gameplay and roleplay sandbag that can be freely deployed without risk or consequence. Because of our metaknowledge rules, someone who is using a chameleon projector cannot be interacted with, because you are required to SEE them activate it (something entirely within their control) in order to start testing for it. With this update, you can see that an item is a chameleon hologram by examining it closely--an in-character action you can take when you know someone came into a room and that they must still be around somewhere.

## Why It's Good For The Game

Metaknowledge policy should not be abused to pull yourself out of the playing field and effectively end any ongoing gameplay (such as a chase with security) under penalty of admin action. The projector is still a powerful item, but it now has *any* counterplay at all that doesn't run afoul of policy.

## Changelog
:cl: Cameron!
balance: Chameleon users can now be revealed by close examination (examine twice).
/:cl:


Gandalf2k15:
I'm not sure this is good for gameplay as it basically ruins the item. If you really wanna give it a tell, maybe use the examine system.

PR #3037, opened 2021-02-02
Xenomorph Balancing V: The Station Strikes Back

KathrinBailey:
## Preface
**This PR is in response to the modular security rework, which sees them using primarily ballistics over laser projectiles. Because security can't deal 40 damage in one hit, health is lowered.**
**This is also our modular version of https://github.com/tgstation/tgstation/pull/55937, but less severe. We might be waiting a bit for TG's rework, so here's... this?**

## About The Pull Request
• Adds documentation to my long quest of finding out why xenomorphs take double burn damage.
• Adds additional documentation to 50% 2x burn 1.5x burn damage fireloss fire xeno xenomorph beno <- these are search tags so you don't suffer like me.
• Xenomorphs now take 1.5x burn damage, down from 2x.
• Xenomorph health reverted to original TG values, now that security aren't carrying lasers.
• Xenomorph Sentinel sneak is now slightly less blindingly obvious.
•**Reworks Neurotoxin Spit!**
•**Reworks Xenomorph Tackle!**

## Reworked Neurotoxin Spit
• Deals 50 stamina damage instead of 5 toxin damage.
• You still spend 10 seconds stunned, however, only 5 seconds are hardstun (paralysis) and the last 5 a knockdown.
In terms of power level, it's a little lower. You do take a hefty amount of stamina damage, however, you're only completely vulnerable to half of the time than before. This gives you a chance to crawl towards a weapon.

## Reworked Tackle/Disarm Intent
Ten second hardstun? No more! You used to do deal 100 deciseconds (10s) of paralysis whenever you tackled a mob with nothing in their hands, else you'd disarm one item from their hand(s).
Now you do three things:
• Add one second of paralysis (hardstun);
• Add five seconds of knockdown;
• Add your attack damage (random number between two lower/upper bounds), but as stamina damage. This checks for armor, meaning if you're in a riot suit, you'll be tough to stamcrit.

## Why It's Good For The Game
What's the design philosophy for this heavy nerf?
It's frustrating to play against, and ideally, xenomorphs should be powerful against single opponents, but not many. Unfortunately, you could click on one person three times whilst dodging, then they're out of the fight for ten seconds, plus the time required to pick up their gun. You can then move onto a second opponent. Tag the third to reset the 10s timer, then go for the third. You can potentially get three or four people permanently locked into hardstuns. This is too powerful.
So now, instead, an attacking xenomorph bashes you **hard** and knocks you done, stunning you for a second. You're on the ground for five seconds, and take stamina damage based on how powerful the xenomorph normally attacks. If a xenomorph doesn't chain-attack you, then you have a chance to crawl away and grab your weapon to keep attacking them. If the xenomorph decides to give you _special_ attention, then they can pretty much beat you into exhaustion.

## Sentinel Sneak/Invisibility
image

## Changelog
:cl:
balance: Xenomorphs are weaker, now that the station are running around with ballistics.
balance: To stop laser weapons completely decimating the xenomorphs like before, they only take 1.5x burn damage instead of 2x.
balance: Xenomorph tackles are no longer ten second hardstuns. They now deal 1s hardstun, 5s knockdown, and their melee damage is applied as stamina damage.
balance: Xenomorph sentinels are now extra sneaky! They're less blatantly obvious outside their nest, but still easy to spot.
/:cl:


Gandalf2k15:
Please update from master :3

KathrinBailey:
I'm not interested in spending what little free time I get fixing this given the host wants roll chance for xenos to be very low.
Not wasting hours on something that'll only happen in-game maybe once every three or four days.

PR #3043, opened 2021-02-03
Removes Clowns and ERP

NullDagaf:
hahaha you got bamboozled it actually removes the clown ERT and the ERP ERT

i don't need to say why it's good

### DIFF
--- a/code/datums/ert.dm
+++ b/code/datums/ert.dm
@@ -72,6 +72,8 @@
 	mission = "Assist in conflict resolution."
 	polldesc = "an unpaid internship opportunity with Nanotrasen"
 
+//SKYRAT EDIT REMOVAL START
+/*
 /datum/ert/erp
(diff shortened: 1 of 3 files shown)


SpookyTheFox:
I rather like those, and they are adminbuse only to begin with. They should stay.

Draganfrukts:
Cringe PR, I should not explain why!

NullDagaf:
> I rather like those, and they are adminbuse only to begin with. They should stay.

Yes, they're exclusive to adminbussing. No, I don't trust admins not to abuse them.

Gandalf2k15:
Eh, no, we're not removing admin bus tools, no reason to, if staff can't be trusted with the tools, then we deal with them. Never blame your tools.

NullDagaf:
> if staff can't be trusted with the tools, then we deal with them

I'll hang that quote on a wall.

PR #3060, opened 2021-02-04
[MIRROR] Kilostation update: HFR, Public Mining, other fixes

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56493
--------------------

## About The Pull Request

This addresses a number of features missing from Kilostation. 

Kilostation has a very compact Atmos, with no nearby large maintenance areas to take over for construction. Like Meta and Delta, it needs an area to build the HFR in. The incinerator has been upgraded with an expanded construction area and the components of an HFR. This is across from the gulag shuttle, which docks just to the north.  The area available is slightly larger than on Meta, but much less than Delta.

KiloStation-1-HFR

Then there's public mining. So it turns out Kilo **does** have a public mining dock - it was set inside aux base construction, but the docking port wasn't set to generate a shuttle. If it loaded a shuttle here, then working on the aux base would require sending it to lavaland first, and getting squished if someone sends it back. This area is behind either cargo, or engineering access, which defeats the purpose of a public shuttle. So Kilo is getting a new one. I've rerouted a bit of maint just above the holodeck and below Medbay. This area is actually accessible to anyone.

KiloStation-1-PubMining

This fixes a number of minor piping issues, including reroutes required for the new additions. It also adds Pubby-style trash chutes to tool storage and the morgue, because this was apparently a rather popular thing.

## Why It's Good For The Game
Kilo gets space to build an HFR and mining access is restored to the general public.

## Changelog
:cl:
add: A Public mining shuttle dock has been installed near the holodeck on Kilostation
add: Kilostation's incinerator has been expanded to accommodate the HFR and provide Atmospherics with an additional work area
add: Adds a trash chute to tool storage and the morgue on Kilostation
del: Removed the nonfunctional public mining dock from Kilostation aux base construction
fix: A number of small piping issues on Kilostation have been fixed
/:cl:


Gandalf2k15:
Requires manual override!

PR #3063, opened 2021-02-04
[MIRROR] Replaces intents with combat mode

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56601
--------------------
## About The Pull Request

This PR removes intents and replaces them with a combat mode. An explanation of what this means can be found below

### Major changes:
- Disarm and Grab intents have been removed.
- Harm/Help is now combat mode, toggled by F or 4 by default
- The context/verb/popup menu now only works when you do shift+right-click
- Right click is now disarm, both in and out of combat mode.
- Grabbing is now on ctrl-click.
- If you're in combat mode, and are currently grabbing/pulling someone, and ctrl-click somewhere else, it will not release the grab (To prevent misclicks)

### Minor interaction changes:
Right click to dissasemble tables, racks, filing cabinets (When holding the right tool to do so)
Left click to stunbaton, right click to harmbaton
Right click to tip cows
Right click to malpractice surgery
Right click to hold people at gunpoint (if youre holding a gun)

## Why It's Good For The Game

Intents heavily cripple both the code and the UI design of interactions. While I understand that a lot of people will dislike this PR as they are used to intents, they are one of our weakest links in terms of explaining to players how to do specific things, and require a lot more keypresses to do compared to this.

As an example, martial arts can now be done without having to juggle 1 2 3 and 4 to switch intents quickly.

As some of you who saw the first combat mode PR, the context menu used to be disabled in combat mode. In this version it is instead on shift-right click ensuring that you can always use it in the same way.

In this version, combat mode also no longer prevents you from attacking with items when you would so before, as this was something that was commonly complained about.

The full intention of this shift in control scheme is that right click will become "secondary interaction" for items, which prevents some of the awkward juggling we have now with item modes etcetera.

## Changelog
:cl: Qustinnus
add: Intents have been replaced with a combat mode. For more info find the PR here: https://github.com/tgstation/tgstation/pull/56601
/:cl:


Mithrun:
i hate this, what if i wanna examine and just shove instead?

MLGTASTICa:
Great PR

RatFromTheJungle:
@Azarak might wanna glance this over before merge, probably going to fuck up the combat rework hard, namely bootleg martial arts and glancing blows

BrotherHangyul:
Please please please don't merge this

Gandalf2k15:
Updated to work with our stuff. Test merge first.

Cheshify:
/tg/ and subsequently Skyrat HAS/HAD a really good game-feel. I feel like this would just fuck with that, and make it harder for new players as well.

hugmeorelse:
I'm not so sure about this PR. Put this one on hold for now, until we get a gameplan. 

Unfortunately, this will have to be merged at some point. We'll find a way to remedy this or find ways to help players adapt.

Gandalf2k15:
UPDATE: Now prevents any damage being done outside of combat mode with items that have a force.

Erokusan:
The hotkeys are already complicated enough as is, I'd really prefer not to make every interaction a hotkey. This only speeds up one interaction (Which is combat) and drops every other interaction (Seems a lot of things that isn't combat) gets slowed down, mostly because people will do them on accident.
Another thing I noticed, this doesn't change the combat mode we have currently in the game, which means there would be **TWO** combat modes, that would not work explaining it to literally anyone. That's just extra steps of explaining that I personally as a staff rather not deal with

YakumoChen:
current combat mode is just a stripped version of what citadel used. You'd be better just changing this to use the indicator.

Gandalf2k15:
Update: USES INDICATOR.

Arctaisia:
I'd really prefer intents myself, this is meant to be an RP server, afterall, not an unga one. Sure, combat reworks are nice and all, but intents gave so many more options for mechanics, this just streamlines combat. Perhaps, we could keep these mechanics, but only for combat mode, and have intents used when it isn't active, as to promote the use of CI, streamline combat when it happens, and allow for all other mechanics outside of combat. It might be lots of code, but, it would be nice. Though, ofcourse, there's still the problem of compatability with upstream merges, but perhaps we'd just need a bit more coding to change it to work for intents/a mix.

TL:DR, I personally don't like the removal of intents on an RP server in place of combat mechanics, but we may be able to keep both.

And another note, if we keep pushing for more combat mechanics, constantly, while not paying attention to the RP related mechanics, we may start to lose our better RPers, it's already happened slightly with the rebase, and we should be careful that it doesn't happen more.

YakumoChen:
I could probably point out TGcode is "arcade styled" and "combat motivated". That's just sort of the price to pay for using their codebase and actively maintaining off of their changes. Changes will continue down this route because that's what the upstream wants, and unless we split away from the codebase, that's what people will just have to deal with.

Gandalf2k15:
Entering combat mode with a gun in your hand will now toggle its safety off/on. - This was ALEKS IDEA
CI is now a bind again.

Arctaisia:
> I could probably point out TGcode is "arcade styled" and "combat motivated". That's just sort of the price to pay for using their codebase and actively maintaining off of their changes. Changes will continue down this route because that's what the upstream wants, and unless we split away from the codebase, that's what people will just have to deal with.

That's indeed true, my biggest hope is that this will at least go down alright with mechanics that used intents, like surgery, you could use disarm to fail one purposefully, then again- That was less of an RP feature and more an annoyance for doctors.


> Entering combat mode with a gun in your hand will now toggle its safety off/on. - This was ALEKS IDEA
> CI is now a bind again.

Could we also get it so that you can't fire guns on help intent, like bay has? I know it's more security for guns, we already have safety, but it'd help with going to hug someone and not shooting them, pff.

dawsonkeyes:
I wanna bring up that this is likely to cause conflicts with object-specific bound interactions, like flipping a table or other similar actions, so any good coders might want to take a look at that beforehand.

Ironnhawk:
I simply need to ask from a game development standpoint. Is Disarm truly important enough to warrant the Right Click spot. Outside of combat it is seemingly unnecessary and defeats the general standard of utilizing right-click as a context menu as standard in strategy games. I would at least like to see consideration into it being rebound into something more useful for all-around gameplay as having played for over 1k hours, I rarely have a need to disarm.

Useroth:
Ultimately, if we wish to remain a direct downstream, we cannot simply ignore a 300 file PR.
The longer we delay the more conflicting PRs we'll have pile up.

I'm not particularly happy about this change either, considering how used everyone is to the traditional intents, but we'll have to either adjust the code or our habits. 
For now, I need to merge this to maintain code parity with our upstream.

TheEaterOfBeans:
revert reeeeeeeeee

DrPainis:
WHY DO I SEE A "VIEW DETAILS" BUTTON AND NOT REVERT!? REEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE

PR #3072, opened 2021-02-04
Headcrab/Headslug can be spawned via gold core

ghost:
## About The Pull Request
So the headslug has been removed from the gold core spawn pool _(Feb 2019 as i remember)_ due to "Xenobio is too OP", but I think it was only because of two or three grief-ish people on TG, so...can we return _it_ back?

## Why It's Good For The Game

> Headslug can create some unique RP situations about some *flesh-y* bio-weapon and etc.

> Non-grief-ish xenobiologist can use it properly, i think.

I think that in Skyrat community, **normal** xenobiologist can use this ability to create some unique situations, also add some capabilities to both NT/Syndie xenobiologist, and yes, memes about bio-weapon.

## Changelog
:cl:
code: Headcrab/Headslug can be spawned via gold core.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -23,6 +23,7 @@
 	ventcrawler = VENTCRAWLER_ALWAYS
 	var/datum/mind/origin
 	var/egg_lain = 0
+	gold_core_spawnable = HOSTILE_SPAWN
 
 /mob/living/simple_animal/hostile/headcrab/proc/Infect(mob/living/carbon/victim)
 	var/obj/item/organ/body_egg/changeling_egg/egg = new(victim)
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -23,7 +23,7 @@
 	ventcrawler = VENTCRAWLER_ALWAYS
 	var/datum/mind/origin
 	var/egg_lain = 0
-	gold_core_spawnable = HOSTILE_SPAWN
+	gold_core_spawnable = HOSTILE_SPAWN //SKYRAT EDIT ADDITION
 
 /mob/living/simple_animal/hostile/headcrab/proc/Infect(mob/living/carbon/victim)
 	var/obj/item/organ/body_egg/changeling_egg/egg = new(victim)


RatFromTheJungle:
is that the one that lets you become a changeling if you sentient to it? if so, I would advise a hardpass on this due to "haha I'm a changeling i don't have to be scared of x, x, and x antag :)"

SarmentiCampbell:
> is that the one that lets you become a changeling if you sentient to it? if so, I would advise a hardpass on this due to "haha I'm a changeling i don't have to be scared of x, x, and x antag :)"

It is indeed the one that turn you into a changeling, though If i remember, It seem to have a type and it is trackable as a admin, So i'm sure that if peoples don't abuse it and actually make sure to ask the RD before doing such a dangerous thin- oh who am i kiddin', Peoples are gonna abuse it so hard that it would not be funny, and i have a feeling this will just end up making a lot of complaints with like peoples crying that they got cremated because security saw that they were a changeling or something like that.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
```diff
@@ -23,6 +23,7 @@
 	ventcrawler = VENTCRAWLER_ALWAYS
 	var/datum/mind/origin
 	var/egg_lain = 0
+	gold_core_spawnable = HOSTILE_SPAWN
```
SarmentiCampbell: Forgot to add the skyrat edit

MLGTASTICa:
This will just lead to powergame , so why not ? security is in full right to kill / cremate them after all anyway.

dawsonkeyes:
Enabling of powergaming is a big bad, and non-changelings that become slugs immediately die, so you can’t use it to powergame how you likely imagined. The same also goes for sentience potions.

This PR exists to GIVE XENOBIOLOGISTS ACCESS TO FREE CHANGELING POWERS and yet fails to take note of the additional blockers in place to prevent it.

code/modules/mob/living/simple_animal/hostile/headcrab.dm
Gandalf2k15:
```suggestion
	gold_core_spawnable = HOSTILE_SPAWN //SKYRAT EDIT ADDITION
```

Gandalf2k15 (changes requested):

CameronWoof:
but why give traitor the ability to become traitorling

Elli-Skala:
> but why give traitor the ability to become traitorling

they won't, this didnt remove the other barriers stopping exactly this

PR #3073, opened 2021-02-04
[Non-modular] Buffs blob.

NotRanged:
## About The Pull Request

Buffs blob to (hopefully) be more balanced around our population count. Testmerge only for a while is advised.

Main buffs
- Starting points 80 (was 60) 
Allows for a slight bit of an easier start on a station where there is generally no such thing as 'a quiet spot'
- Expand cost 3 (was 4)
You'll want to spend points building, not manually expanding most of the time. As a result of this, attack cost is now 1 (was 2).
- Automatic expansion to reflectors around the core
- Reflectors now always reflect (but still take reduced damage from reflections)
- Reflectors now reflect 'hard' projectiles (though most bullets ignore this anyway...)
Changes made in order to prevent cheesing the core with 5 emitters and a mirror, killing the blob instantly with nothing to do against it.
- Core and nodes get some spore buddies growing out of them. Increased damage to 8-16 (was 2-4).
Gives the core and nodes some natural protection, spore buff to make them more viable over just getting a blobbernaut instead.
- Distance between resource collectors to 3 (was 4)
Practically doesn't change anything balance wise since you can still only have 4 per node, but it's less punishing on the building for newer players who don't know the optimal powergame layouts.
- Reinforced blob cost to 10 (was 15), reflector to 5 (was 15)
Makes them less prohibitively fucking expensive and actually a viable choice
- Relocate core cost 25 (was 80
In a situation where you need to relocate core, you /never/ have 80 points, and even 25 is a big investment at that point. Might actually see use now.

Nerfs:
- Blob win condition to 700 (was 400)
More time for the crew to respond, organize, put up a fight, call for ERT... Because ending the round immediately is stupid.
- At 350 tiles, an automated distress call is sent.
This helps in letting anyone onboard the station know just how truly fucked things are, and gives admins (more) reason to send (larger/deadlier) ERTs.
- Increased additive resource node time scaling to 0.75s (was 0.25s)
Every additional resource node past the first increases gather time by this. Less lategame points.

## **Minimum living player count increased to 60 (was 20)**
No lowpop blobs pls. This checks for living non-afk players (ghosts excluded).

## Why It's Good For The Game

Makes blob less of a pushover on med-highpop. It only ever stood a chance if it went AI sat (bug allowing that has been mostly fixed) or went unnoticed for like 10 minutes before being found.

## Changelog
:cl:
balance: Blob has been buffed. View PR (Skyrat-tg/pull/3073) for details.
/:cl:


CameronWoof:
Blob doesn't need to be stronger. It already wins sometimes and that's more than enough for an event that just turns the game off. Ideally, blob winrate hovers somewhere around 20%, not 50%.

NotRanged:
> Blob doesn't need to be stronger. It already wins sometimes and that's more than enough for an event that just turns the game off. Ideally, blob winrate hovers somewhere around 20%, not 50%.

Except that the winrate doesn't hover around 20% (especially now that the human EMP bug is fixed), and blob only ever really wins when it spawns in the AI sat or goes unnoticed for a long time. As soon as it's detected, 12 sec officers and 20 assistants with welders descend upon it like a horde of bees.

I also don't want blobs to 'win' (hence the increase in victory cap), I want them to put up more of a fight, right now they just roll over and die once spotted.

If any of the buffs in this PR are too much, feel free to comment on that.

MLGTASTICa:
Good changes , sounds like blob will be challanging to defeat. Might be interesting if you made the spores actually apply reagents with actual effects however, and gave the blob types other weaknesses.
Also instead of buffs , what if you  made it scale with the amount of players ? you added defines which can make that easily with some multipliers

YakumoChen:
I would be careful with this. Security is is big and powerful now but also deeply ammo dependent. They can put up a fight but once they start relying on their protolathe instead of rechargers to keep the fight going I have doubts they'll perform well in a long and protracted fight.
They are pretty limited in laser weapons now and I can see cargo having a hard time just ordering 20 cases now as well.

SpookyTheFox:
Blob is in a good spot for balance right now, and has a non-trivial winrate, including a bunch of recent victories even after the EMP fix. These changes would severely unbalance it late game, where it does not need any help at all.

In particular, attack cost was a major problem in previous blob buff attempts, and needs to stay high. Reflectives also should be _more_, not less expensive.


I need to ammend my previous statement, after yet another blob win no one could stop. blob is much much too strong right now. It is extremely difficult to fight, and it's built in penalties for spacing the station essentially have no effect. It is generally not able to be fought by the station at large with the atmos problems we have had as of late, I think a better solution would be to disable blob until it can be properly reworked in light of atmos problems and weapon balance changes.

KathrinBailey:
I'm gonna ask that everyone reads the changes **fully** before passing judgement.
A notable change is shifting the blob to **high amounts of living population** only to have a chance of rolling. This will not roll during lowpop. This will allow the blob to be balanced around something consistent. Having 50 players instead of the usual 70+ makes a big difference.

This PR will be **TESTMERGED** first. There will be plenty of opportunities to see how it plays it, but let's remember a few things first:
• The crew rarely utilises the correct tools to fight the blob. Emitters are often just ignored entirely.
• Borgs and IPCs still run head-first into EMP blobs.
• Blob, before this PR, has only one viable tactic. Blobbernaut spam. Everything else is very out of meta.

We'll probably go in the direction of nerfing blobbernauts slightly as we bring up previously never-used tactics to a point where they're actually viable. For those worried that it's going to steamroll, it needs almost twice the points to win. That's plenty of time to send an ERT and for the balance to be adjusted for the next blob roll. x_x

CameronWoof:
I think the increased point requirement is mostly placebo. Blob wins long before the game says it wins. Now it will just take longer for the game to admit that it's over.

NotRanged:
Blobbernaut: health and attack damage reduced by 25% (200 to 150, 20u to 15u), cost reduced (40rsc to 25, 60 to 50 for factory)

edit:
Only health reduced, cost back to 40, factories get 1 extra spore

Gandalf2k15:
After looking this over I think it's suitable for a test merge after the next update!

LimblessLayabout:
Cost reductions are good, but at the end of the day, every blob-patch we make, I bring up the issue that blobs just have no way to generate resources or otherwise _act_ once under siege, leading to a death spiral where you're best served logging out or ghosting.
It's miserable gameplay.

If your resource nodes get targeted, you don't have any other input to the game. Your existence at the controls ceases to have meaning. 
You **have** to grant some active means of obtaining resources _during a fight_ or the blob is just gonna keep being a misery machine. Damage based enrage. Eating corpses. Special target objects. Fucking **anything**. Let the blob be a participant or just NPC this thing already and give up.

Gandalf2k15:
Test merge period completed. Observations show an increase in blob effectivity.

PR #3158, opened 2021-02-07
More special examines: facehugger, ash flora

YakumoChen:
## About The Pull Request
* Adds some various examines for facehuggers and their variants. Scientists get a very clear warning about what the real deal is. Aliens get funny examine text if they happen to look at Lamarr or a toy.
* Added special examines for lavaland plants, as a small bonus for any botanists who change jobs and can't remember what each one does.
<s>* Captain and Bartender both get an extra-obvious message when examining a beer nuke, that also hints at the beer-splosion</s> removed because you can't do this with machines yet

## Why It's Good For The Game
A proof of concept. Special examines have more uses beyond just hiding antag items. They can be used to give useful or insightful information for a relevant role.

## Changelog
Very minor changes so not really needed.

### DIFF
--- a/code/modules/mob/living/carbon/alien/special/facehugger.dm
+++ b/code/modules/mob/living/carbon/alien/special/facehugger.dm
@@ -22,9 +22,9 @@
 	layer = MOB_LAYER
 	max_integrity = 100
 	item_flags = XENOMORPH_HOLDABLE
-	special_desc_requirement = EXAMINE_CHECK_JOB
-	special_desc_jobs = list("Scientist, RD")
-	special_desc = "This alien is an extremely dangerous life form capable of creating a xenomorph. You would know well not to approach without full body biological protection."
+	special_desc_requirement = EXAMINE_CHECK_JOB //SKYRAT EDIT
+	special_desc_jobs = list("Scientist, RD") //SKYRAT EDIT
+	special_desc = "This alien is an extremely dangerous life form capable of creating a xenomorph. You would know well not to approach without full body biological protection." //SKYRAT EDIT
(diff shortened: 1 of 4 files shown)


GoldenAlpharex:
Changelog is always needed, even for minor changes, otherwise people won't know what's up. That's... Kind of the goal of a changelog.

YakumoChen:
Not true. This PR isn't likely to change player behavior at all and is not important enough to warrant a changelog message.
Most players aren't even likely to see the effects of this PR in their playtime.

GoldenAlpharex:
> Not true. This PR isn't likely to change player behavior at all and is not important enough to warrant a changelog message.
> Most players aren't even likely to see the effects of this PR in their playtime.

I mean, for the beer nuke, if you're going to do this but not change the description, then I guess it doesn't matter,you're right.

YakumoChen:
Doesn't matter anyway, because I forgot the beer nuke is a machine and the system doesn't function with it yet.

code/modules/mob/living/carbon/alien/special/facehugger.dm
```diff
@@ -22,6 +22,9 @@
 	layer = MOB_LAYER
 	max_integrity = 100
 	item_flags = XENOMORPH_HOLDABLE
+	special_desc_requirement = EXAMINE_CHECK_JOB //SKYRAT EDIT
+	special_desc_jobs = list("Scientist, RD") //SKYRAT EDIT
```
Gandalf2k15: Make sure to include the entire job name, "Research Director"

YakumoChen: Global defines protect against this but alright

PR #3190, opened 2021-02-07
[TM-ONLY] Disables jukeboxes as orderable cargo crates.

Useroth:
Apparently the echo and sound falloffs got really bonked and the jukebox is very loud across the entire z-level.
Let's disable them while we work it out.


Cheshify:
Actually, I think it's intended and not a bug

Useroth:
previously we had a really nice "echo" system that made the jukebox from afar sound really muffled and be barely audible

Low3:
I swear like a week ago they worked properly.

Gandalf2k15:
Might as well FM this as it's stupidly broken.

PR #3194, opened 2021-02-08
Removes the lrp suicide verbs.

Rohesie:
## Title: Suicide verb removal 

MODULE ID: SUICIDE_VERB 

### Description:

Removes the "comedy"-based LRP suicide verb, in the same manner as most RP servers.

### TG Proc/File Changes:

Modified:
- `code/modules/client/verbs/suicide.dm`

Removed:
- `/mob/living/carbon/human/verb/suicide()`
- `/mob/living/brain/verb/suicide()`
- `/mob/living/silicon/ai/verb/suicide()`
- `/mob/living/silicon/robot/verb/suicide()`
- `/mob/living/silicon/pai/verb/suicide()`
- `/mob/living/carbon/alien/humanoid/verb/suicide()`
- `/mob/living/simple_animal/verb/suicide()`

### Defines:

- N/A

### Master file additions

- N/A

### Included files that are not contained in this module:

- N/A

### Credits:

Me.

### DIFF
--- a/code/modules/client/verbs/suicide.dm
+++ b/code/modules/client/verbs/suicide.dm
@@ -21,6 +21,8 @@
 		if(mmi.brainmob)
 			mmi.brainmob.suiciding = suicide_state
 
+//SKYRAT EDIT REMOVAL BEGIN - SUICIDE_VERB
+/*
 /mob/living/carbon/human/verb/suicide()
 	set hidden = TRUE
 	if(!canSuicide())
@@ -214,6 +216,8 @@
 
 		death(FALSE)
 		ghostize(FALSE)	// Disallows reentering body and disassociates mind
+*/
+//SKYRAT EDIT REMOVAL END
 
 /mob/living/proc/suicide_log()
 	log_message("committed suicide as [src.type]", LOG_ATTACK)
--- a/modular_skyrat/modules/suicide_verb/readme.md
+++ b/modular_skyrat/modules/suicide_verb/readme.md
@@ -1,0 +1,45 @@
+https://github.com/Skyrat-SS13/Skyrat-tg/pull/3194
+
+## Title: Suicide verb removal <!--Title of your addition-->
+
+MODULE ID: SUICIDE_VERB
+
+
+### Description:
+
+Removes the LRP suicide verb, in the same manner as most RP servers.
+
+
+### TG Proc/File Changes:
+
+Modified:
+- `code/modules/client/verbs/suicide.dm`
+
+Removed:
+- `/mob/living/carbon/human/verb/suicide()`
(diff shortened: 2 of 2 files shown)


StarHorder:
i'd rather see it disabled until the round has ended, as that is where the vast majority of its usage is seen. I'm all for not letting people do it in round, though.

code/modules/client/verbs/suicide.dm
```diff
@@ -21,6 +21,8 @@
 		if(mmi.brainmob)
 			mmi.brainmob.suiciding = suicide_state
 
+//SKYRAT EDIT REMOVAL BEGIN - SUICIDE_VERB
+/*
 /mob/living/carbon/human/verb/suicide()
```
Gandalf2k15, line 26: Can we perhaps have some feedback to the user about it being disabled followed by a message to the admins?

NotRanged: I don't see any reason to act on a message of 'x tried to access the suicide verb but couldn't lol' so please no extra adminlog spam

Same with the primary armament token redeeming message on improper alrt, haven't seen anybody act on it yet, because why would we -- they didn't do anything wrong.

SpaceVampire:
This is how it was handled on old-base, and I think its the best way to handle it.
https://github.com/Citadel-Station-13/Citadel-Station-13/pull/10614
I was going to port this myself but I was too lazy.

Rohesie:
I personally see no need to have it enabled at any point, really.
It's better if the game sets the proper tone for what the expectations are, instead of baiting new (and sometimes old) players with the wrong ideas.

code/modules/client/verbs/suicide.dm
Rohesie: NotRanged has a point here, I'd tend to agree with it.
Verdict, @Gandalf2k15 ? Will you want this change done?

SpaceVampire:
I mean, I think having it be EoRG only is fine.
Its no different than anything else that happens during that part of it.
Some of the suicide stuff is funny.
That is how it was on the old-base, and I don't recall any issues coming from it.
I also enjoy having stuff as a config option in general.

Rohesie:
I'll leave it up to the maintainers if they want to have this after the round end.
I personally think that if RP is what they're after this feature can be safely eschewed, and that there's no need to cater for this kind of thing as the round ends.
Also noting that not everything after the round ends is eorg, or at least there's plans for that to be the case.

code/modules/client/verbs/suicide.dm
Gandalf2k15: It's fine as is.

Gandalf2k15:
We don't need to cater for roundend stuff. We'll be fine merging this as is.

PR #3219, opened 2021-02-09
MAJOR: Non-modular map hardset 09.02.2021

Gandalf2k15:
This PR resets all non-modular maps to upstream perfection.

All modular maps will require an update.


KathrinBailey:
*cries*

PR #3226, opened 2021-02-09
Changes beakers to be like oldbase

Swiftfeather:
## About The Pull Request

Beaker set to 60u, large beaker set to 120u, x large beaker set to 150u, bartending shaker set to 120u as well.

## Why It's Good For The Game

On oldbase these were the beaker values, making a lot of mathematical things a lot easier, since 120 and 60 are both divisible by 3, 4, 5, and 6, they can have mixtures fit the beaker sizes a lot better. This adds more versatility to the beakers and makes an allround better experience for those mixing things.

## Changelog
:cl:
tweak: beakers and shakers now hold more.
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/drinks/drinks.dm
+++ b/code/modules/food_and_drinks/drinks/drinks.dm
@@ -569,7 +569,7 @@
 	icon_state = "shaker"
 	custom_materials = list(/datum/material/iron=1500)
 	amount_per_transfer_from_this = 10
-	volume = 100
+	volume = 120 //SKYRAT EDIT: Original value: (100)
 	isGlass = FALSE
 
 /obj/item/reagent_containers/food/drinks/flask
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -130,12 +130,14 @@
 
 /obj/item/reagent_containers/glass/beaker
 	name = "beaker"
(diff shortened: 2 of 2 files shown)


MLGTASTICa:
Maybe buff bluespace beakers from 300u to 360u ? They are quite expensive and if we make the affordable ones bigger so should be the bluespace one.

Swiftfeather:
> Maybe buff bluespace beakers from 300u to 360u ? They are quite expensive and if we make the affordable ones bigger so should be the bluespace one.

Bluespace beakers, even on oldbase, only held 300u. I do not think it needs to be increased as this would allow people to make even bigger bombs.

MLGTASTICa:
> > Maybe buff bluespace beakers from 300u to 360u ? They are quite expensive and if we make the affordable ones bigger so should be the bluespace one.
> 
> Bluespace beakers, even on oldbase, only held 300u. I do not think it needs to be increased as this would allow people to make even bigger bombs.

If you're not aware TG does not allow bluespace beakers to be put in normal grenades , they also nerfed water/pottasium and its extremly weak now.

Arctaisia:
360u means a max bomb capacity of 2520u. 👀

Gandalf2k15:
Seems like a logical movement from my past PR to change chemistry, might as well complete the journey.

Gandalf2k15 (approved):

PR #3266, opened 2021-02-10
Adds Monophobia as a choosable quirk.

NullDagaf:
Adds Monophobia as a choosable quirk. Apparently some people like to not like being alone.

~~please gitkraken work~~


Gandalf2k15:
Seems acceptable, let's see if anyone has any reeing to do.

NullDagaf:
> Seems acceptable, let's see if anyone has any reeing to do.

Guess not.

PR #3273, opened 2021-02-10
[Modular] Yet More DS-1 Tweaks

unit0016:
## About The Pull Request
Inspired by my previous DS-1 Tweak PR and people deriding the place for being only good for ERP and Assault Ops, but even more drastic. Opens up the bar to cargo, reorganizes medbay and gives chemistry beakers, moves atmospherics and power generation, changes power generation to RTGs, Removes the syndiecomms keys, replaced with more DS-1 keys. Warden's records consoles are gone in favor of an AI restorer and syndicate network camera console. Spreads more materials out across the entire station. Renamed the NanoMed to SyndiMed. Added NanoDrug.
I debated making the prisoners functional ghostroles too rather than catatonic monkies, but decided against it. I feel it'd ruin the chill vibe of the place.
image

## Why It's Good For The Game
People don't really like DS-1 in it's current state, with it being treated as a syndicate-themed ghost cafe. While I like that it's disconnected from the main game, it's probably best to give the place things to do for those who.. don't. This partially helps, by adding more materials so more things can be printed from the lathe.

## Changelog
:cl:
tweak: DS-1 Has been opened up somewhat, with new materials inside.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -5553,6 +5553,11 @@
 /obj/item/borg/upgrade/transform/syndicatejack,
 /obj/item/borg/upgrade/transform/syndicatejack,
 /obj/item/borg/upgrade/transform/syndicatejack,
+/obj/item/encryptionkey/headset_assault,
+/obj/item/encryptionkey/headset_assault,
+/obj/item/encryptionkey/headset_assault,
+/obj/item/encryptionkey/headset_assault,
+/obj/item/encryptionkey/headset_assault,
 /turf/open/floor/iron/dark,
 /area/cruiser_dock)
 "Cm" = (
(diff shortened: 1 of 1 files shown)


Grunnyyy:
I think working ghost prisoners is something many would like

Sure, there are disadvantages to it, but at the same time... There can be prisoners with guards interacting without the consequences of a station. 

It'd also give a purpose to the whole thing, because one of the reasons why it's said to be, well, a place to ERP in, is because of being only a chill place. Many have criticized it being just Syndie themed ghost Cafe and with the amount of assops we have had in recent times it really doesn't justify its existence much, to be honest, since it's Lava land Syndies without free access to materials

YakumoChen:
I'd rather have this be Prison Hotel or something than put the Luxury Prison Module back on the station.

Gandalf2k15:
Seems like a good change yep!

PR #3280, opened 2021-02-11
Coronal Mass Ejection Event

Gandalf2k15:
## About The Pull Request

This basically adds a randomly occurring CME event that can happen in 3 variants; random, minimal, moderate and extreme. Each with slightly different stats for the CME.

What the CME actually does is sends an alert and a time to impact so that players can prepare and after that time is up, every 30 or so seconds a small EMP bubble will form somewhere on the station and then burst, sending out a small pulse in the area. 

The waves last anywhere from 3-10 minutes depending on the severity, and you should avoid the big donuts of doom!

There are a few cool sound effects and visual effects included, so hopefully, it's quite intriguing. I was contemplating making it a station-wide EMP but this seems more fun!

image - Extreme flares.

Videos to come!

https://streamable.com/whe2sq

The above video is edited to make testing easier, in reality, bubbles can take 10-30 seconds to burst, and the warning time is way bigger, 5-10 minutes, depending on wave type.

## Changelog
:cl:
add: A new random event: Coronal mass ejections!
add: A toggle for admins to stop the events system from spawning events automatically.
fix: Gamma loop.
/:cl:


GoldenAlpharex:
That sounds cool.

Swiftfeather:
RIP silicons

Gandalf2k15:
Added a random mode!

CameronWoof:
I think the idea is fine, but huge chat-spanning messages popping up every few seconds for 3-10 minutes is going to be very irritating. I don't think the individual pulses need warning messages (especially since it doesn't look like there's anything to do/prepare for if you're not literally in the room), but the event can still have one. 

EMPs of this magnitude are going to require a lot of time and busywork to fix. Not sure how much fun this is really going to be.

Gandalf2k15:
> I think the idea is fine, but huge chat-spanning messages popping up every few seconds for 3-10 minutes is going to be very irritating. I don't think the individual pulses need warning messages (especially since it doesn't look like there's anything to do/prepare for if you're not literally in the room), but the event can still have one.
> 
> EMPs of this magnitude are going to require a lot of time and busywork to fix. Not sure how much fun this is really going to be.

You get a very long time to prepare, 5-10 minutes, you can prevent most important things from being damaged by it, and the low end ones only last about 3 minutes.

CameronWoof:
Your video doesn't illustrate this very well. The event triggers. EMPs go off in certain locations every few seconds for 3-10 minutes? Is that right?

Gandalf2k15:
> Your video doesn't illustrate this very well. The event triggers. EMPs go off in certain locations every few seconds for 3-10 minutes? Is that right?

The video has some wacky var edits for testing, in reality, it's a lot more slow paced.

CameronWoof:
What kind of preparations do you expect people can or will make?

Gandalf2k15:
> What kind of preparations do you expect people can or will make?

If you dismantle a machine, it is completely protected, moving things into a safe area, e.g. mining, and stuff. And any synths can either run around the station avoiding them, or escape to somewhere safe. Honestly I'm curious to see how the crew handle it.

Might also make it so you can surround somewhere in grilles so it's protected.

CameronWoof:
Going to echo my initial reaction which is that this seems like a lot of busywork and not very much *fun*, but I guess we'll see how it goes.

Gandalf2k15:
> Going to echo my initial reaction which is that this seems like a lot of busywork and not very much _fun_, but I guess we'll see how it goes.

Could say this for any engineering event, rods, meteors, etc. But this shouldn't be THAT destructive.

Elli-Skala:
may i leave the sudgestion to try making the bubble last abit longer and able to be anomaly neutralized?

unit0016:
If this isn't able to be prevented like Elli suggested, make sure it can't affect the dorms / AI sat, please. Otherwise pretty solid event.

Gandalf2k15:
Noted changes: make the alarm sound less obnoxious and only occur once, make it not harm dorms. Event occurrence rate needs to be tweaked.

code/controllers/subsystem/events.dm
```diff
@@ -60,7 +62,8 @@ SUBSYSTEM_DEF(events)
 	set waitfor = FALSE //for the admin prompt
 	if(!CONFIG_GET(flag/allow_random_events))
 		return
-
+	if(manual_mode)
+		return
```
Gandalf2k15, line 67:
```suggestion
//SKYRAT EDIT ADDITION
	if(manual_mode)
		return
//SKYRAT EDIT END
```

Gandalf2k15:
Seems good enough for a full merge now.

PR #3319, opened 2021-02-12
re removes changeling head slugs from gold slime pool 

Elli-Skala:
## About The Pull Request

re removes the changeling head slug from the gold slime pool, it really shouldnt be there

## Why It's Good For The Game

the headslug is supost to be a changeling power to escape via vent crawling as a last resort. being in the gold pool will lead to players without the changeling tag becoming a slug that dies if it bites a dead body due to more tg blockage to stop non changelings from becoming changelings 

## Changelog
:cl:
del: changeling headslug once again removed from the pool
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/headcrab.dm
+++ b/code/modules/mob/living/simple_animal/hostile/headcrab.dm
@@ -22,7 +22,6 @@
 	speak_emote = list("squeaks")
 	var/datum/mind/origin
 	var/egg_lain = 0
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
Why remove it if it punishes people that try to powergame ? No reason to not have in it at all.

SarmentiCampbell:
It do not punish peoples, there is no second barrier and the headslug just turn peoples into antags, which is bad and should be removed

Gandalf2k15:
Yes after being made aware of their true purpose I agree with this.

PR #3362, opened 2021-02-13
Small medbay and chemistry changes

scrubs2009:
## About The Pull Request
Replaces the door to the paramedic dispatch on metastation with a non-see through one for privacy. Adds a medical records computer to icebox station as unlike the other maps there was none that regular medbay staff could access. 
On metastation each chemistry station now has disposals unit like on delta. In the spot where the upper unit used to be I've added a blood pack of O- blood to be used to help restore blood lost from chemists during synthflesh production.
On Iceboxstation I've removed several small internal windows inside of their chemistry lab to allow the chemists access to each of the two chemistry fridges without having to push the other chemist out of the way.

## Why It's Good For The Game
The new medical records computer adds the same RP opportunities provided by one to Icebox station as it has previously been lacking it. The non-see through door was added so paramedics could better relax when waiting for injured to run out and treat.
This also makes having 2 chemists in the department much easier and they'll end up getting in each other's way much less. 

## Changelog
:cl:
add: Added medical records computer to icebox station medbay. Added new disposals unit and blood bag to chemistry on metastation
del: Removed 2 internal windows blocking chem fridge access on icebox.
tweak: Replaced glass paramedic dispatch door on metastation with a non see-through one. 
/:cl:

### DIFF
--- a/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
+++ b/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
@@ -25827,6 +25827,9 @@
 /obj/effect/turf_decal/trimline/blue/filled/line{
 	dir = 4
 	},
+/obj/machinery/computer/crew{
+	dir = 8
+	},
 /turf/open/floor/iron/white,
 /area/medical/medbay/central)
 "bwO" = (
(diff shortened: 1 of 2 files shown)


KathrinBailey (changes requested):
Conflicts. Also, some of these changes (such as med data computers) have been made already.

KathrinBailey:
Stale

PR #3393, opened 2021-02-14
Changeling Rebuff

Gandalf2k15:
## About The Pull Request

Rebuffs changeling to be able to compete with security. This does the following, flat and simple:
Chems: 75
Chems storage: 125
Chem recharge rate: 2
Total chem storage: 125
Genetic points: 15
Total genetic points: 15

Adernals: 
Now completely clear stamina loss.

True changelings have had their delay in transformation removed and had a double health buff. This form is intended to be the final stage, and security SHOULD have to try to kill it. Not just click and it's over.

This is constantly changing, check files for stats.

## Changelog
:cl:
balance: Rebalanced changeling.
/:cl:


CameronWoof:
Reel this way back. These are buffs on the order of double or even triple effectiveness *and* a way to get out of stamcrit 4 or 5 times over in a fight. This is absurd.

MLGTASTICa:
They're a changeling ,  the fact that a security officer could easily win against one in a one on one is more than enough proof they needed buffs desperately.

SpookyTheFox:
A good start. I never recall having a chemical shortage on the old base, but on TG it seems to happen all the time. This gets them up to speed an not so badly limited on using existing abilities, but it may still be necessary to improve their single most important stat in the game, time/clicks to horizontal/kill another space man. Arm blade is still on the gold standard of 20 damage, which compares poorly with the weapons likely to be used against them. Health and regen in horror form is as well balanced around 20 damage red jolly ranchers, with substantially fewer adversaries than the reality for a high population server.

SpaceVampire:
I would buff arm-blade damage as well. Given that I am pretty sure you can get implants that are more powerful then it.
Buffing horror-form also wouldn't be an awful idea, given that the damage output is a lot higher now.
I think everything else they have is mostly fine, but those two powers could really use tweaks to make em more viable.

RatFromTheJungle:
A buff to chameleon to make it more then "in 45s I'll be invisible"
and the armblade going up to 30 with 25% AP would be great additions to their 1-1 combat, while still leaving them weak to secs heehoo baton spam

SpookyTheFox:
> 
> 
> A buff to chameleon to make it more then "in 45s I'll be invisible"
> and the armblade going up to 30 with 25% AP would be great additions to their 1-1 combat, while still leaving them weak to secs heehoo baton spam

Ling chameleon has been on my radar for a while. Outside the scope of this PR, but it deserves to be addressed at some point. On the old base, it at least had no cost to turn on and off. On TG, it's actually worse than the genetics version. It consumes chemicals to toggle it in either direction, and is subject to server tick rate for how slowly it fades you. Considering the traitor options are both instant and let you move around, making it at least quickly cloak you in place would be nice. I was going to take a stab at improving that particular ability, as well as a re-spin of ling chat and ling space adaptation, but I need to get my local test working again first. I'll do it probably within the week if no one else beats me to it.

YakumoChen:
Monsterously overtuned. Absolutely ridiculous. They can horror form, get killed after an amazing amount of fight, and INSTANTLY. INSTANTLY get back up and in horror form as if ahealed.

Gandalf2k15:
That was an unintended side effect. I've now made it so they have to wait 3 minutes(about 5 minutes with TD) to reform. SO that should give officers a decent chance to contain them.

YakumoChen:
3 minutes after they die or 3 minutes after they horror form? Because it pretty consistently took longer to do the former, and nothing stops you from just horror forming off camera and spring the free instant revive the moment you're down.

Gandalf2k15:
> 3 minutes after they die or 3 minutes after they horror form? Because it pretty consistently took longer to do the former, and nothing stops you from just horror forming off camera and spring the free instant revive the moment you're down.

3 minutes AFTER they die.

Gandalf2k15:
Alright, I'm gonna FM this as it seems to have settled down and there aren't really any issues anymore. Balance seems reasonable. We can change it down the line, also TM's bad.

PR #3416, opened 2021-02-15
doubles the wizard starting spell points 

Elli-Skala:
## About The Pull Request

Gives the wizards 20 starting spell points to be in line with other antag's double power

## Why It's Good For The Game

helps wizards some what fight against our buffed security using the same changes as other antags

## Changelog
balance:  wizard starts with 20 spell points 
/:cl:

### DIFF
--- a/code/modules/antagonists/wizard/equipment/spellbook.dm
+++ b/code/modules/antagonists/wizard/equipment/spellbook.dm
@@ -617,7 +617,7 @@
 	throw_speed = 2
 	throw_range = 5
 	w_class = WEIGHT_CLASS_TINY
-	var/uses = 10
+	var/uses = 20
 	var/temp = null
 	var/tab = null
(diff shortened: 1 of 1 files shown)


Swiftfeather:
Amazing, I totally felt that our wizards were very underpowered due to sec buffs, this will even the playing field more and all the future wizard rounds after this comes out will be much more interesting

SpaceVampire:
I unironically support this PR.

SarmentiCampbell:
this will be very helpful, yes

Cheshify:
We have Wizards?

Oscari-pog:
We dont have wizards, but in case they are enabled, or a admin runs a event, they wont get memed on by rubbers anymore.

Gandalf2k15:
We don't have wizards right now, if we ever get them, we can look into buffing them if we need to.

PR #3439, opened 2021-02-16
Memento Mori is now better, and more punishing in ways.

Vizoplez:
## About The Pull Request
Memento Mori is now impossible to remove without beheading the target.

## Why It's Good For The Game
You can currently game memento mori through a decently fun meme. For example, if you equip it, and get clonescanned; something can cause the traits itself to bug out - and transfer to your clone after you remove it (likely because removing it doesnt actually remove the traits).

Also, its dumb that others can just yeet you from the round by grabbing your funny necklace pendant, and I've been stripped as a shaft miner prisoner with it on.
Not gamer.

## Changelog
:cl:
balance: Memento Mori can no longer be removed willingly, or by force, truly taking its place as a cursed item. No more suiciding to get out of brig and into your autocloners, prisoners.
/:cl:

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -240,6 +240,9 @@
 		ADD_TRAIT(user, TRAIT_NODEATH, "memento_mori")
 		ADD_TRAIT(user, TRAIT_NOHARDCRIT, "memento_mori")
 		ADD_TRAIT(user, TRAIT_NOCRITDAMAGE, "memento_mori")
+		//Skyrat Edit. Once you put it on, it wont be coming back off, not without great effort.
+		ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
+		//Skyrat Edit End
 		icon_state = "memento_mori_active"
 		active_owner = user
 


SpaceVampire:
Honestly, this just reminds me that cloning is really stupid, and didn't need to be ported over to this code-base, especially with all the issues our copy of it seems to have.
But I really don't see any issue with people being able to remove it, as that is one of the points of balance behind it.
If you don't like the chance to get dusted, you don't gotta put it on and use it in the first place.

Vizoplez:
> Honestly, this just reminds me that cloning is really stupid, and didn't need to be ported over to this code-base, especially with all the issues our copy of it seems to have.
> But I really don't see any issue with people being able to remove it, as that is one of the points of balance behind it.
> If you don't like the chance to get dusted, you don't gotta put it on and use it in the first place.

Being dusted is still possible; being beheaded /will/ dust you and disable cloning. Its why its a cursed artifact - and not something you should be able to pickup without consequence. Cloning is somewhat good in certain instances (Such as space dragons going on a vore spree for more than 5 minutes, mass decaying organs). It just puts more emphasis on Memento being an item that makes you _harder to kill_; not easier in most circumstances.

Draganfrukts:
> Someone has a memento mori as a prisoner out of all jobs
>

lmaonigga

Vizoplez:
> > Someone has a memento mori as a prisoner out of all jobs
> 
> lmaonigga

this would be equally hilarious and frustrating for them, memento prevents succumb/hardcrit too, good luck in the zone, stalker.

Cheshify:
Just a question, gibbing/dusting via megafauna isn't affected by this?

Gandalf2k15:
I don't think we need this change, it's fine as is.

PR #3446, opened 2021-02-17
[MIRROR] Hovering your mouse over things now shows their name on the top of the screen

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56729
--------------------
feature I saw on other servers that I think is nice, because looking at the bottom left to see the name sucks

:cl: Mothblocks, Qustinnus
add: Hovering your mouse over things now shows their name on the top of the screen
/:cl:

image

image

image

this is a pref, you can disable it.


SpookyTheFox:
I did a quick local test of this, and as I suspected, the screentips do show invisible people such as cloaked lings and the stealth implant box on mouse over. Having them exposed on right click was already dubious, but you can easily accidentally see them now.

The old way, the gray bar at the bottom of the screen still did this, but it was a little less on the nose.

PR #3476, opened 2021-02-18
Nerfs security into the ground

Gandalf2k15:
## About The Pull Request

I put security on their high horse, and I'll rip them right back down.

After playing a few rounds as antagonist it's become very clear to me that I have overbuffed security by a mile. I was putting this off for a bit, but I have to.

All security rounds, aside from 12mm, have had a siginficant debuff.

I've also done away with that nasty hitscan gun.

## Changelog
:cl:
balance: Security weapons are no longer instant fight winners. Also yeeted the hitscan lasergun.
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/bullets.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/bullets.dm
@@ -15,7 +15,7 @@
 
 /obj/projectile/bullet/advanced/b6mm
 	name = "6mm bullet"
-	damage = 15
+	damage = 10
 	speed = 0.5
 
 /obj/item/ammo_casing/b6mm/rubber
@@ -30,7 +30,7 @@
 	name = "6mm rubber bullet"
 	icon_state = "bullet_r"
 	damage = 5
-	stamina = 10
+	stamina = 5
 	ricochets_max = 6
 	ricochet_incidence_leeway = 0
 	ricochet_chance = 130
@@ -66,7 +66,7 @@
 
 /obj/projectile/bullet/advanced/b9mm
 	name = "9mm bullet"
-	damage = 20
+	damage = 15
 	speed = 0.7
 
 /obj/item/ammo_casing/b9mm/hp
@@ -80,10 +80,10 @@
 /obj/projectile/bullet/advanced/b9mm/hp
 	name = "9mm hollowpoint bullet"
 	icon_state = "bullet_h"
-	damage = 20
-	wound_bonus = 30
-	embedding = list(embed_chance=75, fall_chance=3, jostle_chance=4, ignore_throwspeed_threshold=TRUE, pain_stam_pct=0.4, pain_mult=5, jostle_pain_mult=6, rip_time=10)
-	armour_penetration = -30
+	damage = 15
+	wound_bonus = 15
+	embedding = list(embed_chance=30, fall_chance=3, jostle_chance=4, ignore_throwspeed_threshold=TRUE, pain_stam_pct=0.4, pain_mult=5, jostle_pain_mult=6, rip_time=10)
+	armour_penetration = -60
 
 /obj/item/ammo_casing/b9mm/rubber
 	name = "9mm rubber bullet casing"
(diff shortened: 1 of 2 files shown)


GoldenAlpharex:
While we're at it, can we get rid of the "holding at gunpoint" mechanic? It's just something dumb that somehow makes it so your accuracy is increased and it doesn't actually do much for roleplay. There's already some "holding at gunpoint" mechanics on /tg/, you just need to be at point-blank.

DiscountDirtyDan:
> While we're at it, can we get rid of the "holding at gunpoint" mechanic? It's just something dumb that somehow makes it so your accuracy is increased and it doesn't actually do much for roleplay. There's already some "holding at gunpoint" mechanics on /tg/, you just need to be at point-blank.

While I'm not the most educated to comment on a gameplay perspective I would like to say from a strictly _**roleplay perspective**_ its removal would encourage people to use emotes more rather than relying on mechanics for things, which I think is cool. Plus, it lets me point at people in the halls without locking a gun onto them, which is a good benefit imo.

Arctaisia:
> > While we're at it, can we get rid of the "holding at gunpoint" mechanic? It's just something dumb that somehow makes it so your accuracy is increased and it doesn't actually do much for roleplay. There's already some "holding at gunpoint" mechanics on /tg/, you just need to be at point-blank.
> 
> While I'm not the most educated to comment on a gameplay perspective I would like to say from a strictly _**roleplay perspective**_ its removal would encourage people to use emotes more rather than relying on mechanics for things, which I think is cool. Plus, it lets me point at people in the halls without locking a gun onto them, which is a good benefit imo.

Yep, it's a mechanic that came from Bay/Polaris, for the sole intemt of, well, holding people at gunpoint for RP. We really need to keep it.

GoldenAlpharex:
> Yep, it's a mechanic that came from Bay/Polaris, for the sole intemt of, well, holding people at gunpoint for RP. We really need to keep it.

My problem with it is if it gives a mechanical advantage to the players, which it shouldn't.

MLGTASTICa:
I believe the issue here is not the weapon damage themselves , but the bonus that comes from the combat update , it makes bullets deal 80% more damage , if your bullet had 15 now it gets 12 stamina damage ontop.

KathrinBailey:
Keep in mind security are still rocking up to the fights with either 65% melee or 65% bullet damage reduction.
Having such low damage just makes each seccie tankier than a blood-red hardsuit, but now they do similar damage to some chucklefuck assistant.

LocalizedDownpour:
If there is a way to separate like...Sec 9mm/10mm from Traitor 9mm/10mm that would be ideal.

RatFromTheJungle:
> 
> 
> Keep in mind security are still rocking up to the fights with either 65% melee or 65% bullet damage reduction.
> Having such low damage just makes each seccie tankier than a blood-red hardsuit, but now they do similar damage to some chucklefuck assistant.

elaborate? i skimmed their stats and their gear appears to be mostly around 35-40 focused on primarily chest/head (assuming the full loadout)

CameronWoof:
Security can no longer contest horror lings. Either pull back on these nerfs or nerf/disable horror form.

BrotherHangyul:
Agree with Cameron, horror form is very OP. I understand it's a last resort and meant to be powerful, but being able to re-transform in the span of ~15 seconds is incredibly overpowered.

Gandalf2k15:
> Agree with Cameron, horror form is very OP. I understand it's a last resort and meant to be powerful, but being able to re-transform in the span of ~15 seconds is incredibly overpowered.

It's 3 minutes from horror form death

CameronWoof:
The problem is that horror lings don't die. They just stay horror form, vent crawl, and fleshmend with their new infinite chem regen.

Gandalf2k15:
I think debuffing security isn't really something I wanna do, we'll continue powercreep instead.

KathrinBailey:
> I think debuffing security isn't really something I wanna do, we'll continue powercreep instead.

**Time to add the new Xenomorph caste, 40mm Xeno that has an underslung grenade launcher on each arm.**
**Blob Factories will also now spawn goliaths that are three times faster than usual!**

PR #3487, opened 2021-02-18
[NON-MODULAR] Disables OCD Device

tf-4:
## About The Pull Request

Removes organic resources disturber from the uplink. Can't really be role locked as we don't have head antags and no other roles make a lot of sense. Can't be hugely balanced around price, as either people buy it and it's a pain in the ass, or it's too expensive so a terrible idea to buy.

## Why It's Good For The Game

It causes LRP and memey situations, as well as headaches for staff. It adds nothing of particular value RP wise.

## Changelog
:cl:
del: removes organic resources disturber from the uplink
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -1968,6 +1968,8 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	item = /obj/item/autosurgeon/organ/syndicate/laser_arm
 	//restricted_roles = list("Roboticist", "Research Director") //SKYRAT EDIT: Removal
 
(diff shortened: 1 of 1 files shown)


SpaceVampire:
I agree with this.
The random event can still roll.
But its not fun or special if it can be forced every shift.
My other idea was to make it roll-locked to cargo tech.
But honestly, removing it fully is fine if not better.

Gandalf2k15:
I agree with this. I was originally unaware of the purpose of the OCD, but now that I'm aware of its purpose I'm for this change. Its completely counterproductive and actually useless in terms of it being a traitor tool.

Gandalf2k15 (approved):

PR #3503, opened 2021-02-19
[NONMODULAR] Removes spread infestation

Vizoplez:
## About The Pull Request
Players cant be trusted with it. You did this to yourselves.

## Why It's Good For The Game
Spiderspam is annoying, uncreative and promotes worse RP. Its a NRP/LRP ability at best that contributes nothing to the round.

## Changelog
:cl:
del: Removes changeling spider infest. You did this to yourselves.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/spiders.dm
+++ b/code/modules/antagonists/changeling/powers/spiders.dm
@@ -1,3 +1,4 @@
+/* Skyrat Edit: Bye! You couldnt be trusted.
 /datum/action/changeling/spiders
 	name = "Spread Infestation"
 	desc = "Our form divides, creating a cluster of eggs which will grow into a deadly arachnid. Costs 45 chemicals."
@@ -12,3 +13,4 @@
 	..()
 	new /obj/structure/spider/eggcluster/bloody(user.loc)
 	return TRUE
+*/


YakumoChen:
🧂

SpaceVampire:
If you want it to be less common, you could always make it so that you actually need to absorb someone for it, instead of just sucking a monkey/legion person.

Vizoplez:
> If you want it to be less common, you could always make it so that you actually need to absorb someone for it, instead of just sucking a monkey/legion person.

This has its own whole host of issues involving with how /tg/ now treats humonkeys, which is why they can get antagonist status such as valentines. Unfortunately, unviable for a quick fix at this point.

Overall the ability itself is poorly designed, and currently changelings near round-remove people by absorbing.

SpaceVampire:
> Overall the ability itself is poorly designed, and currently changelings near round-remove people by absorbing.

I honestly don't see an issue with the whole near round-removing by absorbing. Its supposed to be really bad if it happens to you, and if they are doing it to take your place, then they wouldn't want the body to be found anyway. But I'll trust that it won't be as easy for a fix, even if I think that locking stuff like this and horrorform behind having to do more things first would be better for the overall flow and balance of the game.

SpaceVampire:
Oh also. Up-voting your own PR is gayyyyyy.

CameronWoof:
Kamikaze ghost-roles with no sense of self-preservation whose only goal is to grief the crew is not a role we need available. 

Especially with the chemical changes, egg spam is just an irritating way for changelings to generate a deathcount without actually being able to link it back to them. Being able to shit these out under racks in maintenance and be on the other side of the station when they hatch just isn't fun gameplay on a server where it's functionally impossible to catch a changeling without them doing something incredibly dumb.

SpaceVampire:
I think that fighting the spiders is fun as they aren't actually that strong.
But I do agree that being able to link it back to the changeling in some way would be good.
Its a valid issue to take with it. I feel like people just want a way they can add some chaos to the shift.
Without personally standing up into the line of fire with a high risk of just getting cremated over not being robust enough.
If it was just the chemical changes, you could re-balance the cost/cooldown on the power.
To limit it to about the same numbers as before. I dunno, I just like spiders.
I feel like these were a lot better to fight then the broodmother ones.
Due to being unable to snowball via eggs into the endless horde.

Vizoplez:
> I think that fighting the spiders is fun as they aren't actually that strong.
> But I do agree that being able to link it back to the changeling in some way would be good.
> Its a valid issue to take with it. I feel like people just want a way they can add some chaos to the shift.
> Without personally standing up into the line of fire with a high risk of just getting cremated over not being robust enough.
> If it was just the chemical changes, you could re-balance the cost/cooldown on the power.
> To limit it to about the same numbers as before. I dunno, I just like spiders.
> I feel like these were a lot better to fight then the broodmother ones.
> Due to being unable to snowball via eggs into the endless horde.

They can self heal for 50 HP after a 20 tick timer, which equates to... about 7 seconds depending on TIDI.

Thats their entire health-bar.

Gandalf2k15:
Obligatory request to do this in another way than just /* :)

CameronWoof:
Can you talk about why this power is good, what value it brings to the table, or have any response to the problems it creates?

Obligatory request to give feedback in another way than just 
image

:)

PR #3563, opened 2021-02-21
[NONMODULAR] Removes AI lawset station trait

Vizoplez:
## About The Pull Request
Title

## Why It's Good For The Game
We disabled this in server config for a reason; we dont want AI lawsets being randomised to potentially dangerous lawsets, such as antimov and/or asiimov depending.

## Changelog
:cl:
del: Randomised lawsets due to station traits are no longer possible.
/:cl:

### DIFF
--- a/code/datums/station_traits/neutral_traits.dm
+++ b/code/datums/station_traits/neutral_traits.dm
@@ -21,7 +21,7 @@
 	show_in_report = TRUE
 	report_message = "For experimental purposes, this station AI might show divergence from default lawset. Do not meddle with this experiment."
 	trait_to_give = STATION_TRAIT_UNIQUE_AI
-*/	SKYRAT EDIT: End - DISABLES LAWSET RANDOMISATION
+*/	//SKYRAT EDIT: End - DISABLES LAWSET RANDOMISATION
 
 /datum/station_trait/ian_adventure
 	name = "Ian's Adventure"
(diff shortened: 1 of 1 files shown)


SpaceVampire:
Looking at the code I am like 90% sure that that isn't an issue. It seemed to call the random lawset based on the lawset weights in the config. The really deadly ones all have a weight of zero and thus will never be rolled. So you are most likely to get one of the fairly standard lawsets with a small chance to roll a quirky one. I think its fun to have as a rare station trait, as the lawsets in general are unlikely to be used outside of it.

SpaceVampire:
Wait, so are you just pushing ahead with this anyway, without saying anything here about it?
All the possible lawsets are still pretty safe, and I think it could be fun to see another one rarely.
If there was something you really didn't like, you could just remove it from the possible list via configs.
Instead of removing the event all together.

SpaceVampire:
If the AI isn't an antag, its pretty clear that it shouldn't just be killing people even if it did end up on something like Asimov. Its not allowed to just kill non-humans because it feels like it, even on TG. If it really gets out of control its not too hard to reset the lawset back to normal as well. I think that having some amount of conflict, on the rare chance that this lawset rolls isn't that big of a deal. All the other options Corporate, Paladin, Robocop, etc, feel fine. Really I feel like this is just a weird removal, given that its already an uncommon event, of which the outcome you are really worried about is already a sub-section of that.

Gandalf2k15:
Station traits are great because they only happen every so often, removing this one means that there will be more occurrences of the other ones. I haven't really seen any issues with the trait, granted, I never play AI. However as Vampire said I don't see why we should babysit the shitter AI's.

PR #3610, opened 2021-02-23
CQC+

Gandalf2k15:
## About The Pull Request

Adds a new traitor item, CQC+, this costs a whopping 30 TC but combines the best of CQC with sleeping carp. Considering it costs most of your TC, you'd expect it to be good. And it is, you'll be able to do all of the CQC moves without the disadvantage of being downed in 3.2 seconds because "FIRE AT WILL". 

Part of the traitor buff.

## Changelog
:cl:
add:  Added a new martial arts book, CQC+, capable of standard CQC moves as well as projectile deflection.
/:cl:

### DIFF
--- a/code/__DEFINES/melee.dm
+++ b/code/__DEFINES/melee.dm
@@ -7,4 +7,5 @@
 #define MARTIALART_MUSHPUNCH "mushroom punch"
 #define MARTIALART_KRAVMAGA "krav maga"
 #define MARTIALART_CQC "CQC"
+#define MARTIALART_CQC_PLUS "CQCPLUS" //SKYRAT EDIT ADDITION
 #define MARTIALART_PLASMAFIST "plasma fist"
--- a/modular_skyrat/master_files/code/datums/martial/cqcplus.dm
+++ b/modular_skyrat/master_files/code/datums/martial/cqcplus.dm
@@ -25,6 +25,4 @@
 	name = "old but gold manual"
 	martialname = "close quarters combat plus"
 	desc = "A small, black manual. There are drawn instructions of tactical hand-to-hand combat. This includes how to defelct projectiles too."
-	greet = "<span class='boldannounce'>You've mastered the basics of CQC.</span>"
-	icon_state = "cqcmanual"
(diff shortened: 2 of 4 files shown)


BrotherHangyul:
**ahem**

YO! KIRYU-CHAN!

Jokes aside, I could see this being OP but at the same time, sec's already pretty strong. I can see good gimmicks coming with this.

CameronWoof:
Please list the exact effects of this.

Ziiro:
A quick glance at the code it appears to just give you 85% block chance (up from baseline CQC's 75%), and the throw mode bullet deflection. ~~No extra moves~~. Seems like it would combo better with Gloves Of the North Star, actually.

Seems a little underwhelming at a baseline for 77% of your TC.

edit: duh inherits CQC's moves, my bad.

CameronWoof:
Is that bullet deflection specifically, or does that include all projectiles? 

85% melee block and immunity to all projectiles seems incredibly strong.

Ziiro:
The projectile deflection requires you to be in throw mode, which requires a lot of micromanagement if you're actually fighting - because it toggles off every time you click or do anything else. It's actually super annoying.

Swiftfeather:
> The projectile deflection requires you to be in throw mode, which requires a lot of micromanagement if you're actually fighting - because it toggles off every time you click or do anything else. It's actually super annoying.

We did discuss having bullet reflect at all times but determined A) that would actually be broken, and B) sometimes you don't want to reflect bullets

The requirement to be in throw mode is what has always made sleeping carp not MEGA powerful, because it really is a pain in the ass to fight and manage throw mode, so for most of our player base, this makes it so that you can either deflect bullets OR fight, being difficult to safely do both.

Gandalf2k15:
Yeah as stated in the comments above I honestly think 35 TC is a bit much. Unless we make it permanently deflect bullets... we'll see how it goes in the test runs but honestly I don't think it's we'll priced. A reduction or buff may be in order. I guess that's the entire point of test merging something.

Gandalf2k15:
Doesn't seem like it's been a problem really.

Draganfrukts:
Tbh I'd keep throw mode reflect *but* make the counter-attack chance a disgusting chance like 95%

Gandalf2k15:
Lowering TC cost to 30.

Gandalf2k15:
Full merging this as the TM period is completed.

PR #3612, opened 2021-02-23
Lowers speed bonus from Stabilized pink extracts, balances stabilized sepia.

Swiftfeather:
## About The Pull Request

A simple change that lowers the value of the light pink to provide only half as much speed bonus. The slowdown has gone from -0.5 to -0.25 (Yes, it doesn't have a speedup, it's just a negative slowdown :^) ) It's still a significant speed increase, and better meshes with the fact that it's original intention was supposed to be "amublance" since it also injects 5u of ephedrine into people in crit nearby. 

Also lowers the sepia's potential 50% speed increase from (-0.5) to (-0.2) so that it is not as stackably powerful with stabilized pinks. It ALSO actually adds a real slowdown chance (the chance is 50/50 still, but the slowdown becomes ~~30~~ 20%, AKA 0.2) from sepia, as previously it gave NO slowdown effect despite its examine text saying it randomly adjusts, it was just, a 50% chance of adjusting to 50% speed, or a 50% chance of adjusting to the same speed, which... isn't that random.

- Stab pink speedup reduced from (-0.5) to (-0.25)
- Stab sepia now has a chance to slowdown rather than do nothing. 
- Stab sepia's speedup reducced from (-0.5) to (-0.2). If the speedup isn't rolled, gets a slowdown of (0.2)

## Why It's Good For The Game

On our server we move so fast already (when the server isn't pinned) and if a xenobiologist decides to make these for the entire security team, it kinda sucks tbh. This doesn't make the extracts obsolete in any regard, they still have a useful speed increase and the light pink's ephedrine into crit people is still useful.

Sepia is also nerfed 

## Changelog
:cl:
balance: Stabilized light pinks do not give as much speed bonus.
balance: Stabilized sepia also has movespeed reduced, plus a chance to slowdown rather than do nothing.
/:cl:


SpaceVampire:
Having the sepia make you slower then it makes you faster seems pretty odd. Is there something I am missing, or did you pretty much make it so that there isn't any reason to use it not?

Swiftfeather:
> Having the sepia make you slower then it makes you faster seems pretty odd. Is there something I am missing, or did you pretty much make it so that there isn't any reason to use it not?

In the item's description it says it randomly alters your speed, but it only moved it up, it had a 50% chance of a 50% speed increase, and a 50% chance of no speed change, a downside to the upside seems much more balanced.

They also already had the code there to adjust just one number.

SpaceVampire:
Yeah, but you made the bad roll from it, worse then the good roll.
That was my main point. Why would you want a waste a slot on something that has more overall downsides then upsides?
Compared to before where it made you faster, or just did nothing, this is pretty major nerf.

Swiftfeather:
> Yeah, but you made the bad roll from it, worse then the good roll.
> That was my main point. Why would you want a waste a slot on something that has more overall downsides then upsides?
> Compared to before where it made you faster, or just did nothing, this is pretty major nerf.

It adds risk to using the item.

SpaceVampire:
> It adds risk to using the item.

My issue wasn't there you added any downside to it. Its that it seems like you made the downside WORSE then the upside. When that shouldn't be the case for something you have to invest a lot of time into creating, that also takes up a slot in your backpack to be used. That being said, I'm not even sure why there needed to be "risk" in the first place, given that all of the other stabilized extracts are just good, but I feel that not having the speed up and slowdown be even at the very least is pretty lame.

EDIT: I mean, this is "lets take two items that were strong together, nerf one of them, and then nerf the other one so much that its not even worth making anymore." I feel like it just comes across as unnecessary.

SpaceVampire:
I'm just really not sure why I would ever use sepia with the new changes to it. Can you give a reason as to why exactly you think the new design of it is better? Maybe there is something I am missing, but it went to something that was useful with the downside of not always being active, to something that isn't that impactful or all, or actually a downside. Because you really only went into it for the stuff about the pink one, which while I still don't like being nerfed, I can at least understand.

Oscari-pog:
Very good PR, no more security running around at lightspeed

Gandalf2k15:
So obviously this is a balance change and there's not much to review codewise. Looking over it a few good points were brought up, mainly that it obviously sort of removes the reason to have the sepia slimes. But i also agree that having lightspeed security is bad. I suggest a test merge. But codewise, this is fine.

Gandalf2k15 (approved):

Swiftfeather:
> So obviously this is a balance change and there's not much to review codewise. Looking over it a few good points were brought up, mainly that it obviously sort of removes the reason to have the sepia slimes. But i also agree that having lightspeed security is bad. I suggest a test merge. But codewise, this is fine.

You make sense and since I made this pr I've sort of seen that with the sepias, I can reduce the slowdown from 20% to 10 or 5%,to make the downside not so severe

RatFromTheJungle:
> 
> 
> > So obviously this is a balance change and there's not much to review codewise. Looking over it a few good points were brought up, mainly that it obviously sort of removes the reason to have the sepia slimes. But i also agree that having lightspeed security is bad. I suggest a test merge. But codewise, this is fine.
> 
> You make sense and since I made this pr I've sort of seen that with the sepias, I can reduce the slowdown from 20% to 10 or 5%,to make the downside not so severe

another thing to note, is a stable red removes the slowdown for sepias entirely, so them having a large slowdown is the exchange for not using up that spare storage space

YakumoChen:
I think what people are trying to say is that there is no reason to ever use Sepia over Light Pink.
Light Pink always gives a better speed boost regardless of what state Sepia is in, where Sepia has high risk for little to no reward comparatively. You could, in theory, stack them with the red slime one and get ~-.5 speed but...

It would make more sense for Sepia to have a wider range of speed states so that sometimes you can go FASTER than light pink, with the risk of also getting slowdown instead at other times, so that the incentive to use it isn't just flat out outweighed by Light Pink being objectively superior at all times.

RatFromTheJungle:
> 
> 
> I think what people are trying to say is that there is no reason to ever use Sepia over Light Pink.
> Light Pink always gives a better speed boost regardless of what state Sepia is in, where Sepia has high risk for little to no reward comparatively. You could, in theory, stack them with the red slime one and get ~-.5 speed but...
> 
> It would make more sense for Sepia to have a wider range of speed states so that sometimes you can go FASTER than light pink, with the risk of also getting slowdown instead at other times, so that the incentive to use it isn't just flat out outweighed by Light Pink being objectively superior at all times.

(you can use both, at the same time)

YakumoChen:
And? They both have different features. If the light pink one is going to be objectively better than find a different use for sepia, or remove it.
If you're going to keep sepia, red, AND light pink for a stupid speed boost hogging your backpack space, more power to you.

Or remove the speed boost from light pink altogether, or give it a different buff like stamina regen.

On a different note, have you considered PRing this upstream?

Swiftfeather:
> On a different note, have you considered PRing this upstream?

Yes, upstream has different server framerate or something, base speed feels slower to the player, and therefore the stacked speed increases don't feel as... absurdly hyperspeed there, I have tested it yes. Their rounds are also not nearly as long as ours, they often don't have time to make it as far as pinks and sepias.

YakumoChen:
I think you're going to have to remake the PR, you fucked up a merge somewhere.

Swiftfeather:
> I think you're going to have to remake the PR, you fucked up a merge somewhere.

I certainly fucking did.

PR #3630, opened 2021-02-24
Removes forced brain damage speech from dumbness trauma

Azarak:
## Changelog
:cl:
del: Removed forced brain damage speech from dumbness trauma
/:cl:


RatFromTheJungle:
for the best

PR #3689, opened 2021-02-27
[NonModular] Sepia/lightpink rebalance

Swiftfeather:
## About The Pull Request

REMAKE OF https://github.com/Skyrat-SS13/Skyrat-tg/pull/3612

lightpink down to 25% speed buff from 50%, sepia speed buff down to 20% from 50%, wth a possible slowdown of 10%

I totally FUCKERED my last PR https://github.com/Skyrat-SS13/Skyrat-tg/pull/3612 of this, like, utterly annihilated it. All that's left is a smoldering crater. Don't go there.

- Lightpink speed buff reduced from 50% to 25%
- Sepia speedbuff reduced 50% to 20%
- Sepia, instead of doing nothing if it doesn't roll the speedbuff, now applies a 10% slowdown.

## Why It's Good For The Game

No lightspeed sec, or other people, adds a potential, and slight, downside if you don't roll the speed buff on the stabilized sepias. It's not as bad as the possible good.

## Changelog
:cl:
balance: Balanced stabilized Light pink and sepia extracts to not create lightspeed characters
/:cl:

### DIFF
--- a/code/modules/movespeed/modifiers/status_effects.dm
+++ b/code/modules/movespeed/modifiers/status_effects.dm
@@ -5,7 +5,7 @@
 	multiplicative_slowdown = 3
 
 /datum/movespeed_modifier/status_effect/lightpink
-	multiplicative_slowdown = -0.5
+	multiplicative_slowdown = -0.25 //SKYRAT EDIT: Oiriginal value (-0.5)
 	blacklisted_movetypes = (FLYING|FLOATING)
 
 /datum/movespeed_modifier/status_effect/tarfoot
--- a/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
+++ b/code/modules/research/xenobiology/crossbreeding/_status_effects.dm
@@ -683,11 +683,11 @@
 /datum/status_effect/stabilized/sepia/tick()
 	if(prob(50) && mod > -1)
 		mod--
-		owner.add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/status_effect/sepia, multiplicative_slowdown = -0.5)
+		owner.add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/status_effect/sepia, multiplicative_slowdown = -0.2) //SKYRAT EDIT: Original value (-0.5)
 	else if(mod < 1)
 		mod++
 		// yeah a value of 0 does nothing but replacing the trait in place is cheaper than removing and adding repeatedly
-		owner.add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/status_effect/sepia, multiplicative_slowdown = 0)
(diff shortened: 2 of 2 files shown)


SpaceVampire:
Shouldn't the max speed of Sepia be higher then the guaranteed speed that you get from pink? Otherwise it seems like its just strictly worse. You don't ever get the free healing chem on crit people thing. I would like to see you make it special in a way that wouldn't just make people Pick every single time over it, only maybe bothering if they wanted to stack em.

RatFromTheJungle:
> 
> 
> Shouldn't the max speed of Sepia be higher then the guaranteed speed that you get from pink? Otherwise it seems like its just strictly worse. You don't ever get the free healing chem on crit people thing. I would like to see you make it special in a way that wouldn't just make people Pick every single time over it, only maybe bothering if they wanted to stack em.

stacking them is the meta, sepia is rarely if ever used alone

Swiftfeather:
> Shouldn't the max speed of Sepia be higher then the guaranteed speed that you get from pink? Otherwise it seems like its just strictly worse. You don't ever get the free healing chem on crit people thing. I would like to see you make it special in a way that wouldn't just make people Pick every single time over it, only maybe bothering if they wanted to stack em.

The reason the max speed is still at that is that they can still be stacked, coming out to a pretty solid speed buff, still making it _worthwhile_ to do, just a bit more tricky.

SpaceVampire:
> The reason the max speed is still at that is that they can still be stacked, coming out to a pretty solid speed buff, still making it _worthwhile_ to do, just a bit more tricky.

I'm just saying that the Max-Speed for the Sepia should be higher then the speed of the Lightpink. Given that you won't have it active all of the time. I just don't want it to be strictly worse, like it is at the moment.

Vizoplez (changes requested):
change title to [NONMODULAR]Sepia/Light Pink Speed Rebalance

YakumoChen:
> 
> 
> > The reason the max speed is still at that is that they can still be stacked, coming out to a pretty solid speed buff, still making it _worthwhile_ to do, just a bit more tricky.
> 
> I'm just saying that the Max-Speed for the Sepia should be higher then the speed of the Lightpink. Given that you won't have it active all of the time. I just don't want it to be strictly worse, like it is at the moment.

That's exactly what I pointed out in the last PR. There's no reason to use Sepia by itself since regardless of how it's used it's always strictly worse than light pink.

I still think you should just remove light pink's speed boost and give it some other buff, like stamina regen.

YakumoChen:
If nothing else, switch the lightpink buff to 10 or 20% and the Sepia buff to 25%

Gandalf2k15:
I'm happy with this after checking it over. We need to do SOMETHING about them. If this invalidates sepias then we can change them again.

Gandalf2k15 (approved):

PR #3739, opened 2021-03-01
[NONMODULAR]Brings Noc regen in line with starlight and fleshmend

Vizoplez:
## About The Pull Request
Changes the symptom threshold from 8 to 7 like every other tier 6 healing symptom.

## Why It's Good For The Game
Consistency and more freedom to avoid 'minmax' diseases, mcbruh.

## Changelog
:cl:
tweak: Nocturnal regen is now in line with every tier 6 healing symptom, in terms of thresholds.
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -184,14 +184,14 @@
 	level = 6
 	passive_message = "<span class='notice'>You feel tingling on your skin as light passes over it.</span>"
 	threshold_descs = list(
-		"Stage Speed 7" = "Doubles healing speed.",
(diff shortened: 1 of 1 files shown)


SpaceVampire:
Do you think that upstream would accept this?
I know it doesn't really matter, but I would be curious.
I do agree with the change.

Vizoplez:
> Do you think that upstream would accept this?
> I know it doesn't really matter, but I would be curious.
> I do agree with the change.

They likely wouldnt due to a massive stigma involving both nanites and virology, and any conceived 'buff' being rejected out of fear of making wounds less effective once more, atleast; thats what I've seen.

Gandalf2k15:
Aye aye.

Gandalf2k15 (approved):

PR #3755, opened 2021-03-02
[SEMI-MODULAR][SOMEWHAT WIP]Pipe tripping

Vizoplez:
## About The Pull Request
Fuck you, Gandalf

## Changelog
:cl:
add: Pipe tripping mechanic
/:cl:


modular_skyrat/master_files/code/datums/traits/skyrat-traits.dm
```diff
@@ -0,0 +1 @@
+#define TRAIT_NOTRIP "TRAIT_NOTRIP"
```
Gandalf2k15: Denied, no newline

Vizoplez: I dont care

Vizoplez:
so yes the element 'technically' works but fails to actually trip people, p sure this is an invoker issue

Vizoplez:
also, obligatory monkey AI defeating CI. Again

Arctaisia:
Time to line the halls with pipes

CameronWoof:
Why?

Swiftfeather:
Fuck you! >:(

Vizoplez:
@CameronWoof 
image

CameronWoof:
Oh. In that case, I think the problem is that at no point in this code do I see it actually calling a proc to trip anything. It just outputs text.

Vizoplez:
> Oh. In that case, I think the problem is that at no point in this code do I see it actually calling a proc to trip anything. It just outputs text.
image

SpookyTheFox:
> 
> 
> > Oh. In that case, I think the problem is that at no point in this code do I see it actually calling a proc to trip anything. It just outputs text.
> > image

have them like 
Paralyze(5)
add_confusion(50)
dropItemToGround(M.get_active_held_item())
dropItemToGround(M.get_inactive_held_item())
	
when they trip

Vizoplez:
> > > Oh. In that case, I think the problem is that at no point in this code do I see it actually calling a proc to trip anything. It just outputs text.
> > > image
> 
> have them like
> Paralyze(5)
> add_confusion(50)
> dropItemToGround(M.get_active_held_item())
> dropItemToGround(M.get_inactive_held_item())
> 
> when they trip

I'll most likely just make it apply a small stam debuff and a slight slowdown by default, namely for balance concerns

MLGTASTICa:
> so yes the element 'technically' works but fails to actually trip people, p sure this is an invoker issue

use COMSIG_ATOM_CROSSED , that should work with elements , since elements themselves are not on the SPOT of the object they are attached to unless you make them be.

KathrinBailey:
**YOUR INTEGRATION TESTS HAVE BEEN RUMMAGED BY MONKEY BUSINESS**
image
**PAY THE BANANA TAX OR NO COMPILE FOR YOU, OOH OOH OOH!!**

Azarak:
This is a bad idea and bad implementation. Because of how we have piping done on maps, and that maintenance doesnt have any catwalks or such, it's going to be a chore to get around

PR #3768, opened 2021-03-02
[NON-MODULAR] Makes the teleprod weight class bulky

tf-4:
## About The Pull Request

Changes the teleprod from normal weight class to bulky

## Why It's Good For The Game

Brings it in line with the other prods, as well as it making very little sense that the most powerful one is not bulky while the others are. It is a beyond free escape tool as it stands, this will keep its power while stopping it from having absolutely zero downsides.
As it stands, it's basically impossible to get a teleprod away from someone that is using it, as they can just pull it out their bag and zap themselves before you can react. Now it's possible for them to actually drop it, letting you have a little more counterplay.

## Changelog
:cl:
balance: Teleprods are now bulky
/:cl:


Gandalf2k15 (approved):
Seems reasonable to me. And it makes sense. Same sized items

Issue #3530, opened 2021-02-20
Xenos cant wag their tails 


Swiftfeather:
## Round ID:

Something like 

## Reproduction:

Xenocrew with the exitable quirk can wag their tails when headpat, but since their tails do not wag they just disappear in stead

YakumoChen:
There isnt a sprite for it, probably because wag calls for a sprite someone forgot to add, or didnt add a duplicate in the wag dmi to prevent it vanishing.

PR #3788, opened 2021-03-03
[MODULAR] Fixes xeno tails vanishing when wagging

Swiftfeather:
## About The Pull Request

This rectifies issue https://github.com/Skyrat-SS13/Skyrat-tg/issues/3530. 

Please note: Xeno hybrids still don't have a wag animation, we still need one, if any spriters are interested.

## Why It's Good For The Game

Xenos tails will not disappear when they wag

## Changelog
:cl:
fix: Xeno tails no longer vanish if they wag them. (Still not animated though)
/:cl:


unit0016:
Can we get one for the monkey tail, too?

Swiftfeather:
> Can we get one for the monkey tail, too?

Monkeys aren't real

GoldenAlpharex:
So you just fixed it by creating an empty file...?

Swiftfeather:
> So you just fixed it by creating an empty file...?

No, I had to add another sprite state that is a copy of the xeno tail state and add _wagging to it so that it knows to path to that state instead of going to a state that doesn't exist causing the tail to vanish.f

also .dmi files are the icon files, they don't show up on github.

GoldenAlpharex:
> > So you just fixed it by creating an empty file...?
> 
> No, I had to add another sprite state that is a copy of the xeno tail state and add _wagging to it so that it knows to path to that state instead of going to a state that doesn't exist causing the tail to vanish.f
> 
> also .dmi files are the icon files, they don't show up on github.

Yeah, I know it's icon files, I'm just a bit curious as per to why it's saying that it's empty on GitHub though, like, usually it just says that it doesn't support the format of binary files or something generic like that.
I checked the check and it works, huh, didn't know you could actually have a visual there, that's good to know for the future.

PR #3795, opened 2021-03-03
[MODULAR] Ninja kit fix thing

Swiftfeather:
## About The Pull Request

So.... the ninja sword was set to 75 block_chance, where the base vibro blade is only 40... you know why? Because when wielded we get this little thing called 
```
if(wielded)
	final_block_chance *= 2
```
which... clearly very much is not what the creator of the kit knew. As our wielding antags were sitting at a sweet 150% block chance. Amazing.

## Why It's Good For The Game

Lmao

## Changelog
:cl:
fix: Reduces ninja traitor kit  blade to be less than 100% block chance when wielded
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/uplink/uplink_items.dm
+++ b/modular_skyrat/master_files/code/modules/uplink/uplink_items.dm
@@ -308,7 +308,7 @@
 	name = "Cyborg Ninja bundle"
 	desc = "Become a force of nature with this customized kit featuring next-generation syndicate technology in an efficient package."
 	item = /obj/item/storage/backpack/duffelbag/syndie/loadout/ninja
-	cost = 25
+	cost = 45
 
 /datum/uplink_item/loadout_skyrat/darklord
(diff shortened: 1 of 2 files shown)


SpaceVampire:
I feel like it could still be higher, if clearly not that high.
Because with the current stamina bullets, it feels like getting hit just a couple times is enough to screw you.
Maybe you could it like a 2/3rds chance at 33% or the actual 40% the base one is?

SarmentiCampbell:
No need to up the price, just change the block chance.

Swiftfeather:
> I feel like it could still be higher, if clearly not that high.
> Because with the current stamina bullets, it feels like getting hit just a couple times is enough to screw you.
> Maybe you could it like a 2/3rds chance at 33% or the actual 40% the base one is?

I set it to 25 to get a wielded block of 50, but I might change that to 30 and reduce weapon damage

Gandalf2k15 (approved):

PR #3875, opened 2021-03-04
[NONMODULAR]Employs interns less.

Vizoplez:
## About The Pull Request
Interns now roll less frequently due to their high chance of hiding priority alerts, such as CC Updates, blobs and radstorms

## Why It's Good For The Game
Intern man rolls too often and causes serious damage to RP and the round itself.

## Changelog
:cl:
tweak: Intern station trait now rolls much, much less frequently. 5 to 1 reduction.
/:cl:

### DIFF
--- a/code/datums/station_traits/neutral_traits.dm
+++ b/code/datums/station_traits/neutral_traits.dm
@@ -52,15 +52,13 @@
 	report_message = "Something seems to be wrong with the PDAs issues to you all this shift. Nothing too bad though."
 	trait_to_give = STATION_TRAIT_PDA_GLITCHED
 
-/* SKYRAT EDIT BEGIN: - Removes intern trait
 /datum/station_trait/announcement_intern
 	name = "Announcement Intern"
 	trait_type = STATION_TRAIT_NEUTRAL
-	weight = 5
+	weight = 1 //SKYRAT EDIT: - CHANGES INTERN WEIGHT
 	show_in_report = TRUE
 	report_message = "Please be nice to him."
 
 /datum/station_trait/announcement_intern/New()
 	. = ..()
 	SSstation.announcer = /datum/centcom_announcer/intern
-*/ //SKYRAT EDIT END.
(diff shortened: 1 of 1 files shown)


Swiftfeather:
Intern good.

chefferz:
Agreed, it's LRP and not amusing in the slightest. I've seen it mask super important alerts like maintenance radiation as well.

Vizoplez:
> Agreed, it's LRP and not amusing in the slightest. I've seen it mask super important alerts like maintenance radiation as well.

This was exactly my concern - it hides CC announcements, fucks with alert text, ecetera; it obscures vital information which can lead, on some stations, to rapid, unavoidable death

SpaceVampire:
> Agreed, it's LRP and not amusing in the slightest. I've seen it mask super important alerts like maintenance radiation as well.

How did you see that here, when we don't have the maintenance radiation event?
You can also still guess the event in a decent number of cases, because of the heading.
I think removing some information is fun, and I am pretty sure the Captain still gets a print-out.
Like, that is clearly part of the point of it. 

EDIT: Yawet, thumbs up on your own PR is still cringe, same as last time.

MLGTASTICa:
Its a good event , just learn to deal with it.

tf-4:
It's LRP, WAY too frequent, and not amusing after the first few times.
I'd rather it was removed, but if not at least tone down the frequency it rolls

SpaceVampire:
Given that the main issue seems to be. "It happens too often."
Just lower the current weight value of five.
Make it like, two or something.

EDIT: Actually yeah. I think that should be done period.
I still support having the event, but it would be more special if it happened less often.
I'll just make my own quick web-edit PR doing that, unless you wanna do it here.
@Yawet330

Gandalf2k15:
Yeah just lower the weight substantially. 2?

RatFromTheJungle:
the intern should roll once every few days, not every round or two, (considering how drastically he does mess with announcements, namely blob/dragon/anything else) i'd go for 1.

PR #3880, opened 2021-03-05
Replaces generic ambience with ERIS style sounds and removes the LRP ones from the ambience pool

Gandalf2k15:
## About The Pull Request

I've had a few complaints and suggestions to change the generic ambience on the station(hallways and generic areas).

This rectifies that by porting some ERIS's style ambience and removing some of the LRP ones from the pool.

Due to the fact these are shorter, the max time for the next ambience sound is now 60.

## Changelog
:cl:
soundadd: Ship ambience is now much more like ERIS ambience.
/:cl:


gonenoculer5:
My only issue with this PR is Ambient 10, which might need a volume adjustment before its added, even on medium volume it can be a bit jarring. Other than that, good change. I hope this doesnt remove the AI satelite space jam ambience, because that one was cool.

code/game/area/areas.dm
```diff
@@ -88,7 +88,7 @@
 	///Used to decide what the minimum time between ambience is
 	var/min_ambience_cooldown = 30 SECONDS
 	///Used to decide what the maximum time between ambience is
-	var/max_ambience_cooldown = 90 SECONDS
+	var/max_ambience_cooldown = 60 SECONDS //SKYRAT EDIT CHANGE - ORIGINAL: 90
```
GoldenAlpharex, line 91: So, after spending quite a lot of time in-game since that was test-merged, I can safely say that 60 seconds is not a long-enough cooldown. Ambience is overlapping waaaaay too often. I'm not sure if you want me to make a quick PR to revert this or if you'd rather do it yourself.

PR #3882, opened 2021-03-05
[Modular] Removes the explosive walls from Interdyne

unit0016:
## About The Pull Request
Cloning is gone, accidentally round removing people 4noraisin should be too. Besides, these are literally only in place to make the explosion from when mining raids the base (which is against the rules anyways, honk) cooler and more complete.
This doesn't remove the self-destruct bomb.

## Why It's Good For The Game
RR bad. RR on Ghostroles is especially bad. Accidental RR On Ghostroles is e s p e c i a l l y bad.

## Changelog
:cl:
del: Interdyne is no longer laced with explosive walls. Only a self-destruct remains.
/:cl:


Draganfrukts:
Like a dog filled with tannerite except it's a wall

YakumoChen:
what's the point of the self destruct bomb then if it doesn't remove the base off the map?

PR #3904, opened 2021-03-05
Disables flipping and spinning, removes dabbing, adjusts blepping

NullDagaf:
Title.
Adds a check for if an emote is "blacklisted" - if it is, you cannot use it intentionally. Makes flip and spin blacklisted. Dabbing was already fully modular, so I just removed it. Also makes it so blepping emote doesn't end with "Blep."

also now replaces spinning with twirling
because salt

### DIFF
--- a/code/datums/emotes.dm
+++ b/code/datums/emotes.dm
@@ -28,6 +28,7 @@
 	//SKYRAT EDIT ADDITION BEGIN - EMOTES
 	var/sound_volume = 25 //Emote volume
 	var/list/allowed_species
+	var/blacklisted = FALSE
 	//SKYRAT EDIT ADDITION END
 
 /datum/emote/New()
@@ -48,6 +49,10 @@
 	. = TRUE
 	if(!can_run_emote(user, TRUE, intentional))
 		return FALSE
+	//SKYRAT EDIT ADDITION START
+	if(blacklisted && intentional)
+		return FALSE
+	//SKYRAT EDIT END
 	var/msg = select_message_type(user, intentional)
 	if(params && message_param)
 		msg = select_param(user, params)
--- a/modular_skyrat/modules/emotes/code/emotes.dm
+++ b/modular_skyrat/modules/emotes/code/emotes.dm
@@ -391,6 +391,12 @@
 	message_AI = "shows an image of a random blepping animal."
 	message_robot = "bleps their robo-tongue out."
 
+/datum/emote/living/carbon/twirl
+	key = "twirl"
+	key_third_person = "twirls"
+	message = "twirls around on their foot!"
+	mob_type_allowed_typecache = list(/mob/living/carbon))
+
 /datum/emote/living/bork
 	key = "bork"
 	key_third_person = "borks"
--- a/modular_skyrat/modules/emotes/code/emotes.dm
+++ b/modular_skyrat/modules/emotes/code/emotes.dm
@@ -173,13 +173,6 @@
 	vary = TRUE
 	sound = 'modular_skyrat/modules/emotes/sound/voice/weh.ogg'
 
-/datum/emote/living/dab
-	key = "dab"
-	key_third_person = "dabs"
-	message = "suddenly hits a dab!"
-	emote_type = EMOTE_AUDIBLE
-	hands_use_check = TRUE
-
 /datum/emote/living/mothsqueak
 	key = "msqueak"
 	key_third_person = "lets out a tiny squeak"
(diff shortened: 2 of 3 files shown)


code/datums/emotes.dm
```diff
@@ -28,6 +28,7 @@
 	//SKYRAT EDIT ADDITION BEGIN - EMOTES
 	var/sound_volume = 25 //Emote volume
 	var/list/allowed_species
+	var/blacklisted = FALSE
```
Gandalf2k15, line 31: Make this a config!

SpaceVampire:
There is nothing wrong with flipping and spinning, they are already on a cooldown so you can't spam it.

Swiftfeather:
Flipping and spinning is good. They have a cooldown. No need to flatout remove.

code/datums/emotes.dm
NullDagaf: HOW THE FUCK DO CONFIGS WORK

Draganfrukts:
Did you change desword code to prevent flipping and spinning

NullDagaf:
> Did you change desword code to prevent flipping and spinning

No, but only because I don't want to be yelled at for commenting things out.

NullDagaf:
aaa why is git yelling at me

NullDagaf:
> There is nothing wrong with flipping and spinning, they are already on a cooldown so you can't spam it.

They shouldn't be in, in the state they are in, in the first place.

modular_skyrat/modules/emotes/code/emotes.dm
```diff
@@ -394,9 +387,9 @@
 /datum/emote/living/blep
 	key = "blep"
 	key_third_person = "bleps"
-	message = "bleps their tongue out. Blep."
```
NotRanged, line 397: why remove this though

NotRanged: removing flipping and dabbing is fine, but keep spinning

NullDagaf: > why remove this though

Just makes it not end with Blep.

SpaceVampire:
> aaa why is git yelling at me

109952116-64bebf00-7cd6-11eb-8cbb-afd6294f97a5

The monkeys know that you have sinned.

Arctaisia:
Honestly, removing dabbing is great, I'm fully behind that. Disabling flipping and spinning is a bit out there, as they can be used in other situations that make sense ICly, like a show. One way we could combat those is add an automatic /me to both, like what I know polaris code has, and make it so doing both while running can cause you to trip over unless wielding a d-esword. Blepping is fine as is, mlemming is the exact same, it ends in mlem, these two are VOREstation emotes that Tali added while they were still here, so I'm not sure why we really need to remove the blep, or even mlem at the end of both, as I think it adds some cuteness to the emotes.

NullDagaf:
> Honestly, removing dabbing is great, I'm fully behind that. Disabling flipping and spinning is a bit out there, as they can be used in other situations that make sense ICly, like a show. One way we could combat those is add an automatic /me to both, like what I know polaris code has, and make it so doing both while running can cause you to trip over unless wielding a d-esword. Blepping is fine as is, mlemming is the exact same, it ends in mlem, these two are VOREstation emotes that Tali added while they were still here, so I'm not sure why we really need to remove the blep, or even mlem at the end of both, as I think it adds some cuteness to the emotes.

It doesn't remove blepping at all. Just cuts the "Blep" part at the end.

NullDagaf:
> > aaa why is git yelling at me
> 
> 109952116-64bebf00-7cd6-11eb-8cbb-afd6294f97a5
> 
> The monkeys know that you have sinned.

fuck the monke

GoldenAlpharex:
I don't see what's wrong with spinning and flipping, especially since we can already basically spin by holding ctrl and then just mashing WASD.
I don't really see anyone abusing flip either, so I'm not sure what the point of removing that is, we already have a cooldown on emotes, so...

comet-husky:
I absolutely support the opinions above, we need to keep spins!

lolman65:
we should at *speen for the speen meme

NullDagaf:
> I absolutely support the opinions above, we need to keep spins!

see commit above

Swiftfeather:
> It doesn't remove blepping at all. Just cuts the "Blep" part at the end.

That's the best part of the blep emote, please don't even try to remove it

NullDagaf:
Bleh. Closing because someone more wise than me said to redo it.

PR #3921, opened 2021-03-06
[MODULAR] Makes the Security Brain Surgeon into the Security Paramedic

SpookyTheFox:
## About The Pull Request
The Security Medic currently bridges three different jobs even on minimal access. This PR attempts to pare that down to two. This means mostly removals, but there are a few replacements. I will detail them below.

First up, is the access this role gets. Even on a skeleton crew, they do not need to perform the jobs of the chemist, the virologist, or the geneticist. The are a medic role, not a doctor, so I have removed their access to surgery. This is two fold. It cuts off access to the doctor's gear lockers, which contain items they shouldn't have access to at round start, and also to the actual surgery rooms, for the same reason. They really should not do a doctor's job at all. If they insist on it, they can wait until a few minutes into the round  for research, do ghetto surgery, or trawl maintenance for those tools like everyone else that's not a medical doctor.

Next, is the item changes. The Security Medic is not a doctor, and should not be doing revivals. Thus, their defibrillator has been removed. Without surgical access, they can't steal the ones in medbay either. If they absolutely insist on doing another role's job, or power gaming a malfunctioning defib, they can get it from the medical fab, and find a cell on their own. They do not need to spawn with an item even the CMO doesn't spawn with.

Similarly, their surgical duffel bag has been removed. The Security Medic should not be doing surgery. That is for doctors. Further, it gives security too easy access to implant searches and pacification. It should require a true medical expert to perform those procedures, as well as almost any other surgery.

The Advanced First Aid kit has been removed. Everyone else has to buy that from the vending machine at high cost, the Security Medic should too. Not even ERT gets that one for free. Besides, the sutures and regenerative mesh in an ordinary gray first aid kit is faster in most cases. The Advanced First Aid kit has been replaced with a brute kit.

I also replaced their latex gloves with nitrile ones. They are a little better.

## Why It's Good For The Game
Division of on station jobs, and making the entire rest of medbay not obsolete. Roles, especially security ones, should not have unchecked access to do every job for themselves. See above for the oversights with the Security Medic, and how this PR addresses them.

## Changelog
:cl:
balance: Security Medic adjustment. Replaced the latex gloves with nitrile. Replaced the advanced first aid kit with a brute one. Removed extraneous skeleton crew access, and surgical normal access. Removed compact defib. Removed surgical bag from locker. Made actual doctors at least plausibly valuable at round start.

/:cl:


modular_skyrat/modules/sec_haul/code/security_medic/security_medic.dm
```diff
@@ -33,14 +33,14 @@
...
-	gloves = /obj/item/clothing/gloves/color/latex
+	gloves = /obj/item/clothing/gloves/color/latex/nitrile
 	shoes = /obj/item/clothing/shoes/combat/peacekeeper
 	glasses = /obj/item/clothing/glasses/hud/health/sunglasses
 	suit = /obj/item/clothing/suit/armor/vest/peacekeeper/security_medic
-	l_hand = /obj/item/storage/firstaid/advanced
+	l_hand = /obj/item/storage/firstaid/brute
 	head = /obj/item/clothing/head/beret/sec/peacekeeper/security_medic
 
-	backpack_contents = list(/obj/item/melee/classic_baton/telescopic, /obj/item/storage/box/gunset/security_medic, /obj/item/defibrillator/compact/loaded)
```
Gandalf2k15, line 43: Leave the defib in as it's part of their job to revive security in the field. Imagine they're on an away mission down under icebox, this will save them.

SpookyTheFox: No. That's the whole idea. 

They have to care enough about their job to print one and load a cell like everyone else if they want to bypass the role that is intended to have access to defibs. Do you know which medical roles walk off the shuttle with a defibrillator? None of them. None at all. The only compact defib at round start is in the CMO's locker, and he has to at least go into his office and retrieve it. Medical doctors have access to a limited number of full size defibs in their equipment lockers, and paramedics do not have access to those.

Why even have a research tree if we are going to spawn in the items it locks anyway? Why have any sense of round progression at all? Why not just start every round with all tech unlocked, and spawn every job with it's advanced tools? The station would only need one or two medical staff total if we just gave them every end game tool and access to all equipment.

Going further, why does the brig medic need to do revivals? Can they not cart people back to medbay? Medics should not be doing that job at all, but this PR leaves the door open for them to obtain a compact defib the same way anyone else who wants one can get one, by printing it from the medical department tech fab, after it has been researched.

Without a meaningful skills system, the best way to make players at least pretend to stick within the lanes of their job is to limit their access to other job's equipment, especially access to late game items at round start.

YakumoChen: they could also ask cargo nicely for the 5000 credit combat defib they can buy, if they really insist.

YakumoChen:
Please improve your changelog detailing actual changes and not some vague things in a PR that shouldn't have an add: changelog regardless (this PR doesn't directly impact Medical Doctor roles as implied in the current cl)
Preferably you should be only using "balance:" as you are neither removing nor adding actual content to the game.

SpookyTheFox:
> 
> 
> Please improve your changelog detailing actual changes and not some vague things in a PR that shouldn't have an add: changelog regardless (this PR doesn't directly impact Medical Doctor roles as implied in the current cl)
> Preferably you should be only using "balance:" as you are neither removing nor adding actual content to the game.

Done.

YakumoChen:
You should mention the balance changes are applied to Sec Medic specifically. The Changelog is what players see when they boot the game, and Changelogs don't directly reference PRs.

The line about doctors should still be removed as it isn't directly relevant to the PR (Medical Doctor is not changed in any way) and it's too vague.

modular_skyrat/modules/sec_haul/code/security_medic/security_medic.dm
SpookyTheFox: As an aside, I think that combat defib should not be for sale in cargo, and neither should the beam. Those are supposed to be an ERT and antag item. That combat defib should also get buffed so that it actually stunlocks people like it did on the old base.

YakumoChen: Ehhh, the beam is plenty expensive and I think the crates are CMO only unless you personal buy them, something doctors like to do with bounties. It keeps medbay productive (cargo mains LOVE this) on very slow shifts and gives them a goal to work towards.

The beam is also actually pretty tempered in its power level (it really doesn't heal that fast) and you have to be very careful with it because crossing the beams is a bad time.

YakumoChen:
An example of an appropriate changelog

:cl: SpookyTheFox
balance: Removed several accesses from Security Medic and rebalanced their round start gear to be closer to a paramedic.
/:cl:

Gandalf2k15 (approved):

PR #3927, opened 2021-03-06
[Modular] Allows you to craft blindfolds into fake blindfolds with wirecutters

Swiftfeather:
## About The Pull Request

Adds a recipe for this to the clothing tab crafting menu

I TRIED to do this much more complicated by making it so if you use wirecutters on the blindfold it'll cut it into the fake blindfold like hwen you wirecutter books, but couldn't get that shit working. This works tho. ezpz

## Why It's Good For The Game

I just think it's neat. THe item already exists, why not let people do it? I'll probably add to loadout later.

## Changelog
:cl:
add: You can use the crafting menu and wirecutters to make blindfolds into fake blindfolds
/:cl:


Swiftfeather:
Gonna PR this to tg first. They might take this.

Oscari-pog:
If someone uses this to roleplay a blind character without actually being blind I will spaz

Gandalf2k15 (approved):

PR #3935, opened 2021-03-06
[NON-MODULAR] Removes bananium drop from abandoned crates

NullDagaf:
## About The Pull Request

Title. Replaces it with space cash.

## Why It's Good For The Game

Bananium bad.

## Changelog
:cl:
balance: Changes the abandoned crates' 79/80th drop from bananium to space cash.
/:cl:

### DIFF
--- a/code/modules/mining/abandoned_crates.dm
+++ b/code/modules/mining/abandoned_crates.dm
@@ -188,7 +188,7 @@
 			/*
 			new /obj/item/stack/sheet/mineral/bananium(src, 10)
 			*/
-			new /obj/item/coin/titanium(src)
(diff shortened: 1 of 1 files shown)


Gandalf2k15:

From now on, please do that.

NullDagaf:
> After discussing these removals with high command we've come to the conclusion that simply having clown disabled in config is a much easier way. Due to the fact we'll be having a different more Action-oriented server soon. And also... we don't want our downstreams to have such harsh removals that they can't undo, say clown.
> 
> While I appreciate the effort, using configuration settings for things is the way we want to go.
> 
> From now on, please do that.

You can't config away a drop.

PR #3939, opened 2021-03-06
[NON-MODULAR] Changes the charged pyrite extract not to spawn bananium

NullDagaf:
## About The Pull Request

This PR changes the charged pyrite crossbreed so it doesn't spawn bananium anymore - instead, it creates a rainbow crayon.

## Why It's Good For The Game

Bananium is bad.

## Changelog
:cl:
balance: replaces bananium spawn with a rainbow crayon
/:cl:


Vizoplez:
Improper title conventions, lack of SKYRAT EDIT comments. Please rectify or close PR
-Intern dev

Gandalf2k15:

From now on, please do that.

PR #3957, opened 2021-03-06
[NON-MODULAR] Removes the "Bananium Shipment" station trait

NullDagaf:
## About The Pull Request

Title. Only removes the datum that rolls it.

## Why It's Good For The Game

Yes, we don't have clowns. But it still rolls and takes the station's trait without even doing anything. So why not just.. remove it?

## Changelog
:cl:
del: removes the bananium_shipment station trait datum
/:cl:


Gandalf2k15:

From now on, please do that.

PR #4029, opened 2021-03-09
Map reset and Blueshield fix

Gandalf2k15:
## About The Pull Request
Resets all maps and transfers over our changes.

Blueshield now gets captains office access, removes the blueshield access define.

## Changelog
:cl:
fix: All maps have been reset to TG standard
fix: Blueshield now has captains room access.
/:cl:


GoldenAlpharex:
Did your testing actually show that it fixed it? Because I've been looking into the code a bit and it's a little confusing why the Blueshield's ID just spawns basically without a proper trim, and how it's not a template in the ID console.

GoldenAlpharex:
Or rather, it's one, but it doesn't add any access to anything.

SpaceVampire:
Its worth knowing that the "Ship Pilot" or whatever also got broke by these changes.
While we don't run ShipStation very often, if you could fix it as well, that would be good.

_maps/icebox.json
```diff
@@ -3,7 +3,7 @@
 	"map_name": "Ice Box Station",
 	"map_path": "map_files/IceBoxStation",
 	"map_file": [
-		"IcemoonUnderground_Below_skyrat.dmm",
+		"IcemoonUnderground_Below.dmm",
```
KathrinBailey, line 6: Is this intentional?

KathrinBailey (approved):
May was well leave the _old maps in there and I'll clean it up later.

_maps/icebox.json
Gandalf2k15: Yeah it is, that map wasn't touched, if we wanna make it modular, we can.

PR #4032, opened 2021-03-09
[Mostly non-modular] Stuff to accomodate multiple servers talking to the same database, including local/global bans.

Useroth:
Don't testmerge until we confirm that the database and config changes are implemented.


Gandalf2k15 (approved):
I don't know much about the database changes, but the code changes are fine!

PR #241, opened 2020-08-07
Changeling horror form! [MODULAR] [READY]

Gandalf2k15:
## About The Pull Request
Adds an ability to changelings called the horror form. It's basically them dropping their flesh disguise for their true form, a horrifying looking thing from the thing.

image

Changeling adaption stats:
	chemical_cost = 75
	dna_cost = 8 //Tier 4
	req_dna = 10
	req_human = 1

The true changeling has a decent attack rate and can vent crawl. It is however quite slow. 

As a true changeling you have the ability to devour people which regenerates your health and damages them, making a lot of mess and noise!

## Why It's Good For The Game

These are genuinely quite fearsome and you should be scared of them. Adds a slight fear factor to the game.

## Known Issues
Upon death of a true_changeling that is a changeling, a runtime will be thrown. This needs fixed, however it's not gamebreaking and can stay as it is for now.

## Changelog
:cl:
add: Added a new ability for changelings; horror form
/:cl:

PR #4063, opened 2021-03-11
[MODULAR] Removes Changeling Horrorform

Swiftfeather:
## About The Pull Request

Reverts #241

Removes horrorform in its entirety.

## Why It's Good For The Game

Currently, Changelings are very powerful. Horror form is almost weaker than they are in normal ling combat form. With their increased chemical storage and generation, and additional DNA points, they have no reason to use horrorform any longer. Horrorform used to be necessary if a changeling really wanted to go loud and hard against security, but with our new modifications to both sec and lings, it has fallen useless, and was never really a great system to begin with.

## Changelog
:cl:
del: Removed changeling horrorform.
/:cl:


SpookyTheFox:
I see no reason to remove it. On the contrary, I would like to see it buffed. Horror form is usually a changeling's one and only chance against security's unnaturally high DPS guns. Remember that per server policy, security is not required to de-escalate with lings at all, so they need to retain this ability for use when cornered and facing certain death. If you want to remove this, work on changing security policy first, and seeing where else you can mechanically address what lings can do against multiple security officers, since they can't get a martial art, and all their other abilities are largely negated by both high population and Armadyne security changes.

Horror form is also something of a signature ability of changelings on Skyrat, and it would be a shame to loose it for that reason alone. On top of that, changeling has already lost hivemind, transformation sting, spread infestation, and even the entire team changeling game mode, our only not-traitor antag is feeling like a red headed step child right now.

Gandalf2k15:
Yea let's not remove this, we should rebalance it.

KathrinBailey:
Slapping testmerge only because I also would rather this be rebalanced.
No issues with TMing it until someone fixes it unless someone can throw up a rebalance PR within the next few days.
Seems fair since people are having huge issues with it at the moment.

PR #4115, opened 2021-03-13
[MODULAR] Tramstation Fixes

Cheshify:
## About The Pull Request

- Prison is properly decaled
- Engineering's delivery has a floor tile
- Armoury is properly kitted out

## Why It's Good For The Game

Tramstation fixes  "good"

## Changelog
:cl:
fix: Tramstation's armoury and prison is now properly retrofit and kitted out
/:cl:

### DIFF
--- a/_maps/map_files/tramstation/tramstation_skyrat.dmm
+++ b/_maps/map_files/tramstation/tramstation_skyrat.dmm
@@ -435,12 +435,12 @@
 /turf/open/floor/plating,
 /area/engineering/engine_smes)
 "aaT" = (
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
There's two wires missing from between the CE's Office's right door and Atmospheric's power grid, which means that most of the network is disconnected. If you could add that, that'd be great.

In general, it feels to me like there's not enough vents / scrubbers, the rooms are huge and there's like, only one set for most of them. I'm thinking mostly of engineering again, where each room, no matter how big, seems to only have a pair of vent/scrubber. It makes dealing with depressurization quite a pain in the ass.

PR #4116, opened 2021-03-13
[Modular] Cybersun Hotfix

unit0016:
## About The Pull Request
So it's come to my attention that cybersun's gotten shafted again due to updates. Here's how that's fixed:
Agent IDs are now the new, shitty versions.
Removed the machine frames that implied how to build a cloner. For obvious reasons.
Since the chameleon skillchip is deleted, and after a bit of internal debate - I've replaced it with a five roboticist skillchips. Because. Borgs.

## Why It's Good For The Game
Fix.

## Changelog
:cl:
fix: Cybersun Industries no longer implies the creation of technology you cannot create.
/:cl:

### DIFF
--- a/modular_skyrat/modules/mapping/_maps/RandomRuins/SpaceRuins/forgottenship_skyrat.dmm
+++ b/modular_skyrat/modules/mapping/_maps/RandomRuins/SpaceRuins/forgottenship_skyrat.dmm
@@ -416,7 +416,6 @@
 /turf/open/floor/carpet,
 /area/ruin/space/has_grav/syndicate_forgotten_ship)
 "bm" = (
-/obj/structure/frame/machine,
 /obj/structure/window/reinforced/spawner/east,
 /obj/effect/turf_decal/trimline/blue/filled/line{
 	dir = 5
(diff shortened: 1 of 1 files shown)


SpaceVampire:
I'm pretty sure they don't need skill-chips at all.
If they are supposed to start with one, it would make more sense to just edit the spawner for them to to include it.
Instead of placing down an item, that then needs another station to install it.

MimicFaux:
I think having the chips on the table is fine. Cybersun is all about their augments, and so having the extra step of building an implanting station is on theme.

SpaceVampire:
> I think having the chips on the table is fine. Cybersun is all about their augments, and so having the extra step of building an implanting station is on theme.

It just feels really unnecessary in the first place imo.
Its not like, a real balance issue or anything.
It just feels wiiierd. "Because borgs" isn't a great reason to have em.

Gandalf2k15 (approved):

unit0016:
> > I think having the chips on the table is fine. Cybersun is all about their augments, and so having the extra step of building an implanting station is on theme.
> 
> It just feels really unnecessary in the first place imo.
> Its not like, a real balance issue or anything.
> It just feels wiiierd. "Because borgs" isn't a great reason to have em.

Cybersun already has robotics access to do cyborg stuff. The roboticist skillchips are there so they don't accidentally link to the AI while disabling the camera, as was originally intended.

PR #4137, opened 2021-03-13
[Nonmodular]Disables Distributed Neuron blob strain

Vizoplez:
## About The Pull Request
Requested by staff somewhat, either a change or an outright disable during the time we figure out how to adjust it to be less gamer

## Why It's Good For The Game
DN is currently the strongest strain in the game. This is a fact. Every blob spore can be player-controlled; while not an issue on /tg/ - we get much higher instasnces of engine sabotage, welderbombing, general balance issues involving some of our maps (such as Tramstation); even against multiple ERTs and otherwise admin interference, it can still /easily/ win short of a literal Deathsquad with 12 people.

This is due to how it spreads. Distributed Neurons has a reagent that /literally changes whoever it hits/ if they're meeting the following pre-requisites

1) They're dead or unconcious
2) The blob has 5 resource points

This makes it extremely unbalanced; see - the thing is, the reagent also instantly kills anyone unconcious or in hardcrit, meaning if a blobbernaut or blob zombie were to hit someone, they get knocked unconcious from an explosion or go into crit, it instantly kills and zombifies them, allowing a new ghost, or the same person; to immediately take control of that new blob zombie.

There is no way to counter this against an experienced blob player, with the increased difficulty in making X-Ray weapons, beam rifles, ecetera; overall making it harder for players to counter this strain; it was decided that, after two rounds of similar occurances, that it would be removed for playabilities sake. It will likely be changed, and possibly PR'd upstream.

## Changelog
:cl:
del: Distributed Neurons is gone, for now. Good-bye zergo blob.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/blob_mobs.dm
+++ b/code/modules/antagonists/blob/blob_mobs.dm
@@ -141,7 +141,7 @@
 	humanize_pod(user)
 
 /mob/living/simple_animal/hostile/blob/blobspore/proc/humanize_pod(mob/user)
-	if((!overmind || istype(src, /mob/living/simple_animal/hostile/blob/blobspore/weak) && !is_zombie)
+	if((!overmind || istype(src, /mob/living/simple_animal/hostile/blob/blobspore/weak)) && !is_zombie)
 		return
 	if(key || stat)
 		return
--- a/code/modules/antagonists/blob/blob_mobs.dm
+++ b/code/modules/antagonists/blob/blob_mobs.dm
@@ -120,7 +120,7 @@
 	if(istype(linked_node))
 		factory = linked_node
 		factory.spores += src
-		if(linked_node.overmind && istype(linked_node.overmind.blobstrain, /datum/blobstrain/reagent/distributed_neurons) && !istype(src, /mob/living/simple_animal/hostile/blob/blobspore/weak))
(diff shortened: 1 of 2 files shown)


SpaceVampire:
Can you just roll back some of the blob changes in general?
Making it so that it always felt like such a total stalemate, with the stupid goal of 700 points, isn't fun.

SpaceVampire:
To add onto that. The blob grows so slowly, its never going to hit that 700 point total. All it does is make people yell for ERT, which then delays the whole thing, and it just keeps going on for ages. The blob was weak on the weak side before, sure, but having it be a simi-minor event that lets people smash something, is better then this horrible nightmare it is now.

Vizoplez:
thats what I get for phone editing

SpookyTheFox:
Absolutely disable DN. Despite there being antag policy specifying to not "bumrush the SM" using DN, administrative staff generally fail to react, such as the recent incident with tramstation. Also, the free player controlled mob spam is obnoxious.
I'd go further, and eliminate the majority of the most recent major blob buffs. Reflectives should be struck entirely, and factories are way over producing and can be built much closer together than they should be.

Vizoplez:
It compiles just fine, its chameleon breaking integration tests

hugmeorelse:
I can't reopen this, because the original repo is deleted.

PR #4222, opened 2021-03-17
Player panel improvements [be antag & character name]

Gandalf2k15:
## About The Pull Request

This adds a few creature comforts to the player panel in-game, namely the addition of your currently selected character name, and an option to disable/enable preround antag and midround antag.

Also adds OOC status to statpanel.

Examples:
image

image

image

## Changelog
:cl:
qol: You can now see your selected character name and toggle antag candidate status from the player panel.
/:cl:


FlamingLily:
But with this change, will anyone want to traitor cryo any% speedrun anymore?

SpaceVampire:
Had some reports of the "Be Antag" not actually working, if you want to just... double check?
It was during the dynamic test if that matters.
I think it also resets every shift, so having it save might be useful.

MLGTASTICa:
Now , all we need is the "be special"  button that lets us randomly roll traits and genetic mutations round-start

PR #4225, opened 2021-03-18
Changeling nerf + Horror form equilibrium 

Gandalf2k15:
## About The Pull Request

So it's obvious changelings are too powerful, it's time to tone them down a bit by doing the following:
They only have 100 chem charges now.
They recharge at 1.5 instead of 2.

Horror form:
We've toned this down a bit, the following has been changed:
Time to return to a human is now 5 minutes.
Max health is now 750, with a starting health of 500.
Melee lowered to 40.
Upon the death of one of these creatures, they will reach equilibrium and explode in a shower of gore and spray bone fragments everywhere, including the changeling within. This is death, real death. Choose this wisely.

https://streamable.com/p3wa1z

Also adds a bonebang grenade, a special nade that can be bought for 5 TC.

## Changelog
:cl:
balance: Changelings are no longer god.
balance: Horror forms will now gib on death, watch out if you go this route!
/:cl:


Arctaisia:
I like the stakes of going horror form, means that lings will have to be strategic when they use it, like when they're backed into a corner!

SpookyTheFox:
I don't support this at all. This does not offer a credible solution for one to many combat, that is balanced against security with red alert guns. Rather than nerfing this, try taking administrative action against the players that abuse it. And five minutes is worse than it was when introduced.

PR #4232, opened 2021-03-18
Nerfs Double Bladed Energy Sword TC cost

BrotherHangyul:
## About The Pull Request

Nerfs the cost of the DE-sword in the traitor uplink from 16 to 34 (about 75% of your TC), making it on-par with the cost of the formerly great CQC+.

EDIT: 16 to 25 at request of TF4.

## Why It's Good For The Game

Stops people on going from "epic gamer" rampages and forces people to be more resourceful with their TC.

Disclaimer: Making this PR on behalf of IsaacTheSharkWolf#3811, I don't really believe in the points I make here.

## Changelog
:cl:
balance: The Syndicate has upped the cost of the double bladed energy sword due to production costs.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -412,7 +412,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			all energy projectiles, but requires two hands to wield."
 	item = /obj/item/dualsaber
 	player_minimum = 25
-	cost = 34
+	cost = 25
 	exclude_modes = list(/datum/game_mode/nuclear/clown_ops)
 
 /datum/uplink_item/dangerous/doublesword/get_discount()
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -412,7 +412,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 			all energy projectiles, but requires two hands to wield."
 	item = /obj/item/dualsaber
 	player_minimum = 25
-	cost = 16
+	cost = 34
 	exclude_modes = list(/datum/game_mode/nuclear/clown_ops)
 
 /datum/uplink_item/dangerous/doublesword/get_discount()


SpaceVampire:
I mean, if you don't don't really believe in the points you probably shouldn't do it.
Even if it did need some kind of balance, I think 34 TC is clearly insane.

BrotherHangyul:
34 TC is the (rounded upwards) 75% equivalent of 45 TC.

Vizoplez:
good change

GoldenAlpharex:
But how much is a normal e-sword?

YakumoChen:
Double sword doesn't even deflect the vast majority of sec weaponry

BrotherHangyul:
> But how much is a normal e-sword?

8 TC.

gamermomentum:
Part of the reason CQC+ was so batshit was that you literally could do nothing about it without pacifying them or hard disabling their hands. 

DEsword, on the other hand, requires both your hands to use and can be dropped. 
And it reflects basically none of sec's current weapons, once more unlike CQC+, it CAN block bullets but it's not a 100% chance. 
34 TC is super crazy for this item because it's only megastrong offensively. Try something like 25-30 if you really wanna nerf it that hard. 

Also I say all this as someone who mains sec and hasn't played antag here in ages.

tf-4:
34 is wacko. 25 would be better. I agree with a nerf, just not *this* nerf

BrotherHangyul:
wait what 

i mean sure i'll lower it to 25?

chefferz:
you do realize you can put two normal eswords together to form a DEsword right...?

chefferz:
Also, the DEsword is stupid easy to counter, All it takes is a couple officers and you're dropped. This sounds like a salt PR, to be honest.

KathrinBailey:
Look at lowering total TC if you think traitors have access to too much. That's a better direction.
Considering you can't deflect security's projectiles because they're bullets instead of laters, it doesn't need a TC cost increase at all.
Also, consider you're going to be one of maybe two antagonists who actually antagonize against seven seccies and an AI who will report you to command and security for standing in a room with your PDA out for too long.
Also, consider we have immensely buffed health compared to TG.

PR #4243, opened 2021-03-19
Coronal Mass Ejection no longer starts fires

YakumoChen:
## About The Pull Request
see title. Plasma has been removed from CMEs. It just makes hot oxygen now.
CMEs no longer shock you and set you on fire when you neutralize them
removes the announce every time someone neutralizes a CME since it's obnoxious

## Why It's Good For The Game
Fuck this event.
Everything about this event is designed to punish you. 
CME spawns? Have a fire.
Ignore CME? Have a big old EMP
Remove CME? _Have another fire._ Enjoy being on fire, idiot, because **FUCK YOU**

## Changelog
:cl:
qol: CME no longer creates fires, especially when you're trying to neutralize them.
qol: less CME earmurder
/:cl:

PR #4264, opened 2021-03-20
Anomaly neutralizers don't murder you with CMEs

YakumoChen:
## About The Pull Request

title

## Why It's Good For The Game
see #4243

## Changelog
:cl:
qol: Anomaly neutralizers don't kill you anymore
/:cl:


Gandalf2k15:
No. We literally just nerfed them, deal with the consequences of touching a literal ball of fire. Stop making PR's right after the previous one was closed.

Vizoplez:
based

hugmeorelse:
> No. We literally just nerfed them, deal with the consequences of touching a literal ball of fire. Stop making PR's right after the previous one was closed.

I think this should be reconsidered. Please don't close PRs like that without allowing others to place their input like my own.

YakumoChen:
Neutralizing a CME gives you 5 fire stacks AND stuns you, which is enough to give you extremely significant damage by the time you can get up or outright crit you if you don't have a fire extinguisher on hand.

As I said before, we should not have dangerous events completely punish you for daring to interfere with them, or else nobody will want to prevent high amounts of station damage from being dealt

There is also NO WARNING for neutralizers punishing you, and precedent for neutralizers on actual anomalies is that doing so is safe outside of the direct damage anomalies can deal to you normally.

It's extremely poor design and doesn't contribute positively to the round to make every CME neutralization be lethally deadly.

PR #4275, opened 2021-03-20
[Non-modular]Examining a syndicate bomb no longer tells you how much it has left before it blows

MLGTASTICa:
## About The Pull Request
Makes it so syndicate bombs don't display their timer on examine anymore
## Why It's Good For The Game
If you know how much a bomb has left , there is no element of urgency or fear. If the player doesn't know how much the bomb has left , security will be encouraged to get a bomb suit as fast and possible and defuse it , and all non-security jobs will run.
Also makes it viable for antags to set bombs with 60000 seconds timer and complete their stuff while everyone is panicking because they don't know how much it has left until it blows.
## Changelog
:cl:
balance: Syndicate minibombs no longer show how much time they have left until they blow.
/:cl:

### DIFF
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -63,7 +63,7 @@
 			if(0 to 5)
 				volume = 50
 				for(var/i,i<3,i++)
-					addtimer(CALLBACK(src, .proc/play_fearsome_ping), i*10)
+					addtimer(CALLBACK(src, .proc/play_fearsome_ping), i*5)
 			if(5 to 10)
 				volume = 40
 			if(10 to 15)
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -98,7 +98,7 @@
 
 /obj/machinery/syndicatebomb/examine(mob/user)
 	. = ..()
-	. += {"A digital display on it reads "[seconds_remaining()]"."}
+	// . += {"A digital display on it reads "[seconds_remaining()]"."} Skyrat change - commented out to make people fear it more.
 
 /obj/machinery/syndicatebomb/update_icon_state()
 	icon_state = "[initial(icon_state)][active ? "-active" : "-inactive"][open_panel ? "-wires" : ""]"
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -62,6 +62,8 @@
 		switch(seconds_remaining())
 			if(0 to 5)
 				volume = 50
+				for(var/i,i<3,i++)
+					addtimer(CALLBACK(src, .proc/play_fearsome_ping), i*10)
 			if(5 to 10)
 				volume = 40
 			if(10 to 15)
@@ -80,6 +82,9 @@
 		timer_set = initial(timer_set)
 		update_appearance()
 		try_detonate(TRUE)
+
+/obj/machinery/syndicatebomb/proc/play_fearsome_ping()
+	playsound(loc, beepsound, 80, FALSE)
 
 /obj/machinery/syndicatebomb/Initialize()
 	. = ..()
--- a/code/game/machinery/syndicatebomb.dm
+++ b/code/game/machinery/syndicatebomb.dm
@@ -98,7 +98,7 @@
 
 /obj/machinery/syndicatebomb/examine(mob/user)
 	. = ..()
-	// . += {"A digital display on it reads "[seconds_remaining()]"."} Skyrat change - commented out to make people fear it more.
+	// . += {"A digital display on it reads "[seconds_remaining()]"."} SKYRAT EDIT : - commented out to make people fear it more.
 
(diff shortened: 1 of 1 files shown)


Vizoplez:
PR this upstream.

Yawet330:
Furthermore; this breaks our SKYRAT EDIT policy. Please correct.

code/game/machinery/syndicatebomb.dm
```diff
@@ -98,7 +98,7 @@
 
 /obj/machinery/syndicatebomb/examine(mob/user)
 	. = ..()
-	. += {"A digital display on it reads "[seconds_remaining()]"."}
+	// . += {"A digital display on it reads "[seconds_remaining()]"."} Skyrat change - commented out to make people fear it more.
```
Vizoplez: should be SKYRAT EDIT: - Commented out for fear factor

SpookyTheFox:
Seeing how much time there is left, and gaming it, is intentional. Ever watch an action movie with a bomb scene? It's part of the drama and excitement. There are other bomb types that you can't see the timers for, like C4/X4, toxins bomb, etc. The syndicate one has an iconic design and the timer is part of the experience.

There are already multiple in game means of making the timer not correspond with actual time until it goes off, such as attaching a secondary timer, a signaler, another explosive, shooting it, etcetera.

code/game/machinery/syndicatebomb.dm
Vizoplez: or along those lines; it just needs to include 'SKYRAT EDIT' so we can ctrl F it

MLGTASTICa:
> Seeing how much time there is left, and gaming it, is intentional. Ever watch an action movie with a bomb scene? It's part of the drama and excitement. There are other bomb types that you can't see the timers for, like C4/X4, toxins bomb, etc. The syndicate one has an iconic design and the timer is part of the experience.
> 
> There are already multiple in game means of making the timer not correspond with actual time until it goes off, such as attaching a secondary timer, a signaler, another explosive, shooting it, etcetera.

knowing when a bomb is going to blow doesn't inspire a lot of fear.

code/game/machinery/syndicatebomb.dm
MLGTASTICa: Done

GoldenAlpharex:
> knowing when a bomb is going to blow doesn't inspire a lot of fear.

Yes it does? Makes you freak out because you know you don't have the time to defuse it :^)

Gandalf2k15:
I'm a big fan of this. How about when the bomb has less than 10 seconds remaining it gives some sort of signal to show this? So people can run like fuck.

NullDagaf:
> I'm a big fan of this. How about when the bomb has less than 10 seconds remaining it gives some sort of signal to show this? So people can run like fuck.

Doesn't sound very realistic, nor is there a real IC purpose for this. Besides, it'd just lead to the same problem of people not caring, until they hear/see that signal, as you can very much run away in these ten seconds.

With that in mind, if it made a noise right before it explodes, sort of like the nuke, that would be good. Maybe one second, or a bit shorter.

Gandalf2k15:
A 3 second beep beep beep would be good.

YakumoChen:
> > I'm a big fan of this. How about when the bomb has less than 10 seconds remaining it gives some sort of signal to show this? So people can run like fuck.
> 
> Doesn't sound very realistic, nor is there a real IC purpose for this. Besides, it'd just lead to the same problem of people not caring, until they hear/see that signal, as you can very much run away in these ten seconds.
> 
> With that in mind, if it made a noise right before it explodes, sort of like the nuke, that would be good. Maybe one second, or a bit shorter.

Knowing when to run from the bomb is better gameplay however. Dying to a syndicate bomb is more or less full round removal and we should make at least a little mitigation for that

NullDagaf:
> Knowing when to run from the bomb is better gameplay however. Dying to a syndicate bomb is more or less full round removal and we should make at least a little mitigation for that

Okay, disregarding for a second the fact that round removal is not a bad thing.

Have you considered that people **should not be near a bomb in the first place?** There shouldn't be any "mitigation" against not being a dumbass - if you're dense enough to stare at a bomb, don't be surprised it blows up in your face.
Also, that's why I said "one second or less" - yes, it will telegraph it's explosion, but you won't have enough time to run away, or do much of anything. All that will do is give a little bit of flavor, **_and nothing else._**

Gandalf2k15:
> > > I'm a big fan of this. How about when the bomb has less than 10 seconds remaining it gives some sort of signal to show this? So people can run like fuck.
> > 
> > 
> > Doesn't sound very realistic, nor is there a real IC purpose for this. Besides, it'd just lead to the same problem of people not caring, until they hear/see that signal, as you can very much run away in these ten seconds.
> > With that in mind, if it made a noise right before it explodes, sort of like the nuke, that would be good. Maybe one second, or a bit shorter.
> 
> Knowing when to run from the bomb is better gameplay however. Dying to a syndicate bomb is more or less full round removal and we should make at least a little mitigation for that

We are not catering for the dumb here

Gandalf2k15 (approved):

PR #4324, opened 2021-03-22
[MIRROR] Thermomachine rework

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/57831
--------------------

## About The Pull Request
Freezers/heaters now works as heat pumps, that moves heat from one source to another, there are now 2 ports on the machine, one green (that will work as the current one, where you connect the pipes and cool down/heat up the contents) and an optional red port that allows the heat movement.
IF no pipe is connected to the red port, or it doesn't have any gases in it, the thermomachine will use the air around it to move the heat around, cooling or heating up the air depending on the settings
Update: Freezers in map are now connected to the waste loop and the CO2 is getting filtered back into waste to be used as heat collector too, atmos now should manage this loop more actively.
TO DO
- [x] Better checks for heat movement
- [x] Maybe heaters don't need the red port connected (they could use much more power to work tho, gimme your thoughts)

## Why It's Good For The Game
freezers now don't delete heat

## Changelog
:cl:
balance: READ ME thermomachines now have 2 ports, the green one is used as you normally do, by cooling/heating the gases connected to it, the red port is optional but it helps move the heat from the green port to the red one (and vice-versa), if no pipe is connected to the red port or there is no gas in it, the heat will be moved from the enviroment; 
balance: ALL the station freezers are now connected to the waste loop, where the CO2 gets collected and used to store that heat (atmos should now manage the waste pipenet more actively); 
balance: the best way to now use freezers is to connect their red port to the waste system (there is at least one pipe in every room)
balance: Rotate the machine without emptying the content will release the gas onto the atmosphere
/:cl:


Sethnov:
why this not merged yet?

PR #4344, opened 2021-03-23
[NONMODULAR] Perhaps maybe makes it so that we might not always likely roll ice box station, possibly.

Swiftfeather:
## About The Pull Request

This disables preference map voting, so those who are not present, and not actively participating in the vote, will not have their map preferences taken into account for map votes.

This... should work, it's in the config.

## Why It's Good For The Game

Ice box station 5 times in a row, anyone?

## Changelog
:cl:
fix: Perhaps maybe fixes icebox rolling endlessly
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -322,8 +322,6 @@ GLOBAL_LIST_EMPTY(the_station_areas)
 	var/players = GLOB.clients.len
 	var/list/mapvotes = list()
 	//count votes
-	//SKYRAT EDIT REMOVAL BEGIN - No autovote
-	/*
 	var/pmv = CONFIG_GET(flag/preference_map_voting)
 	if(pmv)
 		for (var/client/c in GLOB.clients)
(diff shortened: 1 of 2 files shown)


DuffCreeper:
I believe it's a config option in the config.txt at due to this
```
## Map voting
## Allows players to vote with their preffered map setting
## When it's set to zero, the map will be randomly picked each round
PREFERENCE_MAP_VOTING 1
```
Also think icebox is a default map hence why its always voted

MLGTASTICa:
> I believe it's a config option in the config.txt at due to this
> 
> ```
> ## Map voting
> ## Allows players to vote with their preffered map setting
> ## When it's set to zero, the map will be randomly picked each round
> PREFERENCE_MAP_VOTING 1
> ```
> 
> Also think icebox is a default map hence why its always voted

If you check the code yourself it has no config check for mapvote , its also specified it  100% disables map voting , which is not what he aims for.

Swiftfeather:
I have a config edit where there shouldn't be one 👎

PR #4346, opened 2021-03-23
[NONMODULAR] Perhaps maybe makes it so that we might not always likely roll ice box station, possibly, AGAIN

Swiftfeather:
## About The Pull Request

This disables preference map voting, so those who are not present, and not actively participating in the vote, will not have their map preferences taken into account for map votes.

I tested and map vote still worked, and didn't auto vote for my character's preference.

## Why It's Good For The Game

Those who are not present or participating in a vote no longer sway votes.

## Changelog
:cl:
fix: Perhaps maybe fixes icebox rolling endlessly
/:cl:


Gandalf2k15:
Wait no this is literally a config

Gandalf2k15:
We'll set this in config.

PR #4347, opened 2021-03-23
[NON-MODULAR] Gives Engie Borgs Light Replacers

DuffCreeper:
## About The Pull Request

Honestly dunno why they don't get light replacers

## Why It's Good For The Game

Allows engie borgs to replace broken or build lights without wanting to kill themselves via lack of hands

## Changelog
:cl:
add: Added light replacers to the engie borg
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -255,7 +255,8 @@
 		/obj/item/stack/sheet/rglass/cyborg,
 		/obj/item/stack/rods/cyborg,
 		/obj/item/stack/tile/iron,
-		/obj/item/stack/cable_coil)
+		/obj/item/stack/cable_coil,
+		/obj/item/lightreplacer/cyborg) // Skyrat Edit - Surprised Engie borgs don't get these
 	radio_channels = list(RADIO_CHANNEL_ENGINEERING)
 	emag_modules = list(/obj/item/borg/stun)
 	cyborg_base_icon = "engineer"


GoldenAlpharex:
They didn't have one because that's the point of Janiborgs.

Engiborgs shouldn't be fucking off to make rooms in maints or space, and being able to put down lights, as simple as it sounds, just makes it so they have to depend on others even less. They already don't care about radiation, breathing, fire, pressure, and with this, they mostly don't have to rely on others to do basically anything if they want to just go and do their own thing, which shouldn't be something borgs do.

Cyborgs were meant to be unable to entirely replicate a job, to give the real players an actual advantage over them, because otherwise you're just a better version of that job who's only drawbacks are that you have to charge and are weak to flashes.

MLGTASTICa:
> They didn't have one because that's the point of Janiborgs.
> 
> Engiborgs shouldn't be fucking off to make rooms in maints or space, and being able to put down lights, as simple as it sounds, just makes it so they have to depend on others even less. They already don't care about radiation, breathing, fire, pressure, and with this, they mostly don't have to rely on others to do basically anything if they want to just go and do their own thing, which shouldn't be something borgs do.
> 
> Cyborgs were meant to be unable to entirely replicate a job, to give the real players an actual advantage over them, because otherwise you're just a better version of that job who's only drawbacks are that you have to charge and are weak to flashes.

Engiborgs do care about radiation , and we do not follow TG principes of balance. Cyborgs should be well equipped in doing their jobs.

SpaceVampire:
I do think that this really falls under Janiborgs, given that they already have the tool.

KathrinBailey (approved):
yeah requiring a janiborg isn't really smart or sensible design.
we don't ban engineers for taking the job of janitors when they replace light bulbs.
people still call the fucking CE to replace lights, so...

PR #4348, opened 2021-03-23
Fixes for integrated circuit printer

hecksadecimal:
## About The Pull Request
There was a json parsing error happening because `proc/stripped_multiline_input` html encodes text passed into it as part of its sanitization steps. This was the cause behind the inability to clone circuits. Additionally there was a missing newline in the UI that I put into place while I was at it.

## Why It's Good For The Game
Without functional circuit cloning creating interesting and complex circuits is severely discouraged. 
It can take nearly an entire round to set up something truly unique. 

## Changelog
:cl: 0xEFF
fix: The integrated circuit printer can now clone circuits as originally intended.
fix: Integrated circuit printer UI given a linebreak where it was needed.
/:cl:

### DIFF
--- a/modular_skyrat/modules/integrated_circuits/code/integrated_electronics/core/printer.dm
+++ b/modular_skyrat/modules/integrated_circuits/code/integrated_electronics/core/printer.dm
@@ -167,7 +167,7 @@
 	else
 		HTML += " Load Program"
 	if(!program)
-		HTML += " [fast_clone ? "Print" : "Begin Printing"] Assembly"
+		HTML += " [fast_clone ? "Print" : "Begin Printing"] Assembly<br>"
 	else if(cloning)
 		HTML += " <A href='?src=[REF(src)];print=cancel'>Cancel Print</a>"
 	else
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
While you're at it, think you can fix the default iron value for parts, so that you can recycle them as it's meant to be? You can't put the parts back into the printer, when it says I think in the description that you should be able to.

hecksadecimal:
> 
> 
> While you're at it, think you can fix the default iron value for parts, so that you can recycle them as it's meant to be? You can't put the parts back into the printer, when it says I think in the description that you should be able to.

Yeah, I'll look into that and submit another PR @GoldenAlpharex

ghost:
@hecksadecimal it works 99%, but refuses to copy some punctuation in text strings, if you think its possible to fix that (specifically exclamation points)

hecksadecimal:
> 
> 
> @hecksadecimal it works 99%, but refuses to copy some punctuation in text strings, if you think its possible to fix that (specifically exclamation points)
@OrionTheFox 

Might have to bypass filtering entirely if that's the case. I was hoping it wouldn't be necessary, but I guess it's not a problem if the input isn't ever displayed anywhere, only parsed.

I'll include it in my next batch of fixes.

PR #4403, opened 2021-03-25
Nightstick batong upgrade + stunsticks(in cargo)

Gandalf2k15:
## About The Pull Request

So we've recently removed stun batongs, which is of course something that was intended from the beginning, but I didn't realize security was constantly buying them. 

This simply renames the batong and removes it's brute damage as well as cutting the cooldown in half.

Adds the stun stick(TBA an upgrade kit) which is a peacekeeper version of the stun baton.

## Changelog
:cl:
balance: The standard security baton has been buffed and renamed. They also no longer deal lethal damage when you whap people with them.
add: A new stunstick is available from cargo.
/:cl:


unit0016:
where

weheretis
wherr is charge
batong

dominuslucian:
thank god

RatFromTheJungle:
75% of the original cooldown would be ideal, with 50% conga lining to baton rush will be an easy choice to make the second you can print shields, due to practically no antag kit being able to stand a chance against that

SilvernKell:
cooldown change pretty strong but thANK YOU for removing the brute damage. That really didn't work out imo we got harmbaton for that and it really screwed over borgs. 
Thank you

unit0016:
do remember to make a variant for the boomerang stunbaton too, as it's still got the batong stats

PR #4414, opened 2021-03-26
[SEMI-MODULAR] Reverts the fire alarm sound to default and stops it looping

DuffCreeper:
## About The Pull Request

The fire alarm will use the default sound that TG uses and won't loop over an hour if someone doesn't fix the reason it's shitting itself

## Why It's Good For The Game

Stops people losing their mind and makes hearing the same sound less, tedious since it doesn't loop for the 100th time

## Changelog
:cl:
add: Fire alarms now no longer loop their sound to save people from headaches
/:cl:


zeratino:
👌

FlamingLily:
Honestly, I don't agree with this change, I think the looping and annoying is good cause it means people are incentivised to both avoid and fix the assumed breach.

However, I do think the temperature that triggers the fire alarm should be lowered, so that it only triggers when not even a winter coat can protect you from the cold.

Gandalf2k15:
Do this by lowering trigger thresholds instead :P

Gandalf2k15:
See above statement, reopen when done 👍

PR #4445, opened 2021-03-27
[MIRROR] [Ready] Fermichem part 2.3 Adds a new reagent: Eigenstasium

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/56918
--------------------

## About The Pull Request

Adds Eigenstasium, a chemical loosely themed around eigenstates.

#### Normal effects:
If the purity of the reagent is over 90%, it will teleport the player to the creation location on consumption (ingestion only)
Next, it will save the location you took the reagent, and create a blue eigenstate version of your character to signal where you will return to.
When the reagent leaves a patients bloodstream, it will teleport the player back to the consumption location.
Dying does not stop processing, you will still return to the consumption location even if dead. (though using it on a corpse, or OD/addiction effects won't proc while dead)

#### Overdose effects:
Causes the stability of your character to split - Teleporting you to a random location close to where you are every second. Afterwards applys a status effect that will slowly go through the following stages:
Stage 1 - warns you and makes you hungry
Stage 2 - teleports 1 item on you at random to a random nearby location slowly.
Stage 3 - teleports in alternative reality versions of yourself who yell at you.
Stage 4 - replaces your character with an alternative reality version of yourself.

At stage 4 the status is removed, or the status can be reversed (but not removed) by taking stabilising agent, bluespace or eigenstatium. Taking 20u of stablising agent whenever you you feel destablised will be enough to hold it off for 10 or so minutes.

#### Turf effects:
Spraying this on multiple lockers on the same tile links the lockers together, creating eigenlockers. See below:

Spraying:

https://user-images.githubusercontent.com/33956696/107873080-7c5d1180-6ea7-11eb-9612-9ca32c640564.mp4

Using:

https://user-images.githubusercontent.com/33956696/107875868-32c9f200-6eba-11eb-8df8-27defd3a8629.mp4

#### Impure effects:
The associated impure reagent will cause the players items in their hands to swap. (Should work for multiple hands if that ever happens)

### Reaction effects:
The reaction has a few tricks to it, but I'll let you figure them out.

#### Overheat:
If the reaction is overheated, it will teleport items around the beaker randomly

#### Below 40% purity:
This will teleport mobs around randomly instead

### Other stuff:
Fixes up all the procs to have the same inputs across the reaction specific procs so it's less confusing and less runtime causing.

## Why It's Good For The Game

Adds an interesting new reagent that uses a lot of the new mechanics to show what sorts of things you can do with it. Also provides a challenge to those who have a good grasp on the system. Plus eigenlockers are really fun.

## Changelog
:cl:
add: Adds eigenstasium, a new reagent that uses the new mechanics in depth!
fix: Fixed a few things to do with reactions on the backend
refactor: refactors status effects to have a var for processing speed
/:cl:


GoldenAlpharex:
That looks sick as fuck.

PR #4446, opened 2021-03-27
Brings TC back down to 30

Azarak:
## About The Pull Request
I disagree with it being any higher, especially with the state of our community and the amount of antags we're having. The point of higher amount of TC was so traitors can still afford a firearm while getting stuff for their gimmicks, or so they extra TC to fallback on in case they get stuff confiscated. 45 TC allows people to turn themselves into ultimate killing machines, and it's quite tempting to do so, so say no more

## Changelog
:cl:
balance: Traitors now get 30 TC in the uplink
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -78,7 +78,7 @@ GLOBAL_LIST_INIT(heretic_start_knowledge,list(/datum/eldritch_knowledge/spell/ba
 
 /// How many telecrystals a normal traitor starts with
 //#define TELECRYSTALS_DEFAULT 20 //ORIGINAL
-#define TELECRYSTALS_DEFAULT 45 //SKYRAT EDIT CHANGE - ORIGINAL: 20
+#define TELECRYSTALS_DEFAULT 30 //SKYRAT EDIT CHANGE - ORIGINAL: 20
 /// How many telecrystals mapper/admin only "precharged" uplink implant
 #define TELECRYSTALS_PRELOADED_IMPLANT 10
 /// The normal cost of an uplink implant; used for calcuating how many


gonenoculer5:
Someone else is gonna do it so i'll say it: _Salt PR._

Weh.

Azarak:
> 
> 
> Someone else is gonna do it so i'll say it: _Salt PR._
> 
> Weh.

Didn't even participate in the round, and this is a change I proposed for some time already. Keep it constructive

DiscountDirtyDan:
> Someone else is gonna do it so i'll say it: _Salt PR._
> 
> Weh.

Sometimes you need salt to get things done

FlamingLily:
> 
> 
> Someone else is gonna do it so i'll say it: _Salt PR._
> 
> Weh.

That doesn't necessarily make it a bad PR, keep criticism constructive

StarHorder:
absolutely needed. I want my roleplay server back.

SpookyTheFox:
Keep in mind that this breaks some of the packs that are supposed to be a specific fraction of available TC, or purchased or excluded based on what was already bought. Several items that were already adjusted need to be re-balanced again.

ideally, TC would scale with antagonist objectives, expected adversaries, number of antagonists, et cetera. There is no good way to do this automatically however, and manual intervention is undesirable.

From what I have seen, the higher TC count has been used as intended for gimmicks, loadouts, and enough firepower to not instantly die to the first two officers to rush someone while typing, so I would say the 45TC has thus far worked as intended. I fear repealing it will likely unbalance interactions again, and lead to more antagonists using the tools available in their departments such as mechs, bombs, death plants, and whatnot rather than the much more easily balanced traitor uplink items.

Also, this proposal is coming at the same time as adding yet _more_ guns to security, a group of players not renowned for their self control. Sad to say it, but I don't think this is the right time to reduce the base stat of skyrat's most common antagonist.

gonenoculer5:
> Keep in mind that this breaks some of the packs that are supposed to be a specific fraction of available TC, or purchased or excluded based on what was already bought. Several items that were already adjusted need to be re-balanced again.
> 
> ideally, TC would scale with antagonist objectives, expected adversaries, number of antagonists, et cetera. There is no good way to do this automatically however, and manual intervention is undesirable.
> 
> From what I have seen, the higher TC count has been used as intended for gimmicks, loadouts, and enough firepower to not instantly die to the first two officers to rush someone while typing, so I would say the 45TC has thus far worked as intended. I fear repealing it will likely unbalance interactions again, and lead to more antagonists using the tools available in their departments such as mechs, bombs, death plants, and whatnot rather than the much more easily balanced traitor uplink items.
> 
> Also, this proposal is coming at the same time as adding yet _more_ guns to security, a group of players not renowned for their self control. Sad to say it, but I don't think this is the right time to reduce the base stat of skyrat's most common antagonist.

I actually pretty much entirely agree with this. 45 TC is fine, imo, especially when sec is about to lose non-lethals. 35-45 TC imo is the ideal, since it gives antags plenty to work with while also leaving room for big gimmicks like it was intended to. Someone could easily get Contractor and do simple things like bribe people to get contracts done, and easily rack up 40+ TC through contracts, without much effort at all. 

Also @Azarak , this PR needs to change existing bundles to be lower priced, especially the robohand bundle since its 45 TC and would be otherwise unobtainable at roundstart like it was intended to be.

KathrinBailey:
The CQC++ Johnny Desrador robofuck arm still being expensive is fine if we take the policy approach of "if you make ambitious ambitions, you can get more TC than the fucker who wastes their antag on breaking two windows."

KathrinBailey (approved):
Thankfully the bane of two seccies to each antag, regardless of if that antag ever intends to do anything like (in reality, one or two loud antags to eight seccies) is over.
I'd like going forward for there to be a design document regarding the direction of what goes into the uplink and what doesn't.

Oscari-pog:
> The CQC++ Johnny Desrador robofuck arm still being expensive is fine if we take the policy approach of "if you make ambitious ambitions, you can get more TC than the fucker who wastes their antag on breaking two windows."

Not sure why dissing Johnny was needed in that one

YakumoChen:
I have been saying for awhile we need less antag power and more antagonists and antagonist teamwork. Good change.

tobjv:
We'll just have to test it and see the results.

Gandalf2k15:
Do not merge until kits have been updated, and respective uplink items.

Gandalf2k15:
I'm gonna do this myself and lower it to 35.

PR #4474, opened 2021-03-28
[NON-MODULAR] Disables Space Adaptation

SpookyTheFox:
## About The Pull Request
Speculative fix and formal proposal to remove the space adaptation mutation. Disables it in the same way as x-ray vision, laser eyes, etc.

## Why It's Good For The Game
Just like x-ray/thermal, there's been many many instances of non-antagonists, especially security, getting this mutation and then using it to their advantage. The ability to operate in space is something that is supposed to be carefully controlled and limited. Traitors and changelings pay full price for space-proofing, the entire security department should not get it for free. There's no sense in such a large portion of the station taking an ability that just removes the whole "space" aspect of being on a space station, and doing it nearly every round. In addition, space adaptation totally negates several effects it should not, such as pressure changes from not-space, and almost completely counteracts chemicals like Frost Oil and Hercuri. Like most of the other genetics goodies, it was never intended to be so widely given out, and it exists to benefit unassisted antagonists such as revs and blood brothers, who are reliant on their department's late game items, not an uplink. We do not have those game modes, and don't need the items and abilities intended for them.

## Changelog
:cl:
config: Locked space adaptation.
/:cl:

PR #4482, opened 2021-03-28
[NON-MODULAR] Disables Self-Respiration

SpaceVampire:
## About The Pull Request

Removes Self-Respiration from the game, by making it in a tier of its own that is not accessible. This allows for the possibility of balancing it in a future PR, but I think just removing from the pool is the best way at the moment.

## Why It's Good For The Game

There's been many many instances of non-antagonists, especially security, getting this symptom and then using it to their advantage. The ability to breathe is something that is supposed to be carefully controlled and limited. This is why you only have limited internals, the entire security department should not get it for free. There's no sense in such a large portion of the station taking an ability that just removes the whole "breathing" aspect of being on a space station, and doing it nearly every round. In addition, Self-Respiration totally negates several effects it should not, such as having to worry about most blood-loss, and almost completely counteracts chemicals like Lexorin while making stuff like Salbutamol useless. This is a RP server after all, there is no reason to take a dangerous virus when internals exist.

## Changelog
:cl:
del: Self-Respiration
/:cl:


SarmentiCampbell:
goddamnit, don't encourage them

Sigmar-alkahest:
really? fucking really?

tf-4:
Good PR, nobody can self respirate in real life so doing it in game is pretty LRP

It failed about every check but that's fine we can handle that later on.

GoldenAlpharex:
Here we go again with removal PRs.

Gandalf2k15:
If you can do this the proper way we can consider it.

GoldenAlpharex:
Also, please, if you think something is not balanced for regular gameplay, stop involving security and the "non-antagonist" crew argument because it's just incredibly irritating that everyone can just justify it as "boohoo crew is getting too strong, now security can kick my ass easier"

GoldenAlpharex:
> If you can do this the proper way we can consider it. But right now we're better off moving genetics to science.

We're talking about virology here.

Gandalf2k15:
> > If you can do this the proper way we can consider it. But right now we're better off moving genetics to science.
> 
> We're talking about virology here.

Dumb, wrong PR lmao

GoldenAlpharex:
Also, I'd like to point out, if breathing was so important to the game, then why are we having like four races that don't need to do it?

YakumoChen:
Ah yes here comes the slippery slope.

Next up: Remove nanites because they allow passive healing
remove viro because of beneficial viruses
remove medibots because they don't require strong medbay interaction

where do we draw the line?

tf-4:
> 
> 
> Also, I'd like to point out, if breathing was so important to the game, then why are we having like four races that don't need to do it?

good point, we should look at adding oxygen damage back to synths and stuff along those lines.

SpaceVampire:
> 
> 
> Also, I'd like to point out, if breathing was so important to the game, then why are we having like four races that don't need to do it?

I agree, but that is outside the scope of this current PR.

Nerev4r:
> > Also, I'd like to point out, if breathing was so important to the game, then why are we having like four races that don't need to do it?
> 
> good point, we should look at adding oxygen damage back to synths and stuff along those lines.

Literally why. It's a robot. Why would it take oxygen damage.

SpaceVampire:
For those that think we should replace it with something, we will eventually come up with new symptoms.

hecksadecimal:
Everything about this concept hurts me on a deep psychological level.

Nerev4r:
> For those that think we should replace it with something, we will eventually come up with new symptoms.

What's the guarantee that 'eventually' will actually come and that genetics/viro won't be left behind due to lack of low-risk things to distribute to the average crewmate?

tf-4:
> 
> 
> For those that think we should replace it with something, we will eventually come up with new symptoms.

I've been thinking about this, maybe a self-righting symptom that makes you immune to slips could be fun.

Owai-Seek:
I posted in the other PR (https://github.com/Skyrat-SS13/Skyrat-tg/pull/4474), but I'll post it here as well:

```
Nanites: Vulnerable to EMP, shocks, and power failures. (Missed Cloud Syncs)

Viruses: Powerful, but there's only one (IMO) -stupidly- powerful virus trait. (Self Resp), which is often combined with space adaptation.

```

As for the claim on the other PR of "Don't worry, we'll add more!", I'd like to say, that's complete bollocks, but I think Genetics has enough to stand on it's own. At least when it comes to unique and fun/beneficial things, without Space Adaptation/Thermals/Xray.

Virology has some gimmicky stuff, but most of it amounts to "Self Resp." "Heal People with X." and "Antag Stuff, most of which can't be used."

While I understand why people would want -both- clipped from the game. (An eye for an eye, I suppose.) I feel genetics has more unique content comparatively.

Stalkeros2:
Bruh???????????

KathrinBailey:
I'd rather this be postponed for a while so we can see how the space adaptation removal plays out.
You can easily fill up internals and with two small inventory slots you can pretty much spend the entire round with two tanks. You'll only need to refill both tanks maybe once or twice if you get yellow ones and do the porta-pump fill. This negates breathing fairly easily with very little preperation and inventory space required. If we made a scale of OP-ness, space adaptation being an 8, self-resp would probably be a 3 in the most ideal situation?

softcerv:
Considering the strength of Genetics and Nanites, I wouldn't really consider Self-Resp too much of an issue comparatively. The are even a few cases where having Self-resp can be detrimental, like trying to preform CPR.

YakumoChen:
> 
> 
> For those that think we should replace it with something, we will eventually come up with new symptoms.

eventually means never, right? Just like that thermal vision gene replacement. The most temporary solution is often the one that sticks.

Gandalf2k15:
I think we can hold off on this for now. Give it a few weeks and then we can see about removing it, if it's really that big of an issue.

PR #4500, opened 2021-03-29
Deltastation Permabrig Remap

Aurelien3189:
Revamps the Deltastation permabrig to offer prisoners more things to do: Cells have windows and chairs, added a kitchen with griddle, freezer, and other equipment, expanded mess hall eating area and added library. Also added prisoner lockers, and made the security locker room bigger.
dreamseeker_7dqthoQYtr
dreamseeker_xJoIzjmcVL


Gandalf2k15:
The problem with big changes like this is that we have to reset the maps every other week, so as long as you keep these changes in an easily selectable SQUARE area, it will likely be transferred over. Otherwise it won't be if it's not all in the same area. When doing map resets we copy/paste the following: Cryo, Blueshields desk and security areas. If security areas are too complex and need cutting, we'll simply redo them.

Aurelien3189:
> The problem with big changes like this is that we have to reset the maps every other week, so as long as you keep these changes in an easily selectable SQUARE area, it will likely be transferred over. Otherwise it won't be if it's not all in the same area. When doing map resets we copy/paste the following: Cryo, Blueshields desk and security areas. If security areas are too complex and need cutting, we'll simply redo them.

You mean you manually copy and paste changes through the map editor? I could show the area with the changes, just wasn't told how map edits were standardised.

EDIT: Seems some gitkraken shenanigans made the map changes unreadable, so I'll close this and make another PR

PR #4545, opened 2021-03-31
[Modular] how ya'll mfs look using jump-to-area

unit0016:
## About The Pull Request
fixes jump to area leading to DS-1
why people are using jump to area is beyond me but I got complaints so here it is

## Why It's Good For The Game
feex good

## Changelog
:cl:
fix: Jump to area no longer leads to DS-1.
/:cl:


tf-4:
god thank you, this was so painful

PR #4617, opened 2021-04-03
[MODULAR] Adds new wizard spell, Magic Stimpack!

Swiftfeather:
## About The Pull Request

Adds a new (functioning) spell whose sole purpose is to give a few u of stimulants to the caster, and get them out of tight spots. This wouldn't be necessary if lethal ammunitions didn't do such an absurd amount of stam damage, but here we are. Only injects 3u, and gets you on your feet, but that's about it.

## Why It's Good For The Game

Lethal rounds do a shit ton of stamina damage depending on the gun they come from. Kinda sucks, this helps wizards get out of binds. Not as powerful as a stimpack, by any means. This won't make wizards walking tanks like a full stimpack would, but that's not what this PR is for. This PR is just so wizards have SOME way to counteract crazy high ranged stamina damage.

## Changelog
:cl:
balance: Wizards now have a stimulant spell.
/:cl:


Gandalf2k15 (approved):
ok

PR #4623, opened 2021-04-03
[modular] added brig medic and sarg to locked item lists in loadout

thestubborn:
you cant be a sergeant cowboy

or wear scrubs as a brig medic

this fixes this, enhancing cowboy roleplay by about 15%


thestubborn:
ah i forgot about that, hold on, new pr

PR #4655, opened 2021-04-05
[NON-MODULAR] Blob can't delay the end of the round for an hour.

SpookyTheFox:
Commented out both the set and clear for blob's hostile environment shuttle lock.

## About The Pull Request
This is because we have blob victory set to a ridiculous 700 tiles, when the battle is pretty clearly decided somewhere around the 200 mark. This results in long drawn out stalemates that mostly result in administrators nuking the station or just ending the round. 

## Why It's Good For The Game
If players are not having fun, their shuttle vote should be respected and the round should end a reasonable amount of time later, not make everyone wait for an hour.

## Changelog
:cl:
qol: Blob no longer holds up the round ending.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/overmind.dm
+++ b/code/modules/antagonists/blob/overmind.dm
@@ -65,7 +65,7 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 	color = blobstrain.complementary_color
 	if(blob_core)
 		blob_core.update_appearance()
-	SSshuttle.registerHostileEnvironment(src)
+	// SSshuttle.registerHostileEnvironment(src) SKYRAT EDIT removal
 	. = ..()
 	START_PROCESSING(SSobj, src)
 
@@ -202,7 +202,7 @@ GLOBAL_LIST_EMPTY(blob_nodes)
 	GLOB.overminds -= src
 	QDEL_LIST_ASSOC_VAL(strain_choices)
 
-	SSshuttle.clearHostileEnvironment(src)
+	// SSshuttle.clearHostileEnvironment(src) SKYRAT EDIT removal
 	STOP_PROCESSING(SSobj, src)
 
 	return ..()


gonenoculer5:
Or, or, heres an idea, **change the win value for blob.**

YakumoChen:
Lower the win amount. Letting people shuttle away let's them just avoid the blob instead of fighting it if it's starting to win

There are a lot admins can do about blob too to influence the fight,  including calling ERTs or allowing the nuke.

Shuttling away from the blob doesnt make sense IC either. It's a biohazard that would spread if allowed to go to Centcomm.

SpookyTheFox:
> Lower the win amount. Letting people shuttle away let's them just avoid the blob instead of fighting it if it's starting to win

Ok, you make that PR.

> There are a lot admins can do about blob too to influence the fight, including calling ERTs or allowing the nuke.

They do, pretty much every time. It doesn't un-break the round. It usually doesn't help at all. And situations that require continuous admin intervention to make the game playable are bad, even as controlled events.

> Shuttling away from the blob doesnt make sense IC either. It's a biohazard that would spread if allowed to go to Centcomm.

Let's not pretend there's an IC reason for the emergency shuttle lock.

Guidesu:
Honestly i don't think civilians should fight blob...
Like taking idea from Bay for example or Vesta.
Only military or trained personal should fight against biohazard and only in extreme situations should an civilian go it-self fight against the blob.
Risking civilian lives on an biohazard is an no go.
Evacuation is way better even if the biohazard wons the station.
Fast evacuation means the biohazard will not be on CC

Guidesu:
So good change!

SpaceVampire:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/4661
I just changed the win amount along with nerfing it, because that was the real issue.

YakumoChen:
> Ok, you make that PR.

I think blob is fine, so no.

> They do, pretty much every time. It doesn't un-break the round. It usually doesn't help at all. And situations that require continuous admin intervention to make the game playable are bad, even as controlled events.

If the blob is good it deserves greentext. If the blob is consistently winning every time then nerf it.

> > Shuttling away from the blob doesnt make sense IC either. It's a biohazard that would spread if allowed to go to Centcomm.

That _is_ the reason for the shuttle lock. It's an extremely dangerous biohazard that locks the shuttle as part of a quarantine. The shuttle lock has existed for the entirety of since Blob was introduced as a game mode, for that explicit reason.

> Risking civilian lives on an biohazard is an no go.
> Evacuation is way better even if the biohazard wons the station.

That's how you get a blob to spread. Say it 'infects' a civilian or rides aboard the shuttle as it leaves, then you have Centcomm infected with The Blob.


The Blob is a round ending antagonist. If the round is going to end by nature of The Blob, it should be done on The Blob's terms.

FlamingLily:
Bad PR, if the blob is going to win with no way to stop it, the Captain should take some goddamn initiative and prime the nuke, followed by evacuating the crew to lavaland. There's no reason to break quarantine ICly, and nuking the station also ends the round.

Nuking the blob counts, or should count, as a blob victory.

SpookyTheFox:
> 
> 
> Bad PR, if the blob is going to win with no way to stop it, the Captain should take some goddamn initiative and prime the nuke, followed by evacuating the crew to lavaland. There's no reason to break quarantine ICly, and nuking the station also ends the round.
> 
> Nuking the blob counts, or should count, as a blob victory.

That is an acceptable solution. However. By the time a captain realizes what they are up against, they usually can no longer do that, and it's still subject to admin intervention. Being unable to even do it using the in character method, and needing admins to trigger the nuke to end the round, speaks volumes about the current state of blob.

There are some alternatives to this PR being proposed now, I look forward to seeing what solutions are offered up.

SpookyTheFox:
@SpaceVampire has put forth a reasonable sounding solution that I would like to try first.
https://github.com/Skyrat-SS13/Skyrat-tg/pull/4661

PR #4657, opened 2021-04-05
[SEMI-MODULAR] Fixes mutant bodyparts that share the hair color not updating from regenerative jelly exposure

GoldenAlpharex:
## About The Pull Request
This PR makes it so being exposed to regenerative jelly will now affect all bodyparts that are sharing their color with the hair of the character, instead of just the hair and facial hair.

Tested the changes on slimepeople and it still doesn't change their hair color, which means this is working how I intended it.

It also modularizes `update_mutant_bodyparts()` and adds an optional argument to dictate whether the update should be done regardless of the usual check for the bodypart difference to avoid useless overlay modifications, which means that more features changing mutant bodyparts in the future may benefit from this without having to do something more hacky.

I haven't tested the effects of this on anthro characters, but then again, most of the things already don't work on anthro characters.

## Why It's Good For The Game
This is bad
Pre-Regenerative-Jelly-Fix
This is good
Post-Regenerative-Jelly-Fix

## Changelog
:cl:GoldenAlpharex
fix: Regenerative jelly should now affect hair-related bodyparts.
code: Made handle_mutant_bodyparts() have an optional bypass argument to allow for ease of bodypart modifications in the future.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_update_icons.dm
+++ b/code/modules/mob/living/carbon/human/human_update_icons.dm
@@ -52,9 +52,13 @@ There are several things that need to be remembered:
 /mob/living/carbon/human/update_hair()
 	dna.species.handle_hair(src)
 
+// SKYRAT EDIT REMOVAL - FIXING CUSTOMIZATION(?) (moved to modular)
+/*
 //used when putting/removing clothes that hide certain mutant body parts to just update those and not update the whole body.
 /mob/living/carbon/human/proc/update_mutant_bodyparts()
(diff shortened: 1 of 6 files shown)


gonenoculer5:
why not PR this upstream?

gonenoculer5:
> why not PR this upstream?

nevermind im retarded

GoldenAlpharex:
Another day with this fix not merged. I cry with my mis-matching hair.

PR #4687, opened 2021-04-06
Sec locker room upgrade (Metastation)

Aurelien3189:
Slight changes and expansion to the metastation security prep rooms to be more friendly to traffic

image
1. Moved EVA equipment from security prep to its own room, using the Gear Storage door used on Icebox station's EVA storage. Also adds bio and bomb suit lockers. Adds practice laser carbines
2. Expanded locker room width by one tile. All lockers are on one side, toolboxes and tables moved on the other side, with added crowbars and hand labeller
3. moved the shutters to the armoury so that they'd enter to the two tile wide space, for better traffic flow

### DIFF
--- a/_maps/map_files/MetaStation/MetaStation_skyrat.dmm
+++ b/_maps/map_files/MetaStation/MetaStation_skyrat.dmm
@@ -3286,13 +3286,16 @@
 	},
 /area/security/prison)
 "ank" = (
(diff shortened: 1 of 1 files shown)


Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/4695

PR #4770, opened 2021-04-09
[MIRROR] Better HFR building

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/57901
--------------------

## About The Pull Request
The PR adds a quick and easy way to build the HFR
there will be boxes that when properly positioned and hit the center with a multitool you'd create the HFR, rotate the side parts to finish the build and hit with a multitool the interface like normally. (can't do much about it because pipe code is hell)
There will also be a buyable crate from cargo with CE secured lock
The normal way to build it is still available.

The boxes
image

The built HFR
image

## Why It's Good For The Game
Less pain to techs when building the machine

## Changelog
:cl:
add: Add a better way to build the HFR with special deployable boxes, set them up in the right position (check the image on each box, it'll tell you where that box should be in a 3x3 grid.. or just examine the box)
/:cl:


NullDagaf:
no
just please no

PR #4856, opened 2021-04-12
Icebox sec tweaks

Aurelien3189:
makes a few tweaks to Icebox's prison and sec prep room, such as a sink in botany.

image

1. Adds a dish drive to the prison kitchen, for that vanity
2. Adds a sink, crowbar and wrench to hydro, as well as some EZ nutrient.
3. Replaces library's 'visitor console' with a console that can actually print books
4. Adds windows to the shooting range/equipment room and rearranges the locker room a little.
5. [can be removed] Adds an airlock to the locker room. Personally I wasn't sure what the bridge was for so I figured that adding an airlock there, next to prep and the EVA equipment would let it see some use. correct me if I'm wrong, though.

### DIFF
--- a/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
+++ b/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
@@ -38584,16 +38584,6 @@
 	},
 /turf/open/floor/iron,
 /area/cargo/sorting)
-"cNP" = (
-/obj/machinery/door/airlock/external{
-	name = "Security External Airlock";
-	req_access_txt = "63"
-	},
-/obj/effect/mapping_helpers/airlock/cyclelink_helper{
-	dir = 4
-	},
-/turf/open/floor/plating,
-/area/icemoon/surface/outdoors)
 "cNR" = (
 /obj/structure/cable,
 /turf/open/floor/plating,
@@ -39217,7 +39207,9 @@
 	id = "prisonlibrarycurtain";
 	pixel_x = -24
 	},
-/obj/machinery/computer/bookmanagement,
+/obj/machinery/computer/bookmanagement{
+	pixel_y = 8
+	},
 /turf/open/floor/wood,
 /area/security/prison)
 "ddy" = (
@@ -40629,6 +40621,9 @@
 	pixel_y = -6
 	},
 /obj/item/reagent_containers/glass/bottle/nutrient/ez,
+/obj/effect/turf_decal/siding/wood{
+	dir = 1
+	},
 /turf/open/floor/plating/dirt/planet{
 	icon_state = "oldsmoothdirt"
 	},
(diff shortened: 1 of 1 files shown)


Grunnyyy:
Generally not the best idea to give prison a Wrench and a Crowbar. 

Why do they need one for Hydro?

Aurelien3189:
> Generally not the best idea to give prison a Wrench and a Crowbar.
> 
> Why do they need one for Hydro?

It's incredibly common for prisoners to redecorate the prison, particularly the floors or even entire rooms. A wrench may come in handy to move furniture around to make things more convenient than the current layout, and I feel like offering that flexibility, short of tearing down walls, is a good way to inspire better designs.

YakumoChen:
I'd remove the air lock, it's kind of unneeded to add an actual entrance to that area.

The bridge is there so that antagonists can break into the armoury from the outside, but I consider the airlock possibly making it a bit _too easy_ for security to properly defend that space. The space is meant to be most advantageous to antagonists, not security.

KathrinBailey (changes requested):
1.The bridge is solely decorative, there's no access point outside the armoury because it makes repelling armoury attackers far too easy. Plus, you can just run out into the snow, it ain't space no more. Get rid of those airlocks.
2. Remove the glass surrounding the security hardsuits, it makes it a bit _too_ easy to steal those hardsuits.
3. Change the pixel_y of that library computer so it sits upon the table, it's a nitpick but...
4. Add back the grilles on the other side of the armoury. They're actually there as a little noise alarm, if you bash the grilles in it's really loud.
5. I don't know if you actually need that EZ nutrient, it might be a red herring. Tell me what unqiue variables the dirt plots have, it might have self-sustaining on.
6. Check the North of the prison and ensure you don't have a crowbar and wrench in there already, there's a little loot room in there.

Aurelien3189:
> 1.The bridge is solely decorative, there's no access point outside the armoury because it makes repelling armoury attackers far too easy. Plus, you can just run out into the snow, it ain't space no more. Get rid of those airlocks.
> 2. Remove the glass surrounding the security hardsuits, it makes it a bit _too_ easy to steal those hardsuits.
> 3. Change the pixel_y of that library computer so it sits upon the table, it's a nitpick but...
> 4. Add back the grilles on the other side of the armoury. They're actually there as a little noise alarm, if you bash the grilles in it's really loud.
> 5. I don't know if you actually need that EZ nutrient, it might be a red herring. Tell me what unqiue variables the dirt plots have, it might have self-sustaining on.
> 6. Check the North of the prison and ensure you don't have a crowbar and wrench in there already, there's a little loot room in there.


image

1. I was curious what the idea behind the design was, never seen the armoury broken into that way but I like the opening. Went ahead and removed the extra airlock.
2. Windows removed!
3. Upped the pixels a little
4. Grills re-added!
5. Figured to leave an EZ nutrient in as late joiners may come across empty plots where weeds have sucked out all the nutrients. It'd be a safeguard so that prison botany isn't stopped by weeds or cooks using up their ingredients, gardening being one of the main things that keep prisoners busy. 
6. There is a crowbar but it's locked behind electric grilles, figured that was a special loot spawn but I haven't touched it.

Grunnyyy:
Tools generally shouldn't readily be available, especially not effortless. Considering it's still a prison and an almost breakout ready kit with jsut the two of those. Prisons are not only recreational for the prisoner role.

KathrinBailey (approved):
Alright, lovely.
I forgot the sinks so good job re-adding them. I'd prefer the infinite sinks, buuut, I suppose people will complain if we give too many """luxuries""".
Conditional approval, remove the construction tools you added to botany and I'll merge.
Further down the line I'll add random chance tool spawners and random chance false wall spawners. I did that on oldbase because it helped prevent characters getting ahead of others based solely on OOC map knowledge. I'll do it here soon if I don't forget (I'll forget).

KathrinBailey:
woah okay :(

PR #4827, opened 2021-04-11
Battery Component

Gandalf2k15:
## About The Pull Request

Introducing the battery component - A component that works much like the cell system that you might find on Eris. However, this one uses components and is made by me. Something I've wanted to do for a while.

Basis:
Think of the battery component like the battery compartment of a flashlight/multitool/whatever.  It holds the battery and processes information relating to the battery. If the battery inside the compartment has no charge then the item won't function, simple as. No battery? No function. 

If the tool the component is attached to is inside a robot, it will instead use the robot powercell.

Items use batteries in two different ways - Passive and action-based.

Passive example:
When an item has its battery inserted, it is idle until you start using the item. For flashlights, this is simple, when on, it succs power until there's no power to power the light, then it shuts off. 

Action-based:
In the case of a multitool, it will try to USE power from the battery when the multitool action is attempted, if it can't use the power from the battery, then it will not work.

I have categorized items into a simple system for defining how much power they use:

> /// The minimal amount of power an item can use.
> #define POWER_CELL_USE_MINIMUM 1
> /// For items that barely use any power at all.
> #define POWER_CELL_USE_VERY_LOW 10
> /// For items that generally wouldn't use very much power.
> #define POWER_CELL_USE_LOW 30
> /// For items that use a average amount of power.
> #define POWER_CELL_USE_NORMAL 50
> /// For items that use a high amount of power.
> #define POWER_CELL_USE_HIGH 70
> /// For items that use A LOT OF POWER.
> #define POWER_CELL_USE_INSANE 100

To-do:
- [x] Get the component working with: Flashlights
- [x] Get the component working with: Multitools
- [x] Get the component working with: Analyzers
- [x] Get it working with ALL power tools(They get a pretty big buff with this)
- [x] Get it working with health analysers
- [x] Get it working with megaphones
- [x] Get it working with pipe painters
- [x] Get it working with PDA's
- [x] Get it working with slime scanners
- [x] Decal painters now use cells instead of ink!
- [x] Working with radio jammers
- [x] Testing the component

## Changelog
:cl:
add: Added a new battery component, which will relate to multiple pieces of equipment, such as a flashlight. Said flashlight will now require a cell to function, and you'll need to recharge said cell in a cell recharger. You can also buy cells from a multitude of cell vendors. To remove a cell, ctrl+shift+click on the item you want to remove it from. To reinsert, simply slap the thing with a cell.
/:cl:

PR #4900, opened 2021-04-14
Item use power and cell component refactor

Gandalf2k15:
## Changelog
:cl:
add: A way for atom items to use power.
refactor: A new cell component refactor for any uses where manual cell interaction is required.
/:cl:

## About The Pull Request
This PR does two things:
It Introduces a new item proc for sending signals to anything listening(the cell component for instance) to use x amount of power and return it to the proc, depending on the outcome. This opens up avenues for countless opportunities in regards to items that would normally need a lot of backend changes to use a cell. (attack self etc)

This also includes a cell component, which aims to refactor all instances of crappy jerry-rigged cell interactions and enables any item to use a cell and interact with it using simple callbacks. It enables you to produce items that require cells without the need to process cell insertion/removal, cell display, cell use etc. 

Instructions on making an item use this component:
If the item is action based then you simply `AddComponent(/datum/component/cell)` in component initialize, and then you can hook into it by using the new item proc, item_use_power. This returns a bitflag of the component status, or, if there isn't one, it returns none:

#define COMSIG_ITEM_POWER_USE "item_use_power"
	#define NO_COMPONENT NONE
	#define COMPONENT_POWER_SUCCESS (1<<0)
	#define COMPONENT_NO_CELL  (1<<1)
	#define COMPONENT_NO_CHARGE (1<<2)

It's very simple to use, here's an example:
if(!(item_use_power(power_use_amount, user) & COMPONENT_POWER_SUCCESS))
	return FALSE

Including user in the parameters will send an automatic message to them as to why the process failed(be it a cell running out of power, or removed.). Power is automatically drawn from the internalized cell.

We also enable passive power usage, to do this, simply process the item like you normally would, and add the item_use_power check there too. If you'd like an automatic callback for the cell being removed, we offer this too, simple add a callback referring to the proc you'd like to trigger when the cell is removed.

All items now have a power rating, the cell component will automatically use these to draw the assigned amount of power from the cell when item_use_power is called.

//General defines for items that use power and the cell component.
/// The minimal amount of power an item can use.
#define POWER_CELL_USE_MINIMUM 1
/// For items that barely use any power at all.
#define POWER_CELL_USE_VERY_LOW 10
/// For items that generally wouldn't use very much power.
#define POWER_CELL_USE_LOW 30
/// For items that use a average amount of power.
#define POWER_CELL_USE_NORMAL 50
/// For items that use a high amount of power.
#define POWER_CELL_USE_HIGH 70
/// For items that use A LOT OF POWER.
#define POWER_CELL_USE_INSANE 100

## Items this actually applies to
Power drills / JOL - These have two states, powered and unpowered, when unpowered, they still work like normal, however their toolspeed is equal to a normal crowbar.  When powered, their toolspeeed becomes 0.5! Fantastic. Uses power when on.
~~Testiculator~~Electrical welder - These are basically super welders in the fact they are incredibly fast and don't technically have an open flame. However of course they use a battery. These replace experimental welders. These don't cause fires and are relatively safe!
image
Flashlights - These now have a cell, however, they have 3 different modes, low, med and high!
Radio jammers - These have been given a cell and had their range upped from 12 to 20.
Decal painters/airlock painters - These have been given a cell instead of ink, and are now uncapped.


Gandalf2k15:
Update on the usage situation:
Power drills / JOL - These have two states, powered and unpowered, when unpowered, they still work like normal, however their toolspeed is equal to a normal crowbar. When powered, their toolspeeed becomes 0.5! Fantastic. Uses power when on.
Multitools have had their cell requirement removed.

Owai-Seek:
I just want to point out that the last PR about this had -20 downvotes, with tons of comments, and this one got reopened in its place. People very much don't like needing batteries in things. Even Useroth said this might be better to do upstream because it touches so many things.

https://github.com/Skyrat-SS13/Skyrat-tg/pull/4827

As for my opinion on it, IMO, it's way too far reaching. Every single small item in the game seems to have a battery attached now, and it feels really silly. I can see using this to like, limit the amount of things that can be abused, like jaws/flashlights, but for small stuff, it seems like, way too much.

I could see mildly interesting things being done with this if we still had a laser weapon focus, and you wanted to scrounge together/steal a battery for self defense, but as it is now, this just feels... really unnecessary?

Gandalf2k15:
> I just want to point out that the last PR about this had -20 downvotes, with tons of comments, and this one got reopened in its place. People very much don't like needing batteries in things. Even Useroth said this might be better to do upstream because it touches so many things.
> 
> #4827
> 
> As for my opinion on it, IMO, it's way too far reaching. Every single small item in the game seems to have a battery attached now, and it feels really silly. I can see using this to like, limit the amount of things that can be abused, like jaws/flashlights, but for small stuff, it seems like, way too much.
> 
> I could see mildly interesting things being done with this if we still had a laser weapon focus, and you wanted to scrounge together/steal a battery for self defense, but as it is now, this just feels... really unnecessary?

Go ahead and post in the discord, thanks.

Sethnov:
ok so let's talk about this, first thing I don't like is when you hit something with a tool it uses power, even when it does nothing. so slapping a glass window with jaws of life uses power or a pipe.

Gandalf2k15:
> ok so let's talk about this, first thing I don't like is when you hit something with a tool it uses power, even when it does nothing. so slapping a glass window with jaws of life uses power or a pipe.

It was changed not to do this, in older builds, it did. In this build, you need to turn it on for it to be in the fast tool mode.

Sethnov:
> > ok so let's talk about this, first thing I don't like is when you hit something with a tool it uses power, even when it does nothing. so slapping a glass window with jaws of life uses power or a pipe.
> 
> It was changed not to do this, in older builds, it did. In this build, you need to turn it on for it to be in the fast tool mode.

yes but in fast tool mode, it would use power if you just example hit something, using power when there was no action involved, like hitting a pipe with jaws of life or hand drill or any object would use power when being hit. different from putting up a window or wrenching a machine, as in intended use.

Gandalf2k15:
> > > ok so let's talk about this, first thing I don't like is when you hit something with a tool it uses power, even when it does nothing. so slapping a glass window with jaws of life uses power or a pipe.
> > 
> > 
> > It was changed not to do this, in older builds, it did. In this build, you need to turn it on for it to be in the fast tool mode.
> 
> yes but in fast tool mode, it would use power if you just example hit something, using power when there was no action involved, like hitting a pipe with jaws of life or hand drill or any object would use power when being hit. different from putting up a window or wrenching a machine, as in intended use.

Yeah thats intended, you're supposed to switch it off/on when you're using it.

Sethnov:
> > > > ok so let's talk about this, first thing I don't like is when you hit something with a tool it uses power, even when it does nothing. so slapping a glass window with jaws of life uses power or a pipe.
> > > 
> > > 
> > > It was changed not to do this, in older builds, it did. In this build, you need to turn it on for it to be in the fast tool mode.
> > 
> > 
> > yes but in fast tool mode, it would use power if you just example hit something, using power when there was no action involved, like hitting a pipe with jaws of life or hand drill or any object would use power when being hit. different from putting up a window or wrenching a machine, as in intended use.
> 
> Yeah thats intended, you're supposed to switch it off/on when you're using it.

is it possible to add a hotkey for it? I imagine it be tedious to turn it on and off a lot. which I just forget about the power mode. make life easier for my fingers.

Gandalf2k15:
Could use a hockey for it sure

PR #4947, opened 2021-04-16
makes cell chargers uniform with other machines

werewolfboyfriend:
## About The Pull Request
before
bad
after
good
nitpicky, but I recolored the cell chargers so they are the same shade of grey as other machines.

## Why It's Good For The Game
not deviating from an otherwise established color palette is good

## Changelog
gonna be honest, I don't think this is a big enough deal for a changelog


ghost:
> ## Changelog
> 
> gonna be honest, I don't think this is a big enough deal for a changelog

counterpoint: you can mark it as Fix for those sweet sweet good-boi-points

PR #4952, opened 2021-04-16
Removes assaultops

Gandalf2k15:
## About The Pull Request

Honestly when I made this it was coded with haste and is pretty shit. It barely works. I'll leave in the shuttle and flavour items. But the gamemode needs to go. See:
https://github.com/tgstation/tgstation/pull/58470 - Means it would need entirely redone, something I am not interested in doing.
I don't want to exert the effort of redoing this either.

Don't merge until we get confirmation of upstream gamemode refactor.

## Changelog
:cl:
del: Assault ops has been removed.
/:cl:


unit0016:
Should mention this nukes the DS-1 specific comms too, which'd.. cause problems with the fact DS-1 is still there.

SpaceVampire:
Wouldn't it make sense to remove DS-1 if we are removing this?
It would have even less of a point with it gone.
Better to just focus on the two other syndie roles.

hugmeorelse:
> Wouldn't it make sense to remove DS-1 if we are removing this?
> It would have even less of a point with it gone.
> Better to just focus on the two other syndie roles.

Nah, this is only gone temporarily. We want to rebuild it later.

NullDagaf:
Personally, I think assault operatives are good as a concept, and could be re-done to be a quite entertaining gamemode.

With that in mind. If you want to, PLEASE, FOR THE LOVE OF FUCK. Modify the vanilla nuke ops instead of making a whole seperate thing, or at LEAST have them use the default/modified nukie outpost. It seems pretty retarded code/map-wise to make a whole Z level for something that's only used every once in a while, especially when there's already a place that serves BASICALLY the same purpose - as a place for the team to equip; and at the same time, it seems quite dumb that the syndicate would make TWO seperate teams, both with BASICALLY the same objective - assault the station (Only that the nukies have to blow it up, but still.). If you want the "head capturing" to be a thing, it shouldn't be hard to map in a small gay baby jail, or something along those lines - maybe even expand it further, making a small base akin to the DS-1? Possibly replacing it altogether? Pretty please?


Anyway, spazm over. No issues with the PR itself.

NullDagaf:
> DO NOT MERGE label.
> Gets merged.

epic

PR #4971, opened 2021-04-17
[MODULAR] Tweaks projectiles bonus stamina damage.

SpaceVampire:
## About The Pull Request

Changes the PROJECTILE_TISSUE_DAMAGE_STAMINA_MULTIPLIER from 1.6 to 1. Subject to change of course.

## Why It's Good For The Game

I think its clear at this point that our Combat Rework came with a couple of issues. But this has made it almost impossible to balance the damage of guns correctly, when all of the attacks do extra damage. You want to knock someone into stam-crit with your weapon? Well now you have to actually use the bullets made for it, and we can maybe start to balance stuff correctly. Honestly, I think that there need to be a number of changes made to it outside of this, but I am starting small. If we actually want to have drawn out ranged combat, then this helps allow it, but not making every bullet in the game knock you to the ground super early. I uh, will look into changing other stuff later, after we see how this feels. Rubber bullets are still really strong for doing this, and now they have more of an actual use.

## Changelog
:cl:
fix: Every bullet in the game dealing too much stamina damage. 
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/combat.dm
+++ b/code/__DEFINES/~skyrat_defines/combat.dm
@@ -9,7 +9,7 @@
 #define SIMPLE_MOB_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.7 //Fights with simple mobs are usually more sustained, so apply a bit less
 #define BLUNT_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.8 //Brute that isnt sharp, knocks the wind outta you real good
 #define OTHER_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.5 //Burns, sharp implements
-#define PROJECTILE_TISSUE_DAMAGE_STAMINA_MULTIPLIER 0
+#define PROJECTILE_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1
 
 //Glancing attacks happen when someone uses disarm intent with melee weaponry, aiming to disable a person instead
 #define TISSUE_DAMAGE_GLANCING_DAMAGE_MULTIPLIER 0.5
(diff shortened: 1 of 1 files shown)


SpaceVampire:
I'm 100% willing to change around numbers here, or touch other stuff. I think this PR should most likely be test-merged so people can see if the new combat feels better or not. We might also have to adjust some other stuff like changeling power to bring it back in line. But I just feel like it uh, doesn't feel great at the moment, so wanted to say something about it. Give feedback.

RatFromTheJungle:
1.3 is a good, 0 off the bat is sixteen stages past an over-nerf, bordering on making lethals pointless, with how everything else would be doing more damage, and better damage, with more immediate results.

code/__DEFINES/~skyrat_defines/combat.dm
```diff
@@ -9,7 +9,7 @@
 #define SIMPLE_MOB_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.7 //Fights with simple mobs are usually more sustained, so apply a bit less
 #define BLUNT_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.8 //Brute that isnt sharp, knocks the wind outta you real good
 #define OTHER_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.5 //Burns, sharp implements
-#define PROJECTILE_TISSUE_DAMAGE_STAMINA_MULTIPLIER 1.6
+#define PROJECTILE_TISSUE_DAMAGE_STAMINA_MULTIPLIER 0
```
RatFromTheJungle: 1 is a much better starting place then absolute 0, as 1 is much easier to tweak up and down from, whereas 0 leaves no room for feedback other then, "Lethals do a lot less damage now"

SpaceVampire: I agree, after thinking about it.
My first reflex was that only the bullets supposed to deal it, should actually deal it it.
But that is most likely too extreme of a starting place.

hugmeorelse:
> I'm 100% willing to change around numbers here, or touch other stuff. I think this PR should most likely be test-merged so people can see if the new combat feels better or not. We might also have to adjust some other stuff like changeling power to bring it back in line. But I just feel like it uh, doesn't feel great at the moment, so wanted to say something about it. Give feedback.

As stated in another PR, can you please identify who you are on the Discord?

hugmeorelse:
Oh, it's Infrey. :v 

Time to DM them. :)))

Gandalf2k15:
Given that this is a multiplier, won't this be the same as completely nulling the modifier?

SpaceVampire:
> 
> 
> Given that this is a multiplier, won't this be the same as completely nulling the modifier?

No, because its based on the other damage applied.
So uh. If you do 24 brute damage. It does that times the mod in extra stam damage.
So right now its set as a 1:1 ratio. That is how it seemed to work in testing anyway.

SpaceVampire:
Holy shit, that "Close Button" really needs a confirm feature.

PR #5003, opened 2021-04-18
[Semi-Modular] De-Multizify the Iceminer arena so it can load on maps other than icebox

unit0016:
## About The Pull Request
Makes godslayer armor obtainable by making the arena map properly spawn. Rejoice, fair minermans, for your phat gamer loot is now possible to obtain.

## Why It's Good For The Game
Bug bad.

## Changelog
:cl:
fix: The Iceminer arena now can spawn properly on maps other than Icebox.
/:cl:


SpaceVampire:
Can we see an image of what you did exactly?
The multi-z is an important mart of that arena.
I'm not exactly sure how it would work without it.
Does this effect the normal ice-box one, or just like.
The one on the map that sometimes loads on other stations.
That feels really out of place, because it doesn't have the normal icebox design.

unit0016:
> Can we see an image of what you did exactly?
> The multi-z is an important mart of that arena.
> I'm not exactly sure how it would work without it.
> Does this effect the normal ice-box one, or just like.
> The one on the map that sometimes loads on other stations.
> That feels really out of place, because it doesn't have the normal icebox design.

I gave it two one way portals, one in on the south and one out on the north. Sadly I can't specify it to be different per map without making a whole new ZTRAIT, but this is what I did. It's Not **/that/** big of a change when you think about it, the arena still works fine. Multi-z just let you fall down if you didn't want to use the ladder to challenge the minermans.
image

SpaceVampire:
I mean, it does mess with the balance because the ladder is supposed to have a climb delay on it, and the house above was a pretty cool landmark. I guess at the end of the day its not like, awful. But it feels like a downgrade made only to accommodate the new and kind of unnecessary system.

unit0016:
>But it feels like a downgrade made only to accommodate the new and kind of unnecessary system.

Never said it wasn't, honk. Wish I had a better solution that wasn't drastic.

GoldenAlpharex:
> I mean, it does mess with the balance because the ladder is supposed to have a climb delay on it, and the house above was a pretty cool landmark. I guess at the end of the day its not like, awful. But it feels like a downgrade made only to accommodate the new and kind of unnecessary system.

But... Ladders don't have climb delay? That was something _you_ wanted to add.
Not to be rude, of course, but it just shouldn't be a balance concern, unless we want to add a way to click on the portals to get out?

NullDagaf:
But.. why? Lavaland already has the many other bosses (besides the fact that an ICE FILLED ARENA on a LAVA FILLED PLANET is ultra unrealistic. (Biodomes don't count.)) As for the other mining place (whatever it's called), pretty sure it's supposed NOT to have gamer loot.

SpaceVampire:
> But... Ladders don't have climb delay? That was something _you_ wanted to add.
> Not to be rude, of course, but it just shouldn't be a balance concern, unless we want to add a way to click on the portals to get out?

I just wanted to give all ladders a base climb delay. It was already supported as option in the code just for that one.
The idea being that once you enter the area it takes a moment to leave, so that you can't just jump up and down z-levels as an easy way to cheese the boss. If you did something like that to the portals it would help.

> But.. why? Lavaland already has the many other bosses (besides the fact that an ICE FILLED ARENA on a LAVA FILLED PLANET is ultra unrealistic. (Biodomes don't count.)) As for the other mining place (whatever it's called), pretty sure it's supposed NOT to have gamer loot.

From what I understand, this isn't so that it can spawn on lavaland. Its so that it can spawn on the whole "Ice-planet instead of lavaland, so its only one z-level" thing, which already has all of the other ruins and special bosses that come with it. Its also not so it can spawn on rock-planet (or something like that) as well, which is the non gamer loot place you are talking about.

NullDagaf:
> From what I understand, this isn't so that it can spawn on lavaland. 

[Semi-Modular] De-Multizify the Iceminer arena
### so it can load on maps other than icebox

tf-4:
> > From what I understand, this isn't so that it can spawn on lavaland. 
> 
> [Semi-Modular] De-Multizify the Iceminer arena
> ### so it can load on maps other than icebox

You don't know what you're talking about

NullDagaf:
> Yes. You don't know what you're talking about, please stop trying to flame this.
> Random Mining maps comes with an icemoon variant. This is so that it loads properly on that.

Huh. Didn't know random maps are a thing. In that case, fair enough.
Only gripe at this point would be to make the portal appear after killing the DFM, but that's honestly not that big of a deal.

PR #5022, opened 2021-04-19
Cuts back the alert voice

YakumoChen:
## About The Pull Request

Cuts back the length of various alert level dialogues. Most of them now only state the alert level has changed to [COLOR]. Delta's announcement is preserved in its entirety and Down to Red (from Delta) still has critical dialogue intact.
Reverts the captain announcement sound to the original

## Why It's Good For The Game

Many of the alert level sounds are _extremely_ annoying. They drone on for far, far too long and relay information better suited to annoying only in text form. Add to the fact the alert level often changes dozens of times a shift and it is grating very quickly.

The same goes for the captain's announcement. The captain sends dozens of messages a shift, and it already has a unique sound byte associated with it. Having a voice state it's a captain's announcement becomes annoying to hear very quickly.

## Changelog
:cl:
sounddel: trimmed down alert level sounds.
sounddel: Captain's announcement uses a voiceless sound byte once more.
/:cl:

disclaimer: I'm not a sound engineer, the edited sound bytes are the same audio files but chopped off at opportune points.


thestubborn:
can we get the intern to do the alerts

Arctaisia:
I feel the voice makes it much more immersive, which is one thing we're aiming to do with our server as a whole. The reading off of the words in the alert levels makes for an easy differentiable rundown on what's allowed and not, and it means that you can listen along, rather than having to scroll up to see the whole announcement. It might get annoying on occasion, but I feel it adds more realism to each alert sound, I feel like even amber lacking this makes it hard for us to function in security at times, because we might jolt up to do amber things, but turns out it was violet. It also assists new security members somewhat, aswell, allowing them to just listen along as I stated prior. 

It's been something I've enjoyed about the announcements as a while, especially being part of the feedback team, and a security main with my sec-taur, so just dropping a bit of reasoning for my own opinion here, wehh.

YakumoChen:
I respectfully disagree. The alert announcements here convey the same level of information, and it is quite easy to remember what's allowed on an alert after a few shifts for security. 

Sounds are an important part of the game, yes, but they should be used sparingly. 
Especially with voice-type sounds, they can be very disruptive and I feel it's not in good game design to overuse long, droning sounds that are liable to repeat many times over the course of just a single round. 

If you've ever heard an AI spam vox announcements, you'll find that people grow tired of it very quickly, and this isn't much different.

Another example of this is the fire alarm sound, which, like you would expect a fire alarm to be, is _very_ grating to hear after a short time and hard to get away from since the sound carries past the room with it active.

I actually preferred it before when alerts were just a noise and not a voice clip, but I feel this is a good middle ground where it conveys through voice only what's vital to hear and lets the player absorb the rest of the information in a less disruptive way.

YakumoChen:
To highlight how I feel about the sounds best, it's at the point where if I'm not in an active role in the game (like if I'm ERPing in dorms) I'll end up turning the game sound off, because it gets too irritating to listen to the alert voice every time the level changes.

I wouldn't consider it a good thing if a sound is grating to the point I want to turn the game sound off.

Gandalf2k15:
With what arc said, I will try to create the Amber and violet alert announcements. These are fine as they are. Don't change them.

Gandalf2k15:
Oh on a side note, alert sounds were specifically given a toggle to turn them on and off. The volume is also adjustable. See game preferences.

Gandalf2k15:
Closing this as you can change it to preference.

YakumoChen:
@Gandalf2k15 can you toggle the volume of the alert dialogue while keeping the alert sound though? The voice is the only annoying part about station alerts

PR #5117, opened 2021-04-23
[MIRROR] Stabilized light pink extract makes you a pacifist while active

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/58498
--------------------
## About The Pull Request

The stabilized light pink extract (the ambulance one) now also makes you a pacifist while it affects you.

## Why It's Good For The Game

Pretty much all the stabilized extracts are absurd in some way so someone's gotta start to bring them in at some point. (Or I could just remove them all.)

## Changelog
:cl: Melbert
balance: The stabilized light pink extract now makes you a pacifist while active.
/:cl:

PR #5123, opened 2021-04-23
Revert "[NonModular] Sepia/lightpink rebalance"

SpaceVampire:
With this new upstream PR: https://github.com/Skyrat-SS13/Skyrat-tg/pull/5117 We have a better balance for them.
Thus this one isn't gonna be needed anymore if we merge it.
Fun fact, clicking Revert doesn't given you the usual format.
So I am doing this one free-form baby.
So you can pick pink and be a pacifist, or pick Sepia and have it randomly not work.
Boom, balance.

### DIFF
--- a/code/modules/movespeed/modifiers/status_effects.dm
+++ b/code/modules/movespeed/modifiers/status_effects.dm
@@ -5,7 +5,7 @@
 	multiplicative_slowdown = 3
 
 /datum/movespeed_modifier/status_effect/lightpink
-	multiplicative_slowdown = -0.25 //SKYRAT EDIT: Oiriginal value (-0.5)
+	multiplicative_slowdown = -0.5
 	blacklisted_movetypes = (FLYING|FLOATING)
 
 /datum/movespeed_modifier/status_effect/tarfoot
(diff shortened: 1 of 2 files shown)


RatFromTheJungle:
I think the 'reduced' speed should stay, because pacifist or not, people being ludicrously fast 4ever is never an ideal situation, especially here with our higher tidi and slower bullets.

SpaceVampire:
> 
> 
> I think the 'reduced' speed should stay, because pacifist or not, people being ludicrously fast 4ever is never an ideal situation, especially here with our higher tidi and slower bullets.

I feel like it could be at least somewhat buffed, given that now its going to see more use as an actual tool for like, medical staff, because you can't really fight with it, and dragging bodies slows you down already. This is kinda a double-nerf to it now, when it was already fine after that last balance. I mean, if we didn't want it, then we could have not merged the other thing, but that is already done.

Vizoplez:
batons and nonlethal things still work, like grenades

RatFromTheJungle:
> I feel like it could be at least somewhat buffed, given that now its going to see more use as an actual tool for like, medical staff, because you can't really fight with it, and dragging bodies slows you down already. This is kinda a double-nerf to it now, when it was already fine after that last balance. I mean, if we didn't want it, then we could have not merged the other thing, but that is already done.

above from yawet, as well as all non-direct fighting still works, pacifism is an illusion, in reality it does very little, especially when people who want to use this, will be preparing pre-emptively.

Gandalf2k15:
Agreed.

PR #5146, opened 2021-04-24
[MOSTLY MODULAR]  Suits have 2 storage for small or smaller items.

Guidesu:
## About The Pull Request
Honestly, this is more QOL than a tweak.
To my incoming PR this will be kinda handy
You can attach 3 smolls items to coats and stuff.
Since its behave like an mini-backpack you need to drag your suit to your hands
There's 3 storage on the photos but now its two.
image
image
Since its behave like an mini-backpack you need to drag your suit to your hands

## Changelog
:cl: Guidesu
add: Exo-suit like coats/winter coats/lab coats in general any exo suit can store two small itmes

/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/pockets/pockets.dm
+++ b/modular_skyrat/modules/customization/pockets/pockets.dm
@@ -1,5 +1,5 @@
 /datum/component/storage/concrete/pockets/exo
-	max_items = 3
+	max_items = 2
 	max_w_class = WEIGHT_CLASS_SMALL
 	attack_hand_interact = TRUE
 	quickdraw = FALSE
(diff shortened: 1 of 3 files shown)


Guidesu:
Nice my edit is buggy

Sigmar-alkahest:
oh boy! another thing for sec to never check!

but in all seriousness this is actually a really interesting idea, and it honestly makes some sense, considering how people can be at times, it does not surprise me that this was made, people do have a tenancy to cram as much stuff into whatever pockets they have as they can, however, could you perhaps limit it so it only works with suits that actually would have pockets? even with the de-equip, it would be rather unreasonable for a hardsuit to act like this, don't you think?

modular_skyrat/modules/customization/pockets/pockets.dm
```diff
@@ -0,0 +1,13 @@
+/datum/component/storage/concrete/pockets/exo
```
Gandalf2k15, line 1: What does exo mean

Guidesu: exosuit

Guidesu:
@Gandalf2k15

Gandalf2k15:
Fine, I'll allow it.

PR #5190, opened 2021-04-26
[SIMI-MODULAR] Replaces Mayhem in a Bottle with Bloodlust in a Bottle.

SpaceVampire:
## About The Pull Request

Removes Mayhem in a Bottle from the loot pool, replaced instead with a new item Bloodlust in a Bottle.
This was taken from Old-Base where it was loot that a certain tumor mob could drop.
It gives a lesser version of the blood-crawling spell.

## Why It's Good For The Game

Mayhem in a Bottle needs to go, because its just bwoink-bait at the moment.
This was the first thing I came up with to replace it, should be fun while still fitting in with the theme.
I can tweak the spell if needed.

## Changelog
:cl:
refactor: Changed Bubblegum Loot Drops, so that it drops other stuff.
/:cl:


RatFromTheJungle:
who is simi, and how is he modular
(but more seriously, that %bloodcost should go up to like 35% at the minimum, blood crawl 'large pool' or not, is one of the most powerful spells ingame)

SpaceVampire:
> 
> 
> who is simi, and how is he modular
> (but more seriously, that %bloodcost should go up to like 35% at the minimum, blood crawl 'large pool' or not, is one of the most powerful spells ingame)

It never came up as an issue on Old-Base. But uh. I could do that.
I will still find it annoying that Blood isn't actually an issue for Synths at the moment.
As far as I understand, so that doesn't actually balance it out for them.

EDIT: I'll just add making synths work correctly to my endlessly growing list and not worry about it.

NullDagaf:
> I will still find it annoying that Blood isn't actually an issue for Synths at the moment.

Also for plasmamen, don't forget about these.

NullDagaf:
Also I'm.. Fairly certain this is not a refactor, more of a balance change.
Also, the changelog is severely non-descriptive. " Changed Bubblegum Loot Drops." really doesn't tell you that much.

SpaceVampire:
> Also for plasmamen, don't forget about these.

True, but the amount of the Shaft-Miner plasmamen, or really plasmamen in general is so low, that I don't think it is really an issue. I thought about making it so that it wouldn't work if you didn't actually have blood, but then figured that it really wasn't needed. Given that its like, such an edge case, and might break other stuff.

> Also I'm.. Fairly certain this is not a refactor, more of a balance change.
> Also, the changelog is severely non-descriptive. " Changed Bubblegum Loot Drops." really doesn't tell you that much.

Made an attempt at updating it to be more helpful. What do you think about the balance for most people using it as well? I feel like it should be fine, as -150 blood is a pretty large chunk. Along with only being able to use it with large pools if iI did it correctly, which really limits its use as an escape tool. But I am willing to tweak it more if needed.

NullDagaf:
> True, but the amount of the Shaft-Miner plasmamen, or really plasmamen in general is so low, that I don't think it is really an issue. I thought about making it so that it wouldn't work if you didn't actually have blood, but then figured that it really wasn't needed. Given that its like, such an edge case, and might break other stuff.

Fair point. There's also a possibility of getting a skeleton transformation from the ash drake bottle, but again - the likelihood of that occurring is extremely slim, so no real point to make an exception for that.

> Made an attempt at updating it to be more helpful. What do you think about the balance for most people using it as well? I feel like it should be fine, as -150 blood is a pretty large chunk. Along with only being able to use it with large pools if iI did it correctly, which really limits its use as an escape tool. But I am willing to tweak it more if needed.

In my opinion, the large cost of use prevents it from being overly abused - which is a good thing. Not sure how big are "large pools", nor how common they are - but that's still better than climbing into a literal blood drop.
As for limiting it's escape potential, pretty sure a simple windup before you actually climb in would suffice. The blood cost limits it's spammability, but not nessesarily it's.. not sure if "abusability" would be the right word, just it's potential to be used as an "get out of a chase free card". Though the restriction to only big pools may be enough, not sure here.

By the way, I was wondering - what if someone faints while blood-crawling? Not saying that it costing blood to enter is a bad idea, but there should probably be either a warning, or a prevention if you try to blood-crawl with low blood. "As you lower your feet into the blood, an overwhelming faintness darts across your body.." , or something like that.

modular_skyrat/modules/modular_items/code/game/objects/items/miscellaneous.dm
```diff
@@ -7,3 +7,44 @@
...
+			perform(target)
+			return
+	revert_cast()
+	to_chat(user, "<span class='warning'>There must be a nearby source of plentiful blood!</span>")
+
+/obj/effect/proc_holder/spell/bloodcrawl/lesser/perform(obj/effect/decal/cleanable/target, recharge = 1, mob/living/user = usr)
+	if(istype(user) && user.canUseTopic(user, TRUE))
+		if(phased)
+			if(user.phasein(target))
+				phased = 0
```
Gandalf2k15:
```suggestion
				phased = FALSE
```

modular_skyrat/modules/modular_items/code/game/objects/items/miscellaneous.dm
```diff
@@ -7,3 +7,44 @@
...
+		else
+			if(user.phaseout(target))
+				phased = 1
```
Gandalf2k15:
```suggestion
				phased = TRUE
```

PR #5186, opened 2021-04-25
[Modular] Disables Cyborg Upgrade (Shrink)

SpaceVampire:
## About The Pull Request

Makes it so that you are unable to print Cyborg Upgrade (Shrink) from the MechFabs. The actual item still exists for the moment, if someone wanted to spawn it or something.

## Why It's Good For The Game

Because its a massive boost, especially to some of the cyborgs that are already really tiny, and I am tired of seeing people abuse it shift after shift. They are actually almost impossible to click, and they really don't seem to be effected by any slowdowns due to the size, making it utterly unbalanced, and just really silly.
Actuallylikefivepixels
This picture should speak for itself, really.

## Changelog
:cl:
fix: Cyborgs being able to become like actually five pixels tall.
/:cl:

PR #5211, opened 2021-04-26
[Modular] Prevents Shrink Module Usage on Small Borgs

Weblure:
## About The Pull Request

Borgs with small sprites are selectively disallowed from accepting the shrink module. This currently applies to the **default** borg, **Borgis**, **Zoombas**, **Eyebots**, that one **drone** borg, **certain** spider borgs (the weird small ones), and **Omoikane**. If you think there are any small borg types that I missed, please let me know in the comments. This PR is proposed alternative that closes #5186.

**Note: Borgs will not be able to accept the shrink upgrade until after they've selected their base borg module/type.**

## Why It's Good For The Game

Borgs that are <5 pixels in height make some people angery. This is a better alternative to completely removing the shrink module from the game, which Xyel planned to do if no alternative was provided.

## Changelog
:cl:
balance: Prevents already-small borgs from becoming even smaller with the shrink module.
/:cl:

### DIFF
--- a/modular_skyrat/modules/altborgs/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/modular_skyrat/modules/altborgs/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -1,5 +1,4 @@
-mob/living/silicon/robot //For restricting shrink module usage on small borgs
-	var/has_small_sprite = FALSE
+var/has_small_sprite = FALSE
 
 /obj/item/robot_model/proc/dogborg_equip()
 	has_snowflake_deadsprite = TRUE
@@ -29,8 +28,7 @@ mob/living/silicon/robot //For restricting shrink module usage on small borgs
 //ROBOT ADDITIONAL MODULES
 
 //STANDARD
-/obj/item/robot_model/standard/be_transformed_to(mob/living/silicon/robot/R, obj/item/robot_model/old_model)
-	. = ..()
+/obj/item/robot_model/standard/be_transformed_to(obj/item/robot_model/old_model)
 	var/mob/living/silicon/robot/cyborg = loc
 	var/static/list/standard_icons = list(
 		"Default" = image(icon = 'icons/mob/robots.dmi', icon_state = "robot"),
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/altborgs/code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -1,3 +1,5 @@
+var/has_small_sprite = TRUE
```
ZephyrTFA:
```suggestion
/obj/item/robot_model/var/has_small_sprite = TRUE
```

ZephyrTFA (changes requested):

modular_skyrat/modules/altborgs/code/modules/mob/living/silicon/robot/robot_modules.dm
Weblure: I ended up reverting because I'm not sure how to access this variable in robot_upgrades.dm when it's defined in this manner

Weblure:
Yeah, so, as-is, this is tested and confirmed to be working. The only problem is that it has an evil global variable that I'm not sure how to work around.

Gandalf2k15:
No god no not another variable to set

NullDagaf:
Oh god why
#5186 is fine as is, **WHY ARE YOU SPAGHETTIFYING CODE IF THERE'S ALREADY A CLEANER SOLUTION**

KasparoVy:
<3 Because I very much enjoy being a drakeborg of the not-huge variety but understand the ramifications of shrunk gamer sprites

Weblure:
> No god no not another variable to set

No, not really. It's a blacklist, not a whitelist. You only need to set it if you want to exclude the borg you're adding from being able to shrink. Even then, it literally only takes 2 seconds to append the variable to the bottom of your code.



> Oh god why
> #5186 is fine as is, **WHY ARE YOU SPAGHETTIFYING CODE IF THERE'S ALREADY A CLEANER SOLUTION**

Pretending that the alternative achieves the same 'solution' as this is misleading an unconstructive.

FlamingLily:
> 
> 
> Oh god why
> #5186 is fine as is, **WHY ARE YOU SPAGHETTIFYING CODE IF THERE'S ALREADY A CLEANER SOLUTION**

These two PRs are separate solutions to the same issue. They are not the same solution.

#5186 disables the module entirely, which is cleaner, codewise, and does solve the issue of small sprites going extra-small, however, it also prevents the larger sprites from being smaller within reason, and hence, that is why this alternative was provided.

Many people, myself included, would prefer not outright removing the shrink module. Sure, the outright removal fixes the issue of abuse, but now there's a new issue that it can't be used as intended, because it's been outright removed.

This PR, however, solves the main issue without creating the second one. And it may be slightly messier, but is that so severe of an issue? From looking at the code, it doesn't seem that bad.

Weblure:
This probably won't be merged until I fix the spaghetti, so I'm closing it for now

PR #5315, opened 2021-04-29
[NON-MODULAR] Changes the amount TG species get per paycheck to be equal with everyone else

DuffCreeper:
## About The Pull Request

Basically any species that were on TG by default had a payslip modifier of 0.75 (excluding humans) meaning that compared to everyone, if you were a moth, lizard, plasmaman, pod people, felnids and so forth. Were getting payed 25% less than everyone else

Also monkeys get 50% more for some reason which I didn't change cause funi monke

## Why It's Good For The Game

Basically it's kinda weird to have TG species have a negative modifier to their payslip when other species on skyrat that we added have the normal value

## Changelog
:cl:
balance: After a few complaints to the HR department, everyone will now be getting their full paycheck amount from now on
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/ethereal.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ethereal.dm
@@ -11,7 +11,7 @@
 	exotic_blood = /datum/reagent/consumable/liquidelectricity //Liquid Electricity. fuck you think of something better gamer
 	siemens_coeff = 0.5 //They thrive on energy
 	brutemod = 1.25 //They're weak to punches
-	payday_modifier = 0.75
+	payday_modifier = 1 //SKYRAT EDIT - 0.75 original value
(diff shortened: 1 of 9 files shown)


thestubborn:
what if we changed all the other species we added to have the same negative modifier

tf-4:
> 
> 
> what if we changed all the other species we added to have the same negative modifier

this is unironically the real fix

code is fine, though

H0gberg:
racism is no more

NullDagaf:
Or you could just... Comment the variables entirely.
As far as I know, these are only being changed from the default 1, so commenting the changes out would be enough.

> what if we changed all the other species we added to have the same negative modifier

Also that.

Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/5334

PR #5334, opened 2021-04-29
[MODULAR] Changes the amount our custom species get per paycheck to be equal with TG non-humans.

SpaceVampire:
## About The Pull Request

Basically any species that were on TG by default had a payslip modifier of 0.75 (excluding humans) meaning that compared to everyone, if you were a moth, lizard, plasmaman, pod people, felnids(sic) and so forth. Were getting payed 25% less than everyone else. I noticed that our custom ones didn't have this, which isn't fair. So now all non-humans get the proper pay cut.

## Why It's Good For The Game

Basically it's kinda weird to have our species not have a negative modifier to their payslip when other species on TG have the normal value. This PR is pretty much an alternative to https://github.com/Skyrat-SS13/Skyrat-tg/pull/5315, because people didn't seem to like it. I just think its a fun minor lore thing, and having worked with a 0.75 paycheck myself, its not actually an issue.
EDIT: This is also a modular change, so its better for the code-base.
## Changelog
:cl:
fix: Custom species not having pay-check mods.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/akula.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/akula.dm
@@ -25,6 +25,7 @@
 	attack_verb = "slash"
 	attack_sound = 'sound/weapons/slash.ogg'
 	miss_sound = 'sound/weapons/slashmiss.ogg'
+	payday_modifier = 0.75
 	liked_food = GROSS | MEAT | FRIED
 	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | ERT_SPAWN | RACE_SWAP | SLIME_EXTRACT
 	limbs_icon = 'modular_skyrat/modules/customization/icons/mob/species/akula_parts_greyscale.dmi'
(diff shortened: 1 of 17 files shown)


GoldenAlpharex:
While I agree with the modular change, having seen it done in the past, the pay difference between species doesn't seem to be something that's likely to be changed on /tg/'s side, so I doubt that just removing that pay difference would require too much maintaining, while I feel like the paycheck thing will have to be added for every new race we end up adding in the future, which is prone to errors.

Gandalf2k15:
Ensure felinids get a 0.5 pay mod

SpaceVampire:
> 
> 
> While I agree with the modular change, having seen it done in the past, the pay difference between species doesn't seem to be something that's likely to be changed on /tg/'s side, so I doubt that just removing that pay difference would require too much maintaining, while I feel like the paycheck thing will have to be added for every new race we end up adding in the future, which is prone to errors.

I think that are mostly correct about the first part, but we also really don't add new races very often as well. Its just a simple one line things you gotta stick there, so I don't see any errors likely to come from it. The worst case would be them not getting affected for a moment before someone added it. Its really not really a big deal either way, I just prefer modular stuff.

> 
> 
> Ensure felinids get a 0.5 pay mod

Core TG-Race, and thus outside the scope of this PR.

PR #5344, opened 2021-04-30
[Modular] Makes it so you can attack without being on combat mode.

SpaceVampire:
## About The Pull Request

Makes it so you don't need to be on combat mode to attack.
This is the normal state on TG, we changed it for some reason.

## Why It's Good For The Game

Combat mode is linked with other stuff, so with this change you have to be blocking people in order to attack anything, for example. Its really annoying because if I am clicking something with a weapon its because I want to attack it, and it is has created a couple of issues with the secondary effect of attack-based weapons can trigger without the actual attack. This change really just makes group fights annoying, while leading to cases where I am clearly trying to attack something, but can't, because I am on the wrong mode without noticing in the heat of battle. Oh, and it makes the Grasp of Rust fail to connect all the time, because that has a check to make it not target the floor when combat mode is off, but you need it on to use your actual weapon.

## Changelog
:cl:
qol: Made it easier to attack stuff.
fix: Fixed some issues related to weird attack checks.
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -170,11 +170,6 @@
 	if(force && HAS_TRAIT(user, TRAIT_PACIFISM))
 		to_chat(user, "<span class='warning'>You don't want to harm other living beings!</span>")
 		return
-	//SKYRAT EDIT ADDITION BEGIN
-	if(force && !user.combat_mode)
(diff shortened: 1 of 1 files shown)


SpookyTheFox:
Functionality with surgical tools is now changed, I suspect this PR is to blame. There is no longer the malpractice safety for failure to start a procedure with drapes.

PR #5359, opened 2021-04-30
[NON-MODULAR] Heretics, Episode II, We went to far😈 

RatFromTheJungle:
## About The Pull Request
This is episode two of the series, this one will more then likely be focusing on cost adjusting the easier abilities, to keep them in line with the other much harder, and often times, less rewarding ones.

This will be accompanied with mild buffs to their lower tier abilities, so they aren't entirely cursed with their slower ascensions, but the primary attention is toning them(Rust and Ash) back because they are slightly over-bearing compared to the other two.
## Why It's Good For The Game

keeping heretic paths vaguely aligned in power to reward, and more importantly, toning back the path(s) of least resistance hahaha get it because they're called paths.

## Changelog
:cl:
balance: All ascensions cost more, to be in line with the ideal that ascension shouldn't be a semi-common event.
balance: fiery-rebirth is less overkill on a server where most people won't actually try to fight you
balance: kicked rust ascension in the shins, due to their cheap teleport, along with potions making them nigh-unkillable.
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
@@ -504,7 +504,7 @@
 
 /obj/effect/proc_holder/spell/targeted/fiery_rebirth
 	name = "Nightwatcher's Rebirth"
-	desc = "Drains nearby alive people that are engulfed in flames. It heals 10 of each damage type per person. If a target is in critical condition it drains the last of their vitality, killing them."
+	desc = "Drains nearby alive people that are engulfed in flames. It heals a low amount of each damage type per person, favoring burn. If a target is in critical condition it drains the last of their vitality, killing them."
 	invocation = "GL'RY T' TH' N'GHT'W'TCH'ER"
 	invocation_type = INVOCATION_WHISPER
 	school = SCHOOL_FORBIDDEN
@@ -529,11 +529,11 @@
 		target.adjustFireLoss(20)
 		new /obj/effect/temp_visual/eldritch_smoke(target.drop_location())
 		human_user.extinguish_mob()
-		human_user.adjustBruteLoss(-10, FALSE)
+		human_user.adjustBruteLoss(-7, FALSE) //SKYRAT EDIT - ORIGINAL 10
 		human_user.adjustFireLoss(-10, FALSE)
-		human_user.adjustStaminaLoss(-10, FALSE)
-		human_user.adjustToxLoss(-10, FALSE)
-		human_user.adjustOxyLoss(-10)
+		human_user.adjustStaminaLoss(-7, FALSE) //SKYRAT EDIT - ORIGINAL 10
+		human_user.adjustToxLoss(-8, FALSE) //SKYRAT EDIT - ORIGINAL 10
+		human_user.adjustOxyLoss(-6) //SKYRAT EDIT - ORIGINAL 10
 
 /obj/effect/proc_holder/spell/pointed/manse_link
 	name = "Mansus Link"
--- a/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_magic.dm
@@ -504,7 +504,7 @@
 
 /obj/effect/proc_holder/spell/targeted/fiery_rebirth
 	name = "Nightwatcher's Rebirth"
-	desc = "Drains nearby alive people that are engulfed in flames. It heals a low amount of each damage type per person, favoring burn. If a target is in critical condition it drains the last of their vitality, killing them."
+	desc = "Drains nearby alive people that are engulfed in flames. It heals a low amount of each damage type per person, favoring burn. If a target is in critical condition it drains the last of their vitality, killing them." //SKYRAT EDIT - ORIGINAL: "Drains nearby alive people that are engulfed in flames. It heals 10 of each damage type per person. If a target is in critical condition it drains the last of their vitality, killing them."
 	invocation = "GL'RY T' TH' N'GHT'W'TCH'ER"
 	invocation_type = INVOCATION_WHISPER
 	school = SCHOOL_FORBIDDEN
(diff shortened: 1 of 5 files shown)


NullDagaf:
I dunno chief. Considering you gain points mostly through sacrifices, wouldn't that just lead to massive murder sprees? Honestly I'd prefer the abilities themselves toned down/reworked somehow, instead of just "haha number go big".
..don't ask how though. No idea on that part.

NullDagaf:
Also yes, i know, reality thingymajigs are a thing and you can gain points from them too. But they pop up way too infrequently to be used as a reliable method of leveling up.

YakumoChen:
Kind of agree, keep the costs of all paths in line with each other, but if you need to balance them, balance the effects instead.

RatFromTheJungle:
> 
> 
> Also yes, i know, reality thingymajigs are a thing and you can gain points from them too. But they pop up way too infrequently to be used as a reliable method of leveling up.

The current issue, as it stands is that while flesh and void are 'weak and slow'  and take comparatively much more effort, ash and rust are perfect for speedrunning, especially if two work together, and can be rushed to ascension, and the issue lies with ease of  the ascension itself. And do recall that most abilities cost 1 point, and sacrifices just got rebuffed to giving 3 per sacrifice, while the book also starts with two, this at most adds 1 more kill to their list, and as I expect, more then likely won't actually slow them down much, or even force them to particularly kill more then they have been, but might be just enough to delay them into getting caught before becoming out-of-control.

Though, i will add on, that it is more specifically rust, and I will take a fair gander at swapping around it's stuff, and toning back down the cost in kind.

SpaceVampire:
The rust ascension is supposed to be about making you super hard to kill. There isn't any reason to nerf it.
If you want to kill them you need to use either a lot of firepower, or make it so they aren't standing on rust.
Because the moment they step off it, they loose all the healing and immunities already.
It should require actually playing around that, as it isn't too hard to do.
Especially when the other effect of ascending is still broken.

EDIT: Making all of those cost more is good.
But you really don't need to nerf it at the same time, if you are doing that.

RatFromTheJungle:
> 
> 
> The rust ascension is supposed to be about making you super hard to kill. There isn't any reason to nerf it.
> If you want to kill them you need to use either a lot of firepower, or make it so they aren't standing on rust.
> Because the moment they step off it, they loose all the healing and immunities already.
> It should require actually playing around that, as it isn't too hard to do.
> Especially when the other effect of ascending is still broken.


the ability itself, is ludicrously overpowered, ascension or not, but when you tack on the other abilities in the same 'path'? no single antag, 
 **who can teleport, freely,**
Deserves to be healing onpar with adminordrazine, or even past it with the wounded soldier potion, let alone as an antag that typically spawns in sets of 3.

Even a changeling who has killed an absorbed 3-4 other changelings isn't half of overloaded as that. hell, I doubt a wizard with 30 would be as rediculous as that.

They get 3 separate abilities to spread rust, their ascension itself not coating the entire station is not justification to leave them bordering on unkillability, they're still tanky, just not ludicrously so anymore.

Do not forget they are starting with 2 points, and now get 3 per sacrifice, they don't need to **also** be unkillable, with a now laughably easy to get ascension (thanks to being able to load up on more spells) while getting a far-beyond premium choice of spells.

And, upon a triple cross examine, this also brings it inline with the other ascensions, bar the "much harder" to get night-lord.

YakumoChen:
Thats a misleading post.

The rust path does not get any free movement based abilities without diverging heavily, which costs points and delays ascencion.

Rust is about turf control and juggernaut style play as opposed to other paths with movement abilities. Rust doesnt even get any crowd control. Tanking is really all they have.

Sacrificing people on skyrat is hard. Murder is difficult to get away with, much less moving bodies far away.

RatFromTheJungle:
> 
> 
> Thats a misleading post.
> 
> The rust path does not get any free movement based abilities without diverging heavily, which costs points and delays ascencion.
> 
> Rust is about turf control and juggernaut style play as opposed to other paths with movement abilities. Rust doesnt even get any crowd control. Tanking is really all they have.
> 
> Sacrificing people on skyrat is hard. Murder is difficult to get away with, much less moving bodies far away.

"diverging heavily" like 3 points, max, just buy across to ash and pick up the starter ash jaunt, every one of their income sources has been buffed, and lord forgive they team up.

They start with two points, get two per fracture(last i tried them), and 3 per sacrifice, the post is not misleading, they have an almost excess of points available, which just so happens to give them access to far more abilities and 'neutral' items then they're supposed to have, even off just two fractures.

SpaceVampire:
I would rather have the ascension costs raised if you think they are too easy, instead of the cap-stone nerfed.
That being said, its also the only one that doesn't come with any kind of offensive power at all.
Its just about being tanky, and just when you are currently standing on rust.
It already has a number of ways to counter it, by just making the floor something else.
The damage output of the weapons here is a lot more then it is on normal TG as well.

RatFromTheJungle:
> 
> 
> I would rather have the ascension costs raised if you think they are too easy, instead of the cap-stone nerfed.

this is the only one that's a problem, and I don't think bumping it up to be far above 5 would do much good, when nerfing it to be in line is easier, and a safer choice.
> That being said, its also the only one that doesn't come with any kind of offensive power at all.

they can just buy offensive abilities, especially with the free "kill whoever" ticket ascending gives you, at this rate.
> Its just about being tanky, and just when you are currently standing on rust.

Tanky and borderline unkillable with 9 people wailing on you are two **very** different things.
> It already has a number of ways to counter it, by just making the floor something else.

I don't believe any amount of 'counterplay' justifies how ludicrously overloaded this is, especially not one that relies on "lol just magically know they only work on rust" (especially one that's as easily countered as said heretic just teleporting away, and throwing down an entropic plume) and all this healing is only made worse by the flat 50% brute ignoring trait from ascending, which..
> The damage output of the weapons here is a lot more then it is on normal TG as well.

turns all this 'damage outpot' of for example, the ostwind, which does 30 per shot at 3 shots a second, down to 2-4 brute pershot with just their heretic robes.

RatFromTheJungle:
coming back 2 this l8ter

Kroto-Ker:
too* :)

RatFromTheJungle:
> 
> 
> too* :)

u r banned

Gandalf2k15:
reopen when is done

PR #5364, opened 2021-04-30
[Modular] Drops an ICBM On DS-1

unit0016:
## About The Pull Request
Drops an ICBM On DS-1.
There is now a new DS-1. It's a rare space ruin. (0.1 weight compared to every other ruin's 1 weight. Very very unlikely to spawn naturally.) Still has most everything old DS-1 had (minus the sauna, RIP) - except the cruiser.
This now means DS-1 is off the fuckin' Centcom z-level, and thus a lot less of a problem child.
This also plays tetris with the second centcom z-level to open up more space. Honk.
image

## Why It's Good For The Game
Putting a normal ghostrole on a centcom z-level was a bad idea
this still doesn't fix the ghost cafe's possibility to interact with the station, but whatever.

## Changelog
:cl:
add: DS-1's layout have been entirely reworked from the ground up.
add: New DS-1 has been added, now (very rarely) appearing in space. Don't spacetide, kids. Seriously.
del: Old DS-1 has been removed, along with the assault ops cruiser.
fix: ...Which Inadvertently fixed all the issues that DS-1 being on the same z-level as the Interlink and ghost cafe caused.
server: DS-1 being classed as a ruin now means it can be completely disabled, or adminspawned elsewhere, etcetera.
server: Do note that admin spawning the syndicate frigate via the shuttle manipulator is unsupported and WILL delete syndiecomms due to how the shuttles were defined. Be careful!
/:cl:


NullDagaf:
I assume the walls aren't indestructible.

That's... All the commenting you're gonna get from me at 2 AM.

Gandalf2k15:
Maybe replace cybersun with this?

unit0016:
> Maybe replace cybersun with this?

Cybersun's a guaranteed spawn (for... some reason. I thought I changed it but guess not). I feel like a whole syndicate station being nearby getting normalized wouldn't be... The best. Should be rare.

Alternatively I can wait on the liquids PR and have it be an ocean ruin.

Gandalf2k15:
Honestly, I don't think it's that bad. Remove the OP gamer loot(like, l6saws, and it might be good.

hugmeorelse:
Hi. I'm the host. DM me. We'll talk more.

unit0016:
> Honestly, I don't think it's that bad. Remove the OP gamer loot(like, l6saws, and it might be good.

I did. all the assault ops gamer loot is long gone, closest there is now is the three c20rs, energy swords, and energy shields I kept for the armory. Along with the two blood reds for the admiral and CMAA respectively.

KathrinBailey (changes requested):
See if you can work on the generally blocky shape of the exterior and make it less... borg cube-like.
Chuck it in an asteroid if you're feeling adventurous, up to you. Just make it less blocky.

unit0016:
Did exactly that.

unit0016:
okay, well. Linters are failing, and I have no goddamn idea what the problem is

RIP.

PR #5372, opened 2021-05-01
[MODULAR] Improves the code for IPC organs and the charging thereof.

ZephyrTFA:
## About The Pull Request

Does some basic code improvements on the code for ipc organs. Most notably this changes how charging works and **by default this is now SLOWER** but can be easily changed by modifying the defines above the proc.

## Why It's Good For The Game

Because bad code is bad; and good code is good.

## Changelog
:cl: ZephyrTFA
code: Improved IPC/Synth code; note that you will now charge slightly slower now.
/:cl:


modular_skyrat/modules/customization/modules/surgery/organs/ipc.dm
```diff
@@ -170,44 +170,61 @@
...
 			break
-		if(A.cell.charge == 0)
-			to_chat(H, "<span class='warning'>The [A] doesn't have enough charge to spare.</span>")
+		// Is the APC almost empty?
+		if(target_apc.cell.percent() < 10)
+			to_chat(user, "<span class='warning'>[target_apc]'s emergency power is active.</span>")
 			break
-		A.charging = 1
-		if(A.cell.charge >= 500)
-			do_sparks(1, FALSE, A)
```
tf-4, line 201: Is this now removed? Seems like a somewhat notable balance change if it is

ZephyrTFA: All that code did is show sparks each charge cycle, I can add it back if you want

SpaceVampire: I think sparks when charging is good.

Gandalf2k15: Add it back yes.

tf-4: Does do_sparks() not cause fires? I always assumed it did. I don't generally play synths though. I mean, if it doesn't cause fires then it's kinda w/e, but if it does then it's also kind of, dare I say it, power creep 😔

ZephyrTFA: It's added back.

tf-4 (approved):

PR #5398, opened 2021-05-01
[MODULAR] Removes antimagic protection from the ez-craft crucifix

RatFromTheJungle:
## About The Pull Request
Removes antimagic properties from the restrictionless cross, now that two antags use magic.

## Why It's Good For The Game

an item that can be crafted by anyone, with just two sheets of ore, that almost completely negates two antags probably isn't a good balancing idea.

## Changelog
balance: commented out antimagic from the crafting menu crucifix.
/:cl:


Gandalf2k15:
delete the code not comment it out

PR #5434, opened 2021-05-03
[MODULAR] TidalLock Chasms!

Cheshify:
## About The Pull Request
TidalLock.dmm used to have a mass of asteroid rock in the center, separating the Icy and the Hot. Now, there's a massive chasm, qdel aplenty for miners who aren't careful. Luckily, there are a number of bridges with railings, if you don't want to make your own bridge (remember, you can't walk on lattice, only catwalks)
gremblo12

## Why It's Good For The Game

First of all, it actually looks really cool, and removes some stupid atmos diffs, secondly
image

## Changelog
:cl:
add: TidalLock has had a massive earthquake, splitting the planet in half! Be careful where you step.
fix: TidalLock shouldn't have as many atmos diffs.
/:cl:


Cheshify:
[[Thoughts on this would be nice, I know some people don't like qdel, so if you have suggestions please send em]]

KasparoVy:
Either works for me, I don't have a problem with the middle part personally but a chasm seems like a cool idea.
Could we please get a mining shuttle console so we can call the shuttle back down from the station? Currently it looks like only public mining has a way to call its dedicated shuttle down but not actual mining

Cheshify:
> Either works for me, I don't have a problem with the middle part personally but a chasm seems like a cool idea.
> Could we please get a mining shuttle console so we can call the shuttle back down from the station? Currently it looks like only public mining has a way to call its dedicated shuttle down but not actual mining

I'll do it, thanks!

KasparoVy:
Cheers, it's greatly appreciated!

Cheshify:
> Cheers, it's greatly appreciated!

Just did it, appreciate the feedback!

PR #5472, opened 2021-05-05
[MODULAR] Adds Ashwalker/Golem Fingers Quirk

KasparoVy:
## About The Pull Request
Adds negative quirk that works exactly like Ashwalker or Golem fingers which...

Prevents you from using Modular Computers (including tablets): https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/modules/modular_computers/computers/item/computer_ui.dm#L16
Also stops from using laser pointers: https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/game/objects/items/devices/laserpointer.dm#L81
And non-adapted guns: https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/modules/mob/living/carbon/human/human_helpers.dm#L103

## Why It's Good For The Game
Improves character customization, provides another option for a strong negative quirk mindful of Ashwalker/other backgrounds.

## Changelog
:cl:
add: Adds a new 8-point negative quirk preventing you from using most firearms, modular computers/tablets and laser pointers like an Ashwalker or Golem. KAs can be adapted w/ a modkit. You cannot be command, security, engineering, atmos, doctor, virologist, chemist, roboticist or scientist with this trait.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -2,5 +2,6 @@
 #define JOB_UNAVAILABLE_SPECIES 7
 #define JOB_UNAVAILABLE_LANGUAGE 7
 
-#define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE
-#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE
+#define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE
+#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE
+#define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE
--- a/modular_skyrat/master_files/code/datums/traits/negative.dm
+++ b/modular_skyrat/master_files/code/datums/traits/negative.dm
@@ -36,9 +36,9 @@
 
 /datum/quirk/ashwalkertalons
 	name = "Chunky Fingers"
-	desc = "Your digits are thick and tough, making firing non-adapted firearms impossible."
-	gain_text = "<span class='notice'>Your fingers feel thicker and slightly less dextrous. Trigger guards seem unusably small now, and might need modification before you can fire a gun.</span>"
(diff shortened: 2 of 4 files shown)


KasparoVy:
Open to further mechanical options to enforce fair use of the SSD toggle, some suggestions I've received have been:

1. Have it turn off automatically after moving at all, probably using signals/component?
2. Trigger temporary paralysis effect that lasts until 2-3 seconds after AFK indicator is disabled
3. Make the person vanish entirely with the 'void' effect.
4. Warn about attacks made by the player when the AFK indicator is on + a short time after it's toggled off

I opted to just put in the 5 second toggle timer as it was the most straightforward and least invasive/severe, and was the amount of effort I'm able to put in right now

Gandalf2k15:
Really not a fan of the SSD toggle at all, really not a fan, it should only ever trigger on logout. This, even with the 5-second thing, is far too abusable as you can't prove someone was actually SSD.  Meaning it's incredibly easy to cheese and you can get away with it.

KasparoVy:
> Really not a fan of the SSD toggle at all, really not a fan, it should only ever trigger on logout. This, even with the 5-second thing, is far too abusable as you can't prove someone was actually SSD. Meaning it's incredibly easy to cheese and you can get away with it.

I agree with your points, is our posture then to enforce that if you're gonna go SSD, close the client or accept the potential shenanigans? Happy to gut that part from this PR, thought it would be nifty to have but in my opinion too there is no better way than to just have people actually click that X and reconnect when they're back given what you said @Gandalf2k15 

If it is then perhaps a notification/reminder might be welcome, otherwise I'm just gonna make sure the folks I know are up to speed on this to avoid problems

KasparoVy:
@Gandalf2k15 Removed SSD indicator change from this PR, I'm not keen on stirring up more trouble just for a bit of convenience

Sigmar-alkahest:
"Adds negative quirk that works exactly like Ashwalker or Golem fingers, preventing you from using all non-adapted firearms."
YEESSSSSSSSSS! YES THIS IS SOMETHING I WANTED SINCE IT WAS ADDED TO THE GLOVES!

KasparoVy:
Converting to draft pending a change to the quirk's point value since TRAIT_CHUNKYFINGERS...

Prevents you from using Modular Computers (including tablets)- https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/modules/modular_computers/computers/item/computer_ui.dm#L16
Also stops from using laser pointers - https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/game/objects/items/devices/laserpointer.dm#L81
And non-adapted guns: https://github.com/Skyrat-SS13/Skyrat-tg/blob/a00c7131e9d789324edead7be7aebadaca8f1e06/code/modules/mob/living/carbon/human/human_helpers.dm#L103

KasparoVy:
Back to draft, adding this trait to the 'forbidden' list for certain jobs

KasparoVy:
Alright done, happy to revise what jobs are blacklisted -- asked around and checked who uses guns/modular computers often. Only one that comes to mind I might be able to take off the list is doctor but I feel their dexterity should be required

PR #5481, opened 2021-05-05
[SEMI-MODULAR] The Rodstopper! Who needs a Research Director?

Cheshify:
## About The Pull Request

Don't you hate it when your Research Director is off doing something else, and you're stuck with a looping rod that has absolutely no way to be removed? Well now, with the new patented Nanotrasen tech: The Rodstopper, you can halt that nasty space-stick right in it's track. Warning: Nanotrasen is not to be held liable for any damages cause by an unstoppable force meeting an immovable object.

rodstopper

Say hello to your new best friend when the RD is out of town. 

~~Dear god this is probably shitcode but it's modular damn it.~~

## Why It's Good For The Game

Xyel wanted an alternative to the RD's suplex ability, and giving engineering a means of handling a looping rod is all cool with me.

## Changelog
:cl:
add: The Rodstopper, a new machine able to halt looping immovable rods, can be found within the CE's locker.
/:cl:


Cheshify:
I'm aware it's a dreaded codersprite. I also do not care since the thing literally explodes within seconds of being built due to being designed to get hit by a rod.

FlamingLily:
Maybe like, I dunno for fluff have like an emergency closet that contains one board that only unlocks on orange alert?

And maybe have it not be printable, cause it's supposed to be like this SUPER ADVANCED machine.

And like, you **KNOW** some chucklester is going to put one of these at dorms

Cheshify:
> Maybe like, I dunno for fluff have like an emergency closet that contains one board that only unlocks on orange alert?
> 
> And maybe have it not be printable, cause it's supposed to be like this SUPER ADVANCED machine.
> 
> And like, you **KNOW** some chucklester is going to put one of these at dorms

It has a few uses (also stops rod-wizards), and rods can happen more than once. I already had to make the choice between shitcode and modularity, and I'm not going back on that. Also if someone puts one of these in dorms that'll be hilarious. I THINK the fabrication and building cost should be enough to dissuade people from building them en-masse, especially because of how ugly they are.

Edit: Let's talk more in the discord

modular_skyrat/modules/rod-stopper/code/immovable_skyrat.dm
```diff
@@ -0,0 +1,9 @@
+/obj/effect/immovablerod/Bump(atom/clong)
+	var/should_self_destroy = FALSE
+	if(istype(clong, /obj/machinery/rodstopper))
+		should_self_destroy = TRUE
+	. = ..()
+	if(should_self_destroy)
```
Gandalf2k15, line 6: maybe make one of those BOH explosions when it happens

Gandalf2k15: Infact yes, do that instead

Cheshify: That's a whooole lot of station destroyed, but I'll look into it.

Cheshify: Welp, it's in.

hugmeorelse:
BoH bomb explosion I don't think quite works. What if you instead have a way to gradually disintegrate the rod as it passes through, until it becomes nothing? It gives engineers and scientists something to do, and it prevents rods from stopped immediately. We may need a different sprite to illustrate this though.

Cheshify:
> BoH bomb explosion I don't think quite works. What if you instead have a way to gradually disintegrate the rod as it passes through, until it becomes nothing? It gives engineers and scientists something to do, and it prevents rods from stopped immediately. We may need a different sprite to illustrate this though.

Right, I don't think that's _super possible_, this is really just an alternative way to stop a looping rod without admin intervention. Should I contact a spriter for something like that?

NullDagaf:
I think this is just.. Well, it works, but a better solution would probably be to just remove the rods entirely. That, or disable them as an event.

Gandalf2k15:
The reason for having it BoH bomb is because we don't want people cheesing this event, since unless there is only ONE of these boards(in the RD's locker) then it's gonna be easy to cheese rods. Having it BoH bomb means that people will have to think about where they're placing them(in space).

Gandalf2k15:
In fact, yes, have the board spawn in the RD's locker and have it non-researchable.

Cheshify:
> In fact, yes, have the board spawn in the RD's locker and have it non-researchable.

got it, I'll make the change soonish

gamermomentum:
> In fact, yes, have the board spawn in the RD's locker and have it non-researchable.

If the reasoning for this is "if you don't have an RD/your RD can't deal with the rod for X reason" shouldn't it be put somewhere outside of the RD's office? 
Because ostensibly you'd need the Captain/a borg to get it in that situation(without just breaking the locker ofc) and it'd just be kinda unintuitive.

Gandalf2k15:
Make it the CE's locker!

Cheshify:
> Make it the CE's locker!

Sure, when I can

FlamingLily:
> 
> If the reasoning for this is "if you don't have an RD/your RD can't deal with the rod for X reason" shouldn't it be put somewhere outside of the RD's office?
> Because ostensibly you'd need the Captain/a borg to get it in that situation(without just breaking the locker ofc) and it'd just be kinda unintuitive.

The main difference is that without an RD, there is no way to get the skillchip without admin intervention, while if there is a board, sure, you'd still need the locker to be opened, that can be done by anyone with access, RD not required

Gandalf2k15:
Fine.

PR #5513, opened 2021-05-07
[FIX] SQL Changelog cleanup AGAIN

ORCACommander:
## About The Pull Request

Recent Upstream PR pushed a duplicate entry into the changelog

## Why It's Good For The Game

don't need the same log entry twice

## Changelog
:cl:
fix: Cleaned up the SQL Changelog AGAIN
/:cl:


GoldenAlpharex:
If only our changelog was working :^)

ORCACommander:
> 
> 
> If only our changelog was working :^)

yeah I have no idea why that is broken on your end. running a direct downstream I get merge conflicts if i let the compile changelogs action run so i turned it off

PR #5536, opened 2021-05-07
[modular] evac shuttle alignment

werewolfboyfriend:
## About The Pull Request
it's been bugging me how the default evac shuttle doesn't align to the standard departures docking arms, so I adjusted the shuttle as well as the sprites for it, so the doors look like they should go there.
screenshot from sdmm:
Screenshot 2021-05-07 174429
from in-game on meta:
Screenshot 2021-05-07 182706
as you can see, I also adjusted the prisoner side of the brig to be slightly larger, instead of just losing a seat to the door

## Why It's Good For The Game
the shuttle will feel less self-conscious, i think

## Changelog
:cl:
fix: The default evac shuttle has been adjusted to fit its doors to standard departures docking arms.
/:cl:


Gandalf2k15:
amazing <3

PR #5541, opened 2021-05-08
Deltaprison tiny fix

Aurelien3189:
Makes a few additions to the deltastation prison:
image

1) Replaced the book visitor console with the one that can actually print books
2) Added a push broom to the cleaning pile, a plant analyzer to the garden, a trash bin to the cafeteria, and a set of candles for when the lights inevitably go out
3) Made the floor decals in the right side cells match the left side cells (my bad)
4) Put some lattices in front of the grille shields because why not


Aurelien3189:
Huh, that's odd, I've never had that error before. What is it?

PR #5627, opened 2021-05-13
[NONMODULAR] Should fix sechuds wrt. alt titles

FlamingLily:
## About The Pull Request
fix

## Why It's Good For The Game
fix

## Changelog
:cl:
fix: Sechuds and Alt titles
/:cl:


FlamingLily:
Actually there both:
is no bug
this would mean IDs w/o a trim but still labelled as a job would mess up on sec huds

PR #5665, opened 2021-05-13
[NONMODULAR]Slaugher demons no longer RR

Vizoplez:
## About The Pull Request
RR bad, funny bus good

## Why It's Good For The Game
yes

## Changelog
:cl:
tweak: (S)laughter demons no longer gib
/:cl:

### DIFF
--- a/code/modules/antagonists/slaughter/slaughter.dm
+++ b/code/modules/antagonists/slaughter/slaughter.dm
@@ -1,5 +1,5 @@
 //////////////////Imp
-
+//FILE WAS SKYRAT EDITED TO PREVENT ROUND REMOVAL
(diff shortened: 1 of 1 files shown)


NullDagaf:
Besides the "slaughter demons should rr" rants I could go on about.
Enlighten me, how does this affect laughter demons? As far as I know, they specifically cost less, BECAUSE they don't gib.

Unless we don't have these, in which case.. IDK, still adjust their cost then. [insert rant about how slaughter demons should rr]

Vizoplez:
fails due to unrelated error

Funce (approved):
_Re-runs job_
What error?

PR #5441, opened 2021-05-04
[SEMI-MODULAR] Re-adds the stock market for cargo

DuffCreeper:
## About The Pull Request

So uh, this was stupidly easy to bring back. Anyway this basically adds the stock market for cargo to fuck with and either earn money or lose it

## Why It's Good For The Game

So this gives cargo a new (old) gimmick to fuck with and another way to earn money other than people throwing stuff at them.
Nothing is changed about this and should perform exactly the same as the day TG removed it.

You will find the circuitboard can either be printed or be found in the QMs locker roundstart. Reason being so cargo techs dont just yeet all the money into stocks round start, having it in the QMs locker gives them some form of control.

But uh yeah, not sure what else to say about this. Anyone who used it before will know exactly what it is

## Changelog
:cl:
add: Re-added the stock market to cargo. You can find the board in the QMs locker round start or print it. Remember, sell high buy low
/:cl:

PR #5697, opened 2021-05-14
Reverts #5441 (Stock Market)

YakumoChen:
## About The Pull Request

Reverts #5441

## Why It's Good For The Game
image
**No.**

## Changelog
:cl:
del: Removed stock market.
/:cl:


DuffCreeper:
Cargo got rich once so the only option is to remove it?

The code has numerical values that can be changed.
Ontop of that it's a 'simulated' stock market that just uses RNG. If them becoming rich is a recurring problem then knowing how and why would be more beneficial than just removing it. For all we know it can be a result of someone just doing regular stock things. Selling high, buying low, buying the news etc.

It can't be manipulated either from an admin side by varediting to cheat since there's nothing to look at. As someone who has used the stock market before irl, there's plenty of factors that can change a price and without knowing anything about how they got the end result. I can't determine if it's someone just dedicating themselves to it, someone getting insanely lucky or a possible bug causing a stock to spike further than it should.

Arctaisia:
Suddenly removing something because one bad thing happened is not a very mature course of action, as a developer you should be seeing how to refine it to be better. As a staffmember, I expect our developers to know this, and to only remove when something is beyond repair.

YakumoChen:
Getting a million credits is a litmus test for cargo exploits, much less being able to get dozens of them. It's clear just by that screenshot alone that something like the stock market is likely to be wildly imbalanced for cargo. They have several ways of gaining credits legitimately, especially bounties. Something likely to be that wildly unstable is not what cargo should consider healthy for a round.

YakumoChen:
I remember when it existed on TG as well and this was a regular occurrence. It's not any kind of bug, it's just the stock market being wildly abusable to gain far more credits than any one round should have.

DuffCreeper:
> Getting a million credits is a litmus test for cargo exploits, much less being able to get dozens of them. It's clear just by that screenshot alone that something like the stock market is likely to be wildly imbalanced for cargo. They have several ways of gaining credits legitimately, especially bounties. Something likely to be that wildly unstable is not what cargo should consider healthy for a round.

Except for it's a one case scenario, while more are expected to happen the question is, how often. If it ends up being once every.. week etc then it's rare as fuck. Anyway it's not much of an exploit in this case due to as I said it being a RNG system. A stock can only go up or down in simple terms. News affects that but there's no way of being able to exploit the news because the game will essentially RTD for an outcome when it happens.

It needs to be balanced yes but I need something more than a photo of rich cargo. Because there's a lot of factors and without knowing those factors then I can't hit a bullseye on the exact problem.

YakumoChen:
It's from round 1738 if there's any logging (you did add logging, right?), but when I ahelped why it happened the only response I got was "stock market moment".

Or you can ask TG, or Goofball, the one who made this, but you're unlikely to get any good answers from either since goof is a pretty one-track-mind coder and TG removed it almost immediately for being unbalanced as hell.

GoldenAlpharex:
After looking a bit at the code with Duff, we believe we might have found an easy fix for this, so this should hopefully not be necessary.

Lyroy:
If it helps any, I was the QM taking care of the stock market during that round. 

I practically spent all the shift micro-managing them, but most importantly I noticed some of the stocks were just going crazy. For example (as I explained on the ahelp) one that was sitting at 13.3 suddenly jumped to 11000 value, and from there instantly up to 450000 (what I assume is the cap). It was always the same 2 or 3 stocks, fluctuating between comfortably sitting at the cap, going down to around 300k and then shooting back up. The other values seemed to be fine.

YakumoChen:
If you can, that's great, but I really don't believe it's healthy for cargo. If they're getting more than hundreds of thousands of credits in a single shift, that's too many.

GoldenAlpharex:
> If you can, that's great, but I really don't believe it's healthy for cargo. If they're getting more than hundreds of thousands of credits in a single shift, that's too many.

We're going to lower the cap from 450 000 credits to like 2000-3000 credits per share.

DuffCreeper:
#5702 Market is capped to 5k. If it continues to be an issue that becomes constant i'll drop it further to 3k

PR #5777, opened 2021-05-18
[non-modular] Disables broken shuttle catastrophe event

NotRanged:
## About The Pull Request

Doesn't work with our current shuttle, setting occurrences to 0 makes it not roll and fail every now and then.

## Changelog
:cl:
fix: Made shuttle catastrophe unable to roll and fail, wasting an event roll.
/:cl:

### DIFF
--- a/code/modules/events/shuttle_catastrophe.dm
+++ b/code/modules/events/shuttle_catastrophe.dm
@@ -1,8 +1,8 @@
 /datum/round_event_control/shuttle_catastrophe
 	name = "Shuttle Catastrophe"
 	typepath = /datum/round_event/shuttle_catastrophe
-	weight = 0 // Skyrat edit -- disabled due to our custom shuttle
-	max_occurrences = 1
+	weight = 10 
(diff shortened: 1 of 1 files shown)


Gandalf2k15:
Can you see about fixing it?

NullDagaf:
> Can you see about fixing it?

Don't fear removals, for they're the step towards enlightenment.

YakumoChen:
I think I specifically remember commenting that the new shuttle messed with alternate shuttles and it was never fixed

NotRanged:
> Can you see about fixing it?

As in, you want shuttle catastrophe to replace your custom shuttle to a random one again?

SpaceVampire:
I do also dislike that the new current shuttle seems to be the same for each station.
Wouldn't mind just returning to the old system, and only keeping that kind of design for the shuttle that takes people to the station.
Its really hard to tell what is a window, what is a wall, and how the blocks actually come together.
Because every part of it is just called "shuttle wall" for some reason.

Gandalf2k15:
Fix the shuttle event.

PR #5438, opened 2021-05-03
[Non-Modular] Stops borgs from being a glorified departmental comms loudspeaker

DuffCreeper:
## About The Pull Request

Hey turns out they use a station bounced radio which is why they're like a freebie to listening to departmental drama

## Why It's Good For The Game

Well lets say being an AI shell isn't a good time when you're trying to vibe somewhere public and everyone can hear how command likes their burrito with extra beanz or how security thinks the funny gray man is looking sus

Jokes aside it stops comms bleeding out when you know.., you're a robot... communications would be internal...

Also you can still hear comms if you're on the same tile as the borg so for example. You're riding their back or underneath them *dab

## Changelog
:cl:
fix: Fixed borgs being a loudspeaker for people to listen in on departmental stuff
/:cl:

PR #5787, opened 2021-05-19
Revert "[Non-Modular] Stops borgs from being a glorified departmental comms loudspeaker"

Vizoplez:
Reverts Skyrat-SS13/Skyrat-tg#5438

Broke shit, needed to be tested.

### DIFF
--- a/code/game/objects/items/devices/radio/radio.dm
+++ b/code/game/objects/items/devices/radio/radio.dm
@@ -386,7 +386,6 @@
 	subspace_transmission = TRUE
 	subspace_switchable = TRUE
 	dog_fashion = null
-	canhear_range = 0 // Skyrat Edit - Stops borgs being a loudspeaker and contains it to the tile they're on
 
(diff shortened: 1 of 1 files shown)


DuffCreeper:
Would it be a good time to mention that what I did has nothing to do with subspace TX

`var/canhear_range = 3  // The range around the radio in which mobs can hear what it receives.`

All I edited was basically a range of the radio so it's contained to one tile. Either way it works, Subspace TX has to be turned off for borgs to speak on the selected frequency if there's no tcomms
image

Vizoplez:
> Would it be a good time to mention that what I did has nothing to do with subspace TX
> 
> `var/canhear_range = 3 // The range around the radio in which mobs can hear what it receives.`
> 
> All I edited was basically a range of the radio so it's contained to one tile. Either way it works, Subspace TX has to be turned off for borgs to speak on the selected frequency if there's no tcomms
> image

wait- im retard

PR #5789, opened 2021-05-19
[modular] [danger!] Character Customization Colors now uses full depth RRGGBB hex instead of converting to RGB

NotRanged:
## About The Pull Request

Currently, when saving a color for anything in your character customization, it gets reduced from the #RRGGBB value you saved, to the depth of #RGB. For example, saving your wing color as #A46D2E would be stored as #AA6622 instead.

With this PR, colors are properly saved without such losses. This allows for a much, _much_ wider range of coloring and shading on your characters, especially noticeable with shades of black and white.

Danger PR because this makes changes to prefs and color saving, and reverting this might break people's colors. Nothing they can't fix by reselecting, but still.

## Why It's Good For The Game

Better customization.

## Changelog
:cl:
code: Selected colors for your character bodyparts are now stored at full hexadecimal depth, instead of duplicating the first hex character. Basically, it will no longer round to a different shade than the one you selected.
/:cl:

### DIFF
--- a/code/__HELPERS/~skyrat_helpers/unsorted.dm
+++ b/code/__HELPERS/~skyrat_helpers/unsorted.dm
@@ -13,3 +13,11 @@
 	M.Scale(squishx, squishy)
 	animate(src, transform = M, time = halftime, easing = BOUNCE_EASING)
 	animate(src, transform = OM, time = halftime, easing = BOUNCE_EASING)
+
+/proc/validate_mutant_RGB(color)
+	// 19-5-2021: making our customization work with RRGGBB instead of RGB would break every stored character without this. Whelp!
+	// This is because of how colors are now stored as RGB, with length 3. They need to become length 6, otherwise handling mutant matrixed color in species.dm breaks.
+	// It adds '00' as alpha to the end of the color, which would give it length 5, which readRGB() can't handle.
+	if(length(color) == 3)
+		color = sanitize_hexcolor(color, 6) // Turn the 3 length RGB into 6 length RRGGBB by duplicating the values.
+	return color
--- a/modular_skyrat/modules/customization/modules/client/preferences.dm
+++ b/modular_skyrat/modules/customization/modules/client/preferences.dm
@@ -1767,7 +1767,7 @@ GLOBAL_LIST_INIT(food, list(
 					if(new_color)
 						if(!loadout[path])
 							return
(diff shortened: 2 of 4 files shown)


KasparoVy:
@NotRanged I fucking love you, by the gods this has frustrated me to no end-- soooo much flexibility now! Thank you very, very much.

> Danger PR because this makes changes to prefs and color saving, and reverting this might break people's colors. Nothing they can't fix by reselecting, but still.

Might be able to run an SQL query which squashes the colours but that might be too much of a PITA.

Ideally this is merged and never reverted if it works like I think it does 😁

Draganfrukts:
imagen

Gandalf2k15:
I recall doing something like this before and it utterly fucked everything

Funce:
> It fucked everything until I added the helper proc and the extra 0 to the stuff in species.dm

_slaps forehead_
Of course, the Alpha Channel has to be two digits as well. That would explain it

Azarak:
i'd suggest properly converting all the colors to 6 and having savefile conversion instead of using that helper proc every time the colors are needed, it's unnessecary overhead and bad practice, plus it would allow the further use of short hex. The other instances where you've got color inputs to there are /datum/species/ procs that pass colors for all sorts of species

NotRanged:
> i'd suggest properly converting all the colors to 6 and having savefile conversion instead of using that helper proc every time the colors are needed, it's unnessecary overhead and bad practice, plus it would allow the further use of short hex. The other instances where you've got color inputs to there are /datum/species/ procs that pass colors for all sorts of species

Not sure how I'd handle savefile conversion for this, it'd have to make a check and save somewhere every time someone's savefile is loaded until everyone's colors are fixed?

Azarak:
> 
> 
> > i'd suggest properly converting all the colors to 6 and having savefile conversion instead of using that helper proc every time the colors are needed, it's unnessecary overhead and bad practice, plus it would allow the further use of short hex. The other instances where you've got color inputs to there are /datum/species/ procs that pass colors for all sorts of species
> 
> Not sure how I'd handle savefile conversion for this, it'd have to make a check and save somewhere every time someone's savefile is loaded until everyone's colors are fixed?

You'd keep a track of a savefile version to make sure you only run the conversion once. TG has one like that, but not to interfere with tg's conversions it may be wise to create one like that for skyrat stuff

Gandalf2k15:
Do as azarak says, discuss in discord.

PR #5712, opened 2021-05-15
[SEMIMODULAR] Station Names for each map!

FlamingLily:
## About The Pull Request
This PR changes the code for station names. Depending on config settings, station names can be preset (a la "Space Station 13) as they could before, they can be randomly generated, with split configs for standard random generation and holiday random generation, and the main feature of this PR, using the current map's name (in the map's .json) as the station's name.

For instance, if the map is "NSS Journey," this would make the station name, ICly, the "NSS Journey."
To be extra clear, there's two more config flags

HOLIDAYSTATIONNAMES which allows for randomly generating a holiday station name (note that holiday station names will override set names, both with the STATIONNAME flag, and the next flag. Standard name generation will also not generate holiday names if this flag is disabled)
and
MAP_STATIONNAME which takes the station_name value in the map's .json file and sets it to the station name. If this is enabled, it will override the STATIONNAME config. If there is no station_name, name generation will fall back on STATIONNAME or random generation

Note that this does not hinder the captain's ability to rename the station, if they so desire.

Obviously, "Delta station" isn't a very good name for our setting, so, each map's name has been updated accordingly (subject to change, with suggestions)
(Note: NSS = NanoTrasen Science Station, NSB = NanoTrasen Science Base)
Delta - NSS Pollux - courtesy of @Snakebittenn 
Meta - NSS Castor - courtesy of @Snakebittenn
Icebox - NSB Boreas - courtesy of @thestubborn 
Tram - NSS Gallivant
Kilo - NSS Unity (get it... cause it's... small... and cramped... so everyone's forced into unity 🤣 )

I have also edited the station name of each file to read NTS Pollux - (Delta Station). This preserves the original map name, for voting and status display, while providing the new name, too add flavour. NSS Journey has also gotten this treatment, being called "NSS Journey - (Skyrat Box Station)" in map_name, and remaining NSS Journey in the station_name

## Why It's Good For The Game
RP, flavour that doesn't break immersion and is more than random words (but also doesn't restrict that ability, via configs)

## Changelog
:cl:
add: Each map now has a unique name!
add: Delta station is now the NSS Pollux
add: Meta station is now the NSS Castor
add: Icebox is now the NSB Boreas
add: Tram is now the NsS Gallivant
add: Kilo is now the NSS Unity
config: Added new config flag - HOLIDAYSTATIONNAMES - allows/disallows generating random holiday station names
config: Added new config flag - MAP_AS_STATIONNAME - takes the aforementioned map names and makes the station name it, if enabled
/:cl:

PR #5796, opened 2021-05-20
[NonModular]  Better Random Station Names

YakumoChen:
## About The Pull Request
alternative to #5712
This is an attempt to make roundstart station name generation better. Includes holiday name config introduced in #5712

Station names now generate as:
["NS" (Nanotrasen Space/Science)] + [A Ship Type] [A Name] [Number/Numeral]
An example would be "NS Lab Ingenuity 8"
Vetted the naming strings to produce better, more "High RP" names.

## Why It's Good For The Game
Randomly generated names have charm and fighting over permanent map names is kind of silly.
Removes all the "Low RP" name generation. Old ship types had their bad ones cut. New names are mainly based on existing, real space probe / ship names + a few solar bodies and other spaceshipy names and can probably be filled out with even more names.

## Changelog
:cl: FlamingLily, YakumoChen
config: Added new config flag - HOLIDAYSTATIONNAMES - allows/disallows generating random holiday station names
server: Ship name generation changed to be more consistent and better for RP.
/:cl:


Arctaisia:
Doesn't align with what the lore team has, at least in draft, for space station names.

YakumoChen:
can you clarify

YakumoChen:
@Arctaisia I took the "T" out of the prefix, based on what you wrote in the other PR, but otherwise it seems to hold up to the same standard you outlined before.

I could just add another prefix chain if you want different roles to make it come out like "Nanotrasen Science Lab Voyager III" or "Nanotrasen Mining Orbital Jupiter Alpha" 

This is pretty much what strings lists are for. They're easy to edit text files.

tf-4:
Good change. This can be adjusted again ~~if~~ _when_ lore comes out, but this is a good stopgap. Current name gen is awful.

FlamingLily:
Y'know

We could do both

It just hit me, I've made sure all my edits are bound to configs, so if we really wanted, we could merge both and just have the configs set to _not_ pre-set names while we wait for lore

YakumoChen:
I considered that, which is why I didn't put that it closes your PR (although I don't like the renamed map files, which is most of the reason for making this PR)

Gandalf2k15:
We'll be fine without it. I'd rather keep the current layout

PR #5837, opened 2021-05-21
[NONMODULAR] Xenobio tone-back part I, thumbsdown emoji.

RatFromTheJungle:
## About The Pull Request
Tones back some of the power of xenobio, starting with 'immediate concerns.'

## Why It's Good For The Game

Xenobio has an absurd amount of individual power in one job, 'rng' or not (it's very easy to have most of what you'd want by 40m), this is step one to toning it back. (also featuring a stable cerulean clarification for obvious reasons)

## Changelog
:cl:
balance: nerfed goldcore mob spawn rates.
balance: nerfed regenerative cores. (used the legion effect as a stand-in for a better one in episode II)
change: changed cerulean cores 'respawn' text to be more in line with (vaguely) policy (prevents it from circumventing the respawns > '; x killed me' clause.)
/:cl:


ghost:
There was someone asking if the cores could keep some of the blood replenishment, at least to keep it above x amount of blood? It'd keep them more beneficial than just plain damages, but still be a debuff

Tokkihopp:
The first and last issues sound really good to me as someone who admittedly plays a lot of xenobio and should for sure be implemented imo. For the second issue, I agree that an aheal is probably too much, especially for permanent traumas! However, dropping it down to healing 25 health, a stam damage heal, and negating slowdown is a huge drop. As a xenobiologist, I'm not sure I'd ever be interested in creating regen cores if that was their function, and would especially not be interested in sliding one to the CMO, something that always felt like nice inter-discipline teamwork that is probably lost with the proposed change, as they wouldn't be meaningful to a medical staff already equipped with basic healing resources at just 25 damage. I'd personally advocate for cleaving off just the really offending parts, like permenant traumas, and maybe adding a time to activate to it so you had to wait 10 seconds after using it for it to work, to prevent it helping rambo-ing?

Tristrian:
Since feedback is allowed :
Okay for cerulean and gold changes.
Is there already a message in case you die with a slime clone too?
Also how about chilling cerulean effect which is intended to be a temporary clone you control and die with?

Altough for regenerative cores, I want first to clarify to the previous comments it doesn't heal permanent traumas as far I know (deep rooted works though), nor restore missing limbs and top of all, it doesn't revive.  It doesn't remove the mindshield either.

Otherwise I understand the reasons of the nerf (having an instant heal of all damage and wounds makes it very powerful in combat), it sadly nerfes passive usage like in medical, as Tokkihopp mentionned. Those cores had some unique usage to restore blood or heal wounds for example, which made an interdepartemental cooperation (considering how often people say that Xenobio stay in their lab).
Further more, the way how the rework is made encourages the crossbreed to become a "combat instant stim" instead, which Xenobio already has some variants like burning purple.
If the nerf was intended to hammer down the combat usages, a channeled action to use the item sounds like a good idea, on the same vibe a medic would use a patch/syringe/gel to heal someome else. And for combat application, either take cover, retreat, or combo with other extracts to put yourself in safety, or use instant variants instead.

RatFromTheJungle:
> Otherwise I understand the reasons of the nerf (having an instant heal of all damage and wounds makes it very powerful in combat), it sadly nerfes passive usage like in medical, as Tokkihopp mentionned. Those cores had some unique usage to restore blood or heal wounds for example, which made an interdepartemental cooperation (considering how often people say that Xenobio stay in their lab).
> Further more, the way how the rework is made encourages the crossbreed to become a "combat instant stim" instead, which Xenobio already has some variants like burning purple.
> If the nerf was intended to hammer down the combat usages, a channeled action to use the item sounds like a good idea, on the same vibe a medic would use a patch/syringe/gel to heal someome else. And for combat application, either take cover, retreat, or combo with other extracts to put yourself in safety, or use instant variants instead.

I actually really like the idea of a timed use on second thought, and am mildly disappointed I didn't even think about that when doing it.
The only flaw being, I'm criminally stupid, and xenobio code isn't particularly 'helpful' (ancient), but this is what bootlegging was made for.

RatFromTheJungle:
regenerative will be in part II, because I'll have to refractor the whole deal for what I wanna do.

GoldenAlpharex:
> I actually really like the idea of a timed use on second thought, and am mildly disappointed I didn't even think about that when doing it.
> The only flaw being, I'm criminally stupid, and xenobio code isn't particularly 'helpful' (ancient), but this is what bootlegging was made for.

Let's talk more about this when I'm done with work today, I'm sure that we can come up with some more interesting mechanics for the regenerative cores like we talked about together yesterday, I've even got some more ideas. I'll hit you up on Discord about it or something.

Tokkihopp:
I'd be quite happy with either a channeled effect in exange for something similar to what we have now, or breaking all of the different things it heals into different colors for the crossbreeds, both sound like ok approaches that don't neuter its use in medical while toning down the mid combat full heal. 

As an aside, is it reasonable for a xenobio scientist feeling their stab cerulean activate and suddenly being in their locker to, knowing what that core does, assume they've somehow met their demise? "X killed me" is bad and should be clarified against, but it seems like "i have died" is a reasonable assessment for anyone with xenobio knowledge to make after that core activates and warps them to their backup. Not sure if that is what's intended by the current text or not.

Gandalf2k15:
Let's not start balance hell in these trying times.

PR #5857, opened 2021-05-22
[NON-MODULAR][TM-FIRST] Cryopods can now store items.

ZephyrTFA:
## About The Pull Request

Cryopods no longer drop all of your items on the ground when you go into cryosleep.

## Why It's Good For The Game

See above.

## Changelog
:cl:
add: Cryopods can store items; and their control computer can dispense them.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -24,6 +24,9 @@ GLOBAL_LIST_EMPTY(cryopod_computers)
 
 	// Used for logging people entering cryosleep and important items they are carrying.
 	var/list/frozen_crew = list()
+// Skyrat Edit Addition - Cryostorage stores items.
+	var/list/frozen_item = list()
+// Skyrat Edit End
 
 	var/storage_type = "crewmembers"
 	var/storage_name = "Cryogenic Oversight Control"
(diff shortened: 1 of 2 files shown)


capsaicinz:
i absolutely agree with this. i don't understand at all why they thought it'd be a good idea to merge that drop item thing.

GoldenAlpharex:
Just to be clear, can just about anyone pick up those items from the computer...?

ZephyrTFA:
> 
> 
> Just to be clear, can just about anyone pick up those items from the computer...?

Huh, good point.

GoldenAlpharex:
> > Just to be clear, can just about anyone pick up those items from the computer...?
> 
> Huh, good point.

I have no clue how it was originally implemented, but I feel like making it so it's only people of at least the same access level as the one that was on your card would have access to your stored items.

ZephyrTFA:
> 
> 
> Just to be clear, can just about anyone pick up those items from the computer...?

Problem solved.

YakumoChen:
> 
> 
> i absolutely agree with this. i don't understand at all why they thought it'd be a good idea to merge that drop item thing.

Because TG is more action oriented, with short rounds. Most of the people cyroing are likely to be assistants anyway, and it prevents any important items (like theft objectives) from being lost somehow.

It works better on TG because TG plays much different than Skyrat, and it's better for that style of gameplay.

NullDagaf:
> i absolutely agree with this. i don't understand at all why they thought it'd be a good idea to merge that drop item thing.

Round removal of objective items, hard-to-get gear (see - mining), etc.
Although we don't have objectives so it shouldn't be that big of an issue.

Gandalf2k15:
Seems good, tested in game. Well done lad.

PR #5889, opened 2021-05-24
!!TEMPORARY SOFT FEATURE FREEZE, PLEASE READ!!

Useroth:
## SOFT FEATURE FREEZE

For time being, we'd like to put new features on hold and focus on polishing what we've got for about a month. Please take this moment to polish what you have, design new features, and to think about the direction you want to take the server on. We will be holding discussions as well, so we can see what developers and others have in store for the future.

We will continue to merge PRs from our upstream, map PRs, fixes, improvements/refinements, donator PRs, and simple PRs involving cosmetic/non-mechanical changes, i.e. non-mechanical content or anything that doesn't introduce or have new mechanics like clothing, sounds, and so forth. We may also add PRs chosen by the host and head maintainers, especially if they are deemed essential or needed. Please ask the host if you fit under any of these requirements.

We may allow some leniency for any ongoing projects that were nearing release and are known to the maintainers, but this is something that needs to be discussed.

### DIFF
--- a/memes
+++ b/memes
@@ -1,0 +1,1 @@
+whatcha lookin' at?


memes
```diff
@@ -0,0 +1 @@
+whatcha lookin' at?
```
Gandalf2k15, line 1: this should be a define

Funce, line 1: Don't forget to tick any new files in your .dme!

unit0016, line 1: Atomize your PR

NullDagaf, line 1: Put this in the changelog. We don't want undocumented changes, now do we?

Funce:
Funny meme

PR #5905, opened 2021-05-24
[MODULAR] Fixes type path error in the Liquids Module

DuffCreeper:
## About The Pull Request

Apparently some typepaths were set to plasteel and not iron?
image

## Why It's Good For The Game

Fixity fix fix so the thing can compile? I dunno

## Changelog
:cl:
fix: Fixed up type paths in liquids
/:cl:


Useroth:
That's an improvement.

PR #5933, opened 2021-05-25
Fixes maps not compiling due to typepath error

KathrinBailey:


DuffCreeper:
That's weird as fuck. I opened that map and never received type path errors, only got them on Waterkilo

PR #5951, opened 2021-05-25
[MODULAR] Robotic Species rebalance II, 20 positives for 3 negatives is far to good a trade.

RatFromTheJungle:
## About The Pull Request

Returns synths damage de-buffs, due to the fatal flaw of them being drastically overpowered with an ever-growing list of benefits for 3 negatives.
## Why It's Good For The Game
Benefits of being a synth:
Most negative quirks do nothing (up to 30 free points)
Reviving you from the worst state imaginable takes like 15s of welding
Practically infinite access to medical supplies (wires and welders)
Immune to virus'
Immune to almost all the downsides of nanites, while getting all their positive,
Immune to Rads,
Immune to Toxin,
Immune to Oxy damage,
can self implant,
Organ rot immunity,
Immune to clone damage,
immune to genedamage, 
immune to like 90% of a changelings kit,
also immune to literally every negative chem,
normal surgeries still work on you, apparently,
Oh, and you never have to eat, ontop of all that.
Immune to wounds,
Don't bleed, 
limbs can be slapped back on

Downsides of being a synth:
EMPs are a hard counter(practically unused)
Having a high bodytemp does more damage, ( because of their regulating bodytemp from being a child of humans, this also is practically unused)
shocks hurt more

## Changelog
:cl:
balance: returns the incoming damage modifiers to synths(1.6 brute, or 60% more 1.5 burn, or 50% more), and nerfs the heat damage, (2x, down from 3x.), Raises the cold damage to be in-line (1.2 or 20% above baseline).
/:cl:


RatFromTheJungle:
I personally feel 1.2 isn't enough, but I'll start with the lowest to see what people think.

Guidesu:
> I personally feel 1.2 isn't enough, but I'll start with the lowest to see what people think.

even if i wanted synth to be nerfed.
1.2 its KINDA big

Vizoplez:
PR doesnt even list how you nerfed it.

RatFromTheJungle:
> 
> 
> > I personally feel 1.2 isn't enough, but I'll start with the lowest to see what people think.
> 
> even if i wanted synth to be nerfed.
> 1.2 its KINDA big

its 20%, that is **very** low, if an item does 30 damage, that only adds 6

Vizoplez:
> I think that the damage mod should be raised.
> I would maybe try 1.6 and 1.5 to start.
> This is really just a just a bandaid patch as well.
> It doesn't actually help fix the many listed ways.
> That synths just don't need to interact with the game at all.

thats almost double damage from all incoming sources. I'm unsure if you're trolling.

SpaceVampire:
> thats almost double damage from all incoming sources. I'm unsure if you're trolling.

I'm unsure how 50-60% more damage is double damage.
I mean, I know I had to repeat my Calc 2 class, but I didn't think I was that awful.
Its also really not "all incoming sources" when a lot of stuff just can't damage them at all.

EDIT: When I played an IPC back on OrcacleStation, when I was new to SS13 in general.
They had this for their stats.
heatmod = 3 // Went cheap with Aircooling
coldmod = 1.5 // Don't put your computer in the freezer.
burnmod = 1.8 // Wiring doesn't hold up to fire well.
brutemod = 1.8 // Thin metal, cheap materials.
siemens_coeff = 2 // Overload!

To make up for all the other game ignoring upsides.
I get that balance has changed in some ways since then, but still.

Nerev4r:
Personally, I think 1.2 or maybe 1.4 was fine.

RatFromTheJungle:
> 
> 
> Personally, I think 1.2 or maybe 1.4 was fine.

on paper, but in reality, it had next to nil effect in game, it didn't even drop the hits to down with an esword by one, practically was a pseudo change.

FlamingLily:
Don't get me wrong, I wholeheartedly agree with this change, but "EMP" and "NEVER USED" are not really related terms, I see people using EMPs a lot

YakumoChen:
Only time I've ever gotten to use EMPs against a synth was as a grief ghost. Most antags only get limited access to EMP aside from ling (being the only thing lings can actually use against them) and it's not really an ideal solution, especially since if you use ANY augments you also get fucked up.

In a fight it's just not useful anyway unless it's a 1v1

Shocks arent even used and insuls give you complete immunity to them

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/robotic.dm
```diff
@@ -19,9 +19,9 @@
 	mutant_bodyparts = list()
 	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_MAGIC | MIRROR_PRIDE | ERT_SPAWN | RACE_SWAP | SLIME_EXTRACT
 	reagent_flags = PROCESS_SYNTHETIC
-	burnmod = 1.3
-	brutemod = 1.2
-	coldmod = 0.5
+	burnmod = 1.5 // Every 0.1% is 10% above the base.
+	brutemod = 1.6 
+	coldmod = 1.2
```
KasparoVy, line 24: 70% increase, ow

RatFromTheJungle: *60%, and its fair exchange for them ignoring wounds (40% of melee combat), and three of the main damage types (blood, oxy, toxin)

Gandalf2k15:
aaaaaaaa please can we not do it like this, please add more features to being a synth instead of number changes. Like, losing hydraulic fluid(blood) causes you to slow down and have arms flail. Having severe impacts to the head causes your optical sensors to reboot(static screen). I'd really like more things like this than a few numbers being modified.

RatFromTheJungle:
> 
> 
> aaaaaaaa please can we not do it like this, please add more features to being a synth instead of number changes. Like, losing hydraulic fluid(blood) causes you to slow down and have arms flail. Having severe impacts to the head causes your optical sensors to reboot(static screen). I'd really like more things like this than a few numbers being modified.

outside of my skill limit, I'm just doing this in place of someone else sweeping in to add better features, because at the moment, they're leagues above everyone else, in practically every functional situation besides getting EMP'd.
(these numbers are also what practically every other server uses for them, ((even oldbase, I'm psure)) that we removed, causing said unbalance)

If someone wants to sweep in and add stuff like that, and tone down the incoming damage that'd be great, but my 🐒 brain isn't made for it, and synths are currently way-way-way to good compared to everyone else.

NullDagaf:
> aaaaaaaa please can we not do it like this, please add more features to being a synth instead of number changes. Like, losing hydraulic fluid(blood) causes you to slow down and have arms flail. Having severe impacts to the head causes your optical sensors to reboot(static screen). I'd really like more things like this than a few numbers being modified.

­>losing hydraulic fluid(blood) causes you to slow down and have arms flail
"Blood amount is tied to speed." Not sure what you meant by "arms flail".

­>Having severe impacts to the head causes your optical sensors to reboot(static screen)
"Hitting a bodypart damages the organs inside"

These mechanics sound better, if they were tied to base humans instead of robots specifically.

...but yeah, mechanics > number changes.

YakumoChen:
>>Having severe impacts to the head causes your optical sensors to reboot(static screen)
>"Hitting a bodypart damages the organs inside"

>These mechanics sound better, if they were tied to base humans instead of robots specifically.

It's almost like we do have mechanics, specifically designed to to do that, for base humans

Mechanics

which

Synths have immunity to

MLGTASTICa:
Their old values were 10% more brute and 10% more burn , not 50% more burn and 60% more brute. This turns Synths/Robots into CBT to play. You should consider coding them actual features that give them weaknesses and not tweaking numbers to kill off anyone who plays this species.
The quirk issue can be fixed by just giving them a quirk blacklist , which is fairly easy to implement. They're not immune to a virus if it has a specific mutation.
The advantage you mention are mostly niches, Genedamage and clone damage are both extremly rare and never used , Radiation is only gotten by being in engineering and you can always drink vodka to get rid of it.
They can't process chemicals , so they can't powergame to heal 11 brute per tick. Not being able to process is also a downside 
Normal surgeries is a bug and that should be fixed. Not used as a motive to nerf a species.

RatFromTheJungle:
> 
> 
> Their old values were 10% more brute and 10% more burn , not 50% more burn and 60% more brute. This turns Synths/Robots into CBT to play. 

they can be revived by anyone, with just a bit of welding from 95% of things that kill them, and the numbers were taken from the only other 'balanced' tg server that runs them, they weren't even balanced before the 'rebalance' PR, much-much-less so currently

>>You should consider coding them actual features that give them weaknesses and not tweaking numbers to kill off anyone who plays this species. 

nothing was stopping them from doing this, but they've had no encouragement
> The quirk issue can be fixed by just giving them a quirk blacklist , which is fairly easy to implement. They're not immune to a virus if it has a specific mutation.
> The advantage you mention are mostly niches, 

how many niches until they're collectively an absurd advantage? 
>Genedamage and clone damage are both extremly rare and never used , Radiation is only gotten by being in engineering and you can always drink vodka to get rid of it.
> They can't process chemicals , so they can't powergame to heal 11 brute per tick. Not being able to process is also a downside

Nanites are more then enough with them not getting wounds, and they can heal upwards of 50-60 a second with someone else welding them

YakumoChen:
If they're going to be flat immune to almost every unconventional way to damage a player, having the most conventional ways be more effective seems like a good tradeoff.

Wound, toxin, and bleed, and oxyloss immunity alone are extremely good advantages that could easily cripple or kill other players even if the source isn't outright lethal by itself. 
Their combination of advantages also makes them more or less immune to brain damage and traumas as well.

They need more harshly severe drawbacks like what this PR is offering. Merge this PR and the drawbacks here can be rolled back when more deliberate disadvantages are made. Make them more like how they are on Citadel for example, where limbs have damage thresholds that require surgery (and not self surgery)

Gandalf2k15:
Alright I'm not 100% happy with the laz approach, but I'm happy to have it as a placeholder, though we might need to rubber band those amounts. Merging for now 👍

PR #5955, opened 2021-05-25
PC mob spawn text

Gandalf2k15:
## About The Pull Request

Title.

## Changelog
:cl:
add: PC MOB POLICY UPDATE: As a player controlled mob you are expected to play the role to the best of your ability. This means if you're an animal, act like one. You shouldn't display much intelligence if any. This also means if you're engaging in combat you should refrain from mercing people fully. Play not to win but to create a challenge. You're there to replace AI, make others enjoy the situation as well. If your simple mob is not above simple or mute intelligence, using structures such as welding tanks/canisters/boxes to hinder your opponent is entirely forbidden. Do not do this.
/:cl:

### DIFF
--- a/config/skyrat/skyrat_config.txt
+++ b/config/skyrat/skyrat_config.txt
@@ -59,3 +59,6 @@ ROLE_ASSIGN_CHANNEL_ID
 
 ##Asteroid Ruins Budget
 ASTEROID_BUDGET 60
+
+##Player controlled mob spawn text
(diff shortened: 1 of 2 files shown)


Elli-Skala:
will theses rules on using intelligence based on the mob you are, also apply to mobs made via intelligence potions

Gandalf2k15:
> will theses rules on using intelligence based on the mob you are, also apply to mobs made via intelligence potions

Only player controlled.

PR #5975, opened 2021-05-26
[NON-MODULAR] Doing Surgery in the correct area now gives you a boost to its speed.

ZephyrTFA:
## About The Pull Request

This makes it so that if doctors perform surgery inside of surgery and with the correct gear, nitrile or latex gloves, you get a bonus to your speed for the operation, which is currently twice as fast.

## Why It's Good For The Game

We punish people for not doing surgery in surgery by rewarding the people who do surgery in surgery.
I wanted to nerf not doing surgery in surgery with a speed penalty but I got told no.

## Changelog
:cl:
balance: Doing surgery in surgery or robotics with the correct equipment now gives you a multiplier to your operation speed.
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -57,6 +57,7 @@
 	return FALSE
 
 #define SURGERY_SLOWDOWN_CAP_MULTIPLIER 2 //increase to make surgery slower but fail less, and decrease to make surgery faster but fail more
+#define SURGERY_SPEEDUP_AREA 2 // Skyrat Edit Addition - reward for doing surgery in surgery
 
 /datum/surgery_step/proc/initiate(mob/living/user, mob/living/target, target_zone, obj/item/tool, datum/surgery/surgery, try_to_fail = FALSE)
 	// Only followers of Asclepius have the ability to use Healing Touch and perform miracle feats of surgery.
@@ -84,6 +85,12 @@
 
 	fail_prob = min(max(0, modded_time - (time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)),99)//if modded_time > time * modifier, then fail_prob = modded_time - time*modifier. starts at 0, caps at 99
 	modded_time = min(modded_time, time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)//also if that, then cap modded_time at time*modifier
+
+	// Skyrat Edit Addition - reward for doing surgery in surgery
+	if(istype(get_area(target), /area/medical/surgery) && (TRAIT_FASTMED in user.status_traits))
+		modded_time /= SURGERY_SPEEDUP_AREA
+		to_chat(user, "<span class='notice'>You breathe in relief as all the tools and equipment you need are in easy reach!</span>")
+	// Skyrat Edit End
 
 	if(iscyborg(user))//any immunities to surgery slowdown should go in this check.
 		modded_time = time
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -87,7 +87,7 @@
 	modded_time = min(modded_time, time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)//also if that, then cap modded_time at time*modifier
 
 	// Skyrat Edit Addition - reward for doing surgery in surgery
-	if(istype(get_area(target), /area/medical/surgery) && (TRAIT_FASTMED in user.status_traits))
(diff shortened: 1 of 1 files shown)


NullDagaf:
I'm not sure if this is the right way of doing this.
WIth that in mind, it's a solution that works and I can't think of a better one.

code/modules/surgery/surgery_step.dm
```diff
@@ -57,6 +57,7 @@
 	return FALSE
 
 #define SURGERY_SLOWDOWN_CAP_MULTIPLIER 2 //increase to make surgery slower but fail less, and decrease to make surgery faster but fail more
+#define SURGERY_SPEEDUP_AREA 2 // Skyrat Edit Addition - reward for doing surgery in surgery
```
NullDagaf: Make it 0.5...

code/modules/surgery/surgery_step.dm
```diff
@@ -85,6 +86,12 @@
 	fail_prob = min(max(0, modded_time - (time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)),99)//if modded_time > time * modifier, then fail_prob = modded_time - time*modifier. starts at 0, caps at 99
 	modded_time = min(modded_time, time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)//also if that, then cap modded_time at time*modifier
 
+	// Skyrat Edit Addition - reward for doing surgery in surgery
+	if(istype(get_area(target), /area/medical/surgery) && (TRAIT_FASTMED in user.status_traits))
+		modded_time /= SURGERY_SPEEDUP_AREA
```
NullDagaf: ...and use *= instead.

ZephyrTFA: How dare you make me following standard coding practices

code/modules/surgery/surgery_step.dm
```diff
@@ -85,6 +86,12 @@
 	fail_prob = min(max(0, modded_time - (time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)),99)//if modded_time > time * modifier, then fail_prob = modded_time - time*modifier. starts at 0, caps at 99
 	modded_time = min(modded_time, time * SURGERY_SLOWDOWN_CAP_MULTIPLIER)//also if that, then cap modded_time at time*modifier
 
+	// Skyrat Edit Addition - reward for doing surgery in surgery
+	if(istype(get_area(target), /area/medical/surgery) && (TRAIT_FASTMED in user.status_traits))
+		modded_time *= SURGERY_SPEEDUP_AREA
+		to_chat(user, "<span class='notice'>You breathe in relief as all the tools and equipment you need are in easy reach!</span>")
+	// Skyrat Edit End
+
```
NullDagaf: 👀 
""oo empty line change outside skyrat comments plox fix""

NullDagaf: Joking aside, pretty sure no one could care less.

LFSpirits:
I get the motivation behind this. But I don't think this'll change too much in the grand scheme of things for those that tend to forego the surgery rooms. 

Surgery with proper setup (Alien, or even experimental tools, surgery table, and an application of something like sterilizer, alcohol, or even honey) is already pretty fast enough for any other speed bonus to feel pretty negligible to most others. Sure, this would add onto that... But the difference is already pretty great enough as it is. It'd just feel like overkill to straight up DOUBLE the speed on top of all of that.

I'd say the extra speed is good in emergencies and all, but in my experiences, medical would tend to just focus on keeping everything in one place during actual chaos so they can switch from patient to patient and work together without having to move around much and keep an eye on everything.

Most I can see this really benefitting is someone trying to ABSOLUTELY rush something like a brainwash or something else vile as an antag before getting caught, or to quickly get something done that isn't too dire, like implants or something.

ADDITIONALLY. Someone brought up the possibilities of just...getting the blueprints to expand what's "Surgery" in medical. What's stopping that?

ZephyrTFA:
> ADDITIONALLY. Someone brought up the possibilities of just...getting the blueprints to expand what's "Surgery" in medical. What's stopping that?

This isn't how the blueprints function

LFSpirits:
> > ADDITIONALLY. Someone brought up the possibilities of just...getting the blueprints to expand what's "Surgery" in medical. What's stopping that?
> 
> This isn't how the blueprints function

Then carry on with that one. I'm not fully aware of how it works.

> > Most I can see this really benefitting is someone trying to ABSOLUTELY rush something like a brainwash or something else vile as an antag before getting caught, or to quickly get something done that isn't too dire, like implants or something.
> 
> Due to the already existing special description and naming system, brainwashing is already risk free

Checking the surgery computer itself can still show brainwashing present as it's being done if it's the active surgery.

Funce (approved):

PR #6046, opened 2021-05-31
Privateers 2.0 Prototype

KathrinBailey:
## About The Pull Request
• Adds in a new shuttle that can be spawned by admins, designed for adminbussing.
• Adds in the 'Guild' radio headset and some ghost role spawners. This is for prototyping the Privateers without needing to outfit them.
• Introduces Privateers as, at the moment, slavers. Their goal is to capture crew and take them as slaves. As this is a prototype, there's nowhere to take the slaves as of yet.

## Why It's Good For The Game
This will allow for the following opportunities:
1. Adminbus events that can use the shuttle. It's a nice shuttle. It can easily be modified to suit other purposes, like a trader bus, if need be. All that happens is the cells need emptying and loot can be put in the cells, which would turn the cells into display cases.
2.  Allows for the concept of privateers to be tested. Whilst they currently don't have a special place to dock to and a special prison to offload their prisoners, this can easily be made separately and spawned into the world via uploading and placing map templates. I can try and find out if this stuff if fun for players before embarking out do the full thing.
3. Allows for balance of privateers to be tested. Is their loadout too strong? Does their ship need defensive turrets? Is their ability to heal too strong? Do the ship engines cooldown too fast or slow? All questions that can be answered through **testing!**

In some ways, this is a revival of assault operatives, except with my vision for it. You're now focused solely on **capture**. Killing doesn't get you anywhere at all. Your arsenal focuses more on stunning than killing. You have the ability to deal stamina damage through three different types of armor: `energy`, `melee` and `bullet`. 

## Changelog
:cl:
add: Prototype version of Privateers, including a new shuttle for their exclusive use!
/:cl:
image


Sigmar-alkahest:
this seems like an interesting idea, the quantity of thrusters on that ship does seem a bit excessive though...

Gandalf2k15:
> this seems like an interesting idea, the quantity of thrusters on that ship does seem a bit excessive though...

See Space Engineers for exessive thrusters

Funce (approved):
Solid!

Arctaisia:
> this seems like an interesting idea, the quantity of thrusters on that ship does seem a bit excessive though...

Makes it incredibly maneuverable, and the fact that it has front thrusters makes me happy, because otherwise they'd need to do a 180 to decelerate, which isn't always the most practical. Dunno why more don't have forward thrusters, and yeah, as Gandalf said, SE is a good example of why

PR #6100, opened 2021-06-03
[ready] Fixes robots screaming in pain when they don't feel it

FlamingLily:
## About The Pull Request
Surgery pain now only forces a scream if it's an organic

## Why It's Good For The Game
fix oversight

## Changelog
:cl:
fix: Synths screaming in non-existent pain when being operated on
/:cl:


Funce (dismissed):

Funce:
I'd recommend porting this upstream actually. Is that something you can do?

They'll probably be able to coach it into a better method too.

FlamingLily:
> 
> 
> I'd recommend porting this upstream actually. Is that something you can do?
> 
> They'll probably be able to coach it into a better method too.

surgery pain is something I set up, along with the screaming, i did ask upstream if they'd want it but didn't get a response

PR #6112, opened 2021-06-03
Merge the Ruins module into the Mapping module

ghost:
## About The Pull Request
These two modules were practical duplicates, with the Ruins module having a similar file structure and contents, and the Mapping module having a specific place for Ruins already. So, uh, yeah. This combines them.

## Why It's Good For The Game
Modules are only effective if they actually have their contents - sometimes ruins would be put into Mapping, sometimes into Ruins. Why not just.. put ALL ruins (+whiteships) into Mapping, because it already had them prepared?

## Changelog
Dont need the CL for this change, I think?


ghost:
Yes, I've tested it. The moved/renamed stuff still all spawns fine - though I think there'll need to be some serverside config done too so the ruins spawn correctly.

PR #6113, opened 2021-06-03
[NON-MODULAR] Modularizes the Material Glass / Windows Don't Create Instantly

ZephyrTFA:
Modularizes glass.dm entirely

## About The Pull Request

Making windows is no longer instant.
They take four seconds to build now and are in line with making walls.
Directional windows take half as long.
All values are set via defines for ease of balance.

## Why It's Good For The Game

Admins said people were abusing windows being made instantly; this is no longer possible.

## Changelog
:cl:
fix: Making windows is now inline with making walls in that its not magically instant
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/glass.dm
+++ b/code/game/objects/items/stacks/sheets/glass.dm
@@ -9,8 +9,6 @@
 #define BUILD_WINDOW 4 SECONDS // This is the exact same as building a girder
 #define BUILD_WINDOW_DIRECTIONAL_MOD 0.5
 
-
-
 GLOBAL_LIST_INIT(plastitaniumglass_recipes, list(
 	new/datum/stack_recipe("plastitanium window", /obj/structure/window/plasma/reinforced/plastitanium/unanchored, 2, time = BUILD_WINDOW, on_floor = TRUE, window_checks = TRUE), \
 	new/datum/stack_recipe("plasma glass shard", /obj/item/shard/plasma, time = 0, on_floor = TRUE) \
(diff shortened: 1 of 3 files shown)


RatFromTheJungle:
directional windows taking two seconds, means that assembly, with the three more seconds to screw them down, and the 7(?) to crowbar them in(reinforced), ends out at twelve seconds per directional window, with a tidi of 20-30% that goes up 16-17 seconds, making it legitimately equal to building an entire airlock, A single 1/4 directional window. For something that wasn't even particularly abuseable to begin with. (they could be altclicked out of the way)

YakumoChen:
Make regular glass windows take half as long

>with a tidi of 20-30% that goes up 16-17 seconds, making it legitimately equal to building an entire airlock

Windows had three steps to construct before so the time isnt much lower. Walls take about as long to build as well.

There is also precedent for this as putting glass on a grille had a timer as well

Funce (changes requested):
Can this be modularized please? Move the new defines, undefines and list inits under `modular_skyrat`

ZephyrTFA:
> 
> 
> directional windows taking two seconds, means that assembly, with the three more seconds to screw them down, and the 7(?) to crowbar them in(reinforced), ends out at twelve seconds per directional window, with a tidi of 20-30% that goes up 16-17 seconds, making it legitimately equal to building an entire airlock, A single 1/4 directional window. For something that wasn't even particularly abuseable to begin with. (they could be altclicked out of the way)

It's why they are defines that can be easily changed as needed; lets see how it works in practice before adjusting it

Funce (approved):
Bit more than I was thinking, but its probably the more ideal system. Probably. I'll eat it though.

FlamingLily:
Remember that we've got 
https://github.com/Skyrat-SS13/Skyrat-tg/tree/master/code/__DEFINES/%7Eskyrat_defines
https://github.com/Skyrat-SS13/Skyrat-tg/tree/master/code/__HELPERS/%7Eskyrat_helpers
and could probably do with a ~skyrat_vars or ~skyrat_llists in https://github.com/Skyrat-SS13/Skyrat-tg/tree/master/code/_globalvars

PR #6116, opened 2021-06-04
[DNM] [TM-ONLY] FLAVOUR TEXT TEMPORARY REVERT

Funce:
Pulled the trigger a bit fast on these two, gotta get the policy in place first.


NullDagaf:
image

PR #6131, opened 2021-06-04
[NONMODULAR] Rebalances cost of holopara to match holocarp.

Vizoplez:
## About The Pull Request
Holoparasite has arguably been overpriced for yearsa; 18 TC for something that makes you rely on another players skill, attention span - and ability to comprehend orders.

Atop that; having a holoparasite gives security a way to deal 100% damage to you regardless of armor, as holoparasite damage reflect transmits all incoming damage without regards to armor - making things like the elite syndicate hardsuit pointless in practice if you buy one.

Holoparasites themselves are extremely loud, easy to notice and otherwise not worth their cost as they can dust the user in question; which sec will almost always do because 'le uncontainable'.

## Why It's Good For The Game

Holoparasite has been very overpriced for a very long time. This brings it in line with the holocarp fishsticks (objectively, and aesthetically more pleasing to use)

## Changelog
:cl:
tweak: Holopara price reduced.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -449,7 +449,7 @@ GLOBAL_LIST_INIT(uplink_items, subtypesof(/datum/uplink_item))
 	desc = "Though capable of near sorcerous feats via use of hardlight holograms and nanomachines, they require an \
 			organic host as a home base and source of fuel. Holoparasites come in various types and share damage with their host."
 	item = /obj/item/storage/box/syndie_kit/guardian
-	cost = 18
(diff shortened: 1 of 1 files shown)


NullDagaf:
> 18 TC for something that makes you rely on another players skill, attention span - and ability to comprehend orders.

"It's not the item that needs to be buffed, it's the player that needs to be better."

tf-4:
Like I said on discord, the holocarp is not a reskinned holopara and should not be priced the same

Funce:
@hugmeorelse Balance change for you. Code is fine, merge or close at your leisure.

Gandalf2k15:
I'd rather not start a new balance war in these trying times

PR #6173, opened 2021-06-07
[Modular] Partially Re-Souls Floor Tiles

unit0016:
## About The Pull Request
Partially Resouls floor tiles. Most are stock /tg/ again - minus the custom floors and reinforced floors.
There is a difference, though - normal iron floor tiles are slightly darker than /tg/ default, including all hybrids. Weh.
image

## Why It's Good For The Game
Having two variants of textured tiles and half a set of untextured tiles is weird, yo.

## Changelog
:cl:
imageadd: Smooth tiles are now the default.
/:cl:


GoldenAlpharex:
Nooooo, I really, really prefer the slightly textured tiles compared to the smooth one /tg/ has. These ones are just too bland, they look like they're made out of plastic.

Arctaisia:
I do personally prefer our slight texturevones, especially since we don't have the decals that virgo has, so they won't nearly be as detailed as I'd hope they'd be, though they still remind me of virgo. Perhaps we could have these, if we fixed up their colours to align with our ones now, we could have them as a proper smooth tile variant, so both can be used in mapmaking?

I did also work on some sprites to differentiate textured tiles from normal ones too, with colour correction, though I'm too lazy and never did anything with them, would be happy to give you them if needed. Makes them look textured rather than just dimpled.

edit: fat fingers phone typing hngg, spelling fixed...

KathrinBailey:
@Gandalf2k15 This is affecting your port.
Personally, I do find the TG adaptation bland as well and prefer the ported Tau Ceti stuff, but hey ho.

unit0016:
ehh. I'd prefer we commissioned new sprites because upstream wackness but w/e

PR #6203, opened 2021-06-09
Revert "Right-clicking on reagent containers picks previous reagent t…

Funce:
Another one that makes 514 cough. Will revert the reversion once we find out what the heck.


Gandalf2k15:
Ditto, check discord

PR #6265, opened 2021-06-11
Moves multi-cell charger empty to right click

tf-4:
## About The Pull Request

Title. Might move cell removal from the component to right click also but that's outside the scope of this PR. Adds a warning that the action has been rebound to right click when the old binding is used (can remove this in a month or so)

Also small code cleanup with variable names.

## Why It's Good For The Game

Gets rid of an unnecessarily complex keybind.

## Changelog
:cl:
qol: Multi-cell chargers now dump all their cells with right-click instead of Ctrl+Shift+Click
/:cl:


tf-4:
That runtime is not my fault

tf-4:
That one isn't my fault either

PR #6272, opened 2021-06-12
Disables savannah ivanov until further notice.

Gandalf2k15:
Title. Host suggest.

Too OP and also has a bug where MMIs go to Brazil if they die in mech.

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -1523,6 +1523,7 @@
 	)
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 2500)
 
+/* SKYRAT EDIT REMOVAL
 /datum/techweb_node/savannah_ivanov
 	id = "mecha_savannah_ivanov"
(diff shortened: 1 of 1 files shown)


Vizoplez:
I got ahold of the savannah ivanov mech that killed 90% of the nukies. Player skill made it far too good.

NetraKyram:
I'll be real with you, it requires coordination and skill, that shouldn't be disabled, or removed. That's just my two cents on it, don't disable one of few _really good things_ that relies on **coordination and teamwork** to be so good. Fixing the bugs? Certainly, disabling it completely? I don't think so.

hugmeorelse:
> I'll be real with you, it requires coordination and skill, that shouldn't be disabled, or removed. That's just my two cents on it, don't disable one of few _really good things_ that relies on **coordination and teamwork** to be so good. Fixing the bugs? Certainly, disabling it completely? I don't think so.

It single-handedly destroyed a 10-men nuclear operatives team.

Iamgoofball:
Skill issue.

PR #6280, opened 2021-06-12
Amber alert energy weapons, and the return of the energy gun.

Tupinambis:
## About The Pull Request

Energy weapons are now authorized at security level amber, and energy guns are now available at vendors with the use of an energy weapon token.

## Why It's Good For The Game

We have three security tokens for energy weapons... of which the only option is the laser gun. This adds the option for the use of the energy gun which has a disabler setting at the cost of overall efficiency.

Additionally, energy weapons fall behind the lethal ballistics in terms of usefulness and are almost always overlooked in red alert situations. Amber seems to be a good niche for them to fill relative to their power level.

## Changelog
:cl: Tupinambis
balance: Energy weapon tokens are now locked to Amber instead of Red security. 
balance: Energy guns may now be acquired using energy weapon tokens.
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/token_system/gunsets.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/token_system/gunsets.dm
@@ -357,6 +357,12 @@
 	. = ..()
 	new /obj/item/gun/energy/laser(src)
 
+/obj/item/storage/box/gunset/e_gun
+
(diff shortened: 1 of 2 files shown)


dominuslucian:
based

KathrinBailey (approved):
Eh, sure.
You can get superior lethals on green anyway.

PR #6285, opened 2021-06-13
New "pirate" type - Imperial Enclave Audit

Gandalf2k15:
## About The Pull Request

Adds a new "pirate" encounter type - The imperial enclave, a new lore faction(TBA) that is very tyrannical and overbearing. Their command structure follows your average military(see Star Wars for a more accurate depiction - NO, these are not the empire. Just a vague reference.). They can commonly be found in the slave trade and have a very large fleet. 

The Imperial Enclave that you will encounter is simply auditors/debt collectors/enforcers. They are not the best example of the Imperial Enclave's military power, they are very weak in comparison. But still, they pack a punch.

This event follows your standard pirate flow, however, they are more open to "negotiations". 

I've also disabled the round removing from the pirate cargo pad, so you can't sell mobs. Those need to be ransomed manually.

The shuttle that you will encounter is the Imperial Enclave Enforcer-Class Starship, a lightweight heavily armored, heavily armed, shuttle. Thanks to Floof for the map, I changed it up a bit to be more dangerous and be compatible with pirates.

Also added an Automag, because why not.

I've given all pirates a backpack and an m1911, since it was requested by the host.

Silverscales can speak common.

## Changelog
:cl:
add: A new type of "pirate" event has been added! The Imperial Enclave Audit, watch out for those imperial scumbags!
/:cl:


Gandalf2k15:
image

PR #6313, opened 2021-06-14
[MIRROR] Juke Build

Funce:
Original PR: https://github.com/tgstation/tgstation/pull/59390
--------------------
## About The Pull Request
image

Integrates [Juke Build](https://github.com/stylemistake/juke-build) into CBT.

This was an effort to extract CBT functionality into an external repository/package with a refined API and extra features.

## Important
**tgui** and **tgfont** assets no longer block the compilation of DME. They are loaded dynamically on server startup. The side effect of this, is that tgui bundle errors will manifest in Dream Daemon instead of the compiler.

Compiler still shows a warning if you compile DME directly, so hopefully there will be no surprise.

## Features of Juke Build
* Simple and strongly typed API
* Robust dependency model with concurrent execution of targets
* CLI argument (and environment) parsing via Parameters API
* Nice colorized output (at least on Linux)

## Features of CBT (new)
* You can specify which specific target to build via command line arguments.
* You can specify defines by stacking `-Ddefinename` flags, or by using `DEFINE=A,B,C` env var with comma separated values.
  
  * When using flags, order of flags and task names matters (more in [Juke Build docs](https://github.com/stylemistake/juke-build)). Tell me if this is something you all are ok with.
* Parallel execution drastically cuts down build time. It's pretty much the same as if you compiled just dm and nothing else.

## Removals
Removed CBT_BUILD_MODE env var, in favor of an optional `tgs` target, which you can launch via `tools/build/build tgs`, and in all other cases by using the `-D` flags to add defines.

## Did I test it?
Juke Build itself has a tiny unit test, and each piece of the CI pipeline was tested in isolation on my machine.

Not sure what bugs will pop up in practice.

CI integration tests might not need to build TGUI now, but I left that job as is.

## Screenshots
image


Funce:
I'll add the two hotfixes for this too

PR #6315, opened 2021-06-14
[MIRROR] Refactors most spans into span procs

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/59645
--------------------

## About The Pull Request

Converts most spans into span procs. Mostly used regex for this and sorted out any compile time errors afterwards so there could be some bugs.
Was initially going to do defines, but ninja said to make it into a proc, and if there's any overhead, they can easily be changed to defines.

This definitely will need a testmerge once I'm done with local testing.

## Why It's Good For The Game

Makes it easier to control the formatting and prevents typos when creating spans as it'll runtime if you misspell instead of silently failing.
Reduces the code you need to write when writing spans, as you don't need to close the span as that's automatically handled by the proc.

## Changelog
:cl:
refactor: Refactors how spans are done in code.
/:cl:


FlamingLily:
agony

PR #6334, opened 2021-06-15
Security/bullet rebalance

Gandalf2k15:
## About The Pull Request

This does a few things - 

The ammunition vending machines have been replaced with ammunition workbenches.

Sidearms now start with NO bullets and are blue alert locked.

Security now starts with a low-range low-capacity taser as their sidearm.

All lethal bullets have been buffed regarding wounds(by 20%)

## Changelog
:cl:
balance: Armadyne has invested in stocking your stations with ammunition workbenches.
/:cl:


FlamingLily:
I'm not SURE about the blue alert. While I agree in theory, It's almost 100% going to turn into *round starts, blue alert*

SpaceVampire:
Why not make it amber-alert instead then?
That might be more reasonable, as if it is only blue-alert.
Then you can just get by with all of non-lethal take-down tools.

FlamingLily:
> 
> 
> Why not make it amber-alert instead then?
> That might be more reasonable, as if it is only blue-alert.
> Then you can just get by with all of non-lethal take-down tools.

Eh, then I feel like (the super rare chance that) when you're doing a manhunt with someone you don't *know* is lethally armed (which imho is only worth a blue) you should have a side-arm just in case.

Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/6318 - Requires this

_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
```diff
@@ -109448,7 +109448,7 @@
 /obj/effect/turf_decal/tile/red{
 	dir = 8
 	},
-/obj/machinery/vending/security_ammo,
+/obj/machinery/ammo_workbench,
```
Gandalf2k15:
```suggestion
/obj/machinery/vending/security_ammo,
```

PR #6356, opened 2021-06-17
Flipped Tables and Turnstiles no longer always runtime

ZephyrTFA:
## About The Pull Request

Flipped Tables and Turnstiles have been the bane of many coders and maintainers for the longest time because they continually runtimed for little reason. I looked into it and found the reason.

Turns out we weren't adding the element correctly!

## Changelog
:cl:
fix: Flipped Tables are no longer coder bait
fix: Turnstiles are no longer coder bait
/:cl:


Funce:
Oops, my bad on this one!

Funce (approved):

PR #6376, opened 2021-06-18
Adds various Departmental warning tapes

ghost:
## About The Pull Request
image
(Image outdated, but only in textures - now they look nicer and more consistent)

Originally just a port from TGMC (police/engi tape only, with painful textures), now this adds Police, Engineering, Atmospherics, Medical, and Biohazard tape for properly partitioning off hazardous areas.
These block passage for those without access - but, those WITH access can 'lift' the tape with an empty hand, letting anyone pass; and, they're easy to break through, with anyone on combat mode/walking will damage it to the point of no longer restricting access. (This is reversible by using the tape on the damaged portions)

## Why It's Good For The Game
My crime scene roleplay hnnngh

Balance-wise, these are the same as if not worse than holo-barriers; yes you can place them faster, but they can be removed just as fast and damaged easily too.

Also just look at these cool things I did! Rightclick will un-roll the tape, deleting a full line! I made RIGHT CLICK WORK! Im learning CODE!
https://gyazo.com/3849ab771bcbc6ffaa2434bbd0cdb29a

## Changelog
:cl:
add: Nanotrasen has been hard at work shipping out new (old) warning tape rolls for departments with common hazards!
/:cl:


OrionTheFox:
pleas someone read thru the code, i left comments explaining how stuff should work, some of which are "if you think its better I can add x" and i'd rather add/delete those before its merged

Gandalf2k15 (approved):

PR #6378, opened 2021-06-18
[NON-MODULAR] changes map voting to exclude both box maps if one would be excluded

dankmemerino147:
## About The Pull Request

see title. If either icebox or journey is put on the blocked_maps list, the other one will be as well. (the blocked_maps list handles excluding maps from voting and rotation)

## Why It's Good For The Game

icebox and journey are fundamentally the same map; not having this change can, and has, lead to streaks of only box being played, which is part of what the blocked_maps list is designed to prevent.

"well actually icebox is multi-Z and journey isn't" fuck off immensely. the fact that the phrase "box maps" is understood to refer to both of them illustrates that they are the same map for >90% of intents and purposes.

## Changelog
:cl:
fix: box maps are treated as one map for map voting/rotation purposes
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -122,12 +122,12 @@ SUBSYSTEM_DEF(persistence)
 				run++
 		if(run >= 2) //If run twice in the last KEEP_ROUNDS_MAP + 1 (including current) rounds, disable map for voting and rotation.
 			blocked_maps += VM.map_name
-		//SKYRAT EDIT START - this way to do this sucks, but i don't expect icebox or journey to get renamed anytime soon
-		if(VM.map_name == "Ice Box Station")
-			blocked_maps += "NSS Journey"
-		if(VM.map_name == "NSS Journey")
-			blocked_maps += "Ice Box Station"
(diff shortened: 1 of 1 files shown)


unit0016:
I for one would just like Icebox gone while keeping Journey but I'm ~~absolutely~~ probably biased seeing as it's my work 🤡

dankmemerino147:
i have no clue why that runtime is happening, it's not affected by anything i changed

FlamingLily:
Nonmodular, actually, but commented correctly so it's fine.

dankmemerino147:
> 
> 
> Nonmodular, actually, but commented correctly so it's fine.

oh i'll change it then

SpaceVampire:
Playing on those maps doesn't feel like the same thing at all.
The location and design totally matters.
Why is people voting for what they want to play an issue?

ZephyrTFA:
These maps play and feel completely different. I don't see the need for this

Gandalf2k15:
Yeah the maps are completely different. This is an indirect way to kill icebox. People vote for what they want. This is also an indirect way to restrict their vote.

PR #6397, opened 2021-06-20
makes vine mutations make gases at 293 instead of 100

Elli-Skala:
## About The Pull Request

sets the temp on n2, o2, and miasma vine to 293 (about 20c) instead of 100

## Why It's Good For The Game

clearly a over site, as rather its meant to stabilize o2 and n2,  or make sickening miasma. but not just make the vines passively freeze everyone around it even without a breach ( even more so with 2 supost to be GOOD mutations...people make good space vines right?)
## Changelog
:cl:
fix: fixed space vine gases being freezing cold 
/:cl:


Elli-Skala:
problems i still see but dont know how to fix is i dont think vines limits how much gas they make (fine for miasma but not the posative n2 and o2) and also n2 and o2 most likely dont have the right mix

PR #6425, opened 2021-06-21
[NON-MODULAR] Fixes heretic books created by the ritual coming with charges.  

SpaceVampire:
## About The Pull Request

Makes it so that the ritual to make a new book, only grants you a new book instead of also coming with points.

## Why It's Good For The Game

While this may not stop Heretics from ascending round-start with three monkeys, I think it is a step in the right direction.
I mean, not that anyone really abused this yet, but I noticed and yeah it is dumb.

## Changelog
:cl:
fix: No free charges with the book ritual.
/:cl:

### DIFF
--- a/code/modules/antagonists/eldritch_cult/eldritch_book.dm
+++ b/code/modules/antagonists/eldritch_cult/eldritch_book.dm
@@ -144,3 +144,6 @@
 
 /obj/item/forbidden_book/debug
 	charge = 100
+
+/obj/item/forbidden_book/ritual  //SKYRAT EDIT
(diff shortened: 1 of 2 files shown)


Cheshify:
Pretty sure this could be upstream, since it IS an exploit

SpaceVampire:
> Pretty sure this could be upstream, since it IS an exploit

I mean, I guess I could try?

SpaceVampire:
I'm just going to wait for it to get merged up-stream as I PRed it there already.
Thus this isn't needed anymore.

PR #6451, opened 2021-06-22
Removes Late-Join cyborg slots

SpaceVampire:
## About The Pull Request

Removes Late-Join cyborg slots.
If you wish to play a cyborg after round-start.
You can do it by getting borged, or by ghosting and using a PosiBrain.

## Why It's Good For The Game

The reason for this is two-fold. Currently if you late-join as a cyborg you won't be law-synced and linked with the current AI correctly. We have had a number of PRs that attempted to fix this issue, but it is still causing problems.
Second of all. If you want to join as a cyborg after the start of the round. You should just do it via the methods that already exist for it. We have the ectoscopic sniffer to detect ghosts, you can poke Robotics to print brains with it.
Making more cyborgs is a core part of their job.

I decided to go ahead and leave the amount of round-start slots at three instead of the normal one.
Because if a couple of people want to do it then when the tech for brains isn't ready, that is fine.
But with how long our rounds are, if you join late, all the stuff you need to spawn should already be there.

## Changelog
:cl:
balance: Late-Join cyborgs not being linked to the AI correctly.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/cyborg.dm
+++ b/code/modules/jobs/job_types/cyborg.dm
@@ -2,7 +2,7 @@
 	title = "Cyborg"
 	auto_deadmin_role_flags = DEADMIN_POSITION_SILICON
 	faction = "Station"
-	total_positions = 3	// SKYRAT EDIT: Original value (0)
+	total_positions = 0
 	spawn_positions = 3	// SKYRAT EDIT: Original value (1)
 	supervisors = "your laws and the AI" //Nodrak
 	selection_color = "#ddffdd"
--- a/config/jobs.txt
+++ b/config/jobs.txt
@@ -52,4 +52,4 @@ Security Medic=1,1
 Corrections Officer 1,1
 
 AI=1,1
-Cyborg=3,3
+Cyborg=0,3


ZephyrTFA:
Yea; that changelog entry is just for GBP farming. Please use the correct changelog label

Gandalf2k15 (approved):
Looks good code wise

RatFromTheJungle:
Balance wise, this is probably for the better, I wouldn't be against someone making a PR that moves posibrains 'earlier in the research tree at the same time, however.

SpaceVampire:
> Yea; that changelog entry is just for GBP farming. Please use the correct changelog label

https://github.com/Skyrat-SS13/Skyrat-tg/pull/3234

The person who added the slots marked it as a fix.
So I must do the same, to keep the universal balance.
(Also GBP doesn't work here, and neither does our changelog.)

YakumoChen:
People forget you can also just go up to the counter and get your brain ripped out as a late join assistant :)

GoldenAlpharex:
> The person who added the slots marked it as a fix.
> So I must do the same, to keep the universal balance.
> (Also GBP doesn't work here, and neither does our changelog.)

GBP actually works, it's just kept on a [separate branch](https://github.com/Skyrat-SS13/Skyrat-tg/blob/gbp-balances/.github/gbp-balances.toml).

SpaceVampire:
> GBP actually works, it's just kept on a [separate branch](https://github.com/Skyrat-SS13/Skyrat-tg/blob/gbp-balances/.github/gbp-balances.toml).

More in that it uh. Follows _Whose Line Is It Anyway?_ rules here.

Gandalf2k15:
Thou shalt not cheat the GBP. Though I'm fine with this being merged.

FlamingLily:
I don't agree with this, since posibrains rely on research and on other players, this effectively means latejoining as a Borg would now rely on other players.

It's a better path of action to try and fix the latejoining syncing with AI

SpaceVampire:
> I don't agree with this, since posibrains rely on research and on other players, this effectively means latejoining as a Borg would now rely on other players.

I mean, its not especially different from wanting to play some job only to find that it is full because you didn't get it on the ready up. With the length of our rounds there are pretty much always posibrains ready, and we have a device that can allow ghosts to ping for em. You can always join in person and just ask to be borged directly as well. I find that being able to just skip all of that is kind of anti-RP in a number of ways.

> It's a better path of action to try and fix the latejoining syncing with AI

It really just keeps breaking, and even if it didn't. Cyborgs are special to the point that having one able to just randomly join at any moment with full access to binary is really annoying if you are trying to do something related to them, because you can't control their creation. They are really supposed to be balanced around requiring stuff to make once the shift is in action.

Funce:
Please discuss the necessity and severity of this change elsewhere. Technical review only.

hugmeorelse:
> Please discuss the necessity and severity of this change elsewhere. Technical review only.

Thank you, Funce. You can create a thread specifically for this on the forums for that. You may even link it here.

Gandalf2k15:
Doot doot merge train incoming

PR #6466, opened 2021-06-23
Update stale.yml

Gandalf2k15:
Updates the workflow to be far less lenient so I'm actually compelled to make things move.


RatFromTheJungle:
ok i dont know what this does but I'll allow it

PR #6472, opened 2021-06-23
Cyborgs have Late-join slots again (and spawn on-station when late-joining), and fixes linking to the AI properly on latejoin.

FlamingLily:
## About The Pull Request
This reverts #6451, while addressing the issue cited in the removal, namely 

> Currently if you late-join as a cyborg you won't be law-synced and linked with the current AI correctly. We have had a number of PRs that attempted to fix this issue, but it is still causing problems.

Has been fixed. When latejoining, a cyborg will be spawned on the AI satellite, the same way prisoners are spawned in their cells. They are then linked to the first located AI on the z level.

## Why It's Good For The Game
The reason I disagree with #6451 is as follows. Firstly, the issue that it was created to avoid has now been fixed, secondly, the changes make it difficult to play cyborg on low population shifts. You require at least one roboticist, let alone one who is working and doing their job, *and* at least one scientist who is actively doing RnD beyond what mining will yell at them for. Sometimes, science will just not research posibrains because they're running on a skeleton crew. Also, the topic of "It stifles RP to be able to skip the borging process," came up, but I would argue the opposite. If you possess a posibrain, you are shoehorned into one single RP path of being a posibrain cyborg. If you want to be an MMI cyborg, you either have to join at roundstart, or go up to the desk and get borged, neither of which are the easiest to RP out depending on the circumstances.

The only argument that I agree with is that "If you're a traitor doing something with a binary key, a latejoining borg can easily rumble you." But I would count that as such an edge case because a) Binary key is rarely used, B) that goes against validhunting rules in silicon policy and C) If you have subverted the AI and any unlinked borgs, chances are you've already gotten an upload terminal for both, and a new borg wouldn't matter all that much. Even then, I feel that an easy fix would be to make new borgs be announced over the binary channel, for instance, so when a traitor is using binary, they can know when to be quiet for a moment.

And as for spawning on station, the logic there, aside from it making the code easier, is to say "hey, why would a station asset that was presumably made on station, and presumably enters storage at the end of a shift. be shipped in from a transit hub and not the cargo shuttle, or just taken out of on-station storage."

## Changelog
:cl:
add: Latejoin Cyborg Slots, again.
add: Latejoin cyborgs now spawn on-station.
fix: Latejoin Cyborgs being linked to the AI, again.
/:cl:

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -738,12 +738,15 @@ SUBSYSTEM_DEF(job)
 		destination = pick(GLOB.jobspawn_overrides[M.mind.assigned_role])
 		destination.JoinPlayerHere(M, FALSE)
 		return TRUE
-	//SKYRAT EDIT ADDITION
-	if(M.job)
-		if(M.job == "Prisoner")
-			destination = locate(/obj/effect/landmark/start/prisoner) in GLOB.landmarks_list
-			destination.JoinPlayerHere(M, buckle)
-			return TRUE
+	//SKYRAT EDIT ADDITION -- ONSTATION LATEJOINS
+	if(M.job && M.job == "Prisoner")
+		destination = locate(/obj/effect/landmark/start/prisoner) in GLOB.landmarks_list
+		destination.JoinPlayerHere(M, buckle)
+		return TRUE
+	if(M.job && M.job == "Cyborg")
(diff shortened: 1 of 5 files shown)


Gandalf2k15:
There are also balance concerns over the slots. I don't want too many cyborgs roundstart.

Gandalf2k15:
We already agreed to remove these.

PR #6539, opened 2021-06-26
[FIX] prevents ghosts from un-flipping tables

SpookyTheFox:
made only the living qualify for un-flipping tables

## About The Pull Request
Ghosts were able to un-flip tables. I found what I think let them do it, and changed it. 

## Why It's Good For The Game
Ghosts had a narrow case where they could directly interact with the round, potentially in a combat situation.

## Changelog
:cl:
fix: prevented ghosts from un-flipping tables
/:cl:

### DIFF
--- a/modular_skyrat/modules/tableflip/code/flipped_table.dm
+++ b/modular_skyrat/modules/tableflip/code/flipped_table.dm
@@ -64,7 +64,7 @@
 
 /obj/structure/table/CtrlShiftClick(mob/living/user)
 	. = ..()
-	if(!istype(user) || !user.can_interact_with(src))
+	if(!istype(user) || !user.can_interact_with(src) || isobserver(user))
 		return
 	if(can_flip)
 		user.visible_message("<span class='danger'>[user] starts flipping [src]!</span>", "<span class='notice'>You start flipping over the [src]!</span>")
--- a/modular_skyrat/modules/tableflip/code/flipped_table.dm
+++ b/modular_skyrat/modules/tableflip/code/flipped_table.dm
@@ -48,7 +48,7 @@
 	if(get_dir(leaving.loc, new_location) == dir)
 		return COMPONENT_ATOM_BLOCK_EXIT
 
-/obj/structure/flippedtable/CtrlShiftClick(mob/living/user)
+/obj/structure/flippedtable/CtrlShiftClick(mob/user)
 	. = ..()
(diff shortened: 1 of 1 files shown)


SpookyTheFox:
If possible, I'd like someone else to look at this. It feels entirely too easy, like I missed something. But it does work in local test.

lyricalpaws:
That fix should work perfectly, checking if they're living or not will stop the damn ghasts from flipping my tables.

modular_skyrat/modules/tableflip/code/flipped_table.dm
```diff
@@ -48,7 +48,7 @@
 	if(get_dir(leaving.loc, new_location) == dir)
 		return COMPONENT_ATOM_BLOCK_EXIT
 
-/obj/structure/flippedtable/CtrlShiftClick(mob/user)
+/obj/structure/flippedtable/CtrlShiftClick(mob/living/user)
 	. = ..()
 	if(!istype(user) || !user.can_interact_with(src))
```
ZephyrTFA:
```suggestion
	if(!istype(user) || !user.can_interact_with(src) || isobserver(user))
```

modular_skyrat/modules/tableflip/code/flipped_table.dm
```diff
@@ -48,7 +48,7 @@
 	if(get_dir(leaving.loc, new_location) == dir)
 		return COMPONENT_ATOM_BLOCK_EXIT
 
-/obj/structure/flippedtable/CtrlShiftClick(mob/user)
+/obj/structure/flippedtable/CtrlShiftClick(mob/living/user)
```
ZephyrTFA:
```suggestion
/obj/structure/flippedtable/CtrlShiftClick(mob/user)
```

ZephyrTFA (changes requested):
I've known how to fix this for awhile I'm just lazy

modular_skyrat/modules/tableflip/code/flipped_table.dm
SpookyTheFox: Interesting. I know where I can use this for a similar oversight involving ghosts where I can't use mob/living. For my own learning purposes on this particular PR though, why is this method better? I would think that living user would be preferred since it should cover edge cases like dead but not observing mobs messing with tables.

PR #6557, opened 2021-06-27
[MIRROR] Removes yellow siding tiles

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/59820
--------------------

## About The Pull Request
Removes these two fellows: 
image

## Why It's Good For The Game
They're only used at Centcom, and we have decals now to do what these tiles do. 

## Changelog
:cl:
imagedel: removes the yellow siding tiles at Centcom
/:cl:


Arctaisia:
I think we also use these at the interlink, might be best to keep them, at least for now, heck, unless we have our resprites in a seperate file

PR #6561, opened 2021-06-28
[MODULAR] Requested Cordyceps Changes

Cheshify:
## About The Pull Request

changes the Spaceacillin & Convermol cure to Neurine & Modafinil

## Why It's Good For The Game

So like, I know, "why would you do this you're literally the sequel to the devil incarnate wahhh" but Feedback team debated over this for a long time, and we eventually came to the agreement that Cordyceps is supposed to be the most deadly disease possible (aside from like zombies), and the fact that it's cured by two readily available chems is ridiculous. 

So, what this PR does is change up the cures so it requires Neurine (which a well stocked medbay should already have) and Modafinil (which a well upgraded medbay can make easily), meaning that the cure is less dependent on someone knowing that medbay has all the cures, and medbay being prepared & upgraded for just normal day-to-day issues.

## Changelog
:cl:
balance: A new strain of Cordyceps has been found, requiring Neurine and Modafinil to cure.
/:cl:

### DIFF
--- a/modular_skyrat/modules/biohazard_blob/code/biohazard_blob_disease.dm
+++ b/modular_skyrat/modules/biohazard_blob/code/biohazard_blob_disease.dm
@@ -3,8 +3,8 @@
 	name = "Cordyceps omniteralis"
 	max_stages = 5
 	spread_text = "Airborne"
-	cure_text = "Spaceacillin & Convermol"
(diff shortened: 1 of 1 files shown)


lyricalpaws:
Mold my beloved

PR #6636, opened 2021-07-01
Revert "[semi-modular] reagent forging"

ZephyrTFA:
This might be the most broken thing we've ever had. Hard pass.


Useroth:
Cool idea but it needs a bunch more work to fit it into the game.

PR #6653, opened 2021-07-02
[Modular] Two New Posters (+ Custom Poster "Framework")

unit0016:
## About The Pull Request
Integrates a 'framework' (Read: stupid workaround rolled into the aesthetics module) for having our own modular custom posters, and creates two new posters; one commemorating a certain round where DS-2 ~~may have done~~ absolutely did more for the dead station miners than NT medical either did... and another celebrating what Medical was up to that same round.

## Why It's Good For The Game
Aesthetics module is getting on my nerves a wee bit but hey whatever if it works it works

## Changelog
:cl:
expansion: One new contraband poster seems to be in circulation..
/:cl:


Arctaisia:
Can I kiss you yet for this? <3

We also need one for the crocin pool incident of 2561...

unit0016:
> Can I kiss you yet for this? <3
> 
> We also need one for the crocin pool incident of 2561...

👩‍🎨

SpaceVampire:
Can you include an image of the posters in question in the PR?

unit0016:
> Can you include an image of the posters in question in the PR?

They're rather understated so I didn't see much point, but here.
Syndicate Medical.
image
Crocin Pool
image

Gandalf2k15:
can you make it not override the base poster and instead add it as a child?

unit0016:
> can you make it not override the base poster and instead add it as a child?

I've tried that, it causes issues with random posters (like the ones found in maint or mapped in) because postercode sucks. I'd have to code it to also inherit the .dmi of whatever poster subtype it pulled and I might be a bit too stupid for that one.

PR #6662, opened 2021-07-03
[MODULAR] Rust Component

ZephyrTFA:
## About The Pull Request

Adds and implements a new component: Rust.
It can be added to anything that has the icon_state "rust" in its icon file

## Why It's Good For The Game

It's good for the game because it simplifies existing methods and makes it possible to expand it in the future

## Changelog
:cl:
code: Rust is now a component instead of a hard path type
/:cl:


ZephyrTFA:
This is upstream now

PR #6663, opened 2021-07-03
Firelocks remastered

Gandalf2k15:
Firelocks now come in two types, standard and onetile, and all reinforced windows now come with firelocks as standard!

Atmos techs can now force open a firelock for repairs.


ZephyrTFA:
Don't forget to set override to TRUE in the RegisterSignal call since you are overriding a signal already present

Gandalf2k15:
Borgs can now ctrl click a firelock to force it open. Firelocks will autoclose after 3 minutes.

PR #6664, opened 2021-07-03
plasmamen quality of life update(s)

CandleJaxx:
so basically y'know how plasmamen can sometimes spawn without envirosuits on prison guard and vanguard? this fixes that

also gives the mining plasma suit and helmet fire resistance as to match other jobs that rely on that resistance so it can be played

### DIFF
--- a/code/modules/clothing/spacesuits/plasmamen.dm
+++ b/code/modules/clothing/spacesuits/plasmamen.dm
@@ -252,6 +252,8 @@
 	icon_state = "explorer_envirohelm"
 	inhand_icon_state = "explorer_envirohelm"
 	visor_icon = "explorer_envisor"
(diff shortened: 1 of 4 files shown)


code/modules/clothing/spacesuits/plasmamen.dm
```diff
@@ -252,6 +252,8 @@
 	icon_state = "explorer_envirohelm"
 	inhand_icon_state = "explorer_envirohelm"
 	visor_icon = "explorer_envisor"
+	armor = list(MELEE = 10, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 100, RAD = 0, FIRE = 100, ACID = 75)
```
Gandalf2k15: Make sure you add a //SKYRAT EDIT comment here;)

PR #6369, opened 2021-06-18
Synthetic wounds

Gandalf2k15:
## About The Pull Request

Continues the work done by https://github.com/Skyrat-SS13/Skyrat-tg/pull/6359 - Yawet, hopefully, they come back to help, Though, this seems to work well so far!

Adds a multitude of wounds to synthetics and buffs their core statistics slightly.

For insight on the wounds, see codde.

## Changelog
:cl:
add: Synthetics can now suffer moderate to severe wounds, watch out!
/:cl:

PR #6669, opened 2021-07-03
Revert "Synthetic wounds"

SpaceVampire:
Reverts Skyrat-SS13/Skyrat-tg#6369

Due to a couple of issues with the code, this PR is here as a hopefully temporary revert until they are fixed.


dekunutkid:
I agree with this. Once you get a wound they simply cannot be removed, and will delete any amount of healing you put into it in order to treat a wound that will never heal, and not heal any of the damage under it.

hugmeorelse:
Not stale. BAD!

KasparoVy:
Fug :(

PR #6693, opened 2021-07-03
[modular][ready]adds an NT rep office to the station

thestubborn:
## About The Pull Request

title

## Why It's Good For The Game

maybe you wanna hold a private meeting in an area you have the power
maybe you just wanna vibe away from command
maybe its better than having to go back and forth if you forgot anything

## Changelog
:cl:
expansion: A run down office has been decommissioned and made into an office for the Rep.
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -3042,11 +3042,11 @@
 /area/maintenance/port/fore)
 "aky" = (
 /turf/closed/wall,
-/area/commons/vacant_room/office)
+/area/command/heads_quarters/captain)
 "akz" = (
 /obj/structure/sign/nanotrasen,
 /turf/closed/wall,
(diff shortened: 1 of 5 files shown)


SpaceVampire:
I always just assumed they would go ahead and use the command meeting room.
Most maps have something along that line, right?
Given that they aren't a normal presence on the actual station. 
(Also post images of the new offices in the PR.)

thestubborn:
> 
> 
> I always just assumed they would go ahead and use the command meeting room.
> Most maps have something along that line, right?
> Given that they aren't a normal presence on the actual station.
> (Also post images of the new offices in the PR.)

I mean, they're a roundstart role now, I see it more as a guest room
also, fine
bitch
image
image
image
image
image

SpaceVampire:
(Just do the funny thing we talked about in DMs instead, and leave those rooms to the whims of the greytide.)
Comes with the added bonus of making it so we don't have to worry about map-resets as well.
_slut_

PR #6737, opened 2021-07-05
[MODULAR] Stamina Update - Three Hops This Time

ZephyrTFA:
## About The Pull Request

Updates the defines for stamina flags.
Stamina damage is capped to MaxHP - (burn + brute); as such the maximum stamina damage you can ever have is MaxHP.
MaxHP is currently 135.
All limbs have a static stamina_modifier that changes how much stamina damage is converted to stamina loss.
This value defaults to 0.75 and is unchanged across most, if not all, boydparts.
This means the most stamina loss you can ever have is 101.25

Currently the only items that can actually put you into stam crit are items that directly set stamina damage instead of using the correct procs.

## Why It's Good For The Game

Resolves an issue with stamina that has been present for five months due to myself and many others making bad assumptions about procs and how they work.
We apologize for the inconvenience.

### DIFF
--- a/code/__DEFINES/~skyrat_defines/combat.dm
+++ b/code/__DEFINES/~skyrat_defines/combat.dm
@@ -28,16 +28,16 @@
 #define STAMINALOSS_REGEN_COEFF 50
 
 //Thresholds for detrimental effects from stamina
-#define STAMINA_THRESHOLD_WEAK 60
+#define STAMINA_THRESHOLD_WEAK 25
 
-#define STAMINA_THRESHOLD_KNOCKDOWN 120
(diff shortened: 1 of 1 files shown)


code/__DEFINES/~skyrat_defines/combat.dm
```diff
@@ -28,16 +28,16 @@
 #define STAMINALOSS_REGEN_COEFF 50
 
 //Thresholds for detrimental effects from stamina
-#define STAMINA_THRESHOLD_WEAK 60
+#define STAMINA_THRESHOLD_WEAK 25
```
Gandalf2k15, line 31: This is more than 50% less than what it was. We should be starting small. Not the other way round.

PR #6753, opened 2021-07-06
[Modular] Tribal bows no longer have safeties.

unit0016:
## About The Pull Request
Highly repurposed branch.
This gives ashie bows no safeties, because this is a serious oversight
and is also a one line fix

a bit like how digitigrade outfits are roughly a 16 pixel change and not that big a deal 😳 

## Why It's Good For The Game
Ashwalkers are SERIOUS about gun safety. We sent a whole security team down there, and sure, their heads were on pikes when we came back - but one guy we cloned rated them the best damn instructors he ever had. He quotes, "Oh god, they were raining down on us. Carl got hit first, right through the shoulder. I watched his arm get split open like meat, all because those damn lizards had their feet shoulder-width apart, had their safeties on only while pointing downrange, and worst of all? Wore proper protective equipment."

## Changelog
:cl:
fix: Tribal bows no longer have safeties.
/:cl:


Cheshify:
I kept the safeties for a reason, like EVERY OTHER ranged weapon has one

hugmeorelse:
Safeties in this case can simply be not preloading the arrow onto the bow. I'm not sure what the point of this PR is.

hugmeorelse:
Safeties in this case can simply be not preloading the arrow onto the bow. If it's about believability, one can use your imagination like everything else in SS13. I don't think this is really needed in my opinion.

hugmeorelse:
Safeties in this case can simply be reimagined as **NOT** preloading the arrow onto the bow if you're concerned about plausibility. It doesn't have to strictly relate to guns. One can use their imagination like everything else in SS13 to perceive it that way. I don't think this PR is really needed in my opinion and will increase lizard on lizard friendly fire!

unit0016:
Host has spoken, ashies will continue to be godly at the range

PR #6757, opened 2021-07-06
[modular][request]gives the BS CC access, and CC Comms

thestubborn:
## About The Pull Request

title

## Why It's Good For The Game

cc rep gets it, so the BS should, as they're also a member of cc

## Changelog
:cl:
expansion: the BS now has CC access and CC Comms
/:cl:


thestubborn:
*shrug

PR #6771, opened 2021-07-07
fixed rogue sprite n stuff donor item

CandleJaxx:
also added shaft miner to job list


SpaceVampire:
There is kind of an issue with your donor items.
Looking at the stats for them, it is possible to take it with jobs where that gear is better then what you would otherwise have.
That was super egregious before it got hot-fixed, because the first PR had an issue armor value on the under stuff, that was better then the gear the Hos is supposed to get. Which again, you could take as a normal officer.
armor = list(MELEE = 20, BULLET = 10, LASER = 10, ENERGY = 10, BOMB = 10, BIO = 100, RAD = 0, FIRE = 100, ACID = 75, WOUND = 10) vs armor = list(MELEE = 10, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 100, RAD = 0, FIRE = 95, ACID = 95, WOUND = 10).
Even after that was fixed, your other helmet wasn't touched.
So it is better then what you are supposed to be getting as a normal sec-officer. Armor = list(MELEE = 10, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 100, RAD = 0, FIRE = 100, ACID = 75)
On top of that, could I ask about this other PR of yours?
https://github.com/Skyrat-SS13/Skyrat-tg/pull/6664
You said "also gives the mining plasma suit and helmet fire resistance as to match other jobs that rely on that resistance so it can be played"
But then went ahead and buffed it to be on par with a security suit.
Where the one normal shaft miners get isn't supposed to come with bonus melee armor.
Seems a bit odd, and outside of the scope of what you said in the actual PR.

CandleJaxx:
new to coding stuff in general so i just copied data from other jobs @.@ i ran it by user earlier but if you could suggest the changes you mean, by all means do

SpaceVampire:
I mean, I would recommend deciding what job you really want to use the stuff armor with, and then making it a subtype of the stuff that job gets already. Otherwise you end up with an issue if you set it to any fixed number, where it would be weaker then the normal Blueshield stuff, but stronger then the normal sec-officer stuff. Until we add some kind of reskin kit there isn't a great way to handle it outside of that for non-cosmetic items. As for the changes in #6664 I'm pretty sure they already had that stuff from their parent, so I am honestly confused about what you were trying to accomplish unless I missed something.

CandleJaxx:
#6664 was two-fold, uhm, so i did know that brig officer and vanguard didn't spawn w/ envirosuits, so i added outfit lines in their job files for plasmamen on one part- shaft miner plasmamen didn't get any fire protection that other risky jobs (sec, blueshield, etc) would have, so i tried to make shaft-miner plasmamen viable

SpaceVampire:
> #6664 was two-fold, uhm, so i did know that brig officer and vanguard didn't spawn w/ envirosuits, so i added outfit lines in their job files for plasmamen on one part- shaft miner plasmamen didn't get any fire protection that other risky jobs (sec, blueshield, etc) would have, so i tried to make shaft-miner plasmamen viable.

The first part of that is fine, because it was pretty clearly an oversight that no one thought about when adding those roles.
But uh. I'm still not sure why claim they didn't have any fire protection.
Do you not understand that /obj/item/clothing/head/helmet/space/plasmaman/mining is a subtype of /obj/item/clothing/head/helmet/space/plasmaman/ and thus inherits the armor values from that already?
	armor = list(MELEE = 0, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 100, RAD = 0, FIRE = 100, ACID = 75)
So when you went ahead and added.
	armor = list(MELEE = 10, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 100, RAD = 0, FIRE = 100, ACID = 75)//SKYRAT EDIT
All you did was buff their melee armor. Which is kind of outside of the scope you were talking about in the PR, and puts miner armor on the same level as security. As you can see, it already had fire protection.

CandleJaxx:
> Do you not understand that /obj/item/clothing/head/helmet/space/plasmaman/mining is a subtype of /obj/item/clothing/head/helmet/space/plasmaman/ and thus inherits the armor values from that already?

ahh yeah no i wasn't aware that that was a thing - i copied an existing armor value further down, fair enough!

Useroth:
The PR by itself is fine.
Other concerns will be resolved in due time. I intend to make reskin kits that will just convert normal suits into the donator looks, while retaining the stats of the original to resolve the discrepancies and make things fair.

CandleJaxx:
thank you love you

PR #6831, opened 2021-07-10
[sprite change] sec plasmaman sprite update

CandleJaxx:
resprites red plasmamen sec uniforms to match the current peacekeeper aesthetic


ghost:
would this not be better as a new item rather than removing the oldsec variant? I know the oldsec stuff sees some use in events and whatnot..

Also I really need to change which account im on jfc.  This is orion hi hello

Gandalf2k15:
Non modular changes to icons aren't acceptable, best thing to do is make a peacekeeper variant and make the changes in the modular folder

SpaceVampire:
Yeah, if you make a non-modular chance to an core icon file, you have to worry about it ever being touched in the future.
It should be handled the way that was suggested above.

Gandalf2k15:
Hi please enact the above changes and reopen this later.

PR #6886, opened 2021-07-12
Rebalances mutants [PENDING TEST]

Gandalf2k15:
## About The Pull Request

Mutants have had an across-the-board nerf due to concerns regarding the sustainability of security and science during an outbreak. This hopefully addresses them.

Most relative values have taken a 20% decrease.


SpaceVampire:
I feel like the main issue, if you want it to be a viable random event.
Is that the "Patient Zero" should be special in some way.
The numbers work fine once an outbreak actually starts.
But at the moment, you are just going to end up in medical if it triggers on you.
With it being very clear what is going on here.

Funce:
Still test merged so regrettably, stale-b-gone

PR #6891, opened 2021-07-13
[NON-MODULAR] Immovable Rods are nicer to people laying down

ZephyrTFA:
## About The Pull Request

Host Request.

If you are hit by an immovable rod while laying down it will only deal enough damage to take you to 20% hp and no lower.
If you are standing you will still get railed; maybe dont stand in the way of an immovable rod

## Why It's Good For The Game

You can do something other than pray against immovable rods

## Changelog
:cl:
balance: Immovable rods are nicer to people who are resting.
/:cl:

### DIFF
--- a/code/modules/events/immovable_rod.dm
+++ b/code/modules/events/immovable_rod.dm
@@ -7,6 +7,7 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 --NEOFite
 */
 
+#define MAXHP_RESTING 0.80 // SKYRAT ADDITION - SEE BELOW
 /datum/round_event_control/immovable_rod
 	name = "Immovable Rod"
 	typepath = /datum/round_event/immovable_rod
@@ -74,7 +75,9 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 	var/loopy_rod = FALSE
 	//SKYRAT ADDITION - RODS DEAL LESS DAMAGE TO MOBS LAYING DOWN
 	/// The percentage of maxhp to apply as damage to a mob laying down on Bump
-	var/maxhp_damage_resting = 0.80
+	var/maxhp_damage_resting = MAXHP_RESTING
+	///Who have we already hit so far
+	var/list/client/already_hit = list()
 	//SKYRAT ADDITION - END
 
 /obj/effect/immovablerod/New(atom/start, atom/end, aimed_at, force_looping)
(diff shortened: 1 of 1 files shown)


jinxynii:
do you still take wounds from being hit while lying down? if so, i don't see much of a point, because even while standing upright, everytime i've been hit by a rod, i've been okay enough to make it back to medbay, even from the bar

FlamingLily:
Is this needed? I don't really understand the rationale behind it...

My understanding of this is that if you've already got a certain amount of damage, it won't damage you, might be wrong though.

Thing is, if you cap the damage, some "funny" guy is going to lay down infront of a looping rod and be like "haha look at me the rod can't damage me anymore because I'm invincible" and that's uh not good.

It also just, removes the threat from immoveable rods, if it works this way. Sure, the idea of "ROD, GET DOWN" is fun, this is no better than just making it not damage laying down people at all, because it doesn't, after a certain point.

I would recommend just making it do less damage to laying down people, but still able to kill.

code/modules/events/immovable_rod.dm
```diff
@@ -72,6 +72,10 @@ In my current plan for it, 'solid' will be defined as anything with density == 1
 	var/dnd_style_level_up = TRUE
 	/// Whether the rod can loop across other z-levels. The rod will still loop when the z-level is self-looping even if this is FALSE.
 	var/loopy_rod = FALSE
+	//SKYRAT ADDITION - RODS DEAL LESS DAMAGE TO MOBS LAYING DOWN
+	/// The percentage of maxhp to apply as damage to a mob laying down on Bump
+	var/maxhp_damage_resting = 0.80
```
Gandalf2k15: Why isn't this a define?

ZephyrTFA: to allow adminbus if people decide they want to just lay under the rod and pretend to be invincible

thestubborn:
> 
> 
> Is this needed? I don't really understand the rationale behind it...
> 
> My understanding of this is that if you've already got a certain amount of damage, it won't damage you, might be wrong though.
> 
> Thing is, if you cap the damage, some "funny" guy is going to lay down infront of a looping rod and be like "haha look at me the rod can't damage me anymore because I'm invincible" and that's uh not good.
> 
> It also just, removes the threat from immoveable rods, if it works this way. Sure, the idea of "ROD, GET DOWN" is fun, this is no better than just making it not damage laying down people at all, because it doesn't, after a certain point.
> 
> I would recommend just making it do less damage to laying down people, but still able to kill.

host request

code/modules/events/immovable_rod.dm
Fikou: vars can use defines

PR #6921, opened 2021-07-15
Armory minidate

Gandalf2k15:
## About The Pull Request

Primary and energy tokens have been removed, secondary tokens will remain with the choice bloat from before. You will spawn with one in your backpack. The vending machine will stay.

All security officers and the warden now spawn with disablers as default, along with their tokens of course.

So what I've done here is created a new way of spawning armory weapons which to me, is a better way than I had it. Instead of having boxes of useless tokens, guns now actually spawn there. However, there's a bit of randomness to them AND they slot into racks very nicely.

I have modified each sec area with the relevant changes as well. There are 4 shotguns and 4 rifles that spawn in the armory at the round start using a unique random spawner which should add a lot of variety to the pool of weapons.

I have also added a new shotgun type, the SAS14. a 14 gauge 5 slug mag-fed shotgun. 14 gauge is our way of balancing shotguns for security without impacting the balance of guns.

image

Some bullet values have been fixed and changed.

## Changelog
:cl:
balance: Guns now start in the armory instead of tokens.
/:cl:


Gandalf2k15:
I'll post screenies soon.

CameronWoof:
Does this mean that Security is no longer able to choose their own sidearms? And that the just-made-available-again disablers are also out?

Gandalf2k15:
> Does this mean that Security is no longer able to choose their own sidearms? And that the just-made-available-again disablers are also out?

The energy spawner isn't done yet, so, no

Cheshify:
How does affect the sarge token that the HOS gets?

Gandalf2k15:
> How does affect the sarge token that the HOS gets?

fuck

capsaicinz:
alright, i commented on this in sooc, but i might as well bring this here as well, in more depth. i think this is an entirely bad idea, forcing security to a specific set of weapons out of their control, and even causing possible _shortages_ in the case of highpop security. all sec players tend to have preferences when it comes to the sidearm they use, and i think that by removing the plenty efficient system of tokens, it kind of ruins any sense of choice there.

i understand the hesitation with armory raids and tokens, but can't this be _just_ as easily fixed in a way that doesn't _further_ mess with security? i've already well expressed my disliking to most of these recent changes to security, thinking it just further complicates things and nerfs non-lethal security into oblivion in a desperate attempt to balance things. the taser guns given to officers are all but useless. the only sensible option is the baton or flash, which both require _extremely close_ proximity, giving no effective, ranged way for security to non-lethally disarm somebody.

i have to say, when we had rubber magazines and firearms at the start of the shift as a way to disarm threats, things were good. why not just replace the rubber with IHDF, if we're so worried about officers capping antags on ||blue||?

IgiariValkyr:
So. Just a list here of current complaints from the whole security team of the round (2383):

- There are "too few guns" for if the alert jumps up (I personally think it's fine. No one needs to be running around with two pistols.)
- There are no ammo pouches.
- Everyone still spawns with tokens, which are unusable due to the removal of the machine. (As a HoS, I can't even do proper sergeant promotions because it requires use of a token. Which we can't use.)
- The only gun security can order from cargo to mitigate their bloodlust is the WT550, a full-auto smg.

Gandalf2k15:
Feedback implemented. Check PR body.

KaiserBlackwood:
I've been playing for a bit, and it appears that we don't spawn with secondary tokens at all. Why say that they will stay, if they're clearly not there? Please, do tell me if I am being mistakenn

Gandalf2k15:
Primary and energy tokens have been removed, secondary tokens will remain with the choice bloat from before. You will spawn with one in your backpack. The vending machine will stay.

All security officers and the warden now spawn with disablers as default, along with their tokens of course.

Gandalf2k15:
> I've been playing for a bit, and it appears that we don't spawn with secondary tokens at all. Why say that they will stay, if they're clearly not there? Please, do tell me if I am being mistakenn

It was just updated

capsaicinz:
be sure to give the sergeant their secondary token back, as i've already gotten a complaint in game of such. and if they could receive the disabler in their shiftstart loadout, too, that would be awesome!

capsaicinz:
> be sure to give the sergeant their secondary token back, as i've already gotten a complaint in game of such. and if they could receive the disabler in their shiftstart loadout, too, that would be awesome!

it may also be useful to give secondary chips to the armory, in case an officer loses their sidearm. maybe just two or three, so nobody tries to dual-wield.

PR #6933, opened 2021-07-15
[NON MODULAR] Nerfs wheely heels, replaces bike crate with a Medical Beamgun crate (still extremely expensive)

RatFromTheJungle:
## About The Pull Request
What the title says, goes over those two.

## Why It's Good For The Game

Wheely's are op as sin, and make you almost to fast to handle, let alone catch (with the tiniest amount of practice with them)
(Bike replacement is up for debate if someone has a better idea, that also isn't just as OP)

## Changelog
no GBP's for this


Draganfrukts:
Just replace the bike with the american baseball bat...

ghost:
Would it be possible to, aside from slowing the bike, make it take wide turns? I.e. it needs to take 1 diagonal before it can go straight again?
Making it too clunky to use reasonably in the halls would leave it to be used elsewhere, even allowing civilian stations to make bike-specific tracks 🤔

Tho if thats not possible then I understand removing the crate yea. I just like the idea of bringing my bike to work lmao

(This is orion btw ive yet to find out why my phone keeps going to this account)

YakumoChen:
The bike isn't really supposed to be ordered from cargo, it shouldn't really be considered in a balance debate.

It serves as a litmus test for cargo balance as a whole. If cargo is somehow obtaining a million credits, something cargo is doing needs to be nerfed.

modular_skyrat/modules/cargo/code/packs.dm
```diff
@@ -249,6 +249,14 @@
 					/obj/item/stack/sheet/cloth/ten,
 					/obj/item/stack/sheet/cloth/ten)
 	crate_name = "advanced art supplies"
+	
+/datum/supply_pack/misc/bicycle // overwritten
```
Funce: As much as I'd love to sneakily overwrite this, comment it out in the main file, set this as a medical supply pack.

Funce (changes requested):
Sneaky, but we can't compromise quality.

modular_skyrat/modules/cargo/code/packs.dm
RatFromTheJungle: all-done

modular_skyrat/modules/cargo/code/packs.dm
```diff
@@ -273,7 +280,7 @@
 					/obj/item/stack/sheet/cloth/ten,
 					/obj/item/stack/sheet/cloth/ten)
 	crate_name = "advanced art supplies"
-
+
```
Funce: Whitespace-

Funce (approved):

PR #7006, opened 2021-07-18
[MIRROR] Job refactor: strings to references and typepaths [This could seriously break things :)]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/59841
--------------------
On first pass:

* `assigned_role` is now a reference to a job instance (formerly string).
* `overflow_role` is now a typepath (formerly string).
* Turned string roles and factions into defines.
* `AssignRole()` now takes a job instance as input (instead of a string).
* Captains can now be jealousy assassination objective targets again.
* Removes the unused `find_target_by_role()`
* `assignedrole` string converted into the `spawner_job_path` type path.
* Lots of job datums created.
* Before we only had a list of the station-faction job instances, in `occupations`. Substituted that for the two lists `all_occupations` and `joinable_occupations`.
* Created several job flags and some variables to better differentiate their properties.

The gist of this PR is to move job logic to the job datum, eventually auto-populating the relevant lists with the proper job instances and diminishing the amount of hardcoded checks and lists.
There's a lot to do but I'm trying to atomize as much as I can and this has grown large enough, so this should be self-contained and work on its own, while leaving more to do for later.

On second pass (AI and borg code was very snowflake, so I was forced to continue refactoring adjacent code):
* Borgs and AI go through the same job setting-up chain than humans. AI has a weird initialization process, so it has a few snowflake checks that can be refactored out later. This meant requiring to spawn the mob earlier in the process, but it simplifies things.
* Had to refactor hardcore random and atomize some prefs transfer code. Behavior should be unchanged. Added a few randomization procs and did some related cleanup.
* Effectively reverted #60156 , as that PR only touched on backup code and does not really do what it was intended. Additionally, we should stop piling up conditional behavior with extra arguments to procs not meant for it, as it makes untangling and atomizing things harder. Those big procs should be divided into smaller parts instead, each doing their on function.
* Did my best to reduce the messy job spawning chain, reducing conditional arguments and procs that took on too many responsibilities.
* Added some logging to cyborg renaming, while improving that code.
* Refactored acting captaincy code.


ZephyrTFA:
Wow this breaks everything I was using for Dispatch. Cool thanks TG

Gandalf2k15:
Merging for those conflicts

PR #7016, opened 2021-07-19
Reverts infinite showers

Gandalf2k15:
## About The Pull Request

This was requested by someone. The PR that made infinite showers was made badly, causing all showers to be infinite, this reverts it so that we can make plumbing useful again.

Also adds the ability to empty the shower using a plunger!

## Changelog
:cl:
add: You can now empty showers using a plunger!
fix: Showers no longer defy the laws of physics and once again require plumbing.
/:cl:


KathrinBailey:
for anyone that doesn't like this you can fix it with varedits on maps via changing refill_rate
image
for anyone who wants infinite sinks, varedit this into a sink:
image

PR #7004, opened 2021-07-18
[NON-MODULAR] Fixes extra job icons, revamps those icons, adds some new ones for the others that didn't have any

Seris02:
## About The Pull Request

Screenshot 2021-07-22 135330

For any in the future: the modular hud file doesn't work because it isn't used anywhere in the code, and make sure you actually add it to the icon-able jobs, and that the name of the iconstates corresponds to the assignment var of the trim.

The NT Rep icon is the same as the CC officer icon, by admin/developer request.
I couldn't fit three arrows for sergeant, so I just made it a little star.
The civil disputes officer icon is meant to be a person.
And (credit to @unit0016 for it) vanguard looks like the gateway since they go exploring.

## Why It's Good For The Game

bug fix

## Changelog
:cl:
fix: fixed extra job icons not showing on the sechud
imageadd: added vanguard, NT rep, civil dispute officer, corrections officer and security medic HUD icons
/:cl:

PR #7057, opened 2021-07-21
[Semi-Modular] #7004 but Modularized and consistent with the ID cards.

unit0016:
## About The Pull Request
you ever get so fed up with something stupid after you've been up for way longer than you ever should so you make a PR on github

yes

Recreates #7004 but makes it as modular as possible; and also making the icons consistent with card.dmi. Bluesec included!
image
image
image
image
~~No don't ask me why our perfectly fine blue centcom was beseetched by goon-green, that's for the nanotrasen rep coders to find out.~~

## Why It's Good For The Game
Same reasons as #7004 

## Changelog
:cl: Seris02 & Blu
qol: Hud icons extend to the rest of the skyrat-specific, onstation jobs.
/:cl:


unit0016:
Drafting this as I pulled the gun a little too early and
as they say
'that shit broke'

unit0016:
I just realized I forgot to add one thing they had that I didn't

brain dumb

Seris02:
Why does the security medic icon have red in it

unit0016:
> Why does the security medic icon have red in it

It's actually unchanged from the file there I've replaced in doing this. No, I have no idea why it's like that.

Seris02:
I'm gonna come right out and say, ahhh
these icons need to be changed
blue sec icons
really fuckin bad idea, they're meant to be a different color for a reason, I don't want to start mistaking regular officer for HoS
and also the security medical icon should be changed to a blue cross if we're actually going to be sticking with it, and the rep colour should be a nanotrasen colour + a different icon.

also, vanguard operative icons are needed too.

Seris02:
All of which is provided in my PR, which I can just change to be modular.

unit0016:
> All of which is provided in my PR, which I can just change to be modular.

Security is blue. The IDs are blue and grey, (The HUDs were too for a bit but that was changed for visibility and eventually just outright removed) The department is blue and grey, etcetera. ~~Also the cross has twinges of blue so this is /probably/ Intentional but Gandalf's on break so there's no point in asking. ~~
**While this is absolutely a problem with mindshields in play**; _(look at that 10ish pixel difference!)_ It's the design decisions that.. exist. ~~Whether they're stupid is another matter, but there's a reason I have my corner and stick to it when it comes to playing the game.~~
image

Seris02:
Yeah I can't fucking tell the difference, I have no clue which one is HoS and which is a sec officer
I'm being completely serious

PR #7088, opened 2021-07-22
Fix integration tests - wounds

Funce:
## About The Pull Request
So baseline wounding brings an element of randomness to the gun_go_bang unit test. The runtimes are probably shrouding the damage being incorrect due to wounds as well. So we'll just... not have them for the test. We can make a different wound test a bit later.

## Why It's Good For The Game
Yeah integration tests get to stop doing themselves

## Changelog
No player facing.


Funce:
Oh shit yeah self merging

PR #6071, opened 2021-06-01
[WIP] [DNM Until FF Is Over] The Feature MegaBundle PR

unit0016:
## About The Pull Request
Adds a bunch of new features - this is my feature freeze megaproject, opening the PR super duper early for feedback and suggestions.
 Or.. projects?
First Feature: Antigravity Boots!
Anti-Magboots, essentially. When worn, you lose your gravitational pull. See here, a (simulated) Nuclear operative after pilfering a pair from cargo (and turning on his oxygen harness.)
image
Second Feature: Buildable Lockable Lockers!
Airlock Electronics can be applied to lockers to add locks. Screwdrivers can remove these on unlocked lockers.
image
Third Feature: The PERSONAL SELF-DEFENSE DISABLER (For Traitors lmfao)
Ports the PSD from that one /tg/station PR from a long time ago. It's a disabler, able to take someone out with 3.3 shots (and two extra spare). It's a traitor item. It can also, like any energy weapon, be shoved into a emitter. You may exploit this knowledge. It's more fun that way.
image
Fourth Feature: Service Airlocks!
Adds dedicated Service airlocks. That's it.
image
Fifth Feature: Ocean Trench Mining Z-level.
An ocean trench mining z-level. And fauna / ruins therein.
Sixth Feature: Interdyne QOL Improvements
A bunch of QOL improvements for Interdyne. Weh.

## Why It's Good For The Game
😳 

## Changelog
:cl:
add: Added new things
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

PR #7133, opened 2021-07-26
[Modular] Anti-Gravity Boots, but real.

unit0016:
## About The Pull Request
Updates and fixes the first feature of #6071, which adds anti-gravity boots for ~~cargo to wear and use and abuse.~~ The QM, as his second special special boy item - now having two, like the CE. /tg/ style traitor objectives haven't been made for these, so they won't count as such if those ever get reimplemented for whatever reason.
Weh.

## Why It's Good For The Game
Helps speed up the manual labor of cargo either through pushing off objects and walls or using a jetpack / thruster implant - the latter is VERY fun In my opinion. Of course it's a high value item and VERY conspicuous, so only the bravest of souls will use it illicitly to it's full potential. Or the QM will just beg the HOP for EVA. I forsee the latter, honestly.

## Changelog
:cl:
add: The QM now has a pair of anti-gravity boots in their locker. Speed out, man!
/:cl:


modular_skyrat/master_files/code/modules/clothing/shoes/antigravboots.dm
```diff
@@ -0,0 +1,51 @@
...
+	desc = "Anti-gravity boots, for those who want to live weightlessly. Control Click to toggle anti-gravity functions."
+	name = "anti-gravity boots"
+	icon_state = "walkboots" //Haha funny reused sprite
+	var/enabled_antigravity = FALSE
+
+/obj/item/clothing/shoes/antigrav_boots/equipped(mob/user, slot)
+	. = ..()
+	if(slot == ITEM_SLOT_FEET)
+		if(enabled_antigravity)
+			user.AddElement(/datum/element/forced_gravity, 0)
```
tf-4:
```suggestion
	if(!slot == ITEM_SLOT_FEET)
		return
	if(enabled_antigravity)
		user.AddElement(/datum/element/forced_gravity, 0)
```

Gandalf2k15 (approved):

tf-4:
Would be better as an antag item imo.

unit0016:
> Would be better as an antag item imo.

It would basically necessitate also buying a hardsuit or breaking into EVA if the boots were traitor items, which would completely balls up the experience for security because near every traitor would be super duper fast, meaning they'd have to set up barricades in hallways and just etcetera etcetera. **I am open to making them high-value items in the QM's locker (just the one pair), like the CE's magboots and blueprints are, however.** (Plus anti-gravity is already obtainable if miners hand off geyser chems to chemistry (but they never do.))

tf-4:
Won't making them printable have the exact same issues?

I think making them a unique QM item is also a good solution

modular_skyrat/master_files/code/modules/clothing/shoes/antigravboots.dm
```diff
@@ -0,0 +1,32 @@
...
+	if(!slot == ITEM_SLOT_FEET)
+		return
+	if(enabled_antigravity)
+		user.AddElement(/datum/element/forced_gravity, 0)
+
+/obj/item/clothing/shoes/antigrav_boots/dropped(mob/user)
+	. = ..()
+	user.RemoveElement(/datum/element/forced_gravity, 0)
+
+/obj/item/clothing/shoes/antigrav_boots/CtrlClick(mob/living/user)
```
tf-4, line 21: could move this to right click?

unit0016: They already have to remove the boots to change it, might as well make it clankier.

PR #7148, opened 2021-07-26
[NONMODULAR] Ninja rebalance, the beginning

RatFromTheJungle:
## About The Pull Request
Nerfing, and somewhat rebalancing ninja to make it viable to midround once more.

## Why It's Good For The Game

Ninja was generally seen as being 'to strong' for a consistent midround, the goal is to bring it back in line with other antags of it's tier. (Nightmare, really really early heretics)

## Changelog
:cl:
balance: 1.5x'd the ninja teleport CD
balance: removed most of the shoes slowdown reduction.
balance: Added a tiny amount of speed up for wearing the suit, so they aren't absolutely fucked in a chase.
removed: The communications hacking objective, things can quickly snowball out of control from a almost-calm round to round-ending chaos within five minutes.
/:cl:


code/modules/ninja/suit/shoes.dm
```diff
@@ -21,4 +21,4 @@
 	min_cold_protection_temperature = SHOES_MIN_TEMP_PROTECT
 	heat_protection = FEET
 	max_heat_protection_temperature = SHOES_MAX_TEMP_PROTECT
-	slowdown = -1
+	slowdown = -0.3 // SKYRAT EDIT - Original: -1
```
tf-4, line 24: won't this just make them faster overall if the suit is set to 0?

RatFromTheJungle: yes, due to their teleport being much longer, and our quite heavy damage slowdown, a single disabler or a lovetap from the weakest pistol would leave them absolutely screwed without a mild kick up in speed. (-0.3 leaves them moving at 'normal' speed up unto about 25 damage, or equal(?) stam.)

Remember for the most part that ninjas can't traditionally heal outside of patches, without removing the entire suit, so their in-combat survivability is all they really have.

tf-4: good enough for me

JohnFulpWillard:
This removes the cool option ninjas have between normal running speed with the ninja suit and being able to run super fast at the cost of not having to wear a suit. Just as a suggestion, maybe the ninja shoes could have -0.5 and the suit could be 0.2, so its still there, but not as good.

tf-4:
> 
> 
> This removes the cool option ninjas have between normal running speed with the ninja suit and being able to run super fast at the cost of not having to wear a suit. Just as a suggestion, maybe the ninja shoes could have -0.5 and the suit could be 0.2, so its still there, but not as good.

Nobody does this. Wearing the suit is infinitely better in basically all scenarios

PR #7160, opened 2021-07-27
[NONMODULAR] Nerfs Bio-reconstruction & Dermal Hardening.

RatFromTheJungle:
## About The Pull Request
(This isn't related to any round in specific, I just forgot to do this when I was going to.)
Pulls back on the raw healing power of nanites, as-well-as the general armor they can provide.
## Why It's Good For The Game
Nanites have been on the top-tier of power for far to long, with far to little risk, and far, far, far to much availability, this is probably just step one.

## Changelog
:cl:
balance: Cuts biorecons healing power in half.
balance: Cuts the armor granted by dermal hardening by about half.
/:cl:

### DIFF
--- a/code/modules/research/nanites/nanite_programs/buffing.dm
+++ b/code/modules/research/nanites/nanite_programs/buffing.dm
@@ -30,15 +30,15 @@
 	. = ..()
 	if(ishuman(host_mob))
 		var/mob/living/carbon/human/H = host_mob
-		H.physiology.armor.melee += 25
-		H.physiology.armor.bullet += 20
+		H.physiology.armor.melee += 15 // SKYRAT EDIT - ORIGINAL: 25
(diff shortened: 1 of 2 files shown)


softcerv:
This PR is a really good step going forward. Although, since the healing is being nerfed by half, you might want to lower the use cost of Bio-reconstruction by a little bit.

RatFromTheJungle:
> 
> 
> This PR is a really good step going forward. Although, since the healing is being nerfed by half, you might want to lower the use cost of Bio-reconstruction by a little bit.

I thought about it at the time, but I also don't want it to end up as accelerated regenv2, that's just always semi-active because of how cheap it is. (it can still be stacked ontop of accelerated regen, to be almost equal to it's power before 1.5 vs 2)

PR #7170, opened 2021-07-27
[NONMODULAR][TM-ONLY] Dynamic tweaking.

Vizoplez:
## About The Pull Request
Title

## Why It's Good For The Game
Some of these were extremely stupid, like midround heretic being over 3x cheaper than roundstart, along with having a higher roll chance.

Discourages midround fishing while encouraging round-start rolling, to help prevent insane threat buildup.

## Changelog
:cl:
tweak: major dynamic adjustments
/:cl:


FlamingLily:
IIRC dynamic is handled via config, so this might not actually do anything

Mothblocks:
You could ask for the history behind the "extremely stupid" numbers from the person who spent a good month tweaking them if you want 🙂 

Midround heretics are cheaper because time and time again, they're shown to be significantly weaker than roundstart heretics. Heretics build their power over time, and so when there's less time for them to start, they're way weaker. Compare to latejoin traitors which have just as much TC to start with as roundstart traitors.

PR #7173, opened 2021-07-28
Revert "[MIRROR] Use rust-g to url_encode (and other builtin overrides)"

Funce:
Reverts Skyrat-SS13/Skyrat-tg#7110

TM while I find out the runtime with this. Maybe a new build with rust? Who knoews


ZephyrTFA:
Yep, they updated their rustg and we didn't

ZephyrTFA:
Silly protogen

Funce:
I can't believe we have to build a new version of rust 😔

PR #7210, opened 2021-07-28
[MIRROR] Remove nanites

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/60473
--------------------

## About The Pull Request
Removes nanites. Icons will be sent to SS13-sprites.

Closes #60251 
Closes #56831
Closes #56569
Closes #54401
Closes #50842 
Closes #50409 
Closes #49924 
Closes #45728 

## Why It's Good For The Game

Nanites are a fundamentally unhealthy and flawed concept.

Getting nanites is a matter of sitting in a tube for about 8 seconds, and, in the case of a competent nanite creator, basically never having to think about it again. In exchange for sitting in a tube, you've now been given a sizable amount of passive healing abilities, revival, and whatever else they decided to cook up.

This is *not* gameplay--this is an outcome.

No part of your experience is altered in any meaningful way other than that you are now passively healing.

The catch is supposed to be that they are high-risk outside of high-reward. This is admitted in [the PR that added it](https://github.com/tgstation/tgstation/pull/37922):

> First, the new setup means that to get/update nanites you need at the very least someone else to operate the console for you, meaning that a lone roboticist still needs to put himself at some risk during injection. Another option is to go for cloud-based nanites; meaning you don't need someone else every time you want a new program for your nanites, but on the other hand anyone can upload some bad stuff while you're not watching the cloud control console.

A noble idea, but sabotaged nanites are *still* not gameplay for the hosts--it is simply an outcome. That is to say, being sabotaged isn't interesting.

It also does not happen enough to where it's worth considering at all. Don't believe me? I collected all the nanites in "suppression" (except for Skull Echo and Happiness Enhancer), "rogue", and "weapon". I then collected rounds this month where Blood Regeneration, Blood Purification, or Bio-Reconstruction were used. No real reason for those 3 other than it seems to filter out people who are just trying to learn nanites.

(Labels were added on in post, since I am still barely figuring out Julia plots--the numbers are accurate, I just logged them and added them on)

With *29* nanites being considered "bad" and *3* being considered "good", little over *20%* of rounds have *any* nanites that aren't exclusively good. This graph doesn't even consider what cloud ID one is using, one bad nanite being used taints the entire round, and it's *still* extremely infrequent. Really good *passive* healing (and revival!) for a 1 in 5 chance of sabotage (which can then just be removed later on) is a weak risk:reward.

So if the risk is not in sabotage (as expected by the original designer), is it in EMPs/shocks? No, not really.

Every time you are EMP'd/shocked, there is a *chance* for something to go wrong. Within that chance, there are 5 equally likely outcomes:

- Deleted
- Deprogrammed (it becomes unactivatable, effectively)
- Toggled
- Triggered
- Swapped with a rogue type

Of all these outcomes, only the *last one* (a 1 out of 5 chance) is likely to actually cause any real damage. The rest effectively just your passive healing *might* not work as well anymore.

...BUT EVEN THEN!

The "rogue types" are not of any possible rogue nanite. They are dependent on what actually errored.

Of the rounds on Terry, these are the glitch replacements:

| Replacement | Count |
| ----------- | ----- |
| Brain Misfire | 11  |
| Dermalysis | 109 |
| Electric Shock | 39  |
| Glitch | 196 |
| Hypoxemia | 125 |
| Necrosis | 214 |
| Nerve Decay | 10  |
| Neuro-Necrosis | 70  |
| Toxin Buildup | 164 |

Of the most common ones:

- Necrosis deals brute damage, which is healed by nanites. It deals 0.75 damage per activate, bio-reconstruction heals 1.
- Dermalysis deals *even less* brute damage--0.25.
- Toxin Buildup deals toxin damage, which is healed by nanites. It deals 0.5 damage per activate, blood purification heals 1.
- THESE ARE WHILE ACTIVE, WHICH MEANS IT MIGHT NOT EVEN PROC IN THE FIRST PLACE, CONSIDERING THAT YOU'VE HAD OTHER NANITES TOGGLED/DEPROGRAMMED.

So there is *still* hardly any risk, even without deliberate sabotage.

And to reiterate--**even in the worst case scenario, this risk is still not interesting. It is an outcome, not gameplay.** Especially when paired with constant, very adequate passive healing (and revival). This is why attempting to "balance" these is a fruitless play--you're not making the game more interesting by making nanites more deadly. 

This means that, in practice, nanites are a way to gain passive healing for extremely little downside, with extremely little care from any of the hosts.

## Changelog
:cl:
del: Nanites have been removed.
/:cl:


hugmeorelse:
I'm locking this pre-emptively. It's a PR from upstream. Discussing it here will not be fruitful unfortunately.

TG seems to be introducing new mechanics instead such as https://github.com/tgstation/tgstation/pull/60338

PR #7276, opened 2021-07-31
HoP Class President outfit

axietheaxolotl:
## About The Pull Request

Gives the HoP a Class President outfit, at the request of me. I wanted this.

## Why It's Good For The Game

Add's more drip, which is always appreciated in my opinion. Also if a school event ever gets added, it'll be helpful then.

## Changelog
:cl:
add: Added Class President outfit to loadout, which is restricted to the HoP.
/:cl:


axietheaxolotl:
I made this pull request when I was on the verge of passing out and didn't properly think it out. Having what amounts to a school outfit with explicitly sexual descriptions, is kinda creepy, I realise that now. I'm gonna keep this open, I'd still like for it to be added as I think it's kinda funny, but I won't be losing any sleep if it doesn't get added.

Gandalf2k15:
I think its best if we try to avoid any situations in which this could raise an eyebrow.

PR #7279, opened 2021-08-01
DecaySS Applies Rust Correctly

ZephyrTFA:
I should've done this a while ago but I forgot, also this is a webedit so it might break horribly


Gandalf2k15:
Was done in a recent mirror

PR #7310, opened 2021-08-02
[MIRROR] Kills misc clothing files (Almost)

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/60481
--------------------

## About The Pull Request

Reorganising the clothing code files to remove all variants of misc from all but code/modules/clothing/under/. Makes an incidental change to make breathmuzzles a subtype of normal muzzles, to clean up some duplicate code.

## Why It's Good For The Game

See #60358 

code/modules/clothing/under/miscellaneous.dm has been intentionally left untouched. Code files in under/ are categorised by typepath, as there is a misc/ type which apart from a couple of special cases is used for clothes which get their sprites from misc.dmi.

A proper solution to this particular file would involve repathing the misc/ clothes, moving their sprites appropriately and putting the code in the corresponding file.

This is out of scope for a PR which is otherwise 99% moving existing core between files unchanged, and this particular file can be tackled separately.

## Changelog
:cl:
code:  Removes (almost) all misc files from the clothing module
/:cl:


Gandalf2k15:
REEE ANOTHER FUGGING PR FOR THE SPECIES DATUMING

PR #7349, opened 2021-08-03
Japanese is a language for Earthlings

YakumoChen:
## About The Pull Request

narrows down Japanese to be available to people who have cultures more likely to be from Earth. Earth is seemingly described as being "old cultures" and Japanese is described as a culturally old language. Whereas where Japanese is applied to now is mostly spacers who aren't anywhere near Earth to begin with.
Edit: On request Japanese was left in the generic default option, so you don't have to go out of your way to find it

## Why It's Good For The Game

it makes sense given the current selectable lore
Also people from Earth should be able to know Japanese
why do vat-grown people get to know an ancient language anyway

## Changelog
:cl:
qol: Japanese is for Earthlings
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/culture/culture/culture_generic.dm
+++ b/modular_skyrat/modules/customization/modules/culture/culture/culture_generic.dm
@@ -1,12 +1,12 @@
 /datum/cultural_info/culture/generic
 	name = "Other Culture"
 	description = "You are from one of the many small, relatively unknown cultures scattered across the galaxy."
-	additional_langs = list(/datum/language/spacer, /datum/language/japanese)
+	additional_langs = /datum/language/spacer
 
 /datum/cultural_info/culture/vatgrown
 	name = "Vat Grown"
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/customization/modules/culture/culture/culture_generic.dm
```diff
@@ -6,7 +6,7 @@
 /datum/cultural_info/culture/vatgrown
 	name = "Vat Grown"
 	description = "You were not born like most of the people, instead grown and raised in laboratory conditions, either as clone, gene-adapt or some experiment. Your outlook diverges from baseline humanity accordingly."
-	additional_langs = list(/datum/language/spacer, /datum/language/japanese)
+	additional_langs = /datum/language/spacer
```
ZephyrTFA: Needs to stay as a list, even if it's only one item

ZephyrTFA (changes requested):

PR #7357, opened 2021-08-03
[NON-MODULAR] Runtime Begone

ZephyrTFA:
Something somewhere is adding null or qdel'd instances to an area's content list; because of the nature of how this code works I am unable to investigate this in any reasonably sane method, so instead we simply verify the target before sending a signal.

image


Gandalf2k15:
we should really be fixing the bad signal senders tbh

FlamingLily:
Is this something happening upstream? Might be good to see if they're aware of it if so

LemonInTheDark:
The list was a source of harddeletes, kyler fixed it recently. Should be all good now

ZephyrTFA:
> The list was a source of harddeletes, kyler fixed it recently. Should be all good now

Closing due to this; should ideally be fixed

PR #7381, opened 2021-08-04
[NON-MODULAR] TTS Devices are now purchaseable from NanoMed.

softcerv:
## About The Pull Request
Adds TTS devices to the medical vendor. 

## Why It's Good For The Game
This gives doctors an option to let patients, that have either lost their TTS device or have gained an impairment that makes speech incomprehensible/impossible, obtain a new TTS device. Currently TTS devices are only obtainable through the loadout menu.  

## Changelog
:cl:
qol: TTS devices are now purchasable by the medical vendor. 
/:cl:

### DIFF
--- a/code/modules/vending/medical.dm
+++ b/code/modules/vending/medical.dm
@@ -17,6 +17,7 @@
 					/obj/item/stack/medical/bone_gel/four = 4,
 					/obj/item/taperoll/med = 2,	//SKYRAT EDIT - Job tape!
 					/obj/item/taperoll/bio = 2)	//SKYRAT EDIT - Job tape!
+					/obj/item/ttsdevice = 3 //SKYRAT EDIT TTS DEVICES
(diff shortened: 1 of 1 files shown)


softcerv:
whoops.

SpaceVampire:
Given that it costs seven points in the load-out, I almost feel like it should be under premium if you are gonna put it there.
As it already almost completely negates the whole "being mute" thing with basically no downside.
Which I already don't think is great, but uh, that is a side-note on its own.
Would be nice to have another thing that could actually be worth spending credits on as well.

softcerv:
> ts in the load-out, I almost feel like it should be under premium if you are gonna put it there.
> As it already almost completely negates the whole "being mute" thing with basically no downside.
> Which I already don't think is great, bu

Sounds reasonable, I'll make that change.

PR #7393, opened 2021-08-05
[MODULAR] Augment Fix

ZephyrTFA:
## About The Pull Request

A recent mirror removed the character_setup var and blindly set all instances of it to TRUE for some reason.
As a result this broke augments horribly; this fixes it.

## Why It's Good For The Game

People pay QP for the augments; maybe we should actually give it to them.

## Changelog
:cl:
fix: Augments now correctly apply on characters
/:cl:


Funce (approved):
Good find.

PR #7410, opened 2021-08-05
[MIRROR] Job refactor 2: less hardcoded lists

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/60578
--------------------
* Creates job department datum singletons, referenced to by the job subsystem.
* Removes several hardcoded job lists. Source of truth for their contents is now in the job and department singletons. Replaces them from dynamically created lists on the job subsystem.
* Documents code a little bit more and improves the name of some variables.
* Assistants and prisoners are no longer included in the service department. They were considered so in some parts of the code and not in others, so I opted for making them simply have no department assigned.
* Fixes obsessed not triggering, because it was assuming `assigned_role` to be a string instead of a job datum reference, as it was before the first job refactor: https://github.com/tgstation/tgstation/commit/a4a5d7744c6024c59a8f37fe93cc61bbe1187aec#diff-a231d19c6da341174bdc59fbd2d3e22c557dde16d03ce2b360cdfc3ba7ef9d76R877


FlamingLily:
this touches stuff I've messed with for alt titles I'll fix it soon

Gandalf2k15:
> this touches stuff I've messed with for alt titles I'll fix it soon

Do it in another PR, Datacore had to be removed.

unit0016:
REAL PUBBYSTATION BREAKS AGAIN

modular_skyrat/master_files/code/modules/jobs/job_types/nanotrasen_admiral.dm
```diff
@@ -0,0 +1,83 @@
+/datum/job/admiral
```
FlamingLily, line 1: based as hell but uh


huh?

PR #7428, opened 2021-08-06
[MODULAR] actually fixes SOOC, probably

Seris02:
## About The Pull Request

Might wanna testmerge this instead, so in case it doesn't work on the live server, again, I don't have to keep making PRs, and can just push to this one.
Again like the last one, tested it locally, but don't know if it'll actually work on the live version.

## Why It's Good For The Game

bug fix

## Changelog
:cl:
fix: fixes SOOC
/:cl:


Azarak:
At this point get rid of the assoc table and make a "can access sooc" variable on a job datum level, then get the job singleton and check

PR #7429, opened 2021-08-06
[FUCK][TM-ONLY] Ashwalkers no longer break save slots. Rejoice.

Vizoplez:
TM-Only because the issue is far more deep-rooted than this


Gandalf2k15:
Don't need this, got a fix already

PR #7446, opened 2021-08-07
gives the sec sarge a sidearm token, finally.

capsaicinz:
## About The Pull Request

see title. long time in the making of fixing making. don't know why it was never added back.

## Why It's Good For The Game

muh bullets.

## Changelog
:cl:
add: sec sarge now gets a sidearm token.
/:cl:


ZephyrTFA:
While this is technically a feature, I think it should probably be allowed past the freeze.

capsaicinz:
> While this is technically a feature, I think it should probably be allowed past the freeze.

i said it in my last update with my disablers- that this may be considered a feature. but it's so small and suggested that it should've been there ever since the armory minidate, that i do think it'd be better off being pushed through. it's making rolling sarge a lot more annoying.

PR #6523, opened 2021-06-25
[MIRROR] Adds Defines for Species IDs

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/59797
--------------------

## About The Pull Request
Simply puts species IDs into defines ("lizard" turns into `#define SPECIES_LIZARD "lizard"`). This does not affect any gameplay, and is purely for making sure future code doesn't go all fucked up because someone accidentally misspells something

I only changed code that was spelt correctly, so if there actually _is_ a typo somewhere, it isn't changed

## Why It's Good For The Game
It's better for developers to work with if they ever decide modifying a specific thing for species in the future

## Changelog
:cl:
code: changed species IDs into defines
/:cl:

PR #6543, opened 2021-06-26
[Modular] Adds species ID defines

DragonTrance:
## About The Pull Request
Finishes #6523 so it also defines species IDs for this code

**DO NOT MERGE THIS BEFORE #6523 GETS MERGED,** there will be compile errors!

## Changelog
:cl:
code: put species IDs into defines
/:cl:

PR #7466, opened 2021-08-08
species id defines

jjpark-kb:
## About The Pull Request
check out https://github.com/Skyrat-SS13/Skyrat-tg/pull/6523 and https://github.com/Skyrat-SS13/Skyrat-tg/pull/6543
closes: #6523 

Simply puts species IDs into defines ("lizard" turns into #define SPECIES_LIZARD "lizard"). This does not affect any gameplay, and is purely for making sure future code doesn't go all fucked up because someone accidentally misspells something

I only changed code that was spelt correctly, so if there actually is a typo somewhere, it isn't changed

## Why It's Good For The Game

It's better for developers to work with if they ever decide modifying a specific thing for species in the future

## Changelog
:cl:
code: changed species IDs into defines
/:cl:


ZephyrTFA:
Actually based, good work

PR #7493, opened 2021-08-10
No conversion cult

Gandalf2k15:
## About The Pull Request

Just a change to cult that needs to be tested. 

The gist of it is that cult conversion is optional and no longer mandatory.

Cult no longer gets the summon narsie objective, and when they complete a sacrifice, it gives them a new target until there are none left.

Needs testing.

# Let me clarify
Upon attempting to convert someone to the cult, they will be brainwashed and put to sleep. Giving you some time to put them somewhere away from your base. T**his means you SHOULD not be killing them if they say no.** - Unless of course they are attacking you.

## Changelog
:cl:
balance: Cult no longer force converts and the narsie objective has been thanos snapped.
/:cl:


ForrestWick:
So what happens if you say no to conversion?

FlamingLily:
> 
> 
> So what happens if you say no to conversion?

The cult says "oh okay" and lets them go /j

https://github.com/Skyrat-SS13/Skyrat-tg/pull/7493/files#diff-03b369e5e6c74160bc7eb992d2b7fc282e6f07d67d6669ef6d5e7f25458dcc9fR263
Looks like the victim takes some blindness and jitter and a lil' brute, before presumably being ritually executed by the cult.

Which is honestly kinda based.

code/modules/antagonists/cult/runes.dm
```diff
@@ -302,7 +322,22 @@ structure_check() searches for nearby cultist structures required for the invoca
...
-				big_sac = TRUE
+				//big_sac = TRUE
+				//SKYRAT EDIT ADDITION
+				var/datum/objective/sacrifice/sacrifice_objective = new
+				sacrifice_objective.team = C.cult_team
+				sacrifice_objective.find_target()
+				C.cult_team.objectives -= sac_objective
+				qdel(sac_objective)
+				C.cult_team.objectives += sacrifice_objective
+				for(var/datum/mind/B in C.cult_team.members)
```
tf-4: B

ForrestWick: B

code/modules/antagonists/cult/cult.dm
```diff
@@ -431,6 +433,6 @@
 			return FALSE
 	else
 		return FALSE
-	if(HAS_TRAIT(M, TRAIT_MINDSHIELD) || issilicon(M) || isbot(M) || isdrone(M) || !M.client)
```
SpookyTheFox, line 434: This one has been definitively identified by multiple people as causing unintended and frustrating situations.
I'd even go so far as to say other antagonist types should be off limits as well, as we also had a heretic cultist.

PR #7494, opened 2021-08-10
[MODULAR][FIX/QoL] Mediguns eject Medicells into hands before the floor.

softcerv:
## About The Pull Request
Changes the proc used to remove the medicells so that they dispense into the users hands first. It was originally intended to work like this, I just didn't know this proc existed. 

## Why It's Good For The Game
Makes Mediguns less cumbersome to use.

## Changelog
:cl:
fix: Mediguns now eject Medicells into the user's hands instead of always on the floor.
/:cl:


Gandalf2k15:
Do both of these things

Gandalf2k15:
Move it to drop location then attempt to put in hanes

Gandalf2k15:
Else it will null it

softcerv:
> 
> 
> Else it will null it

OH, lemme do that real quick then!

PR #7367, opened 2021-08-03
FEATURE FREEZE (SEE ANNOUNCEMENTS IN DISCORD)

Gandalf2k15:
# Only bug fixes/cosmetics past this point. We are in sync with TG.

PS.
We will be going on another PR Freeze. Cobalt and Useroth have recently come back from their vacation. However, there are a few technical issues that has arisen in their absence that needs to be addressed and fixed, hence the PR freeze.

All existing bounties will be respected and still merged if done during the PR freeze. Any PRs that are currently up on the repository will also still be reviewed and merged during the PR freeze. Only PRs related to fixing bugs, map changes, cosmetics, and so forth will be merged during the PR freeze, however, any PRs approved by myself can bypass the freeze. This may include anything like a balance change, however, this will be done sparsely to avoid adding to the current overhead.

We aim to have this PR freeze be shorter than the last, but we'll have to see how this goes.

P.S. Again, if you just missed the PR freeze, you are free to DM me. I may approve PRs to get through the PR freeze if they are of sound quality in terms of code and is :goodjobderp:, but this will be done scarcely as stated before.

PR #7500, opened 2021-08-11
Cyborg tablet software system

twilightwanderer:
## About The Pull Request

Adds a system for installing programs to cyborgs depending on their species.

## Why It's Good For The Game

It so happens that in the architecture of the tgstation build, every cyborg owns a tablet. In this form, these tablets are useless, because they are designed only for one program - **robotact**. At the same time, cyborgs have access to install and run any program on any station computer. And instead of using their tablet and station network, the cyborgs have to drive to the computers every time, which seems a little illogical.

I propose to use a system in which each cyborg is recorded a specific set of programs for remote use depending on its specialization.
For cyborgs at the moment I have made an installation of the following programs:

image

**Enginer** - _power_monitor, alarm_monitor, supermatter_monitor_;
**Medical** - _lifeline_;
**Security** - _secureye_;
**Service** - _chatclient_;
**Peacekeeper** - _lifeline, secureye_;
**Clown** - _chatclient_. (HONK in chat)

For the programs **secureye** and **power_monitor** I have created their variants for cyborgs, which are not available for installation on regular tablets and need a network. The regular versions of secureye and power_monitor are not available for installation on tablets.
This set of programs can easily be changed or added to in the future. I plan to do this if my changes are included in the main repository, porting variants of the conventional console interfaces for use in cyborg tablets. 

## Changelog
:cl:
add: Added system for loading programs to cyborgs when selecting a specific model
add: Added two special programs for cyborgs: power_monitor/integrated and secureye/integrated
code: The following files have been changed: robot.dm, tablet_presets.dm, powermonitor.dm, secureye.dm.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -83,8 +83,31 @@
 	addtimer(CALLBACK(src, .proc/show_playstyle), 5)
 
 /mob/living/silicon/robot/proc/create_modularInterface()
-	if(!modularInterface)
+	// TG ORIGINAL BEGIN
+	//if(!modularInterface)
+	//	modularInterface = new /obj/item/modular_computer/tablet/integrated(src)
+	// TG ORIGINAL END
+	
+	//SKYRAT EDIT ADDITION BEGIN
+	//Creating a certain cyborg tablet depending on his species
+	if (model.name == "Default")
 		modularInterface = new /obj/item/modular_computer/tablet/integrated(src)
+	else if (model.name == "Engineering")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/enginer(src)
+	else if (model.name == "Security")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/security(src)
+	else if (model.name == "Medical")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/medical(src)
+	else if (model.name == "Service")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/service(src)
+	else if (model.name == "Peacekeeper")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/peacekeeper(src)
+	else if (model.name == "Clown")
+		modularInterface = new /obj/item/modular_computer/tablet/integrated/clown(src)
+	else
+		modularInterface = new /obj/item/modular_computer/tablet/integrated(src)
+	// SKYRAT EDIT ADDITION END
+	
 	modularInterface.layer = ABOVE_HUD_PLANE
 	modularInterface.plane = ABOVE_HUD_PLANE
 
@@ -166,8 +189,9 @@
 	var/input_model = show_radial_menu(src, src, model_icons, radius = 42)
 	if(!input_model || model.type != /obj/item/robot_model)
 		return
-
+	
 	model.transform_to(model_list[input_model])
+	create_modularInterface() // SKYRAT EDIT ADD
 
 
 /// Used to setup the a basic and (somewhat) unique name for the robot.
@@ -727,7 +751,8 @@
 
 	ionpulse = FALSE
 	revert_shell()
-
+	create_modularInterface() // SKYRAT EDIT ADD
+	
 	return TRUE
 
 /mob/living/silicon/robot/model/syndicate/ResetModel()
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -83,33 +83,32 @@
 	addtimer(CALLBACK(src, .proc/show_playstyle), 5)
 
 /mob/living/silicon/robot/proc/create_modularInterface()
-	// TG ORIGINAL BEGIN
-	//if(!modularInterface)
-	//	modularInterface = new /obj/item/modular_computer/tablet/integrated(src)
-	// TG ORIGINAL END
-	
-	//SKYRAT EDIT ADDITION BEGIN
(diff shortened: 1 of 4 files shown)


FlamingLily:
Absolutely awesome PR, unfortunately https://github.com/Skyrat-SS13/Skyrat-tg/pull/7367

code/modules/modular_computers/computers/item/tablet_presets.dm
```diff
@@ -110,3 +110,39 @@
...
+	hard_drive.store_file(new /datum/computer_file/program/radar/lifeline)
+
+/obj/item/modular_computer/tablet/integrated/security/Initialize()
+	. = ..()
+	var/obj/item/computer_hardware/hard_drive/small/hard_drive = find_hardware_by_name("solid state drive")
+	hard_drive.store_file(new /datum/computer_file/program/secureye/integrated)
+	
+/obj/item/modular_computer/tablet/integrated/service/Initialize()
+	. = ..()
+	var/obj/item/computer_hardware/hard_drive/small/hard_drive = find_hardware_by_name("solid state drive"
```
ZephyrTFA:
```suggestion
	var/obj/item/computer_hardware/hard_drive/small/hard_drive = find_hardware_by_name("solid state drive")
```

ZephyrTFA:
Instead of giving them the applications from the installing of their module; give them access to the program downloader. Its much simpler and will be 100% modular

twilightwanderer:
> Instead of giving them the applications from the installing of their module; give them access to the program downloader. Its much simpler and will be 100% modular

Thank you for your response. That's what I thought too at first. It would have been the easiest option. But it looks like the program downloader is unavailable for a reason. The fact is that some programs simply can not be installed on the tablet, other programs are simply unnecessary, and the rest have a modifier _available_on_ntnet = FALSE_.

Vizoplez:
> > Instead of giving them the applications from the installing of their module; give them access to the program downloader. Its much simpler and will be 100% modular
> 
> Thank you for your response. That's what I thought too at first. It would have been the easiest option. But it looks like the program downloader is unavailable for a reason. The fact is that some programs simply can not be installed on the tablet, other programs are simply unnecessary, and the rest have a modifier _available_on_ntnet = FALSE_.

This is already possible, this PR isn't needed!

Cyborgs can have datadisks inserted into them with an open panel, which can contain program data. Furthermore, their tablets themselves are actually incompatible with most of the programs listed here, the mainly functional ones are.

NT CIMS (Crystal monitoring)
SECUREYE
Chat Clients
Outbomb Cuban Pete (don't ask)

while the rest are deprecated.

Stalkeros2:
No.

FlamingLily:
> > Instead of giving them the applications from the installing of their module; give them access to the program downloader. Its much simpler and will be 100% modular
> 
> Thank you for your response. That's what I thought too at first. It would have been the easiest option. But it looks like the program downloader is unavailable for a reason. The fact is that some programs simply can not be installed on the tablet, other programs are simply unnecessary, and the rest have a modifier _available_on_ntnet = FALSE_.

Another problem with doing that is that borgs have all access, so it would make it very easy to get access locked programs, I much prefer this method

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -83,8 +83,31 @@
...
 /mob/living/silicon/robot/proc/create_modularInterface()
-	if(!modularInterface)
+	// TG ORIGINAL BEGIN
+	//if(!modularInterface)
+	//	modularInterface = new /obj/item/modular_computer/tablet/integrated(src)
+	// TG ORIGINAL END
+	
+	//SKYRAT EDIT ADDITION BEGIN
+	//Creating a certain cyborg tablet depending on his species
+	if (model.name == "Default")
```
FlamingLily, line 95: This is causing runtimes, model is a null at this point, which makes me theorise that this is called before setting up the model for the borg

FlamingLily:
Infact, thinking about it, this is not a great way to do it. You're effectively changing the internal tablet item whenever there's a model change, which is not a great way to do it. They way I'd do it is

When a cyborg chooses a model, it installs the programs of that model. When it gets a reset, all programs (apart from lawtact or robotact or whatev) get deletes

code/modules/mob/living/silicon/robot/robot.dm
twilightwanderer: Okay. I will do a separate procedure for this purpose.

twilightwanderer:
> Infact, thinking about it, this is not a great way to do it. You're effectively changing the internal tablet item whenever there's a model change, which is not a great way to do it. They way I'd do it is
> 
> When a cyborg chooses a model, it installs the programs of that model. When it gets a reset, all programs (apart from lawtact or robotact or whatev) get deletes

I will think about how this can be implemented and I will try to present an implementation according to your advice in the near future

twilightwanderer:
> 
> 
> Infact, thinking about it, this is not a great way to do it. You're effectively changing the internal tablet item whenever there's a model change, which is not a great way to do it. They way I'd do it is
> 
> When a cyborg chooses a model, it installs the programs of that model. When it gets a reset, all programs (apart from lawtact or robotact or whatev) get deletes

I redid it as you recommended.

code/modules/mob/living/silicon/robot/robot.dm
twilightwanderer: I restored the standard tablet creation and added a procedure for installing programs

jjpark-kb:
due to being stale and the feature freeze, im going to close this for now. Please reopen this when the feature freeze ends!

PR #7502, opened 2021-08-11
fix floor cluwne being so rude

jjpark-kb:
## About The Pull Request
there are now two kinds of clown outfits: the normal one that changes you, and the one that doesnt.
the floor cluwne will now give you the one that doesnt change you (too much)
additionally, some text as a hint on how to leave the maze

## Changelog
:cl:
fix: the floor cluwne won't change your name now
/:cl:


jjpark-kb:
was given permission to self-merge due to being a fix

PR #7515, opened 2021-08-12
[MODULAR] Increases cost of 'Believer' bundle.

RatFromTheJungle:
## About The Pull Request
CQC+'Esword'(lace remover)+Storage implant+Enough to half a kit, for **25** tc is far-far to cheap.

## Why It's Good For The Game
Overpowered and underpriced bundles are terrible for balance.

## Changelog
:cl:
balance: heavily increases the cost of the believer bundle (25 to 35)
/:cl:


Useroth:
I think I agree with the price increase. It's a gimmick oriented, albeit very featureful and stealthy bundle.

tf-4:
Changelog should be "nerfs the cost". A price increase is not a buff

Or just "increases the cost" for minimal ambiguity

PR #7529, opened 2021-08-12
[NON-MODULAR] Gives Cap and HoP seclites

axietheaxolotl:
## About The Pull Request

See title.

## Why It's Good For The Game

It's a flashlight, that's why it's good. Put it on your gun (so you can abuse it never running out of battery), put it in your pocket. Use it to light up maint tunnels, use it to actually see when the power goes out because the engie's inevitably screwed up the supermatter. DON'T use it to validhunt, but like, at the same time, it's a fucking flashlight, how are you going to validhunt with it.

It's a fucking flashlight, don't melodrama about how giving the HoP something as basic as that is going to make them Hopcurity. Don't nothingburger, I beg of you.

Also, you can literally already just print a seclite if you want. This is just saving you the hassle of not having to walk three minutes to do it. It really changes nothing besides save time.

## Changelog
:cl:
add: Seclite to HoP and Cap lockers.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/security.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/security.dm
@@ -31,6 +31,7 @@
 	new /obj/item/clothing/gloves/color/captain(src)
 	new /obj/item/storage/belt/sabre(src)
 	new /obj/item/storage/box/gunset/pdh_captain(src) //SKYRAT EDIT CHANGE - SEC_HAUL
+	new /obj/item/flashlight/seclite(src) //SKYRAT EDIT
 	new /obj/item/door_remote/captain(src)
 	new /obj/item/storage/photo_album/captain(src)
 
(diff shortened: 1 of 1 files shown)


desertmote:
pointless powercreep in my opinion

Aether217:
> pointless powercreep in my opinion

"Power creep, otherwise known as powercreep is a process that sometimes occurs in games where new content (in this case cards) slowly outstrip the power of previous alternatives."

SpaceVampire:
Running around in maintenance isn't really something either of those jobs are supposed to be doing, so it doesn't really make any sense that they should start with one in their locker. I feel like you could use the exact same line of logic to put that pretty much any place. I really don't see this as anything other then bloat, added to lockers that already have too much stuff in them.
If you want a flashlight, just print or request one. There is no reason you need to start with it.

ZephyrTFA:
Seclites exist to help security see in places like maintenence. The captain and HoP shouldn't be exploring maintenence

GoldenAlpharex:
This is a feature during the feature freeze.

axietheaxolotl:
> 
> 
> This is a feature during the feature freeze.

Then don't merge it? It's pretty simple, merge/close it when the feature freeze is over. But, okay.

PR #7530, opened 2021-08-12
[SEMI-MODULAR] Gives CMO, RD, and CE mini energy guns

axietheaxolotl:
## About The Pull Request

Adds a mini e-gun to the lockers of the CMO, RD and CE, as well, removes them from the kilo map file so they aren't spawning with two energy guns round-start.

## Why It's Good For The Game

The telebaton is good for self-defence, but it's not.. great.. Having this, which is already on kilo gives heads more options to defend themselves. It's intentionally not amazing as well, so it's not used for hunting people down, it's to pepper people as you run away, not to be used as a killing weapon.

## Changelog
:cl:
add: Added mini e-guns to the lockers of the CMO, RD, and CE
del: Removed the mini e-guns from lockers in Kilo
/:cl:

### DIFF
--- a/_maps/map_files/KiloStation/KiloStation_skyrat.dmm
+++ b/_maps/map_files/KiloStation/KiloStation_skyrat.dmm
@@ -48286,7 +48286,6 @@
 	},
 /obj/machinery/firealarm/directional/south,
 /obj/structure/extinguisher_cabinet/directional/west,
-/obj/item/gun/energy/e_gun/mini,
 /turf/open/floor/iron/dark,
 /area/command/heads_quarters/cmo)
 "gsh" = (
@@ -49871,7 +49870,6 @@
 	dir = 8
 	},
 /obj/effect/turf_decal/tile/neutral,
-/obj/item/gun/energy/e_gun/mini,
 /obj/structure/extinguisher_cabinet/directional/north,
 /obj/machinery/camera{
 	c_tag = "Research Director's Office";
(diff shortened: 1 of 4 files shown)


desertmote:
same as before, but worse. just power creep in my opinion

axietheaxolotl:
> 
> 
> same as before, but worse. just power creep in my opinion

how is giving command members SELF-DEFENSE weapons, that they already have on one map a powercreep? The weapons is intentionally bad lmao. Also, you're still misusing the term power creep, but okay.

SpaceVampire:
There are a couple reasons why I disagree with this PR.
But I think the most utilitarian is that by making part of the balance come in the form of a map-edit.
You are going to have to ensure that the same thing is done again whenever we end up resetting the maps.
Thus ending up with double guns there if nothing is done.
I don't think it existing on one map, is a reason they should have it on all maps.
The telebaton is good enough for self-defense and department control.
There really isn't any reason that they need an extra ranged weapon for the job.
Especially when we already have a role that is supposed to help protect heads.
I'm not really sure how the term "powercreep" is being misused here.
When this gives extra guns to people who wouldn't otherwise have em.

ZephyrTFA:
This was discussed in the past and was shot down due to the existence of the blueshield.
While the blueshield exists no heads outside the Captain, HoS, and occasionally HoP, will be allowed to have self defense weaponry.
Besides this fact just giving them mini eguns doesn't even change much given antags don't even have many reasons to go after the less important heads.

tf-4:
Powercreep, feature freeze 👎

GoldenAlpharex:
Another feature during the feature freeze.

axietheaxolotl:
Re-opened because people liked this idea???? I dunno

SpaceVampire:
> Re-opened because people liked this idea???? I dunno

If you don't mind me asking, who liked it?
As this PR is currently still sitting at -6 votes.
So just saying that, doesn't add much in the way of extra justification.

GoldenAlpharex:
I don't think this is a good idea, nor see how this could be beneficial to the game. Sorry.

PR #7582, opened 2021-08-15
change to icemoon demon watchers

CandleJaxx:
## About The Pull Request

changes their ranged proc from basalisk/ice to normal basalisk

## Why It's Good For The Game

temp change until i figure out how to make the ice proc not do fire damage- this way it still does -75 temp damage and makes it more focused on melee


Gandalf2k15:
do this upstream

PR #7589, opened 2021-08-16
[MIRROR] Del The World: Unit testing for hard deletes

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/59612
--------------------

## About The Pull Request

Adds a unit test that runs at the very end of all other tests. It creates and then instantly destroys one of each type path. Some are blacklisted because of various reasons.

This won't catch anywhere near everything, but it'll help a lot with regressions, and catches a lot of runtimes too. The really nasty stuff is things that sleep as a part of initialize, since we don't init, wait, then qdel since it's a lot more prone to error

Oh and I went through and fixed all the errors I could, alongside adding descriptive errors on the stuff that's just gonna runtime in init.

This is gonna cause some unit test failures, but hopefully it'll just catch the simple stuff, which should be easy for people to fix.

I'm gonna add a label for fixing harddels, alongside a gbp boost of 5. I'd like people other then me to start at least somewhat caring about these

On the more technical side, @ ninjanomnom asked me to add support for running tests on all objects of whatever type. Wanted to ask @ Mothblocks for their opinion about it first tho.

## Why It's Good For The Game

\<Read the guide\>
[The Guide](https://github.com/LemonInTheDark/tgstation/blob/del-the-world/.github/HARDDEL_GUIDE.md)
\</Read the guide\>

This'll allow us to prevent hard delete regression, and it'll catch a good amount of general errors too since we're guaranteed to hit the init and destroy procs of nearly every typepath, which should help prevent runtime spam from say, spooky material not cleaning up properly or whatever.

~~Speaking of which, there's still some errors left on this pr, I'm hoping they're related to #59608, because otherwise I'm gonna have to actually fix them, which I don't relish.~~ They weren't, but I fixed them. Turns out connect loc is dumb, and pretty much the only solution was a `if(runtime) return` sorta deal.

Oh also I redid how tanks generated their parents list, it was causing warnings because they were doing it by relying on update_parents rather then using on_construct, which should really just be a mapload thing by the by

Lemme see, uh, this increases the awareness of hard deletes and prevents the bi yearly maintainer screed going through a list of recent harddels and just fixing whatever they can get their hands on.

This'll also catch all the bits of code that fail to return qdel/initialize hints, or do other naughty stuff. 

Seriously tho I just want atmos cpu time man is that so much to ask


Gandalf2k15:
OOOH YEAH BOYS TIME TO FIX THE FUCKING HARD DELS ALL HANDS ON DECK!!!

PR #7607, opened 2021-08-17
No more ventcrawling horror-lings

FlamingLily:
## About The Pull Request
Does what the title says

Yes there's a FF. This is urgent.

## Why It's Good For The Game
The "Last resort" "Go out with a bang" "If I go down, so do you" ability should not also allow relatively risk-free escapes. Nothing stops a ling going horror, popping out of a vent, killing, popping back in a vent. That's bad.

## Changelog
:cl:
balance: Horrorform lings no longer ventcrawl.
/:cl:

### DIFF
--- a/modular_skyrat/modules/horrorform/code/modules/mob/hostile/true_changeling.dm
+++ b/modular_skyrat/modules/horrorform/code/modules/mob/hostile/true_changeling.dm
@@ -56,7 +56,6 @@
 	devour = new /datum/action/innate/devour
 	turn_to_human.Grant(src)
 	devour.Grant(src)
-	ADD_TRAIT(src, TRAIT_VENTCRAWLER_ALWAYS, INNATE_TRAIT)
 
 /mob/living/simple_animal/hostile/true_changeling/Life()
 	. = ..()


Gandalf2k15:
Horror lings are supposed to be able to crawl, if you remove their staple escape route, revert the change that kills them infinitely

FlamingLily:
> 
> 
> Horror lings are supposed to be able to crawl, if you remove their staple escape route, revert the change that kills them infinitely

I fundamentally disagree with the notion that a horror ling should be able to escape. Horrorlings should be a "well, my best shot at surviving is to kill everyone here then scamper off and heal, but if I don't, I don't"

They should be a "I'm not going down without a fight" not a "Haha pop out of vent kill kill run away with no punishment"

hugmeorelse:
a

hugmeorelse:
> Horror lings are supposed to be able to crawl, if you remove their staple escape route, revert the change that kills them infinitely

This is true, but I think we can consider giving them something else besides vent crawl to compensate. Vent crawl + horror form can be very frustrating

PR #7641, opened 2021-08-19
[MIRROR] Adds a mapping helper that can cycle multiple airlocks regardless of location (#60675)

Useroth:
https://github.com/tgstation/tgstation/pull/60675


AlinaStarkova:
this will be needed for when this gets merged https://github.com/tgstation/tgstation/pull/60952

i only mention it since i notice it says theres an integration check failure, so just want to make sure that this PR is merged before the one i linked kek, as that would surely cause issues haha

PR #7661, opened 2021-08-20
[MODULAR] Quick-Spawning other people is no longer awkward for the spawning admin

GoldenAlpharex:
## About The Pull Request
It's no longer a game of trust to see if you trust non-admins with spawning themselves in. YOU now have the control, as the spawning admin, of how they're going to spawn in.

## Why It's Good For The Game
I don't want to accidentally have someone in Bluespace Technician outfit when I meant to spawn them as an assistant.

## Changelog
:cl: GoldenAlpharex
fix: Admins are now the one that decide how you spawn when they quick-spawn you, instead of having to awkwardly wait for you to do it yourself.
/:cl:

### DIFF
--- a/modular_skyrat/modules/icspawning/code/_onclick/observer.dm
+++ b/modular_skyrat/modules/icspawning/code/_onclick/observer.dm
@@ -11,13 +11,13 @@
 		outfits["Show All"] = "Show All"
 
 		var/dresscode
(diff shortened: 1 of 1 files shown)


ZephyrTFA:
You can already do this by just using the Select Equipment context menu action

GoldenAlpharex:
> You can already do this by just using the Select Equipment context menu action

I'm well aware, except that doesn't mean fixing it isn't better. Plus, this is actually faster.

PR #7678, opened 2021-08-21
Addition: Security Gun Vendor Circuit

Deek-Za:
This adds a circuit for the Armadyne Weapons Dispenser.

## About The Pull Request

Adds a Weapons Dispenser (Machine Board) [circuitboard/machine/gun_vendor] into game.
Adds Weapons Dispenser circuit into the Warden's Locker.

## Why It's Good For The Game

People have noted it's been destroyed times before and there's no way to replace it. Strangely it does not have a circuit unlike the ammo machines next to it. Due to it being in an area that is highly prone to being bombed this would offer a backup option.
Due to the important nature of the machine, this would add one spare board(Much like departmental lathes) in the Warden's Locker.

## Changelog
:cl: Deek-Za
balance: Prevents security and their respective tokens from potentially being taken entirely out of the fight in the event of a bombing.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/security.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/security.dm
@@ -138,6 +138,7 @@
 	new /obj/item/flashlight/seclite(src)
 	new /obj/item/clothing/gloves/krav_maga/sec/peacekeeper(src) //SKYRAT EDIT CHANGE - SEC_HAUL
 	new /obj/item/door_remote/head_of_security(src)
+	new /obj/item/circuitboard/machine/gun_vendor(src)
 
 /obj/structure/closet/secure_closet/security
 	name = "security officer's locker"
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/crates_lockers/closets/secure/security.dm
```diff
@@ -138,6 +138,7 @@
 	new /obj/item/flashlight/seclite(src)
 	new /obj/item/clothing/gloves/krav_maga/sec/peacekeeper(src) //SKYRAT EDIT CHANGE - SEC_HAUL
 	new /obj/item/door_remote/head_of_security(src)
+	new /obj/item/circuitboard/machine/gun_vendor(src)
```
axietheaxolotl:
```suggestion
	new /obj/item/circuitboard/machine/gun_vendor(src) // SKYRAT EDIT CHANGE
```
specify non-modular changes pls

GoldenAlpharex (approved):
I think this is a step in the right direction, at first I thought "Yeah but this should probably be tied to research" and then I remembered how there's many other boards that can't be printed, and it's not the end of the world. This way, if the weapon distributor boards are all gone, the station can still contact CentCom to ask for a new one. I think this is the right solution for this problem.

PR #7698, opened 2021-08-22
Capping Threat Level to 70 [Test Merge Only]

jjpark-kb:
## About The Pull Request
requested

See comment below.


hugmeorelse:
I checked the math, and it's been verified by Mothblocks.

Anyways, this is temporary. I'm going to go through and edit the events to match this proportionally, so we can revert back to 100 with the same effect.

hugmeorelse:
I checked the math, and it's been verified by Mothblocks. It just clamps the output given by the Lorentzian distribution for the threat level generation to a cap of 70. 

Shifting the center does not cap the max, so please do not suggest that to me. Probability distributions do not operate in that way. **Also, this will have virtually have no effect on most Skyrat rounds.** Due to altered parameters, we have a much lower variance to begin with which means all this does is keep us from having the occasional extreme rounds that only happen once in a few hundred rounds. All it does is keeping the extreme rounds from becoming more extreme. **You will still have the same amount of chaotic rounds.**

Later, this will be changed so it has more of a profound impact on the average round in Skyrat, however, this will be done temporarily as a means of experimenting on a macro-level. I plan on going through each event and adjusting their cost to match proportionally to the effect that's done here. To start, this method is much easier for us to change, allowing us to explore the sweet spot that matches the tempo we want for Skyrat, as we only have to adjust one value rather than having to change the cost of each and every dynamic event. That part comes later.

GoldenAlpharex:
@hugmeorelse @jjpark-kb Was there something that was changed in the dynamic config to not have to do it this way? If so that'd be quite swell.

GoldenAlpharex:
We'll bring this back if needed.

PR #7699, opened 2021-08-22
[MODULAR] Hypospray buffs

softcerv:
## About The Pull Request
Included in this PR are two different buffs.

First, the injection and spray speeds on the standard hypospray have been buffed. This is only for the standard Mk2 Hypospray, and not the CMO one.

- The injection time has been lowered from 2.5s to 2.0s
- The spray time has been lowered from 2.5s to 1.5s

Second, the standard hypovial can now inject it's entire storage (60u) instead of being able to inject only 20u max. It now also has more selection modes (30u, 40u, 50u, and 60u.)  

## Why It's Good For The Game
The standard hypospray is slow and clunky to use and is also outclassed by pills and patches with nitrile gloves. This PR should make the standard hypospray feel more responsive to use, and give doctors more control while using it.

## Changelog
:cl:
balance: Buffs the injection/spray speed on the standard Hypospray.
balance: Standard Hypovials can now inject their entire storage and have more selection modes 
/:cl:

### DIFF
--- a/modular_skyrat/modules/hyposprays/code/game/hyposprays_II.dm
+++ b/modular_skyrat/modules/hyposprays/code/game/hyposprays_II.dm
@@ -1,8 +1,8 @@
 #define HYPO_SPRAY 0
 #define HYPO_INJECT 1
 
-#define WAIT_SPRAY 25
-#define WAIT_INJECT 25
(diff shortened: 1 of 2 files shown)


softcerv:
*scream

PR #7702, opened 2021-08-22
[Modular] Adds nympho quirk to headrole, security restricted quirk lists.

sqnztb:
## About The Pull Request

Adds nymphomania to the HEAD_RESTRICTED_QUIRKS and SEC_RESTRICTED_QUIRKS lists. NT Fleet staff now have to abide by the head bad quirks list as well.

## Why It's Good For The Game

Reduces the chance of heads of staff and security going, "It's a medical condition!~" and ignoring their duties.

## Changelog
:cl:
balance: Head roles and security cannot have nymphomania. NT fleet now under same head role quirk restrictions.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -3,6 +3,6 @@
 #define JOB_UNAVAILABLE_LANGUAGE 7
 #define JOB_NOT_VETERAN 8
 
-#define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE, "Nymphomania" = TRUE, "Masochism" = TRUE, "Sadism" = TRUE
-#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE, "Nymphomania" = TRUE, "Masochism" = TRUE, "Sadism" = TRUE
+#define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE, "Nymphomania" = TRUE
+#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE, "Nymphomania" = TRUE
 #define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE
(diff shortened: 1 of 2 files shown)


Arctaisia:
Thankyouthankyouthankyouthankyou

nevimer:
only nymphmania deserves this, the rest don't make you do anything at all except fill your heart up. in the past, it made you do stuff, but that's not the case anymore.

lyricalpaws:
Right, let's see here. 

`#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE`

This is fine for the nt fleet, in fact I would *like* this for the ntf.

As for blacklisting the erp quirks. 

Nyphomania is fine to blacklist, it encourages erp which isn't the best for command nor sec.

Sadism and Masochism mechanically just increase your arrousal, which just makes you moan, so I don't see why they should be blacklisted. And because I know someone will say it, you do NOT cum from taking damage without manually using the climax verb.

KasparoVy:
> 
> 
> only nymphmania deserves this, the rest don't make you do anything at all except fill your heart up. in the past, it made you do stuff, but that's not the case anymore.

Really good to know, thanks!

sqnztb:
Pulled sadist and masochist from the blacklists - I'd originally been advised they worked differently.

FlamingLily:
This is unnecessary, it's not going to counter people ERPing as heads or sec (something which doesn't happen above green (or blue, for non-hos heads)), all it does is unfairly stifle the actual RP people that actually use these quirks can have

Furthermore, if heads of staff or sec are ERPing during an emergency, thats a staffing matter, and staff need to apply job bans. Mechanically "solving" this only hides the behaviour, not solves it.

If you don't like ERP, this probably isn't the server for you.

~~Infact, you should also add the clown pin to the blacklist because as clowns do, "they encourage LRP"~~

GoldenAlpharex (approved):
We don't really need more Command members that get affected by ERP mechanics while they're in the middle of something non-ERP-related.

Personally, I would've kept Sadism and Masochism, because it's a bit awkward to see a traitor shoot at the Captain, only to then see the Captain moaning instead of screaming in horror, but this is good enough for me.

FlamingLily:
> We don't really need more Command members that get affected by ERP mechanics while they're in the middle of something non-ERP-related.
> 

as I said above, this PR isn't stopping any problem behaviour, only hiding it. Problem players will go back to plugging all their holes with vibrators or doping on crocin, and will still be moaning on the bridge.

Nymphomania already provides a way to *not* get the nympho messages, because it gives a character a constant state of arousal, they can hit climax any time, which resets the stress counter and stops the messages. If you want to stop the nympho messages during a non-erp setting, yell at problem players, but this won't fix anything.

GoldenAlpharex:
> > We don't really need more Command members that get affected by ERP mechanics while they're in the middle of something non-ERP-related.
> 
> as I said above, this PR isn't stopping any problem behaviour, only hiding it. Problem players will go back to plugging all their holes with vibrators or doping on crocin, and will still be moaning on the bridge.
> 
> Nymphomania already provides a way to _not_ get the nympho messages, because it gives a character a constant state of arousal, they can hit climax any time, which resets the stress counter and stops the messages. If you want to stop the nympho messages during a non-erp setting, yell at problem players, but this won't fix anything.

I've seen it happen too many times and it's simply not the place for it. You can roleplay being a nympho, but having the Captain leaving to their office every 25 minutes to hit the climax button is kinda silly.

FlamingLily:
> I've seen it happen too many times and it's simply not the place for it. You can roleplay being a nympho, but having the Captain leaving to their office every 25 minutes to hit the climax button is kinda silly.

I don't see how it's any sillier than the captain leaving to the pharmacy to go get a medication for one of many command and sec available quirks.

Also, medication, **camphor** suppresses nymphomania, so there's another mechanic that exists to prevent this from happening if need be.

It's entirely workable, the issue lies with problem players, not the mechanic

Gandalf2k15:
Oh holy shit why is this quirk even a thing.

PR #7705, opened 2021-08-22
You must now opt into antag, and be antag now actually works.

Gandalf2k15:
## About The Pull Request

Basically what this does is inverts the current state of be_antag to FALSE, meaning when you join, you are set to have antagonist off.

And now the game actually respects if you have this on or off, if you have it off, you will not get it.

If no one opts in, the game will not start, and you will be notified. This means you can opt in and have a near 100% chance of getting antag.

## Changelog
:cl:
balance: Be antag now starts set to off.
fix: Be antag now actually works.
/:cl:


FlamingLily:
LET'S GO!

FlamingLily:
That was merged kinda fast

PR #7723, opened 2021-08-23
Ostwind, Norwind, Vintorez and PCR inhands sprites update.

LeonY24:
## About The Pull Request
Ostwind, Norwind, Vintorez and PCR are now dual-wielded.

## Why It's Good For The Game

New sprites would like much better than old ones, they look much more like a gun and just a bit more realistic.
Here are the changed guns.
image
image
image
image

## Changelog
:cl:
imageadd: DTR-6 Ostwind, DMR Norwind, Vintorez and PCR-9 now have dual-wielded sprites.
/:cl:


FlamingLily:
My concern here is that these items aren't actually two-handed, so you'll have one sprite in both hands, and then a second sprite in one of the hands.

Or worse, dual wielding two primaries.

I very much like the sprites though.

PR #7729, opened 2021-08-23
[MODULAR] Correcting a sin with the VSS Vintorez

Vizoplez:
## About The Pull Request
Title

## Why It's Good For The Game
Needless and stupid nerf to one of the ((least)) damaging primaries in the game.

Come on, it was already only propped up by its mag size. Killing it twice by nerfing that isn't ok.

## Changelog
:cl:
tweak: Vintorez now doesn't have the sinful 15 round mags (which don't exist IRL)
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/guns.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/guns.dm
@@ -691,7 +691,7 @@
 	icon_state = "norwind"
 	ammo_type = /obj/item/ammo_casing/b9mm
 	caliber = CALIBER_9MM
-	max_ammo = 15
+	max_ammo = 20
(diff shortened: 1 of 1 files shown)


Draganfrukts:
The mag size of the VSS was always 15

(And 20 round magazines are used by the AS-VAL but nobody cares about that :trollface:

Vizoplez:
> The mag size of the VSS was always 15
> 
> (And 20 round magazines are used by the AS-VAL but nobody cares about that :trollface:

it takes ASVAL mags

FlamingLily:
This is a balance change, knowing how robustly the FF is enforced, you're more likely to get it merged by labelling it correctly.

PR #7730, opened 2021-08-23
[MODULAR]Fixes shitcode with breaching hammers

Vizoplez:
## About The Pull Request
Title

## Why It's Good For The Game
Needless and stupid bug caused by improper use of changeNext_move

## Changelog
:cl:
fix: Breaching hammers now actually only delay your next action by 5 seconds if you try and breach a door.
/:cl:


Vizoplez:
yes this means misclicking with hammers no longer hardstuns anything other than moving for five seconds


yes this means they're now good for PVP if you're not aimbotting again

Funce (changes requested):
Pretty sure this was placed where it was to include an attack delay and missed that it affected everything.

Add an attack delay for this and we'll be good.

Funce (approved):

PR #7731, opened 2021-08-23
[MODULAR] Nerfs pepperball guns into the fucking floor

Vizoplez:
## About The Pull Request
Title

## Why It's Good For The Game
Pepper balls shouldn't be better than primaries unless you're wearing a security gas mask + glasses

## Changelog
:cl:
tweak: pepperballs no longer inject 50u of reagents (What the fuck?)
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/junior_officer/pepperball_gun.dm
+++ b/modular_skyrat/modules/sec_haul/code/junior_officer/pepperball_gun.dm
@@ -46,7 +46,7 @@
 	embedding = null
 	impact_effect_type = /obj/effect/temp_visual/impact_effect
 	var/contained_reagent = /datum/reagent/consumable/condensedcapsaicin
-	var/reagent_volume = 50
+	var/reagent_volume = 5
 
 /obj/projectile/bullet/pepperball/on_hit(atom/target, blocked, pierce_hit)
 	if(isliving(target))
@@ -60,7 +60,7 @@
 	name = "Pepperball Ammo Box"
 	id = "pepperballs"
 	build_type = AUTOLATHE | PROTOLATHE
-	materials = list(/datum/material/iron = 10000)
+	materials = list(/datum/material/iron = 5000)
 	build_path = /obj/item/ammo_box/advanced/pepperballs
 	category = list("intial", "Security", "Ammo")
 	departmental_flags = DEPARTMENTAL_FLAG_SECURITY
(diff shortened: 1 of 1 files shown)


Draganfrukts:
Wouldn't this be a balance change instead of a tweak? :trollface:

Vizoplez:
no :)

Deek-Za:
Alright so. I'll be honest this seems mostly to be a salt PR. 
I don't think it's a good idea to be making balance changes based off of a round where things went poorly for someone who wasn't prepared and got caught off-guard by a CDO. There's multiple options for protective coverings for this gun after all.
The fact is that this "Nerf" doesn't actually nerf the gun at all by addressing the wrong issue. If you do 5 over 50, it's still gonna trigger the same knockdown effect because that's just how it(The reagent) works. You can see this by using a peppersprayer which only doses 5u at a time, for an example.
I think a discussion about changing the effects of the pepper sprayer is fine, but this PR isn't anything more then a _placebo_.

Vizoplez:
> Alright so. I'll be honest this seems mostly to be a salt PR.
> I don't think it's a good idea to be making balance changes based off of a round where things went poorly for someone who wasn't prepared and got caught off-guard by a CDO. There's multiple options for protective coverings for this gun after all.
> The fact is that this "Nerf" doesn't actually nerf the gun at all by addressing the wrong issue. If you do 5 over 50, it's still gonna trigger the same knockdown effect because that's just how it(The reagent) works. You can see this by using a peppersprayer which only doses 5u at a time, for an example.
> I think a discussion about changing the effects of the pepper sprayer is fine, but this PR isn't anything more then a _placebo_.

This is incorrect. Condensed Capsaicin, while in your system, causes blurred vision, causes you to drop items, causes multiple separate effect. This applies 5u to put it in line with pepperspray, henceforth a /nerf into the ground/. It should not be a nuclear bomb of chems that cause funny effects.

image


it also slows you until its out of your system

Vizoplez:
AKA: 

The knockdown is fine.

The massive lingering effects the gun has is not.

Deek-Za:
> AKA:
> 
> The knockdown is fine.
> 
> The massive lingering effects the gun has is not.

Honestly, Yeah. This is a much better explanation of what this is trying to achieve. 
And that's fair.

FlamingLily:
As I said on the vintorez PR, this is a balance change, you'll get yelled at for mislabelling, but probably get it merged if it's labelled correctly.

Vizoplez:
> As I said on the vintorez PR, this is a balance change, you'll get yelled at for mislabelling, but probably get it merged if it's labelled correctly.

This is less of a balance PR and more of a tweak PR for the above reasons, the pepperballs function identically, its just they have a more reasonable after-effect duration now.

KathrinBailey:
> Wouldn't this be a balance change instead of a tweak? :trollface:

I mean 50u of chems per ball doesn't sound like intended behaviour, maybe a fix? 👀

Gandalf2k15:
50 was too much.

PR #7738, opened 2021-08-24
[Modular] Med Bandolier Buff

Nerev4r:
## About The Pull Request
Nothing unrealistic or too drastic or severe, just some subtle tweaks to improve the atmosphere.

Removes the ability for it to hold beakers. I will accept it no longer holding pill bottles too if that would make it more balanced.
Allows it to hold 14 items instead of 7.

## Why It's Good For The Game
The bandolier is pretty underwhelming at the moment and doesn't really suit as a replacement for the medical belt. This should hopefully even the odds and make it a bit more useful to actually take.

## Changelog
:cl:
balance: Medical Bandolier now holds 2x the items. Can't hold beakers or the syndicate syringe gun anymore.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/clothing/storage/belts.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/storage/belts.dm
@@ -157,15 +157,7 @@
 		/obj/item/reagent_containers/pill,
 		/obj/item/reagent_containers/syringe,
 		/obj/item/reagent_containers/medigel,
-		/obj/item/reagent_containers/glass/bottle/vial,
-		/obj/item/lighter, // This was too funny to pass up.
-		/obj/item/storage/fancy/cigarettes, // Same here.
 		/obj/item/storage/pill_bottle,
-		/obj/item/reagent_containers/hypospray,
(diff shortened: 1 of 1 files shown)


FlamingLily:
Should probably use "balance" instead of "add", and mention that you can't use beakers in 'em

GoldenAlpharex (changes requested):
I think I would prefer this as only containing items that would make sense to be clipped into a bandolier, such as the injectors, the implanters, the medicells, the hypovials, maybe pill bottles (I'm not too sure about this one yet, but I think that I'll trust our playerbase with that, they can already carry a ton in boxes anyway), syringes, I guess individual pills (even if that's really stupid, but I think it's funny), the medigels and the droppers. No syringe gun, no hypospray, no spray canister, etc.
This will make this belt feel more unique and feel like it really has a varied use case, but won't feel necessarily like the go-to for everyone, rather than the general-purpose medical belt.

Nerev4r:
ready probably.

GoldenAlpharex (approved):

PR #7760, opened 2021-08-25
changes to ashwalker lair

jjpark-kb:
## About The Pull Request
it feels like the ashwalker lair is far too much... this tones it back while still preserving that "feel"

## Why It's Good For The Game
balancing

## Changelog
:cl:
balance: ashwalker lair is less, with the same aesthetic
/:cl:

### DIFF
--- a/modular_skyrat/modules/mapping/_maps/RandomRuins/LavaRuins/lavaland_surface_ash_walker1_skyrat.dmm
+++ b/modular_skyrat/modules/mapping/_maps/RandomRuins/LavaRuins/lavaland_surface_ash_walker1_skyrat.dmm
@@ -3,178 +3,50 @@
 /obj/item/stack/sheet/mineral/coal/ten,
 /turf/open/floor/plating/asteroid/basalt/lava_land_surface,
 /area/ruin/unpowered/ash_walkers)
-"bc" = (
-/obj/structure/stone_tile/cracked{
-	dir = 8
-	},
-/obj/machinery/smartfridge/drying_rack{
-	idle_power_usage = 0;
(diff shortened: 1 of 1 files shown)


N8-Toe:
I've been playing alot of ashwalker lately, not sure how it's to much, the large size allows for building new structures which I see happen pretty often. While some of the space does go unused, I've been seeing pretty high ashwalker pop some rounds in the 6 to 8 range, and slimming down the base might make things cramped during those rounds. 

Also no images or descriptions of what is being changed.

jjpark-kb:
> I've been playing alot of ashwalker lately, not sure how it's to much, the large size allows for building new structures which I see happen pretty often. While some of the space does go unused, I've been seeing pretty high ashwalker pop some rounds in the 6 to 8 range, and slimming down the base might make things cramped during those rounds.

You should definitely expand the base and not just stick within the confines that are given to you.

> Also no images or descriptions of what is being changed.

You can view the changes by going to the Checks > MapDiffBot > The map drop down. But I will also post it here
image

CameronWoof:
I prefer the shape of the previous tendril temple; it actually looks like it's large enough to serve as a place of worship rather than just a room the tendril happens to be in. 

I *do* like that the buildings are more ordered, it makes it look like the temple is this larger structure with buildings constructed off of it, but the area seems too small now to be a village, as it's intended to be. It just needs more space. More buildings, maybe some with beds or something that isn't purely utilitarian. The original has some problems but this doesn't address them, it just makes the area even more difficult to build in.

Gandalf2k15 (approved):

PR #7770, opened 2021-08-25
Changes cult sacrifice and what happens to construct on their death

Arturlang:
## About The Pull Request
What if becoming a construct was simply a tempoary measure, instead of a likely permanent end?
This PR adds a component which is added to you whenever you are soul sharded, and makes the offer ritual no longer gib.

## Why It's Good For The Game
This makes being a construct a lot more bearable. Hopefully

## Changelog
:cl:
tweak: Constructs and shades now return to their old bodies on their death.
tweak: The offer rune will no longer gib carbons or humans, instead, simply killing them and popping most of their organs out.
/:cl:


ZephyrTFA:
This is very nice and I approve. 
@Gandalf2k15 consider TMing this as it's a very good idea. 

Also, don't manually change ckeys, use the transfer_to proc on the victims mind datum

softcerv:
God bless you for this. This makes cultists a way better fit for Skyrat

PR #7775, opened 2021-08-25
[MIRROR] Refactor area and turf lighting

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/60954
--------------------

## About The Pull Request
Pt1. of a bacon/fancy/hybrid/whatever lighting port

Replaces the current area and turf ligting system with one that uses an MA overlay instead whos color and intensity can be varedited and changed
Minorly changes the turf lighting as well
Wouldnt mind a small tm to test as i might have missed an area or two

## Why It's Good For The Game

Allows for more flexible area lighting with cleaner code

## Changelog
:cl:
refactor: Area based lighting can be varedited and changed now
/:cl:


jjpark-kb:
we have custom areas that will need to be modified, as well as I guess areas they are purging because they have no use, yet we may still use.

Go over this carefully

PR #7782, opened 2021-08-26
Fixes the Ant Reagent Recipe literally not working how did no one test this

BurgerLUA:
## About The Pull Request

Before: Ants could not be created from sugar because of fermichem fuckery.
After: I fixed the fermichem fuckery and made it so that ants require 10 sugar to reproduce because it's easier than 6 and having that mess with purity.

## Why It's Good For The Game

Bug fixes are good.

## Why it was broken
Because it was a unique recipe where it was removing reagents it was creating, it could break and cause a runtime. Instead of touching /tg/code I just decided to just change the flag of it because no one is paying me to fix this.

Is it an upstream issue? Don't fucking know, don't see any bug reports about it, I see people on reddit making ants on /tg/. Something broke here and again I'm not paid enough to diagnose reagentcode especially for a system I don't like.

## Changelog
:cl: BurgerBB
fix: Fixes the Ant Reagent Recipe.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/recipes/others.dm
+++ b/code/modules/reagents/chemistry/recipes/others.dm
@@ -911,12 +911,12 @@
 
 /datum/chemical_reaction/ants
 	results = list(/datum/reagent/ants = 3)
-	required_reagents = list(/datum/reagent/ants = 2, /datum/reagent/consumable/sugar = 6)
+	required_reagents = list(/datum/reagent/ants = 2, /datum/reagent/consumable/sugar = 10) //SKYRAT CHANGE, SUGAR FROM 6 TO 10.
 	//FermiChem vars:
 	optimal_ph_min = 3
 	optimal_ph_max = 12
 	required_temp = 50
-	reaction_flags = REACTION_INSTANT
+	reaction_flags = NONE //SKYRAT CHANGE, REACTION_INSTANT TO NONE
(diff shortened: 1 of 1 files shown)


FlamingLily:
This would be an upstream issue, I don't think we've touched chem

BurgerLUA:
> This would be an upstream issue, I don't think we've touched chem

It works upstream given the lack of bug reports on the subject and the fact that I've seen posts on reddit of it working.

GoldenAlpharex:
> > This would be an upstream issue, I don't think we've touched chem
> 
> It works upstream given the lack of bug reports on the subject and the fact that I've seen posts on reddit of it working.

To be able to tell that it works upstream, you'd need to test it upstream. You can't just "assume" it works.

BurgerLUA:
> > > This would be an upstream issue, I don't think we've touched chem
> > 
> > 
> > It works upstream given the lack of bug reports on the subject and the fact that I've seen posts on reddit of it working.
> 
> To be able to tell that it works upstream, you'd need to test it upstream. You can't just "assume" it works.

/tg/code is on strike right now on both ways so I'm not going to bother to see if its an upstream issue for sure.

BurgerLUA:
Just tested it. It works on /tg/. Whatever is causing it to break is because of Skyrat code as I've always said before.

BurgerLUA:
Not stale.

GoldenAlpharex:
Eh. I guess we can revert this later if it causes troubles.

PR #7759, opened 2021-08-25
[MODULAR] NSS Journey Prison Overhaul

ForrestWick:
## About The Pull Request

The first in a small series of prison reworks I'm doing. Essentially, this changes the whole prison wing for NSS Journey and makes it a bit more like a proper prison. It's no longer a donut loop and instead has proper deadends and whatnot, as well as removes the curtains from cells. Instead, prisoners can go ERP in little hidden rooms. The pool is also gone and the communal showers are replacing it. It still retains the overall space and feel of current prison but without the more "luxury resort" aspect people complained about. Also they get a cargo bounty pad now

image

## Why It's Good For The Game

A lot of people complain that the prison is too much of a resort instead of a prison, so now the prisoners will have a bit less "luxury" and more of a proper prison feel. This should encourage a bit more proper prison RP as well as make sec players happy too.

## Changelog
:cl:
add: Overhauled the entire prison for NSS Journey to be more like a proper prison instead of a luxury resort; added a cargo bounty pad and console, removed private cells and made hidden ERP rooms instead.
/:cl:

PR #7783, opened 2021-08-26
Removes prison wing curtains from all cells.

FlamingLily:
## About The Pull Request
These make no sense at an IC level nor at an OOC level, actively *hinder* sec gameplay, and especially CO gameplay.

"Ah yes, let's make the place we put our discovered antags have *no way of knowing when they're escaping*"

## Why It's Good For The Game
Because the addition of them in the first place was stupid and they remain stupid to this day.

## Changelog
:cl:
balance: Prisons no longer get *crime obfuscating curtains*. they're prisoners for gods sakes.
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -2881,17 +2881,6 @@
 	},
 /turf/open/floor/iron,
 /area/maintenance/disposal)
-"arq" = (
-/obj/machinery/button/curtain{
-	id = "prisoncell5";
-	pixel_y = 21
-	},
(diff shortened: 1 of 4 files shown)


ghost:
(Orion on his mobile account)

I'm gonna have to disagree with this pr for one reason:
Prison revamps are already planned, one of which is open and doesnt include curtains. Even merging this early as a bandaid would accomplish nothing except more merge conflicts for the prison revamp PRs, and thus this pr's changes overwritten within the month(hopefully).
#7759 is the currently open one, for reference. I think theyre doing them one at a time to avoid constant conflicts.

GoldenAlpharex:
For the sake that there's already a PR up to revamp the prisons, I don't think that we need this right at this moment.

PR #7784, opened 2021-08-26
[MIRROR] Exit on DreamMaker warnings in CI Windows build + Tools reset

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61025
--------------------

## About The Pull Request

Added an option to tools/build/build to treat DreamMaker warnings as errors. Enabled this option only for Windows Build (tools/ci/build.ps1).

## Why It's Good For The Game

Solves #60895. Warnings are treated only as errors if --error is flag is used in build. This is because you don't want local builds to fail because of warnings, sometimes you just want to test. It also is not enabled for all CI jobs. This is because I think it's beneficial to differentiate between map errors, integration error and and "warnings treated as errors" errors. The example is here https://github.com/mokulus/tgstation/pull/3. Only Windows build fails. The exit code is 2 to differentiate from a real error. There is also error message logged "=> Compile warnings treated as errors" and it only shows if there are warnings.

If any change doesn't seem right (for example naming or 2 return code) I'll change it.

## Changelog
:cl:
code: Treat DreamMaker warnings as errors in CI
/:cl:


GoldenAlpharex:
We need to compare all of our building tools with upstream, because there seems to be a lot of stuff missing.

.github/gbp.toml
```diff
@@ -3,18 +3,18 @@ reset_label = "GBP: Reset"
 
 [points]
 "Atomic" = 2
-"Balance/Rebalance" = -4
+"Balance/Rebalance" = -8
```
GoldenAlpharex: Revert this file, this was modified for our needs.

.github/workflows/pr_emoji.yml
```diff
@@ -0,0 +1,19 @@
+name: PR emoji stripper
```
GoldenAlpharex, line 1: Hahaha, the Emoji stripper for PRs, that's a good one.

GoldenAlpharex (changes requested):
From what I can tell, this /should/ work. So long as this is Juke 0.8.1 and not Juke 0.9.0, we should be good to go. Just remove `gbp.toml` from the PR and this should (hopefully) have no impact on us.

.github/gbp.toml
```diff
@@ -3,18 +3,18 @@ reset_label = "GBP: Reset"
...
-"Fix" = 4
+"Fix" = 3
 "Grammar and Formatting" = 1
 "Hard Deletes" = 12
 "Logging" = 1
 "Performance" = 12
 "Priority: CRITICAL" = 20
 "Priority: High" = 15
-"Quality of Life" = 2
+"Quality of Life" = 1
```
Gandalf2k15:
```suggestion
"Quality of Life" = 2
```

.github/gbp.toml
```diff
@@ -3,18 +3,18 @@ reset_label = "GBP: Reset"
 
 [points]
 "Atomic" = 2
-"Balance/Rebalance" = -4
+"Balance/Rebalance" = -8
 "Code Improvement" = 2
-"Mechanic" = -4
+"Mechanic" = -10
```
Gandalf2k15:
```suggestion
"Mechanic" = -4
```

.github/gbp.toml
```diff
@@ -3,18 +3,18 @@ reset_label = "GBP: Reset"
...
 "Feedback" = 2
-"Fix" = 4
+"Fix" = 3
```
Gandalf2k15:
```suggestion
"Fix" = 4
```

.github/gbp.toml
Gandalf2k15:
```suggestion
"Balance/Rebalance" = -4
```

PR #7795, opened 2021-08-26
donator item replacement

CandleJaxx:
out with the old, in with the whatever


SpaceVampire:
Don't make them a subtype of sunglasses, that gives them free flash protection.
They only need to be a subtype of normal glasses.

CandleJaxx:
please god i just want drip help

PR #7799, opened 2021-08-27
Disables CMEs until a rework is made.

FlamingLily:
## Why It's Good For The Game
CMEs are unfun, anti-RP, and round ending. No one cares to fix them because the damage is done when they hit, so it ends the round prematurely, hindering RP.

Meteors are fine because they give ample warning of when, and by their nature, where, they're going to hit, allowing people to hide from them and RP out the "oh god we're going to die." CMEs don't do that.

CMEs have no benefit, no positive upside, and are all negatives. 
They have the same reason for being in the game as an event that suddenly gibs everyone on station with no warning.

I've heard mention of a possible rework, but until then, they should be disabled to prevent any more round-ruining.

## Changelog
:cl:
del: CMEs.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cme/code/cme.dm
+++ b/modular_skyrat/modules/cme/code/cme.dm
@@ -17,7 +17,7 @@ Armageddon is truly going to fuck the station, use it sparingly.
 	typepath = /datum/round_event/cme
 	weight = 10
 	min_players = 30
-	max_occurrences = 1 // Why was this allowed to roll three times bruh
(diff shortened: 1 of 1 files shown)


CameronWoof:
Suggest testmerging until changes are made available.

softcerv:
I really hope they test merge this while they figure out some kind of actual solution to this, CMEs are actually awful right now.

Gandalf2k15:
We have that now

PR #6429, opened 2021-06-21
[NON-MODULAR] Minorly nerfs the stealth implant, now 5 seconds instead of 2 to cloak

FlamingLily:
## About The Pull Request
Maaaaaybe outright removing is a LITTLE too much. And no, this isn't a salt PR, please don't give me that.

## Why It's Good For The Game
But I do feel like this needs a nerf.

## Changelog
:cl:
balance: Stealth implant now takes 5 seconds to cloak, instead of 2.
/:cl:

PR #7808, opened 2021-08-27
[NONMODULAR] Nerfs stealth implant

FlamingLily:
## About The Pull Request
5 seconds to cloak

## Why It's Good For The Game
Stealth implant is an infinite, un-counterable in it's current state get out of jail free card

## Changelog
:cl:
balance: Stealth implants are slower
/:cl:

### DIFF
--- a/code/game/objects/items/implants/implant_stealth.dm
+++ b/code/game/objects/items/implants/implant_stealth.dm
@@ -20,7 +20,7 @@
 	enable_door_overlay = FALSE
 
 /obj/structure/closet/cardboard/agent/proc/go_invisible()
-	animate(src, , alpha = 0, time = 20)
+	animate(src, , alpha = 0, time = 5 SECONDS) // SKYRAT EDIT -- NERFED FROM time = 20
 
 /obj/structure/closet/cardboard/agent/Initialize()
 	. = ..()


Seris02:
5 seconds is a bit too much, how about something like 
2 seconds to cloak
and make it take like 2 seconds to actually activate it

FlamingLily:
> 
> 
> 5 seconds is a bit too much, how about something like
> 2 seconds to cloak
> and make it take like 2 seconds to actually activate it

That's 4 seconds total. 1 second difference.

Seris02:
It's the fact that it splits it that makes a difference. The cloaking can happen even if you accidentally touch a wall or something, then it would take 5 seconds again, which is too much. But if you make it take time to activate instead, then it'll be more fair, and it won't be an instant 'haha now I'm invisible', as well as not having to wait 5 seconds for it to recloak every time you touch a wall.

FlamingLily:
> 
> 
> It's the fact that it splits it that makes a difference. The cloaking can happen even if you accidentally touch a wall or something, then it would take 5 seconds again, which is too much. But if you make it take time to activate instead, then it'll be more fair, and it won't be an instant 'haha now I'm invisible', as well as not having to wait 5 seconds for it to recloak every time you touch a wall.

It only reveals you if you bump into a living atom

SpaceVampire:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/6429
Balance change during a feature freeze again? 🤔 
But really. I feel like it would it be better to increase the activation time instead.
So that it is still as useful for sneaking around once you are inside of it. But less of an escape tool.
Five seconds really is a bit too much for this kind of change. More than double what it was before.

ForrestWick:
The stealth box is also easy to counter since all you need to do is close a few doors

PR #7899, opened 2021-08-31
[MIRROR] The Maintening (maint walls are now maint areas)

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61084
--------------------

## About The Pull Request
image

Floyd mentioned yesterday that for engravings all walls shared between maint and a dept to be mapped as a maint area. as far as i saw, nobody ever actually responded to it to take on the issue, and since i had some time and was bored i went and did it, all maint is now expanded so any walls touching a maint area, is also a maint area

i did some testing and didnt see any issues, nor should there be, what area a wall is shouldnt affect anything since wallmounts are actually mapped onto a floor in the rooms, not on the wall tile itself

heres a small example of basically what this is doing: just maint areas over maint walls
image

while i was at it, i ensured all maint access doors are maint area'd like they should, and i also ensured that any special maint rooms had their doors actually their room area, not the maint one, (so like, the walls between maint and the solars control rooms are maint area'd but the solars door itself is set to the area of the solar room, so it uses power from the room, not maint)

also fixed some minor mapping issues i noticed, showcased below:
(cringe double reinf walls)
image

(?????)
image

(cringe)
image

(why is this maint snaking into the other maint?)
image

(areas like this have their maint doors be area'd to the area that fills the center room, so that the doors all share the same power source rather than cuting off one part of it while the others remain active, in this particular case, viro door stays viro, so the airlock powers together)
image

## Why It's Good For The Game
was a change asked for, its considered "area cleanup"
approved by Maurakas

## Changelog
:cl:
fix: Maint areas have been cleaned up to properly cover maint, minor tiles and airlocks which had space areas were fixed to have an actual maint area
/:cl:


Gandalf2k15:
This will require a map reset.

PR #7900, opened 2021-08-31
Make centcom not full bright

Gandalf2k15:
## About The Pull Request

Reverts a change in a mirror PR


GoldenAlpharex:
I'm already fixing this, so it won't be needed any longer.

PR #7919, opened 2021-09-02
hurtsposals

jjpark-kb:
## About The Pull Request
makes disposals have a chance of hurting you when changing direction.

a large metal chute that makes the objects going down it go down fast... and yet they arent getting banged up?
introduces an element of "last resort" where if you want to escape through the bin, you gotta think carefully

## Changelog
:cl:
add: added hurtsposal (disposals now can hurt you)
/:cl:

### DIFF
--- a/code/modules/recycling/disposal/pipe.dm
+++ b/code/modules/recycling/disposal/pipe.dm
@@ -18,8 +18,8 @@
 	var/flip_type // If set, the pipe is flippable and becomes this type when flipped
 	var/obj/structure/disposalconstruct/stored
 	//SKYRAT EDIT: HURTSPOSALS
-	///whether a disposal pipe will hurt if a person changes direction
-	var/fluffy = FALSE
+	/// Whether a disposal pipe will hurt if a person changes direction. `FALSE` for hurting, `TRUE` to prevent making them hurt.
+	var/padded_corners = FALSE
 	//SKYRAT EDIT: HURTSPOSALS
 
 
--- a/code/modules/recycling/disposal/pipe.dm
+++ b/code/modules/recycling/disposal/pipe.dm
@@ -17,6 +17,10 @@
 	var/initialize_dirs = NONE // bitflags of pipe directions added on init, see \code\_DEFINES\pipe_construction.dm
 	var/flip_type // If set, the pipe is flippable and becomes this type when flipped
 	var/obj/structure/disposalconstruct/stored
+	//SKYRAT EDIT: HURTSPOSALS
+	///whether a disposal pipe will hurt if a person changes direction
+	var/fluffy = FALSE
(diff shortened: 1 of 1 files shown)


FlamingLily:
You should make it do brain damage :)

ghost:
Less-than 50% chance, non-lethal damage... I like it, keeps disposals escape as a valid option, but punishes people who jump in them repeatedly/are unlucky enough to go through the entire loop. 👍

code/modules/recycling/disposal/pipe.dm
```diff
@@ -17,6 +17,10 @@
 	var/initialize_dirs = NONE // bitflags of pipe directions added on init, see \code\_DEFINES\pipe_construction.dm
 	var/flip_type // If set, the pipe is flippable and becomes this type when flipped
 	var/obj/structure/disposalconstruct/stored
+	//SKYRAT EDIT: HURTSPOSALS
+	///whether a disposal pipe will hurt if a person changes direction
+	var/fluffy = FALSE
```
GoldenAlpharex: Needs a meaningful name and the comment needs to make sense with the variable ;)
```suggestion
	/// Whether a disposal pipe will hurt if a person changes direction. `FALSE` for hurting, `TRUE` to prevent making them hurt.
	var/padded_corners = FALSE
```

code/modules/recycling/disposal/pipe.dm
```diff
@@ -80,7 +84,17 @@
...
+	if(dir != P.dir && !fluffy)
+		if(prob(20))
+			for(var/objects_within in H.contents)
+				if(!isliving(objects_within))
+					continue
+				var/mob/living/living_within = objects_within
+				if(living_within.stat == DEAD)
+					continue
+				living_within.adjustBruteLoss(5)
+	//SKYRAT EDIT: HURTSPOSAL
```
GoldenAlpharex: Here, should work with the previous comment I left.
```suggestion
	/// SKYRAT EDIT START - HURTSPOSAL
	if(dir != P.dir && !padded_corners)
		if(prob(20))
			for(var/objects_within in H.contents)
				if(!isliving(objects_within))
					continue
				var/mob/living/living_within = objects_within
				if(living_within.stat == DEAD)
					continue
				living_within.adjustBruteLoss(5)
	/// SKYRAT EDIT END
```

GoldenAlpharex (changes requested):
We're getting there. I think this should be test-merged before a full-merge, though, just so we can find out about any big issues, should they arise.

GoldenAlpharex (approved):
Seems good for a test-merge in my opinion.

Strazyplus:
Suggestion, breakable objects in disposals should break. as in glass.

Gandalf2k15:
LGTM

PR #7975, opened 2021-09-04
some forge fixes

jjpark-kb:
## About The Pull Request
so what if I said deconstructing the forge didnt give you back the goliath hides?
so what if I said that smelting ore would have yielded you almost nothing?
so what if I said I might have forgotten an in_use = FALSE and return somewhere?
(just know I can't deal with it if it runtimes, that'll need a refactor)

bugfixing and putting in intended things.

## Changelog
:cl:
fix: added one more in_use = false to the forge
fix: forge gives back goliath hides upon deconstruction
fix: forge actually gives the amount of ore it should have been
/:cl:


Aphast:
what if man, what if...

PR #7983, opened 2021-09-05
NSS Journey Prison touchups

ForrestWick:
## About The Pull Request

image

Turns out I made a fucking poopy and forgot a sink in the kitchen. I also added a rec room and yard area as well. Also a Lustwish vendor they have to break down a wall to get to.

Fixes and stuff.

## Changelog
:cl:
expansion: The NSS Journey prison now has a rec room, yard area, and a sink in the kitchen, with water tanks too. The secret room also has an addition many prison players are probably craving.
/:cl:

### DIFF
--- a/_maps/map_files/NSSJourney/NSSJourney.dmm
+++ b/_maps/map_files/NSSJourney/NSSJourney.dmm
@@ -1686,6 +1686,10 @@
 /obj/effect/spawner/structure/window/reinforced,
 /turf/open/floor/plating,
 /area/security/processing)
+"akJ" = (
+/obj/structure/table,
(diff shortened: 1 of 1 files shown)


FlamingLily:
This is just re-introducing the same issues again, curtains mean prisoners have an easy way to escape, and the lustwish has actual, real tools, including a screwdriver and a multitool, so it definitely shouldn't be get-able.

ForrestWick:
> This is just re-introducing the same issues again, curtains mean prisoners have an easy way to escape, and the lustwish has actual, real tools, including a screwdriver and a multitool, so it definitely shouldn't be get-able.

This was resolved on discord btw

PR #8024, opened 2021-09-06
[MODULAR] Removes screenshake from M45A5

axietheaxolotl:
## About The Pull Request

title

it doesn't, it's just annoying and needs to be removed.

## Changelog

:cl:
del: Removed M45A5 Screenshake
/:cl:


GoldenAlpharex:
In this instance, considering how it's meant to be a self-defense weapon, yeah, I can see why we'd want to remove it, worst case scenario we can add it back later.

PR #8070, opened 2021-09-09
[MODULAR] Mop recipes

YakumoChen:
## About The Pull Request

You can now craft mops from cloth, a rod, and a bit of coil
You can now craft a tribal mop from wood, cloth, and sinew.
(the tribal mop isn't actually any different)

I absolutely loathe the liquid spills system but what I loathe most about it is how obnoxious is to remove spills. 
Mops are now craftable as a means to partially get around this, since literally nothing else works to get the slip hell off the god damn floor
(this PR is mostly for ash walkers though who don't get mops)

## Changelog

:cl:
qol: You can craft mops in the crafting menu now, with a rod, some cloth, and some cables.
/:cl:


YakumoChen:
I forgot sinew exists so I'll make a proper tribal version first

PR #8073, opened 2021-09-09
Fixes the Foreigner quirk to work correctly given that 90% of players aren't basic human.

Iamgoofball:
## About The Pull Request

See title.

Perk is bugged and leaves you with no languages currently.

:cl:
fix: Fixes the Foreigner quirk to work correctly given that 90% of players aren't basic human.
/:cl:

### DIFF
--- a/code/datums/quirks/neutral.dm
+++ b/code/datums/quirks/neutral.dm
@@ -39,14 +39,16 @@
 /datum/quirk/foreigner/add()
 	var/mob/living/carbon/human/human_holder = quirk_holder
 	human_holder.add_blocked_language(/datum/language/common)
-	if(ishumanbasic(human_holder))
-		human_holder.grant_language(/datum/language/uncommon)
+	//SKYRAT EDIT: Fixes most people not getting Galactic Uncommon.
+	//if(ishumanbasic(human_holder))
+	human_holder.grant_language(/datum/language/uncommon)
 
 /datum/quirk/foreigner/remove()
 	var/mob/living/carbon/human/human_holder = quirk_holder
 	human_holder.remove_blocked_language(/datum/language/common)
-	if(ishumanbasic(human_holder))
-		human_holder.remove_language(/datum/language/uncommon)
+	//SKYRAT EDIT: Fixes most people not getting Galactic Uncommon.
+	//if(ishumanbasic(human_holder))
+	human_holder.remove_language(/datum/language/uncommon)
 
 /datum/quirk/vegetarian
 	name = "Vegetarian"


code/datums/quirks/neutral.dm
```diff
@@ -39,14 +39,16 @@
...
+	human_holder.grant_language(/datum/language/uncommon)
 
 /datum/quirk/foreigner/remove()
 	var/mob/living/carbon/human/human_holder = quirk_holder
 	human_holder.remove_blocked_language(/datum/language/common)
-	if(ishumanbasic(human_holder))
-		human_holder.remove_language(/datum/language/uncommon)
+	//SKYRAT EDIT: Fixes most people not getting Galactic Uncommon.
+	//if(ishumanbasic(human_holder))
+	human_holder.remove_language(/datum/language/uncommon)
```
GoldenAlpharex, line 51: For this, I would recommend checking that it's not already in their background that they know the language, otherwise they will lose it in case it ever gets taken away from them.

GoldenAlpharex (changes requested):

code/datums/quirks/neutral.dm
Iamgoofball: I flipped through backgrounds and nothing seems to give access to Galactic Uncommon except this quirk.

GoldenAlpharex: Yuh, but in case there's ever a chance someone has it one way or another, surely there's a way to check where the language came from, no?

Iamgoofball: Fair

Azarak:
This quirk shouldn't exist in the first place (And was removed when background languages were introduced, so nice job whoever re-introduced it). There are job checks regarding languages and they're checked from the languages from backgrounds, aswell as there is no such thing as a "Galactic Uncommon".

GoldenAlpharex:
> This quirk shouldn't exist in the first place (And was removed when background languages were introduced, so nice job whoever re-introduced it). There are job checks regarding languages and they're checked from the languages from backgrounds, aswell as there is no such thing as a "Galactic Uncommon".

image

Azarak:
> 
> 
> > This quirk shouldn't exist in the first place (And was removed when background languages were introduced, so nice job whoever re-introduced it). There are job checks regarding languages and they're checked from the languages from backgrounds, aswell as there is no such thing as a "Galactic Uncommon".
> 
> image

This is a roleplay server, no? Then what is a galactic uncommon? And why is it called uncommon even if it's claimed to be a second most spoken Human language, that nobody knows about if they're not foreigners? It doesn't have any support by lore either. It's all nonsense and contradictions. 

We also have a sub-common language that's called Spacer, that would fit the role of this language, and it's already in a system that supports proper customization, and not shitty quirks

FlamingLily:
My suggestion: Make Galactic Uncommon into Sol Common, make it a human racial language, and scrap this quirk

Iamgoofball:
All of your suggestions are bad and are against the design of this quirk. Also, job checks for languages is dumb and should be removed.

Iamgoofball:
>This is a roleplay server, no?

this quirk and roleplay server are not mutually exclusive

GoldenAlpharex:
I'd like to know if it's still an issue as of right now.

Iamgoofball:
technically obsoleted by the new languages menu I made

PR #8085, opened 2021-09-10
less destructive rodstopper

jjpark-kb:
## About The Pull Request
makes the rod stopper a little nicer: gives you 5 seconds to run away and less destructive

requested: less destruction means less problems; less RR means more fun to be gained

## Changelog

:cl:
add: rodstoppers now give you 5 seconds to evacuate the area
balance: rodstopper is not so destructive
/:cl:


xXPawnStarrXx:
You could also add something to the examine of the machine and circuit board that warns people it explodes or implodes.

PR #8126, opened 2021-09-12
[NONMODULAR] fuck you dynamic (turns the admin cancel button for midrounds from 10 seconds to 60 seconds)

Seris02:
## About The Pull Request

now you don't have to be like fuck fuck should I click it and you actually get a tiny bit to discuss it

because then we don't have shit like round 2977, where it was all filled with roleplay, until the xenos midround, and I couldn't cancel it, so they wiped the floor with the lowpop station

## Changelog

:cl:
config: changes the admin midround pick time from 10 seconds to 60 seconds
/:cl:

### DIFF
--- a/code/game/gamemodes/dynamic/ruleset_picking.dm
+++ b/code/game/gamemodes/dynamic/ruleset_picking.dm
@@ -1,4 +1,4 @@
-#define ADMIN_CANCEL_MIDROUND_TIME (60 SECONDS)
+#define ADMIN_CANCEL_MIDROUND_TIME (60 SECONDS) //SKYRAT EDIT - ORIGINAL 10 SECONDS
 
 /// From a list of rulesets, returns one based on weight and availability.
 /// Mutates the list that is passed into it to remove invalid rules.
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/dynamic/ruleset_picking.dm
```diff
@@ -1,4 +1,4 @@
-#define ADMIN_CANCEL_MIDROUND_TIME (10 SECONDS)
+#define ADMIN_CANCEL_MIDROUND_TIME (60 SECONDS)
```
RatFromTheJungle: // SKYRAT EDIT - ORIGINAL 10s

Gandalf2k15:
You shouldn't be canceling midrounds. And if you commit that sin, you better be choosing something else.

Seris02:
this changes the entire string, as in the MIDROUND bla bla (CANCEL | SOMETHING ELSE) string, it just gives more time to decide

Gandalf2k15:
Yes and clicking the something else button is what you should be doing. If you're playing a non curated round and then you start fucking with the flow, you take the randomness out of it and may as well just run curated all the time. There are very very very few circumstances that allow you to cancel a naturally occurring midround.

Iamgoofball:
The Cancel button is there for when you're running an event and need to stop a meteor shower interrupting it. Just hit Something Else.

hugmeorelse:
Increasing the time to 60 seconds is fine.

Gandalf2k15:
Anyway 60 seconds is fine however do not use this incorrectly.

PR #8136, opened 2021-09-13
Makes it so cyborgs can get the affection modules again after getting their model reset

GoldenAlpharex:
## About The Pull Request
~~I haven't tested this yet, so don't merge it until I do.~~
Does what it says on the tin. It was very simple, and works as expected.

People are complaining about it, and it makes no sense for it to not be able to work properly.

## Changelog

:cl: GoldenAlpharex
fix: Cyborgs can now be given the affection modules after getting their model reset while they had them.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -648,6 +648,7 @@
 		hasShrunk = FALSE
 		resize = (4/3)
 		update_transform()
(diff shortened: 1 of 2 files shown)


KathrinBailey:
There's still a DNM tag on this, if this is ready convert it back to open and since you can do it yourself, remove the DNM tag.

GoldenAlpharex:
Yeah, I forgot to test this. Now it's tested. It works.

PR #8342, opened 2021-09-23
Reverts NTF roles and the NTF Cargo system pending OVERMAP.

Gandalf2k15:


axietheaxolotl:
haha what if we made this permanent and didn't re add them
haha

Aphast:
poop fart cargo cuck system, keep it down

PR #8426, opened 2021-09-27
[fix]fixes the cossac bug

thestubborn:
cossack now works, both are here

i'd re add more loadout but it causes lag, sorry

I don't want gbps for this


thestubborn:
the issue was in all the chaos, someone had the sec and normal varient share the same path
now they both are seperate

PR #8441, opened 2021-09-28
Ghost Files

Funce:
So we had uhhh, two versions of these files

icons/UI_Icons/contracts/bluespace.png
the other
icons/ui_icons/contracts/bluespace.png

And while some git clients can figure out that these are the same, others complain forever. So here's a PR to remove the extra ones.


GoldenAlpharex:
Very good PR

PR #8458, opened 2021-09-28
Temporarily reverts powerarmor

Gandalf2k15:
Well excuse me for trusting something to be somewhat balanced.


YakumoChen:
Come on. The code wasn't even cleaned up if you take even a brief glance at it (powerarmor.dm is still chock full of  magic numbers and single letter variables)

It wasn't even tested for something that's likely to be a majorly powerful feature??

I even pointed out almost it wasn't going to be balanced at all.

PR #8492, opened 2021-09-30
cargo teleporter

jjpark-kb:
## About The Pull Request
adds the item that can teleport other iterms! (can't be living and cant be anchored!)

qol for cargo techs, being able to deliver things faster to each place

## Changelog

:cl:
add: added the cargo teleporter, item that can teleport items to markers
/:cl:


GoldenAlpharex:
That seems... Really odd of a decision. If people don't want to bother delivering it themselves, they can already use the destination tagger, wrap the crate and yeet it in disposals.

YakumoChen:
power creep

YakumoChen:
A more constructive comment: We already have a lot of items that are made to teleport things (which have high research requirements and or limited uses / availability/difficulty in making/require a large, complicated degree of setup beyond just clicking a spot on the ground you want to teleport to)
We have: 
-hand teles
-bluespace lockers
-quantam telepads
-just the teleporter machine itself
-teleport pads with the console
-circuits using teleport consoles
-The teleport doodad that miners get
-The other teleport doodad miners get that only works outside
The list probably goes on.

We also have _several_ ways to deliver the mail, including TWO (!) automated varieties that you don't need to leave cargo to use.
-Disposals mail
-The hauler bots nobody ever actually uses for some reason 

Both of these seem like this severely intrudes into that space without adding anything significantly more other than being, as most would define, actually just power creep. 
What makes this different other than being any of the above teleport methods but significantly easier to set up? 
Why does the QM get one at round start (mostly nullifying the point of researching them?)
What stops someone from just teleporting a living mob in a locker or crate?

thestubborn:
tbf, its the disposals that no one uses, people tend to use the shit bots instead

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
```diff
@@ -0,0 +1,126 @@
...
+	COOLDOWN_DECLARE(use_cooldown)
+
+/obj/item/cargo_teleporter/examine(mob/user)
+	. = ..()
+	. += span_notice("Attack itself to set down the markers!")
+	. += span_notice("ALT-CLICK to remove all markers!")
+
+/obj/item/cargo_teleporter/Destroy()
+	if(length(marker_children))
+		for(var/obj/effect/decal/cleanable/cargo_mark/destroy_children in marker_children)
```
Fikou, line 20: as anything

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
```diff
@@ -0,0 +1,126 @@
...
+			destroy_children.parent_item = null
+			qdel(destroy_children)
+	. = ..()
+
+/obj/item/cargo_teleporter/attack_self(mob/user, modifiers)
+	if(length(marker_children) >= 3)
```
Fikou, line 26: use a define

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
```diff
@@ -0,0 +1,126 @@
...
+		return
+	to_chat(user, span_notice("You place a cargo marker below your feet."))
+	var/obj/effect/decal/cleanable/cargo_mark/spawned_marker = new /obj/effect/decal/cleanable/cargo_mark(get_turf(src))
+	playsound(src, 'sound/machines/click.ogg', 50)
+	spawned_marker.parent_item = src
+	marker_children += spawned_marker
+
+/obj/item/cargo_teleporter/AltClick(mob/user)
+	if(length(marker_children))
+		for(var/obj/effect/decal/cleanable/cargo_mark/destroy_children in marker_children)
```
Fikou, line 37: as anything

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
```diff
@@ -0,0 +1,126 @@
...
+	. = ..()
+	. += span_notice("Attack itself to set down the markers!")
+	. += span_notice("ALT-CLICK to remove all markers!")
+
+/obj/item/cargo_teleporter/Destroy()
+	if(length(marker_children))
+		for(var/obj/effect/decal/cleanable/cargo_mark/destroy_children in marker_children)
+			destroy_children.parent_item = null
+			qdel(destroy_children)
+	. = ..()
```
Fikou: return ..() at the end of procs

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
```diff
@@ -0,0 +1,126 @@
...
+		if(!ismovable(check_content))
+			continue
+		var/atom/movable/movable_content = check_content
+		if(isliving(movable_content))
+			continue
+		if(movable_content.anchored)
+			continue
+		if(!do_after(user, 1 SECONDS, target = target))
+			break
+		movable_content.forceMove(moving_turf)
```
Fikou: do_teleport?

modular_skyrat/modules/cargo_teleporter/code/cargo_teleporter.dm
jjpark-kb: unsure about what you mean

Seris02:
make it so that it uses the get_all_contents_type(/mob/living) proc to look for living things in contents, otherwise it'll just skip right over a bluespace bodybag hiding in a locker or something.

jjpark-kb:
> make it so that it uses the get_all_contents_type(/mob/living) proc to look for living things in contents, otherwise it'll just skip right over a bluespace bodybag hiding in a locker or something.

feared I would have to do that.

YakumoChen:
I printed one and it wouldn't teleport crates, just left an ash and did nothing

It doesn't have any debugging / sensibility features (chat error when it fails, warm up before teleporting, actual teleport noise, etc) and trying to teleport something that doesn't work (like yourself, for instance) does nothing but refresh the cooldown.

Choosing which mark to send it to is extremely clunky, it should be a toggle you set and not something you choose every time you teleport an object.

The description also doesn't feel clear enough, making marks feels clunky, getting rid of marks is confusing and even more clunky, and its functionality not consistent with other common items that leave markers on the ground (like holo-barrier items)

PR #8506, opened 2021-09-30
[MODULAR] Conversion Kits Framework

faaaay:
## About The Pull Request

This PR adds a consumable "custom_kit" object, which can be used on a specific item to turn it into another specific item. Simple enough, right? The intent behind this is to allow for a greater range of donor items. Under this system, your item could be a conversion kit that allows you to reskin an item that you may not always have access to. For example, it could turn a standard hardsuit into a painted (but functionally identical) variant, or replace your armoured peacekeeper beret with another equally practical hat!

Things to note:

- This does not currently **add** any kits, or change any existing items to use them. It's just laying the groundwork for that in the future.
- The kits' sprite is currently a vendor restock sprite. This is a placeholder with the intent that when specific kits are added, they'll use a variant (probably this sprite with a symbol on it).

It doesn't.
Does it contribute to the RP experience, exactly? Maybe not. But it may allow for a better range of donor/custom items.

## Changelog

:cl:
add: Added generic "conversion kits". In the future, these may be used to implement certain custom items.
/:cl:


modular_skyrat/modules/customization/game/objects/items/conversion_kits.dm
```diff
@@ -0,0 +1,21 @@
...
+    if(!proximity_flag) //Gotta be adjacent to your target
+        return
+    if(isturf(O)) //This shouldn't be needed, but apparently it throws runtimes otherwise.
+        return
+    else if(O.type == from_obj) //Checks whether the item is eligible to be converted
+        var/obj/item/converted_item = new to_obj(get_turf(src))
+        user.put_in_hands(converted_item)
+        qdel(O)
+        qdel(src)
+        user.visible_message("You modify [O] into [converted_item].")
```
GoldenAlpharex: Otherwise it'll runtime as well, because it'll try to use the name of something that's being deleted.
```suggestion
        user.put_in_hands(converted_item)
        user.visible_message("You modify [O] into [converted_item].")
        qdel(O)
        qdel(src)
```

modular_skyrat/modules/customization/game/objects/items/conversion_kits.dm
```diff
@@ -0,0 +1,21 @@
...
+    var/from_obj = null //The base object to be converted
+    var/to_obj = null //The object to turn it into
+
+/obj/item/device/custom_kit/afterattack(obj/O, mob/user as mob, proximity_flag)
+    if(!proximity_flag) //Gotta be adjacent to your target
+        return
+    if(isturf(O)) //This shouldn't be needed, but apparently it throws runtimes otherwise.
+        return
+    else if(O.type == from_obj) //Checks whether the item is eligible to be converted
+        var/obj/item/converted_item = new to_obj(get_turf(src))
```
GoldenAlpharex, line 17: You make it as an item here, probably should be forced to have `to_obj` as an item too.

modular_skyrat/modules/customization/game/objects/items/conversion_kits.dm
```diff
@@ -0,0 +1,21 @@
+/obj/item/device/custom_kit
+    name = "modification kit" 
+    desc = "A box of parts for modifying a certain object."
+    icon = 'modular_skyrat/master_files/icons/donator/obj/kits.dmi'
+    icon_state = "partskit"
+    var/from_obj = null //The base object to be converted
+    var/to_obj = null //The object to turn it into
+
+/obj/item/device/custom_kit/afterattack(obj/O, mob/user as mob, proximity_flag)
```
GoldenAlpharex: Use a more descriptive variable name than `O`.

modular_skyrat/modules/customization/game/objects/items/conversion_kits.dm
```diff
@@ -0,0 +1,21 @@
+/obj/item/device/custom_kit
+    name = "modification kit" 
+    desc = "A box of parts for modifying a certain object."
+    icon = 'modular_skyrat/master_files/icons/donator/obj/kits.dmi'
+    icon_state = "partskit"
+    var/from_obj = null //The base object to be converted
+    var/to_obj = null //The object to turn it into
```
GoldenAlpharex: As far as I know, you don't need to make them equal to null, declaring a variable without giving it anything makes it null too, I think. Try it out and tell me if it works.

GoldenAlpharex (changes requested):
Great idea!

softcerv:
This is dope. I can definitely see using this for mediguns instead of the current upgrade via crafting method thing I am using!

faaaay:
> Great idea!

Thanks for the code suggestions, I've gone through and made those tweaks.

> This is dope. I can definitely see using this for mediguns instead of the current upgrade via crafting method thing I am using!

I didn't think of that, but that'd be a good use for it!

modular_skyrat/modules/customization/game/objects/items/conversion_kits.dm
```diff
@@ -0,0 +1,21 @@
+/obj/item/device/custom_kit
+    name = "modification kit" 
+    desc = "A box of parts for modifying a certain object."
+    icon = 'modular_skyrat/master_files/icons/donator/obj/kits.dmi'
+    icon_state = "partskit"
+    var/from_obj //The base object to be converted
+    var/obj/item/to_obj //The object to turn it into
```
GoldenAlpharex: Might as well make it autodoc-compatible. Also, let's make both as items for now, we can change that later if need be, but might as well make it coherent.
```suggestion
    /// The base object to be converted.
    var/obj/item/from_obj
    /// The object to turn it into.
    var/obj/item/to_obj
```

GoldenAlpharex (changes requested):
Almost there!

GoldenAlpharex (approved):
Nice!

PR #8509, opened 2021-09-30
Collars the singularity

Gandalf2k15:
## About The Pull Request

This rowdy bitch kept eating it's own cage so now we've locked it up tight.

It can no longer eat it's own containment field.
Glass no longer gets yoinked at stage 2, instead it's stage 5.
Railings are very sturdy so they only get pulled at stage 5.

Now you can safely make singularity engines!!!

:cl:
fix: Due to the fact we actually use the singularity as an engine, we have repaired it's ability to consume it's own containment and pull the glass out of your glasses.
/:cl:


modular_skyrat/master_files/code/modules/power/singularity/containment_field.dm
```diff
@@ -0,0 +1,2 @@
+/obj/machinery/field/containment/singularity_act()
+	return
```
jjpark-kb, line 2: is this the machine that generates the field? or the field itself?

Gandalf2k15 (commented):
lgtm

modular_skyrat/master_files/code/modules/power/singularity/containment_field.dm
Gandalf2k15: The field itself

PR #8623, opened 2021-10-05
Fixed Internals for Lizards(And other races too!)

Deek-Za:
## About The Pull Request
This checks your testicles for internal status and if so, prevents them from being shown or able to be exposed.
If not, it simply returns the normal external testicle description.
This removes their visibility and ability to be exposed, while allowing them to function properly.

Internals Testicles have been bugged for a long time now. If you had them you had to go to the expose/hide option and hide your testicles at round start every round. Otherwise your testicles would by default be exposed when unclothed, and the genital description would describe an external pair of testicles that you shouldn't have.
This is good because it'll take a minor inconvenience off the board
## Changelog

:cl: Deek-Za
fix: fixes Internal testicles

### DIFF
--- a/modular_skyrat/modules/customization/modules/surgery/organs/genitals.dm
+++ b/modular_skyrat/modules/customization/modules/surgery/organs/genitals.dm
@@ -159,7 +159,10 @@
 	icon_state = passed_string
 
 /obj/item/organ/genital/testicles/get_description_string(datum/sprite_accessory/genital/gas)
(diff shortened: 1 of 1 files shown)


Gandalf2k15:
You don't need skyrat comments for a skryat file

Deek-Za:
> You don't need skyrat comments for a skryat file

Gotcha.  I was under the impression this originally got developed by cit. I'll tweak it.

PR #8625, opened 2021-10-05
[NON-MODULAR] Removes cargo exports elasticity, unleashes creativity

Lauwma:
## About The Pull Request

This PR disables the cargo export price elasticity by changing the apply_elastic variable. The cargo export price elasticity was a modifier on the export price that made the price of exports go down the more you exported them. 

While it was a good way to prevent cargo from making a million by themselves through miasma collecting (stacking a lot of monkey corpses in a single tile), easy miasma production is now gone and so are other exploits. 

Price elasticity harmed other means of money making and potential antagonism/roleplay opportunities. I'll take an exemple that happened in a recent round, in blueshift. 
I was an antag and basically the gimmick was to make an ungodly amount of cocaine and smuggling it in the shuttle. I made my factory, sold one crate to make a bit of start money to hire some assistants that needed to pay their medical bills (medical was charging money for treatments) and got around 5k for a full crate of bricks. Then I sent another one because I had some spare, almost got caught by the custom and gained 272 credits for the same amount of bricks.
Producing cocaine on an industrial level lost all its point since it wasn't worthy enough and I ended up not hiring anyone because I didn't needed to, I didn't sell anymore cocaine so the custom agent or security never found out and basically the absurd price elasticity harmed a whole antag gimmick .

This change would also help botanists, xenobiologist, space explorers, even security (prisoner plates are affected by price elasticity), make money by ways that include roleplay. 
Custom agent now has to look out for things that are being exported, else a lot of money will be made illegaly. With more money into play, departments will also be able to charge for things. Do note that this change doesn't remove the utility of bounties. In fact most crewmembers don't need a lot of money so they won't invest in making a lot of goods to export, they will just do bounties instead.
In conclusion that mechanic change will push for more interactions, more conflicts, enable some kind of antagonism and make the cargo job less boring. 

This change also need live testing, if there's new op ways to make money safely and easily I'll be happy to nerf them by hardcoding a low price to keep the other things balanced.

## Changelog

:cl:

balance: Nanotrascen economist specialist deduced that a single station couldn't overflow a market by themselves, the export price elasticity has now been removed.

/:cl:

### DIFF
--- a/code/modules/cargo/exports.dm
+++ b/code/modules/cargo/exports.dm
@@ -27,7 +27,7 @@ Then the player gets the profit from selling his own wasted time.
 	var/list/exported_atoms_ref = list() //if they're not deleted they go in here for use.
 
 // external_report works as "transaction" object, pass same one in if you're doing more than one export in single go
-/proc/export_item_and_contents(atom/movable/AM, apply_elastic = FALSE, delete_unsold = TRUE, dry_run=FALSE, datum/export_report/external_report) //skyrat edit
+/proc/export_item_and_contents(atom/movable/AM, apply_elastic = FALSE, delete_unsold = TRUE, dry_run=FALSE, datum/export_report/external_report) //SKYRAT EDIT
 	if(!GLOB.exports_list.len)
 		setupExports()
 
--- a/code/modules/cargo/exports.dm
+++ b/code/modules/cargo/exports.dm
@@ -1,3 +1,5 @@
+//THIS FILE HAS BEEN EDITED BY SKYRAT EDIT
+
 /* How it works:
 The shuttle arrives at CentCom dock and calls sell(), which recursively loops through all the shuttle contents that are unanchored.
 
--- a/code/modules/cargo/exports.dm
+++ b/code/modules/cargo/exports.dm
@@ -27,7 +27,7 @@ Then the player gets the profit from selling his own wasted time.
 	var/list/exported_atoms_ref = list() //if they're not deleted they go in here for use.
 
 // external_report works as "transaction" object, pass same one in if you're doing more than one export in single go
-/proc/export_item_and_contents(atom/movable/AM, apply_elastic = FALSE, delete_unsold = TRUE, dry_run=FALSE, datum/export_report/external_report)
+/proc/export_item_and_contents(atom/movable/AM, apply_elastic = FALSE, delete_unsold = TRUE, dry_run=FALSE, datum/export_report/external_report) //skyrat edit
 	if(!GLOB.exports_list.len)
 		setupExports()
 
@@ -105,7 +105,7 @@ Then the player gets the profit from selling his own wasted time.
 		cost = init_cost
 
 // Checks the cost. 0 cost items are skipped in export.
-/datum/export/proc/get_cost(obj/O, apply_elastic = FALSE)
+/datum/export/proc/get_cost(obj/O, apply_elastic = FALSE) //SKYRAT EDIT
 	var/amount = get_amount(O)
 	if(apply_elastic)
 		if(k_elasticity!=0)
@@ -121,7 +121,7 @@ Then the player gets the profit from selling his own wasted time.
 	return 1
 
 // Checks if the item is fit for export datum.
-/datum/export/proc/applies_to(obj/O, apply_elastic = FALSE)
+/datum/export/proc/applies_to(obj/O, apply_elastic = FALSE) //SKYRAT EDIT
 	if(!is_type_in_typecache(O, export_types))
 		return FALSE
 	if(include_subtypes && is_type_in_typecache(O, exclude_types))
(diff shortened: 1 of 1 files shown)


Lauwma:
UPSTREAM: Might be able to make apply_elastic a proper variable that can be changed in a config maybe, I dunno I'm not coder

YakumoChen:
👎 
This system is in place specifically to prevent cargo exploits. I've seen cargo make essentially "coke conveyors" that automate the process. Systems like these are harmful for gameplay.

Giving the means for easy cargo credits outside of bounty cubes (the intended way for cargo to make dosh) doesn't increase "roleplay value" for cargo, it just gives them silly amounts of funds.

Lauwma:
> 👎 
> 
> This system is in place specifically to prevent cargo exploits. I've seen cargo make essentially "coke conveyors" that automate the process. Systems like these are harmful for gameplay.

Most of the cargo exploits has been patched by upstream, miasma, crate dupes with the advanced cargo console.
Cargo is not supposed to make "coke conveyors" openly, drug distribution (exporting a distribution) is forbidden by the space law. If a non antag cargo starts making narcotic it's more a ruling and policy issue than a code or mechanic one. 

> 
> 
> 
> Giving the means for easy cargo credits outside of bounty cubes (the intended way for cargo to make dosh) doesn't increase "roleplay value" for cargo, it just gives them silly amounts of funds.

Letting the opportunity for antagonists or other department like botany, science and engineering to export adds roleplay, I just gave an example on how it can be done as antagonist, it does rely on cocaine exports but it would of gave rp and conflicts possibilities.

If they gain too much money with a method then I'll be glad to hardcode a price for thoses items. If they gain too much money it's not because they value doesn't decrease but because the base value is too high

SpaceVampire:
This is just going to create a couple of objectively best ways to make money, and your plan of just "lower a hardcoded price" sounds like it will involve playing wack-a-mole with whatever the method of the week is. The system exists so that people don't have to go around doing it, and while you could maybe talk about tweaking the scaling behind it I don't think this is the best way to handle it. Would just be a massive buff to cargo at the moment.

ChungusGamer666:
This will create a ton of cargo exploits and automation. 
This is in no way good for roleplay, or for balance.

Lauwma:
> This is just going to create a couple of objectively best ways to make money, and your plan of just "lower a hardcoded price" sounds like it will involve playing wack-a-mole with whatever the method of the week is. The system exists so that people don't have to go around doing it.

If you look at the export files you can see that most items have hard coded prices in order to prevent abuses and also have consistency. The current system is already wack-a-mole because most cargo changes upstream in the past 6 months were money making nerfs (miasma and gas in general), express crates dupes, reinforced glass resell nerf.
If this system is intended so nobody has to do it I think it's failing its purpose. 

>and while you could maybe talk about tweaking the scaling behind it I don't think this is the best way to handle it. Would just be a massive buff to cargo at the moment.

Another alternative would be to hardcode export prices of items so they doesn't get affected by the plasticity but I feel like it would be even more work and more troubles than just removing or lowering plasticity.

This change shouldn't affect cargo much, in fact cargo shouldn't make factories on their own, botanists, engineers and scientists should. Cargo would only get a 50% share of that when labelled with the price tag. Even if it gives cargo money it's still is fair money since they are working for it. As an antagonist I don't see the point of making cocaine bricks if they end up being worth 0.5 credits for a 4 second craft. 

This change open opportunities for interactions between fields instead of restricting money making to just bounties.
Bounties are make, throw on the cargo desk/disposals and aren't equal in every fields. 

You have to take the setting the server is in when you want to make economy changes. This change don't and shouldn't promote no rp money dump in cargo.

GoldenAlpharex:
I'd much rather have the elasticity itself being tweaked, so that making a factory can still pay out, while incentivizing people to not just make one factory and roll in money for the rest of the shift.

SpaceVampire:
> You have to take the setting the server is in when you want to make economy changes. This change don't and shouldn't promote no rp money dump in cargo.

I feel bad only replying to this part of your argument. But they just. Literally do. There is no way this doesn't end up with Cargo Techs just making a literal money printing machine each shift. If it just works and they don't ever have to readjust to it. There is going to be some kind of new meta, and people are just going to do that. You already see people talking about using the whole Glassblowing or whatever for the reason, as the change to make them sell for a fixed amount was the main appeal of it.

Lauwma:
Alright after taking all the comments into account I think what I'll do is tweak how the price is calculated. Instead of being able to go down all the way to 0.5 credits I'll make so the price cannot decrease under a percentage of the initial price of the item. It will obey this function (not code, just mathematics)

ExportPrice(x,n) = x/2 + ((x/2)/((n+k)/k))

x being the initial price (variable already in the code),
n being the number of this item that has been sold (variable already in the code),
k being how high the number needs to be (not in the code, could be calculated with x as in the more valuable the item is, the more quickly it's price drops)

I'll make a graph when I'll be home but until I code it I'll put this as a draft unless maintainers are willing to try this change as it is

Lauwma:
image

So for x = 200, n being in |R+, k = 30, we get that graph. Sadly, since I'm on phone I can only show you a poor representation of it.

f(200, n) > 100, the price drops to 75% of it's original price after 30 items. After 240 item sold the price doesn't change much and is near the 100 credits.
Note that it's and exemple and we can change a lot of parameters, also I used an inverse function but I can  also use others function.

GoldenAlpharex:
> Alright after taking all the comments into account I think what I'll do is tweak how the price is calculated. Instead of being able to go down all the way to 0.5 credits I'll make so the price cannot decrease under a percentage of the initial price of the item. It will obey this function (not code, just mathematics)
> 
> ExportPrice(x,n) = x/2 + ((x/2)/((n+k)/k))
> 
> x being the initial price (variable already in the code), n being the number of this item that has been sold (variable already in the code), k being how high the number needs to be (not in the code, could be calculated with x as in the more valuable the item is, the more quickly it's price drops)
> 
> I'll make a graph when I'll be home but until I code it I'll put this as a draft unless maintainers are willing to try this change as it is

That's already a much better way of doing it, yeah.

Lauwma:
Closing until done

PR #8648, opened 2021-10-06
Changes to GBP to incentivize bugfixes

GoldenAlpharex:
## About The Pull Request
~Let's penalize people less for balance PRs and give a bit more incentive to QoL PRs again.~

Okay, change of plan. We're going to try and make GBP more incentivized instead. Yes. That means you lose even more GBP for additions.

Why?
All the additions add more strain on the maintainer team, and people clearly don't put enough work into fixing their bugs afterwards, and we're the ones that end up having to do it.

We might revert this later, but right now, I think this could be for the best. We want to have people fix stuff for us too, not just create more bugs with no fixes.

More GBP on the coder's GBP on a stick makes them code faster, or so they say.

Especially when they have less incentive to work on features and more incentive on actually either expanding existing content, or fixing bugs with said content.

## Changelog

 Not player-facing ;)


Funce (approved):
This is fine.

ZephyrTFA:
You should also enforce GBP and not allow people without sufficient GBP to make PRs if you want people to actually care about GBP

gonenoculer5:
The fuck is GBP

Gandalf2k15:
# Time to enforce GBP.

Issue #8424, opened 2021-09-27
Ash Walker / Primal Podperson no longer selectable during Character Creation


Lyroy:
## Round ID:
3132

## Testmerges:
#8350 

## Reproduction:
Open species menu.
There are no ashies or poddies.

PR #8665, opened 2021-10-07
[SEMI-MODULAR] Brings back custom Ash Walkers and Primal Podpeople, also fixes job banned species checks

Lyroy:
## About The Pull Request
This re-adds the always_customizable species (Ash Walkers and Primal Podpeople) that were left out of the new preferences menu.
I had to fix the has_banned_species check from jobs and re-add it to the roundstart job distribution as well as latejoin job options to prevent them from just joining the station.
Also prevents them from learning any language not in their learnable_languages list. It's a bit of a hack but I'm expecting an upcoming backgrounds PR to change how that entire thing works.
Also gave ashies a description and some quirks for their race selection. Podpeople have a very barebones one.

This solves #8424.

The Ash Walker ghost role brings great opportunities for RP, and it's basically been abandoned since people weren't allowed to play as their characters there.
Oh yeah and Podpeople too but nobody plays that hahah...

## Changelog

:cl:
fix: fixes the has_banned_species proc for jobs
fix: The recent strong storm keeping the Ash Walkers trapped and isolated has finally ended. Miners should be cautious, as the Lavaland natives might start showing up again.
fix: Primal Podpeople have finally managed to exit their cryopods again.
/:cl:


modular_skyrat/master_files/code/modules/client/preferences/middleware/languages.dm
```diff
@@ -71,20 +71,20 @@
...
 			preferences.languages[language] = LANGUAGE_SPOKEN
-		qdel(species)
-
 	var/list/selected_languages = list()
 	var/list/unselected_languages = list()
 	for (var/language_name in GLOB.all_languages)
 		var/datum/language/language = GLOB.language_datum_instances[language_name]
 		if(language.secret)
 			continue
+		if(species.always_customizable && !(language.type in species.learnable_languages)) //For the ghostrole species. We don't want ashwalkers speaking beachtongue now.
```
Iamgoofball, line 86: Please make sure this doesn't apply to any of the station species.

Lyroy: I made sure, it doesn't. The only two species with always_customizable = TRUE are ashwalkers and (primal) podpeople.
No idea why we don't just rename it to ghost_role_species really.

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/species/ashlizard.ts
```diff
@@ -0,0 +1,57 @@
+import { Species } from "./base";
+
+const Ashwalker: Species = {
+  description: "The primitive natives of Indecipheres, this race, \
```
gonenoculer5, line 4: Is this the canon name for Lavaland?

Lyroy: It's the most canon one we have afaik, at least according to the trading card game

code/modules/client/preferences/species.dm
```diff
@@ -49,7 +54,7 @@
 
 	var/list/food_flags = FOOD_FLAGS
 
-	for (var/species_id in get_selectable_species())
+	for (var/species_id in (get_selectable_species()+get_customizable_races())) //SKYRAT EDIT CHANGE - ORIGINAL: for (var/species_id in get_selectable_species())
```
GoldenAlpharex: Just to make it more readable.
```suggestion
	for (var/species_id in (get_selectable_species() + get_customizable_races())) //SKYRAT EDIT CHANGE - ORIGINAL: for (var/species_id in get_selectable_species())
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species.dm
```diff
@@ -572,3 +572,29 @@ GLOBAL_LIST_EMPTY(customizable_races)
...
+
+	return GLOB.customizable_races
+
+/**
+ * Generates races available to choose in character setup at roundstart
+ *
+ * This proc generates which species are available to pick from in character setup.
+ * If there are no available roundstart species, defaults to human.
+ */
+/proc/generate_customizable_races()
```
GoldenAlpharex, line 590: I love documented procs, thanks a lot!

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/species/pod.ts
```diff
@@ -0,0 +1,17 @@
+import { Species } from "./base";
+
+const Pod: Species = {
+  description: "An ancient, bio-engineered species of sentient plants, \
+	created by a Precursor Civilization.",
+  features: {
+    good: [],
```
GoldenAlpharex: I feel like the perks of podpeople should be listed here. Also, I'd rename this whole thing to PrimalPodpeople just to make it easier and more straight-forward.

GoldenAlpharex (changes requested):
Looks like some very solid code that I'm happy I didn't have to do myself.

Two questions remain, however:
What does it look like in the Occupations menu when you can't select a job because your race doesn't allow for it?
Could you add a check so that ashwalkers are only used on Ash Walker spawners, and Pod People, on Pod People spawners? Thanks!

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/species/pod.ts
Lyroy: I'm not too familiar with podpeople perks but I'll try! Can't really rename it though, as the species selection page looks for a .ts file named exactly like the species datum and I'd have to change the datum itself as well.

Lyroy:
> What does it look like in the Occupations menu when you can't select a job because your race doesn't allow for it?

Right now it looks like normal, nothing has been changed there, so you're able to set your job priorities, it just won't let you qualify for a job on roundstart or latejoin one which I believe was how it worked with the old prefs menu. I could make each job check for banned species when you open the occupations tab, though.

> Could you add a check so that ashwalkers are only used on Ash Walker spawners, and Pod People, on Pod People spawners?

I don't quite understand this though. Do you mean like add a check when you use a prefs character on a species-exclusive spawner? This is already the case.

gonenoculer5:
> > What does it look like in the Occupations menu when you can't select a job because your race doesn't allow for it?
> 
> Right now it looks like normal, nothing has been changed there, so you're able to set your job priorities, it just won't let you qualify for a job on roundstart or latejoin one which I believe was how it worked with the old prefs menu. I could make each job check for banned species when you open the occupations tab, though.

Originally, the case was that it would grey out every bar on the occupations menu when you selected a non-qualifying race. So if you were an ashwalker for example, you would be unable to take any position on the roster because it was unselectable in the first place. Non-qualifying at roundstart would probably work too, but drawing a vingrete (I think thats the word) over the occupations menu, and saying "Unable to qualify for position due to ineligble species." or something might be a visually pleasant way to handle it.

Lyroy:
> > Right now it looks like normal, nothing has been changed there, so you're able to set your job priorities, it just won't let you qualify for a job on roundstart or latejoin one which I believe was how it worked with the old prefs menu. I could make each job check for banned species when you open the occupations tab, though.
> 
> Originally, the case was that it would grey out every bar on the occupations menu when you selected a non-qualifying race. So if you were an ashwalker for example, you would be unable to take any position on the roster because it was unselectable in the first place. Non-qualifying at roundstart would probably work too, but drawing a vingrete (I think thats the word) over the occupations menu, and saying "Unable to qualify for position due to ineligble species." or something might be a visually pleasant way to handle it.

A vignette isn't a bad idea but I think checking on a job by job basis and displaying a "Bad species" text instead of the priority selector will be better for future-proofing as opposed to just locking the entire screen, in case we ever end up adding species that can't take specific jobs.

I'll commit it later today.

gonenoculer5:
> > > Right now it looks like normal, nothing has been changed there, so you're able to set your job priorities, it just won't let you qualify for a job on roundstart or latejoin one which I believe was how it worked with the old prefs menu. I could make each job check for banned species when you open the occupations tab, though.
> > 
> > 
> > Originally, the case was that it would grey out every bar on the occupations menu when you selected a non-qualifying race. So if you were an ashwalker for example, you would be unable to take any position on the roster because it was unselectable in the first place. Non-qualifying at roundstart would probably work too, but drawing a vingrete (I think thats the word) over the occupations menu, and saying "Unable to qualify for position due to ineligble species." or something might be a visually pleasant way to handle it.
> 
> A vignette isn't a bad idea but I think checking on a job by job basis and displaying a "Bad species" text instead of the priority selector will be better for future-proofing as opposed to just locking the entire screen, in case we ever end up adding species that can't take specific jobs.
> 
> I'll commit it later today.

Sounds good!

GoldenAlpharex:
Perfect, if you're still working on this, I'll make this a draft.

GoldenAlpharex:
I'm going to test this in local tomorrow, if I'm satisfied with it as-is I'll undraft and merge.

Lyroy:
Here's some previews of how it looks right now.

The occupations tab
jobs menu

The join menu
join menu

And what it tells you when you join as an ashie
ash spawn
ash spawn 2

GoldenAlpharex:
Give me like an hour and I'm fixing the conflict and merging this bad boy.

GoldenAlpharex:
Unless you can do it yourself :)

Lyroy:
> Unless you can do it yourself :)

Done! Feel free to merge whenever.

PR #8666, opened 2021-10-07
removes the welding goggles from the loadout

thestubborn:
## About The Pull Request

title, also removed the real blindfold since it shares the same purpose, use the fake one if you want to wear a blindfold

you shouldn't be able to spawn with welding goggles _**as any role**_

## Changelog

:cl:
balance: you can no longer instantly weld as any job
/:cl:


gonenoculer5:
I dont see why the true blindfold should be removed. Welding goggles yes.

thestubborn:
> 
> 
> I dont see why the true blindfold should be removed. Welding goggles yes.

they work the same as welding goggles

GoldenAlpharex (approved):
That must've slipped through, yeah. This wasn't intended.

PR #8675, opened 2021-10-08
Disable Regained Tail Trauma

Deek-Za:
## About The Pull Request

Disables the Tail Trauma event

_My tail is back, but that was traumatic..._

The regained your tail trauma moodlet debuff fires off on **Spawn**(Both roundstart and latejoin) if you have any sort of tail due to changes with how we handle the tails on SR. This negatively impacts the beginning of play for sizeable chunk of people due to it being a furry server. It's also rarely used in a legitimate way during any given round. 

Proposed change until a proper fix is issued since it makes a rougher round start for so many people and this has been an issue for several months.

## Changelog

:cl: Deek-Za
del: Disabled Tail Trauma
qol: Less troubling start for tailed individuals.
/:cl:


GoldenAlpharex (approved):
Code checks out, usually I'd recommend finding a solution but honestly nobody ever gets their tail removed on Skyrat so this is pretty much dead code anyway.

PR #8695, opened 2021-10-08
[MODULAR] Prefs Menu Refactor: Fine, I'll Do It Myself Edition

GoldenAlpharex:
## About The Pull Request
thanos-ill-do-it

Moderate refactor/bugfix PR for the Prefs Menu.

### _What's new?_

- Your first page is now taking more space, which means that you can type in it better and in general see more options at once!
- **Expensive Augments are back!** - That means, augments that used to cost quirk points will cost quirk points again. But that also means that you'll be getting quirk points for taking prosthetics! This took a ton of time and required some quirk backend minor refactoring.
- Augments that cost quirk points will actually display how many points they cost.
- When an augment can't be selected because it would mess up your balance, you'll be able to click it, but it won't actually change the augment at all (blame TGUI limitations on this one, otherwise they'd be greyed out).
- The Augments And Markings tab (or whatever name it was) was renamed to Augments+, so it fits in the button.
- The Augments+ tab now is using scrollable sections, so there's no more visible bug with the character preview looking out of place.
- **Augment styles will now load properly!** No need to change your augments' styles every time you'd be changing character or joining the server!
- **Quirk points are fixed!** (And by that, I mean, you can remove the breathing tube that you couldn't remove before and that would break your quirks if you had it!)
- The quirks balance is now calculated outside of the interface. What that means is that more functionalities could tie in and modify the balance without messing it up in the menu (again)!
- The Other Gender button now works again! (I still don't understand why that icon wasn't working, but honestly this one is a lot clearer to distinguish, I guess)

A better prefs menu makes for better character customization and a better user experience.

Also look at these screenshots:
*Main page*
image

*New look of the Augments+ page*
image

## Changelog

:cl: GoldenAlpharex
fix: Quirks will no longer be in a weird state where they reset and you don't know why anymore. Check your Augments+ tab if that was happening to you, and remove the augments you don't want.
fix: The Limbs Augmentations And Markings page was relabeled "Augments+" and has been revamped visually to work better.
fix: You will no longer wake up on the Interlink with your cybernetics looking wrong because you didn't ensure they were set right properly during your sleep (in the prefs menu).
fix: The breathing tube is now accessible again in the Augments+ tab.
fix: The 'Other' gender is now no longer afraid of itself and will now properly render, allowing you to change gender if you'd like.
code: Quirk balance tallying up is now done outside of the TGUI interface, allowing us to hook other systems that cost points into it.
refactor: Refactored the Augmentations panel to actually look decent and support the usage of quirk points to balance stronger or crippling augments.
/:cl:


Iamgoofball:
:+1:

fixes all the shit I couldn't figure out how to fix

tgui/packages/tgui/interfaces/PreferencesMenu/MainPage.tsx
```diff
@@ -13,7 +13,7 @@ import { filterMap, sortBy } from "common/collections";
 import { useRandomToggleState } from "./useRandomToggleState";
 
 const CLOTHING_CELL_SIZE = 48;
-const CLOTHING_SIDEBAR_ROWS = 9;
+const CLOTHING_SIDEBAR_ROWS = 13.4; // SKYRAT EDIT (Makes the main window prettier) - ORIGINAL: const CLOTHING_SIDEBAR_ROWS = 9;
```
Gandalf2k15:
```suggestion
const CLOTHING_SIDEBAR_ROWS = 13.4; // SKYRAT EDIT CHANGE - ORIGINAL:  9
```

Gandalf2k15 (approved):
LGTM, merge once CI passes

tgui/packages/tgui/interfaces/PreferencesMenu/MainPage.tsx
GoldenAlpharex: Oh yeah I don't know how I missed that.

PR #8699, opened 2021-10-09
Fixing the Sergeant's Hat.

Deek-Za:
Fixed the Sergeant's Hat

## About The Pull Request

Sergeant's hat got messed up. Won't show on certain anthro races and also no armor values, making it worse to wear then pretty much any sec hat you can buy out of the vendor.
Fixed visibility on certain races
Added armor values matching a peacekeeper beret.

It makes things work properly for folk!

## Changelog

:cl: Deek-Za
balance: Sergeant's hat has armor added.
fix: Sergeant's hat is now visible on anthro races.
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/security_sergeant/secseg_clothes.dm
+++ b/modular_skyrat/modules/sec_haul/code/security_sergeant/secseg_clothes.dm
@@ -4,6 +4,8 @@
 	icon = 'modular_skyrat/master_files/icons/obj/clothing/hats.dmi'
 	worn_icon = 'modular_skyrat/master_files/icons/mob/clothing/head.dmi'
 	icon_state = "peacekeeper_sergeant_cap"
+	mutant_variants = NONE
+	armor = list(MELEE = 40, BULLET = 30, LASER = 30,ENERGY = 40, BOMB = 25, BIO = 0, RAD = 0, FIRE = 20, ACID = 50, WOUND = 0)
 
 /obj/item/clothing/under/rank/security/peacekeeper/sergeant
 	name = "peacekeeper sergeant uniform"


GoldenAlpharex:
Could that issue actually be rooted deeper than this? Surely it must come from one of its children not inheriting of the property properly.

Deek-Za:
> Could that issue actually be rooted deeper than this? Surely it must come from one of its children not inheriting of the property properly.

Very possible. You probably got more experience so you can take a gander if you want. But it seems like a lot of sec stuff is linked up in various ways, but this wasn't, maybe since it is unique in terms of hats(There was definitely a dedicated section for flagging stuff that's not affected by anthro status, but it only seemed set for old sec gear in the non skyrat section of code). Ultimately I went the route of whoever setup the secmed's hat went with (The _Mutant_variant=_ code). Then of course altered the armor since that was definitely non-existant in the code.

Update: Someone else looking into the mail code apparently ran into the same issue with Sergeant and secmedic and stuff. Seems this might be similar across most added jobs. So for the sake of code consistency I'd go with how it's done for the other _added_ jobs for now as it's the same across all peacekeeper stuff(Checked, PK Officer, Secmed, junior officer and they all use the same method.) Seems the sergeant's hat just got overlooked.

SpaceVampire:
It feels like it would be better if you just updated the paths to make sense.
It should really get armor from being /obj/item/clothing/head/beret/sec/ at the root.
But the person who did this did /obj/item/clothing/head/sec/peacekeeper/sergeant instead.
Where the correct one would have been /obj/item/clothing/head/beret/sec/peacekeeper/sergeant.
Given we have /obj/item/clothing/head/beret/sec/peacekeeper in the code already.
I'm a bit tired when typing this, and stuck on my laptop but uh.
Hopefully I got all of those paths correctly.
It does feel better than just adding the armor as a patch.
And yeah, this is an issue for some of our other modular stuff as well.

Deek-Za:
@SpaceVampire 
I'm sure it's possible to be able to change everything. The issue is this is how it's rigged up on **all** the Peacekeeper jobs equipment because of yeah, the modular thing. My setup keeps the code in-line with the other PK jobs. Which is why I chose it for code consistency sake(Which the coding guide recommends). Re-addressing the armor value(And subsequently the sergeant's hats invisibility) can possibly be handled by changing the pathing. Just... That would require every piece of PK equipment to be addressed, since they were all done this way, instead of just the one hat that's suffering issues for the moment. 
I think right now, imo, it'd be better to get it functioning properly and like the rest of the PK gear now, and then later tackle the more time consuming process of changing around everything PK related. Possibly with whoever handled the original setup.

Also in retrospect I think maybe why didn't do that pathing like you mentioned is because they might not have ever intended to give it the stats from the beret. Considering the sec sergeant's armor has an increased armor value from the PK officer's uniform (1 additional melee and bullet). 
I just went with the beret's(The hat is *not* a beret, it's a Drill sergeants hat) stats because I don't know what the intended values were since they weren't added. Either way I'll look into other options

Deek-Za:
@GoldenAlpharex So I'm just gonna do a quick summary since you're the reviewer and had concerns. After double and triple checking, there isn't any source that this could inherit any armor values or the mutant variant options from. The only other cap going down this line is the solfed cap(Non-hos version) which was put in for aesthetic and doesn't have armor either.
I considered the other's request about repathing to berets, but they have a more complex setup apparently and it causes a bunch more issues without a lot more tweaking. Plus the hat isn't a beret to begin with, I just was gonna mimic their armor values.
Which brings me to three real options.
1) I leave the code the way I did it. It gives beret level armor(I can buff this value for the sergeant if wished) and allows non-human races to wear the hat properly once again.

2) I do re-path to something else. I considered the option of making it like a helmet, which would bring it more to code of.
image
This would effectively grant it beret level armor but with a bump in fire durability from II to V.

3) I let it sit at the state it is now (Un-armored, invisible for non-humans), I close this PR, and we address it again at a later date.

Just want your opinion on it when ya can.
image

GoldenAlpharex:
> @GoldenAlpharex So I'm just gonna do a quick summary since you're the reviewer and had concerns. After double and triple checking, there isn't any source that this could inherit any armor values or the mutant variant options from. The only other cap going down this line is the solfed cap(Non-hos version) which was put in for aesthetic and doesn't have armor either. I considered the other's request about repathing to berets, but they have a more complex setup apparently and it causes a bunch more issues without a lot more tweaking. Plus the hat isn't a beret to begin with, I just was gonna mimic their armor values. Which brings me to three real options.
> 
> 1. I leave the code the way I did it. It gives beret level armor(I can buff this value for the sergeant if wished) and allows non-human races to wear the hat properly once again.
> 2. I do re-path to something else. I considered the option of making it like a helmet, which would bring it more to code of.
>    image
>    This would effectively grant it beret level armor but with a bump in fire durability from II to V.
> 3. I let it sit at the state it is now (Un-armored, invisible for non-humans), I close this PR, and we address it again at a later date.
> 
> Just want your opinion on it when ya can. image

That sounds reasonable, thank you for doing some looking around. We'll need a refactor of those paths later on, aye.

PR #8702, opened 2021-10-09
Fixes security mail and RPG job titles

sandshark808:
## About The Pull Request

I noticed the new security jobs had no mail goodies, which meant they were drawing from the basic random pool instead. This PR brings them in-line with other jobs (including the new goofsec jobs and blueshield, which already had mail goodies lists).

Security Medic gets the exact same goodies as a Paramedic, since I feel their roles are very much intended to be the same.

Security Sergeant gets the same goodies as a Security Officer with a small chance at a single flashbang.

Corrections Officer gets the same goodies as a Security Officer with a small chance at getting prison contraband. The chance is higher than Prisoners get. It's more of a flavor choice really but they can potentially trade it with prisoners or keep it in their office and give a tantalizing prize for breakins.

I also added RPG titles to each job even though I don't think we even have the event that makes those show up. In case someone ever does it, it'll be there instead of just being blank.

Everyone loves getting mail, and jobs in departments should be drawing from a department-related mail pool instead of giving out the default pool. Ultimately by including the RPG titles in this it's to push a fix/balance change to get these jobs to adhere to the same template as every other job.

tl;dr: missing features bad

## Changelog

:cl:
fix: Security Medic, Security Sergeant, and Corrections Officer now get mail goodies.
fix: Security Medic, Security Sergeant, and Corrections Officer now have RPG titles for use in a special event.
/:cl:


GoldenAlpharex (approved):
Wicked. Thanks!

PR #8742, opened 2021-10-10
[modular] stops tgui from migrating prefs when it shouldn't be and thus causing bad things when making new characters

Seris02:
## About The Pull Request

title

it doesn't

## Changelog

:cl:
fix: stops tgui from migrating prefs when it shouldn't be
/:cl:


Funce (approved):

Seris02:
By the way, before this gets merged, can someone backup everyone's prefs, again

I mean like, it shouldn't go wrong. But byond is weird.

PR #8745, opened 2021-10-10
[MOSTLY MODULAR] Fixes Hypovial Texture

softcerv:
## About The Pull Request
Repaths hypovials to not be a child of bottles, making it so that the bottle overlay does not apply to them. From what I've tested, this shouldn't have any effect on how hypovials actually function. Gives hyposprays fill_icon_thresholds, allowing them to add textures.

The end result is that hypovials properly display their overlays instead of the bottle overlay. 
image

Glitched textures are unrealistic and LRP

## Changelog

:cl:
fix: Hypovials now properly display their intended overlay texture, instead of the chemical overlay textures from bottles.
/:cl:


GoldenAlpharex (approved):
Finally!

PR #8759, opened 2021-10-11
Get Features cache fix

Azarak:
## About The Pull Request
The cache getter was missing from the proc.
It's hilarious how this proc and many other are needlessly overused btw, topic being a quarter of the processing is epic

obraz


GoldenAlpharex (approved):
That really makes you think.

PR #8761, opened 2021-10-11
better master_erp_preferences check

Seris02:
## About The Pull Request

title

it doesn't

## Changelog

no visible difference


ORCACommander:
Validated. Looks like everything is in place for the server toggle now except a certain emote. I'll probly just keep that emote commented out instead of something more dynamic and needing upstream maint

ORCACommander:
thanks @Gandalf2k15

PR #8787, opened 2021-10-13
[MODULAR] Disables a certain emote when a Server Toggle is thrown

ORCACommander:
## About The Pull Request

Its a rough and dirty way to disable it but it hooks into the server wide toggle for exotic content. Ideally this would be eliminating it completely from the keybind menu but that is beyond my skill and have been told it is very difficult and would require /tg/ core file edits

## Changelog

:cl:
code: hooked a certain emote to the server wide disable
admin: Non-ERP enabled server admins may see an increase in boinks
/:cl:


GoldenAlpharex (approved):

ORCACommander:
better system was implemented

PR #8788, opened 2021-10-13
[Semi-modular] Silicon flavor texts

NotRanged:
## About The Pull Request
Re-implements silicon flavor texts. Tested and working, only issue is I can't figure out how to get the icon of a cyborg show up in the examine, so it's just a black screen right now.
image

fun cat fact this is the second time I've implemented silicon flavor texts on this codebase

because this reuses the old key 'silicon_flavor_text' this should grab the flavor texts you had before the prefs change but idk for sure since I have no way to actually test that

## Changelog

:cl:
add: Re-implemented silicon flavor texts!
/:cl:


GoldenAlpharex:
I actually made some verifications on the server, I'm pretty sure the silicon flavor text is still there in its integrity, yeah.

GoldenAlpharex:
I love how that picture is telling a story as well, hahahah.

NotRanged:
NOT THE BEES
**_AAAAAAAAAAAAAAAAA_**

Seris02:
"Burn with me."

ORCACommander:
Oh boy one of red's children has discovered video games....

code/modules/mob/living/carbon/human/examine_tgui.dm
```diff
@@ -54,7 +72,7 @@
 		ooc_notes += "Vore: [e_prefs_v]\n"
 		ooc_notes += "\n"
 
-	ooc_notes += holder.dna.features["ooc_notes"]
+	ooc_notes += preferences.read_preference(/datum/preference/text/ooc_notes)
```
GoldenAlpharex: Will that read the person's OOC preferences or that of the body? Because we want that of the body, so that if a changeling impersonates someone, people can't easily tell "Hey it's not the same person because they don't have the same OOC notes!"

GoldenAlpharex (changes requested):
Just one thing I want to make sure of first.

code/modules/mob/living/carbon/human/examine_tgui.dm
NotRanged: it reads the client here, I'll change it to be contained within the ishuman/issilicon so it doesn't

GoldenAlpharex (approved):
Looks good to me. :+1:

PR #8790, opened 2021-10-13
conveyor sorter

jjpark-kb:
## About The Pull Request
adds a tool that can set down a marker, which will allow people to set certain items to be forced moved into a direction
image

cargo qol, a means of automatically sorting things would be a life saver

## Changelog

:cl:
add: added the conveyor sorter, which can force move things to a direction if in the sorting list
/:cl:


gonenoculer5:
we playin factorio now

PR #8816, opened 2021-10-14
[MIRROR] Buffs eye contact, adds "Shifty Eyes" quirk [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61711
--------------------

## About The Pull Request
So back in #52108 I added eye contact, where looking at someone within one second of them looking at you (assuming you were within 5 tiles and didn't have hidden faces) would give you both a little message about making eye contact. I think it's a really little neat feature, but I basically never see it proc because of how short the time window is for it.

This PR extends the time window for making eye contact to 2 seconds, as well as adding a new trait called "Shifty Eyes" that rarely (10% - 1% for each tile of distance between the person and you) makes someone who looks at you think you both made eye contact. 

Codewise, this PR decouples examine_more and eye contact sharing the same time window. Atoms are removed from recent_examines after 2 second rather than 1 though behavior-wise, examine_more still works the same, examining multiple times in one second still only makes one check for eye contact.

## Why It's Good For The Game
The few times I've seen this happen in game or heard players talk about having it happen, they were caught off guard and thought it was cool in an immersion/paranoia sense. The numbers here might need some tweaking to find a good balance, but I think a moderate buff to how often it happens would be positive.

## Changelog

:cl: Ryll/Shaps
add: Added new quirk, "Shifty Eyes", which rarely causes people who examine you while standing closeby to think that you both made eye contact, even when you didn't.
balance: The time window for making eye contact with someone else has been extended from  examining each other within 1 second to within 2 seconds.
/:cl:


YakumoChen:
Is it really only one second? it feels more like 5 on skyrat.

PR #8821, opened 2021-10-14
fishing changes

jjpark-kb:
## About The Pull Request
ups the chance to get the fishing skill chip
adds a chance to get trash now from fishing
fishing rods require 2 hands now
fixed the primitive rod recipe

making it less crazy and fixing issues

## Changelog

:cl:
fix: primitive fishing rod recipe is fixed
balance: fishing rods require 2 hands now
balance: easier to get the fishing skill chip (higher chance while fishing)
balance: trash is now within the pool to get loot from for fishing
balance: fishing will yield less strange rocks
/:cl:


GoldenAlpharex:
You should also remove at least the strange stones from the lava fishing loot, otherwise it renders Xenoarcheology completely useless, since you can just fish an infinite amount of them anyway, with much less danger.

modular_skyrat/modules/fishing/code/fishing.dm
```diff
@@ -74,13 +75,17 @@ GLOBAL_LIST_INIT(fishing_weights, list(
...
+	if(prob(25))
+		spawning_reward = pick_weight(GLOB.trash_loot)
+		while(islist(spawning_reward))
+			spawning_reward = pick_weight(spawning_reward)
 	new spawning_reward(spawning_turf)
 	atom_parent.visible_message(span_notice("Something flys out of [atom_parent]!"))
-	if(prob(1))
-		new /obj/item/skillchip/fishing_master(spawning_turf)
+	if(generate_fish && prob(70))
+		generate_fish(spawning_turf, random_fish_type())
```
GoldenAlpharex: These probabilities could use being defines (that get undefined at the end of the file to avoid cluttering the global environment) so these values are easier to find and easier to modify in the future.

Alternatively, they could also be attributes of the fishing rod itself (which could then send those values to the component upon having it added to said fishing rod, so it'd be attributes of the fishing component as well), so the fishing rod could influence the odds you have of fishing one thing or another. I think that could be the best course of action, but it would require some more refactoring of how this works.

This would also allow more diversity in fishing rods and possibilities to make them even more unique and with a niche of their own in the future. For instance, I think regular fishing rods shouldn't have a chance to get stuff like ores from just fishing, but a lava fishing rod would make a bit more sense for that.

GoldenAlpharex: Also, this should probably be a big switch statement so you don't get a fish, a fishing_master skillchip and also a special ore reward at once. You should only get two rewards when you have the fishing_master skillchip and that's it.

modular_skyrat/modules/fishing/code/fishing.dm
```diff
@@ -1,12 +1,13 @@
...
-	/obj/item/stack/ore/silver = 1,
-	/obj/item/stack/ore/iron = 1,
-	/obj/item/xenoarch/strange_rock = 10,
+	/obj/item/stack/ore/gold = 2,
+	/obj/item/stack/ore/uranium = 2,
+	/obj/item/stack/ore/titanium = 3,
+	/obj/item/stack/ore/silver = 3,
+	/obj/item/stack/ore/iron = 5,
+	/obj/item/stack/ore/glass = 5,
+	/obj/item/xenoarch/strange_rock = 1,
```
GoldenAlpharex, line 10: I'm still unsure about strange rocks being obtained from fishing.

GoldenAlpharex (dismissed):

modular_skyrat/modules/fishing/code/fishing.dm
jjpark-kb: the components aren't added to fishing rods, but to the sources themselves.

modular_skyrat/modules/fishing/code/fishing.dm
jjpark-kb: lets just keep it in for now, it should be far less occurring.

modular_skyrat/modules/fishing/code/fishing.dm
jjpark-kb: I will do a switch, sure

jjpark-kb: At a later PR, I can refactor again to make the rods more exclusive with what they will fish.

GoldenAlpharex: This was still not made a switch statement.

Funce (approved):

PR #8842, opened 2021-10-15
First attempt at stopping tgchat from eating messages

GoldenAlpharex:
## About The Pull Request
After discussing with stylemistake for a while, we theorized that the ping might be responsible for the chat eating messages (basically failing to display them). Therefore, in this pull request, I'm disabling the ping feature from tgchat, to try and see if it helps.

This is a Test-Merge Only PR for now, as if this ends up being a viable solution, I'll work with stylemistake to properly do it upstream.

Being able to see someone else's 12-lines emote because there's a character limit on Runechat is pretty neat, I don't know about you but I'd say this makes roleplay less annoying.

## Changelog

:cl: GoldenAlpharex
fix: Disabled tgchat's ping feature in an attempt to stop it from failing to display certain messages.
/:cl:


Aphast:
best way to test this would be to have 2 bots on 2 servers saying 1-200 then checking which ones miss the most numbers, otherwise this may be hard to check

YakumoChen:
Its been working well so far for me, normally I have a lot of messages eaten and I can't say I've noticed any of them going missing right now.

chefferz:
Dunno if this matters at all but ping still shows in who.

YakumoChen:
the status ping is a different metric anyway

YakumoChen:
Something is still making it eat messages but it only happens now at high-pop with high lag.

Gandalf2k15:
fine

PR #8898, opened 2021-10-17
Minor change/fix to Hollow Bone drink ingredients

sandshark808:
## About The Pull Request

I was asked on discord to change the recipe of this so it didn't sometimes break automation of Bone Hurting Juice in fermichem. Basically the issue was that BHJ requires milk, and BHJ + Milk is Hollow Bone, meaning if you tried to make too much of it it would turn into the liquor and not the chemical you wanted.

This is a change to make a greater mechanical separation between a chem recipe and a fluff drink recipe to prevent ruining factory chems by mistake.

My reasoning for changing milk to soymilk is because the drink causes 30 disgust to skeletons, meaning they think it tastes awful. I figure soymilk is a bad imitation of the real thing, and even though the drink is no longer toxic it's nasty.

## Changelog

:cl:
qol: Changed Hollow Bone to use Soy Milk instead of Milk, making it harder to accidentally ruin your batch of Bone Hurting Juice.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/food_and_drinks/recipes/drinks_recipes.dm
+++ b/modular_skyrat/modules/customization/modules/food_and_drinks/recipes/drinks_recipes.dm
@@ -141,7 +141,7 @@
 
 /datum/chemical_reaction/drink/hollow_bone
 	results = list(/datum/reagent/consumable/ethanol/hollow_bone = 10)
-	required_reagents = list(/datum/reagent/toxin/bonehurtingjuice = 10,  /datum/reagent/consumable/milk = 15)
+	required_reagents = list(/datum/reagent/toxin/bonehurtingjuice = 10,  /datum/reagent/consumable/soymilk = 15)//Skeletons love milk, and will find soymilk disgusting!
 
 /datum/chemical_reaction/drink/jell_wyrm
 	results = list(/datum/reagent/consumable/ethanol/jell_wyrm = 2)


SpaceVampire:
> My reasoning for changing milk to soymilk is because the drink causes 30 disgust to skeletons, meaning they think it tastes awful. I figure soymilk is a bad imitation of the real thing, and even though the drink is no longer toxic it's nasty.

Pretty sure that isn't correct. It is the Race Drink for both skeletons and Plasmamen. They supposed to like it. Not sure where you got the part about it giving them 30 disgust, because it is also 25 in the code for the other races.

That being said, I still agree it should be changed because that is poorly thought out interaction with the mixing.
I'm also fine with soymilk being the replacement for it. Just uh. Thought that logic was weird.

GoldenAlpharex (approved):
Looks good to me!

sandshark808:
> > My reasoning for changing milk to soymilk is because the drink causes 30 disgust to skeletons, meaning they think it tastes awful. I figure soymilk is a bad imitation of the real thing, and even though the drink is no longer toxic it's nasty.
> 
> Pretty sure that isn't correct. It is the Race Drink for both skeletons and Plasmamen. They supposed to like it. Not sure where you got the part about it giving them 30 disgust, because it is also 25 in the code for the other races.
> 
> That being said, I still agree it should be changed because that is poorly thought out interaction with the mixing. I'm also fine with soymilk being the replacement for it. Just uh. Thought that logic was weird.

I'm dumb and mixed it up with another drink when I was writing the writeup. Skeletons actually love it yeah. Still, the gameplay reason for picking soymilk is that it's just as easy to get as normal milk.

PR #8918, opened 2021-10-18
Fixes CE's teleport access.

scrubs2009:
## About The Pull Request
The CE has access to the main teleporter rooms on every station. Except for Delta and Icebox. This is....weird. Especially considering that he has access to the AI sat teleporter on **every** station. What's even stranger is that he can still get into the Delta station teleporter room by pressing the shutter button. Which he has access to....even though he can't use the door. 

No longer will CE's be the most oppressed gamer minority and have to jaw open teleporter room doors every shift. Or weirder, run to the AI sat and use their teleporter to teleport to the other teleporter. 
 

The CE is a head of staff who has access to teleporters on all stations...except when he doesn't. I mean he still does...but not.
This fixes that!

## Changelog

:cl:
fix: fixed a few things
/:cl:


SpaceVampire:
You should maybe do this upstream, so that it makes it into the next map reset.

EDIT: That changelog is wack as well. What is the deal with the sound thing?

scrubs2009:
> You should maybe do this upstream, so that it makes it into the next map reset.
> 
> EDIT: That changelog is wack as well. What is the deal with the sound thing?

I just missed that one. Apologies. And I'm a github baby. How do I do it upstream?

KathrinBailey:
You need to do it here, just download their repo as you did Skyrat's.
https://github.com/tgstation/tgstation/pulls

On top of that, this isn't the right way to do it. The teleporter doors should require teleporter access **and/or** bridge access. What you're doing is making it so anyone with teleporter access cannot get inside, because the access you're setting is for the bridge.
This is what the req_one var is for. If the other maps do it like this, then they're wrong and the teleporter access should just be given to the CE instead.

PR #8920, opened 2021-10-18
Makes oversized VC only

Gandalf2k15:
Host request and agreed.

### DIFF
--- a/code/modules/client/preferences/middleware/quirks.dm
+++ b/code/modules/client/preferences/middleware/quirks.dm
@@ -59,7 +59,7 @@
 	//SKYRAT EDIT ADDITION
 	var/list/quirks = SSquirks.get_quirks()
 	var/datum/quirk/quirk = quirks[quirk_name]
-	if(initial(quirk.veteran_only) && !is_veteran_player(preferences.parent.ckey))
+	if(initial(quirk.veteran_only) && !is_veteran_player(preferences?.parent.ckey))
 		return FALSE
 	//SKYRAT EDIT END
 
--- a/code/modules/client/preferences/middleware/quirks.dm
+++ b/code/modules/client/preferences/middleware/quirks.dm
@@ -34,7 +34,7 @@
 	for (var/quirk_name in quirks)
 		var/datum/quirk/quirk = quirks[quirk_name]
 		//SKYRAT EDIT ADDITION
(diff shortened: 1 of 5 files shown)


GoldenAlpharex (commented):
This looks like it'd work, I think it would be nice to probably have a system for racial-restricted quirks (aka, tesharis shouldn't be allowed to use oversized)

hugmeorelse:
> This looks like it'd work, I think it would be nice to probably have a system for racial-restricted quirks (aka, tesharis shouldn't be allowed to use oversized)

I don't know. Big chungus Tesharis would be 👀

GoldenAlpharex:
> > This looks like it'd work, I think it would be nice to probably have a system for racial-restricted quirks (aka, tesharis shouldn't be allowed to use oversized)
> 
> I don't know. Big chungus Tesharis would be 👀

It would look jank AF, they aren't allowed to change their size currently, but, no worries, Ranged implemented that, we're good to go once CI passes.

code/modules/client/preferences/middleware/quirks.dm
```diff
@@ -33,6 +33,10 @@
 
 	for (var/quirk_name in quirks)
 		var/datum/quirk/quirk = quirks[quirk_name]
+		//SKYRAT EDIT ADDITION
+		if(initial(quirk.veteran_only) && !is_veteran_player(preferences.parent.ckey))
```
GoldenAlpharex: Otherwise CI won't pass.
```suggestion
		if(initial(quirk.veteran_only) && !is_veteran_player(preferences?.parent.ckey))
```

code/modules/client/preferences/middleware/quirks.dm
```diff
@@ -52,6 +56,13 @@
 /datum/preference_middleware/quirks/proc/give_quirk(list/params, mob/user)
 	var/quirk_name = params["quirk"]
 
+	//SKYRAT EDIT ADDITION
+	var/list/quirks = SSquirks.get_quirks()
+	var/datum/quirk/quirk = quirks[quirk_name]
+	if(initial(quirk.veteran_only) && !is_veteran_player(preferences.parent.ckey))
```
GoldenAlpharex:
```suggestion
	if(initial(quirk.veteran_only) && !is_veteran_player(preferences?.parent.ckey))
```

GoldenAlpharex (approved):

PR #8930, opened 2021-10-19
[MODULAR] Turns the Irish tactical sweater into the Sol tactical sweater, and fixes it in the loadout [option 1]

RatFromTheJungle:
## About The Pull Request
Removes references to a certain based group from a loadout item, and instead, references Sol itself.
Also makes it actually work in the loadout lmao

So humans can rep the homeland

## Changelog

:cl:
fix: Removes references to a specific irish group, and replaces it with Sol, also makes it available in the loadout.
/:cl:


RatFromTheJungle:
democracy has spoken

PR #8977, opened 2021-10-22
Goofmed Bugfix: Screaming no longer kills you. God. Why. Come on.

Iamgoofball:
doctors were suffocating people with pain screams by accident

i don't need to explain why this is bad

### DIFF
--- a/modular_skyrat/modules/emotes/code/scream_emote.dm
+++ b/modular_skyrat/modules/emotes/code/scream_emote.dm
@@ -7,8 +7,6 @@
 	if(!(. = ..()))
 		return
 	if(!user.is_muzzled() && !(user.mind && user.mind.miming))
-		if(ishuman(user))
-			user.adjustOxyLoss(5)
 		var/sound = get_sound(user, TRUE)
 		playsound(user.loc, sound, sound_volume, vary, 4, 1.2)
 


Iamgoofball:
according to gitblame this was an oldbase thing

SpaceVampire:
This seems like it could be another reason for people to want to use painkillers.
Isn't that a good thing?

ZephyrTFA:
Doctors should already be using painkillers or putting their patient to sleep using sleeping gas.
This is a bad change.

Iamgoofball:
You two don't understand the actual problem here. The scream emote should not be killing people given the amount of ways you can force someone to *scream.

ZephyrTFA:
> You two don't understand the actual problem here. The scream emote should not be killing people given the amount of ways you can force someone to *scream.

Its a grand total of 5 oxy damage.
It doesnt apply lose breath stacks, it doesnt apply nobreathe, it just gives a flat 5 oxy.
You are making a problem where there isn't one.

modular_skyrat/modules/emotes/code/scream_emote.dm
```diff
@@ -7,8 +7,6 @@
 	if(!(. = ..()))
 		return
 	if(!user.is_muzzled() && !(user.mind && user.mind.miming))
-		if(ishuman(user))
-			user.adjustOxyLoss(5)
```
GoldenAlpharex, line 11: Have it be stamina loss instead, bam, issue fixed.

GoldenAlpharex (changes requested):

Gandalf2k15:
This was added on citadel to prevent scream spamming. We have a literal hard cooldown on emotes which includes screaming. So damage is no longer required.

PR #8999, opened 2021-10-23
Smoke-B-Gone

SingingSpock:
## About The Pull Request

Stops fires (both atmospheric and welding fuel) from creating smoke by commenting out the lines that cause it. Cigarettes still do make smoke.

Smoke is not a gas. It is a pollutant, a completely different thing. It is not cleared by normal scrubbers, but portable scrubbers will clear it quickly even if set to filtering nothing or something not in the air. It is coded to choke people when there's a certain amount in the air you breathe. It never trips atmos/fire alarms, no matter how much of it exists, and will not interact with fire beyond being created by it. You cannot suck it out into space, it can actually exist in space just fine. The only way to clear it out is to bring in a portable scrubber. In short, it has no actual interaction with the normal atmos system. This makes firefighting way, way harder than it should be because you can't see anything, and if you try to use a scrubber to clear it the scrubber melts. If we want a smoke system to represent the way vision is impaired in a fire then it should be coded as an actual gas that follows the normal rules; as it stands it is simply frustration.

## Changelog

:cl:
del: Fires no longer create smoke pollutant
/:cl:


GoldenAlpharex (approved):
I wouldn't be against this.

Gandalf2k15:
Change the cap by setting a different pollutant cap, don't comment out skyrat files.

code/modules/atmospherics/environmental/LINDA_fire.dm
```diff
@@ -134,7 +134,7 @@
 			volume = affected.reaction_results["fire"]*FIRE_GROWTH_RATE
 			location.assume_air(affected)
 
-	location.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), POLLUTION_ACTIVE_EMITTER_CAP) //SKYRAT EDIT ADDITION
+	//location.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), POLLUTION_ACTIVE_EMITTER_CAP) //SKYRAT EDIT ADDITION
```
Gandalf2k15, line 137:
```suggestion
	location.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), 150) //SKYRAT EDIT ADDITION
```

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
```diff
@@ -133,7 +133,7 @@
 				total_reagents -= burn_rate
 
 	my_turf.hotspot_expose((T20C+50) + (50*fire_state), 125)
-	my_turf.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), POLLUTION_ACTIVE_EMITTER_CAP)
+	//my_turf.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), POLLUTION_ACTIVE_EMITTER_CAP)
```
Gandalf2k15, line 136:
```suggestion
	my_turf.PolluteListTurf(list(/datum/pollutant/smoke = 15, /datum/pollutant/carbon_air_pollution = 5), 150)
```

Gandalf2k15 (changes requested):

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
SingingSpock: > Change the cap by setting a different pollutant cap, don't comment out skyrat files.

The point isn't to have _less_ smoke from fires, it's to have _no_ smoke from fires. The files for smoke/carbon air pollution themselves aren't being touched (even though I think they should at some point be removed and perhaps replaced with another system), they're just no longer being called within the fire systems.

Gandalf2k15: Then the PR will not be accepted. The cap is there to be changed.

SingingSpock: Is there any actual reason for not removing this? Everyone I've spoken to in atmos/engineering hates this system because it makes no sense and wants smoke from fires removed. So far the only reasoning you've given is "don't comment out skyrat files". I could theoretically just change the number to 0 if that'd make you feel better, but then we're just wasting memory calling a proc that has no effect.

YakumoChen:
I'd just remove the whole system. Pollution creates thousands of lists and is generally terrible for code and extremely unoptimized.

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
Gandalf2k15: You're not even removing it correctly. So either way this wouldn't go through. Secondly smoke is a feature and can be tweaked. So do that before anything else.

SingingSpock: Again, you're not giving any actual reasoning, just the equivalent of "don't remove this code because I put it in and I said so". If I'm removing it incorrectly that's one thing, and I'd be happy to be shown the correct way, but removing it is most certainly the aim here and you haven't given any counterarguments. It's a "feature" that is disliked and very poorly optimized, that doesn't interact with the rest of our systems. So can you give an actual good reason **not** to remove it?

Gandalf2k15:
Given the fact you are not willing to change the smoke cap on this and instead want to go straight for removal, this is being closed, if you wish to make a new PR actually doing what was suggested we will consider it. The smoke cap is there to be changed, and I do not see an issue with simply half the amount produced and capping it. So as I said, if you want to make another PR doing so, feel free to.

PR #9001, opened 2021-10-23
Lower power draw of flashlights

Thesingleworstcoder:
It makes absolutely no sense why a power cell capable of sustaining a BSA can't hack the output of a flashlight, so I've dropped the power draw of flashlights down a notch

## About The Pull Request

Literally just dropped the power draw of flashlights down a notch. I don't see how this changes game balance apart from making it so that flashlights are actually slightly useful and not such a constant annoyance.

It makes the power cells in-game more believable, increasing everyone's fun.

## Changelog

:cl:
balance: changed power draw of flashlights
:cl:


GoldenAlpharex (approved):
It was a change I wanted to make for a good little while.

PR #9006, opened 2021-10-23
Are You Ready To Dye? (Barber fixes and QoL)

GoldenAlpharex:
## About The Pull Request
In this PR, I merged the two dye sprays together to make them feel nicer to use, and less confusing, and less code duplicating. There's still a lot, but I'll fix the razors another time. They now give proper feedback to the right people when being used, and also have a lot more uses to make them make sense with the job. If you're going to dye people's hair, might as well let them try to go for a wide variation of options, when you're at it.
Oh yeah and the dye spray's behavior is now fully modular, so upstream doesn't make it stop working if they make changes to it.
I made the razors unable to attack you unless you're in combat mode, to make it more intuitive to use for the barber.
I removed the pipette from the Barber's shop because it honestly didn't really fit there. I removed it from all maps except for Blueshift, to remove a possibility of map conflict with the other Barber-related Blueshift PR currently up.
I also fixed a few grammar issues.

Making the barber's job smoother means that it's going to be better to play as.

## Changelog

:cl: GoldenAlpharex
qol: The hair dye spray now has proper chat feedback to everyone involved.
qol: The hair dye spray's internal tanks have been increased, allowing it to fit no less than four times as much dye in it! Or maybe it's just more efficient at spraying dye. You decide.
fix: The razors will now properly direct you towards the right areas to aim for instead of letting you maim the person you're trying to shave, unless you have combat mode turned on. At that point, maim away.
spellcheck: Fixed the name of a few of the barber's tools.
code: Both hair dyes are now merged together in one, super powerful hair dye spray!
/:cl:


thestubborn:
are the hairdye cans effected by colourist? if not, they should be

GoldenAlpharex:
> are the hairdye cans effected by colourist? if not, they should be

What do you mean?

code/game/objects/items/dyekit.dm
```diff
@@ -1,17 +1,17 @@
 /obj/item/dyespray
 	name = "hair dye spray"
-	desc = "A spray to dye your hair any gradients you'd like."
+	desc = "A spray to dye hair and to also give it any gradients you'd like." // SKYRAT EDIT - Making the dyespray change hair color
```
jjpark-kb: I believe gradient is supposed to be singular.

modular_skyrat/master_files/code/game/objects/items/dyekit.dm
```diff
@@ -0,0 +1,64 @@
...
+/obj/item/dyespray/proc/dye(mob/target, mob/user)
+	if(!ishuman(target))
+		return
+
+	if(!uses) // Can be set to -1 for infinite uses, basically.
+		to_chat(user, span_warning("No matter how hard you shake the can, nothing'll come out, it's empty!"))
+		return
+
+	var/mob/living/carbon/human/human_target = target
+	var/list/dye_options = list("Change hair color", "Apply a gradient")
```
jjpark-kb: should these be set everytime the proc is called? or rather put on the parent so that you could have an item that can do either, or even both.

GoldenAlpharex: Hmmm... Well, if you,re talking about the dye_options specifically, I don't see much else that could be done here, aside from making the list static, which would be an improvement, yes.

code/game/objects/items/dyekit.dm
GoldenAlpharex: That was in the original description, I'll look it up because now I'm curious.

GoldenAlpharex:
```suggestion
	desc = "A spray to dye hair as well as giving it any gradient you'd like." // SKYRAT EDIT - Making the dyespray change hair color
```

thestubborn:
> 
> 
> > are the hairdye cans effected by colourist? if not, they should be
> 
> What do you mean?

the quirk, colourist

it lets you use spray cans more, seems like dye should fall under that, no?

GoldenAlpharex:
> > > are the hairdye cans effected by colourist? if not, they should be
> > 
> > 
> > What do you mean?
> 
> the quirk, colourist
> 
> it lets you use spray cans more, seems like dye should fall under that, no?

Pretty sure we currently have colourist turned off. However, I think you're thinking of Tagger. Eh. I think it'd be a bit much, it's not like you're going to run out of dye *that* fast now.

thestubborn:
> 
> 
> > > > are the hairdye cans effected by colourist? if not, they should be
> > > 
> > > 
> > > What do you mean?
> > 
> > 
> > the quirk, colourist
> > it lets you use spray cans more, seems like dye should fall under that, no?
> 
> Pretty sure we currently have colourist turned off. However, I think you're thinking of Tagger. Eh. I think it'd be a bit much, it's not like you're going to run out of dye _that_ fast now.

I'm a god damn idiot, it is tagger, yeah

Gandalf2k15 (approved):
LGTM

PR #9026, opened 2021-10-25
[MIRROR] HFR frontend rework [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61737
--------------------

## About The Pull Request

Massive overhaul of the HFR frontend.

hfr-ui-redux-8

I'm still getting used to the tgui framework, so any guidance toward becoming mergeworthy would be appreciated.

## Why It's Good For The Game

This makes the interface much more usable, and tries to help introduce the concepts involved in a very complicated system. The current interface is a bit clunky to use, and does not offer much guidance as to what actions are possible, or what the implications of taking them might be. You can't see the readouts at the same time you're tweaking the controls. Many failure states are not warned about, or even made clear when they're actively happening. To this end, this PR provides many very helpful qol changes, listed in the changelog below.

## Changelog

:cl:
code: LabeledList.Item now supports arbitrary nodes as labels. String labels still work as previously.
code: RoundGauge now supports an alertBefore property, for when warnings need to happen below a threshold. This can combine sensibly to form ranges with alertAfter.
refactor: The HFR interface has been heavily rewritten and restructured, fixing many edge cases where data was empty, unset, or took extreme values. Major parts now have their own file. The structure of gas lists is now consistent. Static data is used for more values. Much presentation layer logic has been moved to tgui instead of dm. Some processing that should never have happened, in tgui or anywhere else, is now gone forever.
qol: The HFR UI now displays the recipe specific gases that each fuel will create in a large table inside an expandable area. This can be collapsed when not needed.
qol: The HFR UI now uses a minimum mass for the scale of its gas lists. This should provide a visual cue for when a gas mix is running out, or needs more fuel to start.
qol: The HFR UI now graphs its various gas mixtures on a unified logarithmic scale, and tells you how fast the temperatures are changing in real units after all effects.
qol: The HFR UI now displays reactor status through gauges, with scales set to indicate typical ranges involved, and warnings to indicate when something is going wrong.
qol: The HFR UI now directly displays reaction instability, reaction activity, and the cell charge of the local APC. Reaction activity is the amount of heat output relative to what is permitted by the heat output limiter. This is effectively making explicit the result of a calculation you would otherwise need to do anyway, and makes it easier to notice when a reaction has stalled.
qol; The HFR UI will now helpfully inform you in discerning red text if you have no recipe set, or if any one of the gas mixtures is empty. This should be particularly helpful for noticing when you didn't add any coolant. Please involve coolant when operating a nuclear reactor.
qol: The HFR UI now always lists Plasma, BZ, and Proto-Nitrate in the moderator gas list, and provides tooltips explaining their usage and purpose.
qol: The HFR UI now has many other tooltips, from explaining the function and purpose of the various reactor controls, to the fringe effects of uncommon moderator gases.
code: heat_output_bool is gone forever yes this deserves its own changelog entry
/:cl:


GoldenAlpharex (approved):
Nice.

PR #9049, opened 2021-10-26
makes roundstart podpeople not OP

Iamgoofball:
## About The Pull Request
roundstart podpeople

see title

## Changelog

:cl:
balance: roundstart podpeople don't heal anymore from light
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/podweak.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/podweak.dm
@@ -23,3 +23,6 @@
 
 	learnable_languages = list(/datum/language/common, /datum/language/sylvan)
 	always_customizable = FALSE
+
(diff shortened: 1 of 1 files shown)


tf-4:
Fill out the template pls. Why is this change a good thing, why is it necessary? And use a more descriptive title

GoldenAlpharex:
While I think they were a bit too OP, I don't think this is the right way to go about it.

GoldenAlpharex (changes requested):
If you're going to nerf them, nerf the healing amount. Perhaps that it could also only be the damage type that they take when they're in the dark that they heal when in the light.

softcerv:
a way that I could see tackling it that would make the healing much less passive is require water to be in the body of plant people for them to heal.

jjpark-kb:
we merged already an alternative PR to nerfing podpeople weak

PR #9092, opened 2021-10-28
Moves the Military VOX Announcer voice to Malf-AI only and improves VOX Announcer voice code

GoldenAlpharex:
## About The Pull Request
~~They're of very low quality compared to the normal one and the sounds are just not equalized, everyone's having a bad time every time someone plays a non-standard one, it's just a jumpscare every time. Now, if you want to make my ears bleed, you'll have to be an antag for it.~~ So, I'm very cool and re-enabled the HL one, at 75% volume to make it more bearable, and the military one is now Malf-AI-only, where it's the only moment it makes sense, rather than being the DEFAULT option. It's also been reduced to 50% volume, and probably still is a little too loud.

Furthermore, the default voice is now the default voice again.
Oh, and I also fixed a possible runtime if someone was to not select any of the options when switching voices, which would otherwise make it null and the game wouldn't like that. Now, that won't happen anymore.

That one doesn't fit and it makes no sense as of right now for all AIs to be able to declare the station as "UNDER MILITARY COMMAND", like, no, that's not true, lmao. It opens up for possibly some interesting gimmicks for Malf AIs, as people already know when an AI is malf, it's really hard to hide for too long, it makes it an interesting option for those that want to go loud.

Furthermore, my ears will stop bleeding as those sounds are played, as I drastically reduced the volumes of them, to make them more tolerable.

## Changelog

:cl: GoldenAlpharex
expansion: Malf AIs now get the ability to change their VOX voice to the Military VOX announcer, rather than it being available to all AIs, as that voice was unfitting for regular AIs.
fix: The AI VOX announcer voice is by default the one it's meant to be, the normal one.
fix: The non-standard VOX announcer voices have had their volumes reduced.
code: Improves the VOX announcer voice code quite a bit.
/:cl:

### DIFF
--- a/code/modules/antagonists/malf_ai/datum_malf_ai.dm
+++ b/code/modules/antagonists/malf_ai/datum_malf_ai.dm
@@ -20,6 +20,12 @@
 	owner.special_role = job_rank
 	if(give_objectives)
 		forge_ai_objectives()
+	// SKYRAT EDIT START - Moving voice changing to Malf only
+#ifdef AI_VOX
+	var/mob/living/silicon/ai/malf_ai = owner.current
+	add_verb(malf_ai, /mob/living/silicon/ai/proc/switch_vox)
+#endif
+	// SKYRAT EDIT END
 
 	add_law_zero()
 	owner.current.playsound_local(get_turf(owner.current), 'sound/ambience/antag/malf.ogg', 100, FALSE, pressure_affected = FALSE, use_reverb = FALSE)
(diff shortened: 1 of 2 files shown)


GoldenAlpharex:
I've actually meant to do this for months and months and months, and just forgot, and was hoping someone else would come and fix them, but nobody ever did.

CandleJaxx:
oh thank god, no more tacticool military ais, bless you

YakumoChen:
won't this make the usage a blatant meta-tell though?

GoldenAlpharex:
> won't this make the usage a blatant meta-tell though?

Yeah, but I've thought about it, and I think it'll be more of a fun gimmick than anything. It's already very easy to tell when an AI is malf, let them have their fun while they're at it.

GoldenAlpharex:
Updated the PR body as this has changed significantly from what it was originally, thanks to motivation from a certain head maintainer (I agree, this is a lot better now)

PR #9123, opened 2021-10-29
Nature & Wildlife preservation measure(s)

Aphast:
## About The Pull Request
Locks a door in a ruin to save a kiwi, literally nothing else

think of the kiwis!! otherwise its just because the kiwi kills itself due to a large wander range removing the ability to have it as a pet or save it from the zombies, the ruin is also unmarked so there's no possibility to save it as its luck only

## Changelog

:cl:
fix: Added a locked door inside of a dangerous research facility, to prevent an unnecessary casualty.
/:cl:

### DIFF
--- a/modular_skyrat/modules/mapping/_maps/RandomRuins/SpaceRuins/dangerous_research.dmm
+++ b/modular_skyrat/modules/mapping/_maps/RandomRuins/SpaceRuins/dangerous_research.dmm
@@ -934,6 +934,7 @@
 "qq" = (
 /obj/machinery/door/airlock/science,
 /obj/effect/decal/cleanable/dirt/dust,
+/obj/effect/mapping_helpers/airlock/locked,
 /turf/open/floor/iron,
 /area/ruin/space/has_grav)
 "qE" = (


GoldenAlpharex:
So you're adding an airlock on top of another airlock? Remove the first one, if you're adding a bolted one. Unless I don't understand how these work.

Ebin-Halcyon:
> So you're adding an airlock on top of another airlock? Remove the first one, if you're adding a bolted one. Unless I don't understand how these work.

What he's adding is a mapping helper, There's different helpers for different things. you place that on top of an airlock you want to make do something (Roundstart bolted, Roundstart AI wire cut, etc etc) it's not an airlock itself.

GoldenAlpharex:
> > So you're adding an airlock on top of another airlock? Remove the first one, if you're adding a bolted one. Unless I don't understand how these work.
> 
> What he's adding is a mapping helper, There's different helpers for different things. you place that on top of an airlock you want to make do something (Roundstart bolted, Roundstart AI wire cut, etc etc) it's not an airlock itself.

Fair enough. This is good to go, then.

GoldenAlpharex (approved):

PR #9136, opened 2021-10-29
Idiot-proofs the False 911 Call Reporter.

Iamgoofball:
>bug report about 911 comes in
>it was caused by an admin messing around in the thunderdome
>they spawned something that's not supposed to exist outside of a real 911 call
>then used it anyways


Seris02:
ahahaha lmao nice

PR #9153, opened 2021-10-31
Fixes fat-fingering 911 by moving where the 911 buttons are(911 still requires an ID swipe as always)

Iamgoofball:
so basically, to make it hard to fatfinger dial 911, you have to swipe your card, right
the problem is the location on the screen where the confirm button for raising the alert level
is the exact same location one of the 911 buttons are now when the confirm button is hit
so galaxy brain captains keep double clicking
because the alert level requires a swipe as well

it is now moved down to the bottom of the UI to reduce fat-fingering

:cl:
fix: Reduces the chance you've fatfingered 911.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/CommunicationsConsole.js
+++ b/tgui/packages/tgui/interfaces/CommunicationsConsole.js
@@ -439,27 +439,6 @@ const PageMain = (props, context) => {
             content="Make Priority Announcement"
             onClick={() => act("makePriorityAnnouncement")}
           />}
-          {
-            // SKYRAT EDIT BEGIN
-          }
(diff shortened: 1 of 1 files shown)


gonenoculer5:
@Iamgoofball ..Why not just make it so the 911 requires a card swipe?

Iamgoofball:
It DOES. The issue is people are clicking the button with ID in hand because they just changed the alert level.

gonenoculer5:
> It DOES. The issue is people are clicking the button with ID in hand because they just changed the alert level.

Aaah okay. I misread then. Phenomenal.

Iamgoofball:
made irrelevant by other 911 PR

PR #9172, opened 2021-11-01
Perfume cap can now be toggled by using them inhand

Useroth:
See title.

:cl: Useroth
qol: Perfume bottles can now have their cap toggled by using them in-hand, rather than just relying on alt-click.
/:cl:


GoldenAlpharex (approved):
Very clean code that I absolutely have nothing to do with.

PR #9178, opened 2021-11-01
Fixes to SM gas filters. 

Iajret:
All of the skyrat map had old filter settings to filter nitrogen and that caused some "small" problems. 

## About The Pull Request
Someone probably noticed that setting for first filter in SM cooling loop couldn't be changed. And even worse, if you leave it be it will clog the system and cause delamination. 

Removes unnecessary hate to engineers.

## Changelog

:cl:
fix: That one broken filter in SM Engine room now should work flawlessly on all maps
/:cl:


Iajret:
Just noticed another PR for coils. Hope that's not a problem

KathrinBailey (changes requested):
Fix conflicts first please

PR #9192, opened 2021-11-02
balancing fishing a little

jjpark-kb:
## About The Pull Request
I find that you are actually getting good things from trash... I dont want that!
Made it only get trash from the loot, not just goodies from maint.

Just a little more strange rocks for the poor man, please? Promoting xenoarch should be fine.

We are getting too many fishing skill chips... suffer from less of them!

balancing fishing to be "less" rewarding from the trash loot.

## Changelog

:cl:
balance: fishing and getting trash will result in actual trash items, not goodies
balance: upped the chance to get strange rocks from the ore result in fishing
balance: lowered the chance to get the fishing skillchip
/:cl:

### DIFF
--- a/modular_skyrat/modules/fishing/code/fishing.dm
+++ b/modular_skyrat/modules/fishing/code/fishing.dm
@@ -7,7 +7,7 @@ GLOBAL_LIST_INIT(fishing_weights, list(
 	/obj/item/stack/ore/silver = 3,
 	/obj/item/stack/ore/iron = 5,
 	/obj/item/stack/ore/glass = 5,
-	/obj/item/xenoarch/strange_rock = 3,
+	/obj/item/xenoarch/strange_rock = 2,
 ))
 
 /datum/component/fishing
--- a/modular_skyrat/modules/fishing/code/fishing.dm
+++ b/modular_skyrat/modules/fishing/code/fishing.dm
@@ -7,7 +7,7 @@ GLOBAL_LIST_INIT(fishing_weights, list(
 	/obj/item/stack/ore/silver = 3,
 	/obj/item/stack/ore/iron = 5,
 	/obj/item/stack/ore/glass = 5,
-	/obj/item/xenoarch/strange_rock = 1,
+	/obj/item/xenoarch/strange_rock = 3,
 ))
 
 /datum/component/fishing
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/fishing/code/fishing.dm
```diff
@@ -7,7 +7,7 @@ GLOBAL_LIST_INIT(fishing_weights, list(
 	/obj/item/stack/ore/silver = 3,
 	/obj/item/stack/ore/iron = 5,
 	/obj/item/stack/ore/glass = 5,
-	/obj/item/xenoarch/strange_rock = 1,
+	/obj/item/xenoarch/strange_rock = 3,
```
GoldenAlpharex: Let's start with double first, just to see what it's like. They should still be kinda rare.
```suggestion
	/obj/item/xenoarch/strange_rock = 2,
```

modular_skyrat/modules/fishing/code/fishing.dm
```diff
@@ -71,26 +71,28 @@ GLOBAL_LIST_INIT(fishing_weights, list(
...
+			trashy_list = trashy_list[1]
+			spawning_reward = pick_weight(trashy_list)
+		if(41 to 65)
 			if(generate_fish)
 				generate_fish(spawning_turf, random_fish_type())
-		if(76 to 95)
+		if(66 to 97)
 			spawning_reward = pick_weight(possible_loot)
-		if(96 to 100)
+		if(98 to 100)
```
GoldenAlpharex: I'd honestly almost make it 2 percents, so from 99 to 100. Maybe just 1 percent even. Gotta make it actually rare and desireable.

GoldenAlpharex (changes requested):

modular_skyrat/modules/fishing/code/fishing.dm
```diff
@@ -71,26 +71,28 @@ GLOBAL_LIST_INIT(fishing_weights, list(
...
-		if(51 to 75)
+		if(1 to 40)
+			var/list/trashy_list = GLOB.trash_loot
+			trashy_list = trashy_list[1]
+			spawning_reward = pick_weight(trashy_list)
+		if(41 to 65)
 			if(generate_fish)
 				generate_fish(spawning_turf, random_fish_type())
-		if(76 to 95)
+		if(66 to 97)
```
jjpark-kb:
```suggestion
		if(66 to 98)
```

modular_skyrat/modules/fishing/code/fishing.dm
jjpark-kb:
```suggestion
		if(98=9 to 100)
```

jjpark-kb:
```suggestion
		if(99 to 100)
```

softcerv:
Maybe you could look into magnet fishing as a way for people to still get loot maint loot of some kind? It would also make it easier to balance around people getting loot from it if you had it as a separate type of rod. (Although, I'm not sure how much refactoring that would require)

I would also recommend maybe lowering the rate of fishing up trash a bit more to like 30% or maybe have the Fishing Master Skillchip lower it.

jjpark-kb:
okay, its been awhile, I have thought of something else to do instead. I do not know when the PR will be up, but itll be something instead of this.

PR #9196, opened 2021-11-02
[MODULAR] Moves the departmental guards baton to their backpack on spawn.

RatFromTheJungle:
## About The Pull Request

Moves the departmental guard's spawn-baton to his backpack, which fixes having a suit loadout item deleting it. (assuming you don't use move-to-case, as this breaks GAGS colorable stuff)

not RP related, but batong is important for the guard-job

## Changelog

:cl:
fix: fixes the guard baton being deleted because you brought your wintercoat to work.
/:cl:


GoldenAlpharex (commented):
The code checks out, but I'm debating just changing a bit how our loadout code works to take that kind of thing into account instead.

RatFromTheJungle:
> 
> 
> The code checks out, but I'm debating just changing a bit how our loadout code works to take that kind of thing into account instead.

that would be infinitely better, and I probably would have tried the same but my brain is currently a bit of sticky-tape and silly-glue on a sponge

PR #9215, opened 2021-11-03
must be an ashwalker to use the staff

jjpark-kb:
## About The Pull Request
the idea of the charges was bad. It doesn't feel good to have to get the useless relics.
Additionally, only ashwalkers should be using the staff, and it should only be to turfs right next to them

balancing the ashwalker staff

## Changelog

:cl:
balance: ashwalker staff no longer has charges
balance: ashwalker staff can only be used by ashwalkers
balance: ashwalker staff can only be used on tiles right next to it
/:cl:


GoldenAlpharex (approved):
Looks good, I still don't know what this really does, but I assume you do.

PR #9217, opened 2021-11-03
Making kudzu more utilitarian

Gandalf2k15:
## About The Pull Request

This basically adds a few more mutations(useful ones) and generally cleans up the code.

New mutations:
Carbon recycling - Basically replaced any carbon in the air with oxygen, much like what a real plant would do...
Breach fixing - Fixes any breaches that occur.
Low layer - Changes the kudzu to be the same layer as the floor, to the same effect of xeno grass.

Kudzu also no longer buckles non-human mobs.

It means we kudzu is a little more friendly and a little less round endy. 

Apparently, people actually use this stuff for cultivation and making positive effect mutations!

## Changelog

:cl:
add: KUDZU MUTATION: Carbon recycling - Basically replaced any carbon in the air with oxygen, much like what a real plant would do...
add: KUDZU MUTATION:  Breach fixing - Fixes any breaches that occur.
add: KUDZU MUTATION: Low layer - Changes the kudzu to be the same layer as the floor, to the same effect of xeno grass.
code: Kudzu has been cleaned up just a little.
remove: KUDZU: Removed co2 producing, co2 absorbing, nitrogen producing, and oxygen producing as they null the above. 
remove: KUDZU: Removed teleportation as it ruins the ability to make beneficial vines.
/:cl:


softcerv:
This might sound odd, but maybe you could give botany the ability to make tiles that could explicitly block Kudzu? So like they could make Kudzu healing rooms or stuff without it spreading to the entire station.

modular_skyrat/master_files/code/modules/events/spacevine.dm
```diff
@@ -793,7 +855,7 @@
 		return
 	for(var/datum/spacevine_mutation/vine_mutation in mutations)
 		vine_mutation.on_buckle(src, entangled_mob)
-	if((entangled_mob.stat != DEAD) && (entangled_mob.buckled != src)) //not dead or captured
+	if((entangled_mob.stat != DEAD) && (entangled_mob.buckled != src) && is_human(entangled_mob)) //not dead or captured
```
Gandalf2k15:
```suggestion
	if((entangled_mob.stat != DEAD) && (entangled_mob.buckled != src) && ishuman(entangled_mob)) //not dead or captured
```

modular_skyrat/master_files/code/modules/events/spacevine.dm
```diff
@@ -27,49 +27,53 @@
...
 
 /datum/spacevine_mutation
+	/// The name of the mutation, it will be displayed in the examine of any vine objects.
 	var/name = ""
+	/// How severe the plant it, think of it as the usefullness. Used in probability calculations.
 	var/severity = 1
-	var/hue
+	/// The color of the plant
+	var/plant_color
+	/// What kind of mutation is it, good, bad, really bad?
```
GoldenAlpharex, line 36: Some sexy documentation. Although, for `plant_color`, wouldn't it be better to have a default color, just in case?

modular_skyrat/master_files/code/modules/events/spacevine.dm
```diff
@@ -474,41 +478,41 @@
...
 	name = "radiation pulsing"
-	hue = "#ffef62"
+	plant_color = "#ffef62"
 	severity = 3
 	quality = NEGATIVE
 
-/datum/spacevine_mutation/radiation/on_grow(obj/structure/spacevine/holder)
-	radiation_pulse(holder, 100, 3)
+/datum/spacevine_mutation/radiation/on_grow(obj/structure/spacevine/vine_object)
+	radiation_pulse(vine_object, 100, 3)
```
GoldenAlpharex: Radiation was changed, I don't think this would work properly anymore, and I fear the range of that pulse (which would be a max range of 100 tiles, which I doubt is what's intended :) )

GoldenAlpharex (changes requested):
Two minor things.

tf-4:
I'm not convinced I'd call this a refactor

modular_skyrat/master_files/code/modules/events/spacevine.dm
Gandalf2k15: No color just means the vines don't get an override

GoldenAlpharex (approved):
Looks good, going to test-merge this one first.

Detrender:
I'm not sure if you re-enabled a lot of stuff when I bountied the kudzu rework, or if it was just never disabled, but as it stands beneficial kudzu is not a possibility, I've never seen the teleporting mutation enabled on a server before.

there's a lot of mutation bloat that needs to be cleared up, and some mutations that are far too strong.

twice now I tried to grow it, and twice I rolled one negative mutation which basically made it immune to my hatchet, and once it caused a station-wide emergency with a full blown emergency shuttle early evac. 

both times it was literally under direct sec supervision, or in the brig, and even then it overpowered the sec-officers on standby.

my suggestions:

_1. **The teleporting mutation needs to be disabled, as it is on every other server I've played on.**_

 Some kind of bug or intended feature makes it immune to my hatchet when it has the teleporting mutation, immediately resulting in it overrunning myself, and then immediately after anyone who doesn't know what they're doing, even in the most secure of places with the most prep it's still wayyy to strong.

_2. **disable co2 producing, co2 absorbing, nitrogen producing, and oxygen producing,**_

With "carbon recycling" having been added these are just mutation bloat that ends up making the actual beneficial mutations harder to get, whilst simultaneously resulting in the vines causing over-pressurization which will end up killing the crew rather than helping them. they're seemingly helpful, but not really, they just end up as unintentionally lethal feature bloat.

_3. **breach covering needs to only be on tiles directly beneath the space vine.**_

if you have a breach-covering vine, and it grows near a window or wall of the station, it immediately places floor tiles outside the station over space. This could be a major problem if the vine gets into space, as it could become self-perpetuating. 

space vine takes a long time to grow over a non-floor-tiled space, which intentionally keeps it from becoming a problem outside the station. Breach-covering is meant to patch holes, but not make the vines able to grow outside the station, so long as it still takes forever to grow over a space tile, it shouldn't be a concern, but if it tiles every open space in a small radius, expansion in space could become a major issue.

Detrender:
so, I've realized a few more problems that make beneficial kudzu non-viable.
1. it forces open doors
2. it buckles the crew

consider changing the low-layer mutation to "domesticated"
and adding that with the domesticated mutation kudzu doesn't buckle humanoids
and doesn't force open doors.

without this people in wheelchairs and crew will complain about the kudzu being annoying
and it causes major security/depressurization concerns if it forces open shutters, command doors, doors in sec, or external airlocks.

the other mutations can be dealt with, with a hatchet and a botanist's tank backbpack.

GoldenAlpharex (approved):
Looks good.

GoldenAlpharex:
I'll merge it once I can update the server, as otherwise it won't be test-merged anymore.

PR #9254, opened 2021-11-05
[MODULAR] Re-adds bleeding/burn wounds to slimes, like they had on oldbase.

RatFromTheJungle:
## About The Pull Request
Let's slimes take bleeding wounds again, meaning they are no longer 100% immune to the wounds system.

Jelly-things should probably start bleeding after getting stabbed twelve times.

## Changelog

:cl:
fix: Slimes can get bleeding wounds again.
/:cl:


tf-4:
This is a balance change, not a fix

Seris02:
yeah this isn't really a fix

RatFromTheJungle:
> 
> 
> This is a balance change, not a fix

they took the wounds before, it was undone in someone copy-pasting species traits, it is legally a fix because they're intended to take bleed wounds, it's why they can go so high into the positives for blood from toxins.

RatFromTheJungle:
> 
> 
> This is a balance change, not a fix

true, actually, after checking up stream, i'll have to go a bit more indepth with it too make up for the mighty 'balance' tag

RatFromTheJungle:
1984

PR #9261, opened 2021-11-05
[NONMODULAR] Adjusts the damage-threshold knockout from regen-coma

RatFromTheJungle:
## About The Pull Request

One of the most obtuse and hard to use symptoms has(had) a check which causes you to instantly fall into the regen coma (hard sleeping) at 70 flat damage, meaning it will stun lock you a whole 60 damage away from crit, namely because it was never updated for our much-increased health pool.

Perhaps, a crumb more versatility and use being available in Virology, instead of nocturnal and starlight on every virus.

## Changelog

:cl:
qol: Makes regen coma not instantly hard-sleep you a whole 60 damage early, thus letting virologists actually use the symptom without it KO'ing people at half-health.
/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -334,7 +334,7 @@
 			return power * 0.9
 		if(SOFT_CRIT)
 			return power * 0.5
-	if(M.getBruteLoss() + M.getFireLoss() >= 70 && !active_coma)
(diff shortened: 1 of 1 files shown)


SpaceVampire:
This is a balance change, not a qol.

RatFromTheJungle:
> 
> 
> This is a balance change, not a qol.

negative, this is keeping it to it's intended health% thresh-hold as it was forgotten when we had the health increase and doesn't use a flat percent, I couldn't in good faith call it a balance change when the issue with it being as it is was an oversight, it'd be closer to a fix, but it also doesn't really fit there.

the maintainer's will have to sully their hands with changing it too that if they see fit

GoldenAlpharex (approved):

PR #9268, opened 2021-11-06
[MOSTLY MODULAR]Hardlight Salve Medicell

softcerv:
## About The Pull Request

Adds the Hardlight Salve Medicell to the game, giving the Medigun an alternate way to heal that is far slower but doesn't suffer from the same damage threshold limitations and is able to work on utility only mediguns. (when those are created) 
salve

Ingame Screenshots of all of the items added by this PR:
image
## How the Salve Cell Works
To keep it short, the salve cell works by shooting out a mending globule similar to the one produced by the Sholean Grapes, but far weaker.
globule
It has the following attributes

- Heals 0.25 Brute and Burn per process 
  - This is equal to 1/4th of the healing speed on the original globule.
- Damage is only healed on the body part that it is attached to.
- Only heals 10 damage up to 10 damage per type.
  - This ends up being 20 overall damage max if both burn and brute have more or equal to 10 damage both.
- Non-reusable, once it is removed from someone, it is gone.
  - Done to prevent stockpiling.
- If the part it is trying to heal is at full health, then it will disappear.
  - This is done to prevent this from being used as a tool to effectively become a damage tank.
- Does not work on the dead or non-organics/non-humanoids.
- Globules can be stacked.
- This does not have the same health threshold limitations and penalties that the standard healing Medicells have.
  - Meaning that this will not do clone damage or disgust and will be able to heal people over 50 damage without consequence. 
- The Medicell is classified as a utility cell
  - This means the projectile is able to pass through glass
  - It will also be able to work on utility only Mediguns when they are eventually added

## How the Cell is acquired
Unlike other Medicells that require either materials or both materials and chemicals, this Medicell comes empty and requires aloe to be activated. 
empty
This should also help pave the way for more Medicells that use non-medical items in their construction. 
This PR further expands on the Medigun system, giving users more freedom on how they go about using Mediguns along with further fleshing out the selection of utility cells. Due to the nature of this requiring aloe, it will also require cross-departmental co-operation with botany to obtain.

## Changelog

:cl:
add: adds in the Hardlight Salve Medicell, please read https://github.com/Skyrat-SS13/Skyrat-tg/pull/9268 for more details.
/:cl:


GoldenAlpharex (approved):
The code checks out, my only concern is, won't that make the whole cloning damage from too high damage on bodies less relevant because you could just use this instead?

softcerv:
> 
> 
> The code checks out, my only concern is, won't that make the whole cloning damage from too high damage on bodies less relevant because you could just use this instead?

Regarding this concern, yes you are able to do it, but it would be far less efficient then using chems or most other methods. It would be possible to do this, but the slow healing rate and localized healing would make it so that doctors would only use this in the most dire of scenarios, unlike how brute and burn cells worked before the threshold nerfs.

PR #9274, opened 2021-11-06
[FUCK] Map configs will no longer systematically fail to load due to an incorrect typepath being provided

GoldenAlpharex:
## About The Pull Request
So, clearly, the person that did that refactor to fix some security issues didn't know how maps are loaded.
As such, at the end of every round, the next map would get written to `data/next_map.json`, and the next round, the game would try and load `_maps/next_map.json.json`. I got a feeling they didn't test their PR too much.

This makes it so that if no parameter is specified for `filename`, aka, `filename` is `next_map`, then it handles that differently than the rest. That way, the game will be able to load the correct map after a restart.

And, yes, I tested it, and it works.

This PR is also up upstream, but I included in this one a small change to make it so, upon failing to load a proper map config, it'll load Skyrat's MetaStation, rather than the mess that is our upstream MetaStation.

There's only so many times people will enjoy playing on the same map every single round. Especially when it breaks the Interlink's shuttle reliably.

## Changelog

:cl: GoldenAlpharex
fix: Maps will now correctly be loaded at the start of a round. Hurray!
/:cl:

### DIFF
--- a/code/datums/map_config.dm
+++ b/code/datums/map_config.dm
@@ -16,7 +16,7 @@
 	// Config actually from the JSON - should default to Meta
 	var/map_name = "Meta Station"
 	var/map_path = "map_files/MetaStation"
-	var/map_file = "MetaStation.dmm"
+	var/map_file = "MetaStation_skyrat.dmm" // SKYRAT EDIT - Making our Skyrat MetaStation the default MetaStation.
 
 	var/traits = null
 	var/space_ruin_levels = 7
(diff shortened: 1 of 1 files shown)


Useroth (approved):
Big oof.
I see it's already PRed upstream so we'll be able to merge it seamlessly just fine.

GoldenAlpharex:
Actually I'll test-merge it first, after making more changes. We got some time until the end of the round anyway.

GoldenAlpharex:
Also, looks like I'll have some cleaning up to do on this map anyway.

GoldenAlpharex:
I'll close this once the upstream PR gets merged, the code is better and improved, and I don't feel like duplicating. I'll make the change to the default config in a separate PR.

GoldenAlpharex:
The real fix has been merged, I'll close this and implement my default config change in a separate PR.

PR #9275, opened 2021-11-06
fix run times on lines 192 and 194

ORCACommander:
## About The Pull Request

Got more run times down stream from the last sticky update

Run times should not go brrr

## Changelog

:cl:
fix: fixed erp server toggle runtime
/:cl:


GoldenAlpharex (approved):
Thank you so much.

PR #9279, opened 2021-11-06
[MIRROR] Adds a generic fallback icon for clothes inhands [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/62574
--------------------
## About The Pull Request

Adds a blank name icon state for clothing inhands to be used as fallback for clothes that do not have specific inhand icons

It looks like a laundry bag
Screenshot_9

## Why It's Good For The Game

Inhands for clothes are generally a waste of time and resources as the only time they will be visible is the split second when someone grabs them to put them on, as a result we have a large amount of clothing items that either have no inhands or are reusing something existing even if they don't really look anything alike.

With this we no longer require unique inhand icons for every new type of clothing and we make sure they won't be invisible when they don't have one

People should still be expected to make inhands when appropriate (i.e. high risk items like armors or spacesuits etc.)

## Changelog

:cl:
imageadd: Added fallback icons for clothing inhands
/:cl:


GoldenAlpharex (approved):
I mean, sure, although I liked when I was holding clothes and they didn't show, but, whatever.

PR #9308, opened 2021-11-08
fix delta map reset

jjpark-kb:
## About The Pull Request
bring back some stuff for delta

bring back our delta map content


GoldenAlpharex (changes requested):
Everything's good, apart from the fact you changed the chapel to use tables instead of pews, which I don't think is quite as fitting, but I don't remember if that was actually an us-change or not.

GoldenAlpharex:
Eh, fuck it, I want the map fixed more than I want pews, I'll ask a mapper to bring them back or something, it's an easy fix.

GoldenAlpharex (approved):

PR #9314, opened 2021-11-08
[MIRROR] refactors injecting mapping helpers, adds an element injector [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/62622
--------------------

## About The Pull Request

refactors component, trait and custom icon injector into one path, atom injector, with them being subtypes of that path
adds an element injector
allows for use of custom args in the component injector (also in the new element one)
allows for use of traits not in the global trait list for trait injector (this is because people often dont update the list even though the trait may exist in the code, now it stack traces but doesnt crash the proc)
deletes the areabound and disease mapping helpers for being useless, you can just replicate them by varediting the component injector

## Why It's Good For The Game

better mapping tools, less copypasted code

## Changelog

not player facing


GoldenAlpharex (approved):
I'll pretend I understand what this does. 😎

PR #9326, opened 2021-11-09
[MODULAR] Hardlight Bullets can now pierce windows and grilles.

Iamgoofball:
:cl:
balance: Hardlight Bullets can now pierce windows and grilles.
/:cl:

This was something I did to avoid breaking kayfabe about them being bullets aesthetically, but it's proven to be far too bad in practice for Security due to SS13 map design!

You see, maps with actual designers behind them(Metastation, Kilostation, Deltastation, etc.) compared to Box which just has been fair game for everyone and anyone for 15 years, have assumptions they make about how people fight in them.

One of those assumptions is that security can fire through windows. That assumption is broken on Skyrat as of this time.

Currently, Security can throw down with folks in hallways and shit just fine, but the second they hit maintenance tunnels and departments, they get immediately stonewalled if the guy they're shooting runs behind a door. because bullets do not penetrate glass and grilles like lasers did. They have to shoot the glass or the door down to get in, meaning the guy can immediately fuck right off into maintenance and be 5 miles away by the time the window is blasted down or the door's been opened, completely killing all pressure they had on the target.

So, the armory gun bullets now pass windows and grilles.

### DIFF
--- a/modular_skyrat/modules/goofsec/projectiles/projectiles.dm
+++ b/modular_skyrat/modules/goofsec/projectiles/projectiles.dm
@@ -12,7 +12,7 @@
 	icon = 'modular_skyrat/modules/sec_haul/icons/guns/projectiles.dmi'
 	icon_state = "bullet"
 	name = "hardlight bullet"
-	pass_flags = PASSTABLE // All of the below is to not break kayfabe about it not being a bullet.
+	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE  // All of the below is to not break kayfabe about it not being a bullet.
 	hitsound ='sound/weapons/pierce.ogg'
 	hitsound_wall = "ricochet"
 	light_system = NO_LIGHT_SUPPORT


GoldenAlpharex (approved):
I suppose I'm fine with this.

SpaceVampire:
I understand your logic behind it, but I actually like the current system. I think it works better for TG, but I like feeling like I could stand behind the glass and use it as a moment to talk with someone without the fear of being suddenly lasered faster then I can react to it. It also really isn't the kind of think you think about bullets, and makes it so that they really don't stand out at all compared to other laser weapons. I liked the ones that did exist felt like they had an extra special thing because of it, and it made up for the fact that most of the current gun weapons are better then them.  So you have classic energy weapons and disabler shots feeling useful. Hopefully this makes sense, I refuse to proofread my thoughts.

Iamgoofball:
By the actual CI mechanics, if folks have CI'd up or have drawn firearms, talking is off the table and folks are engaging in fisticuffs galore. Therefore, talking is not a concern in situations where this has come up.

SpaceVampire:
> By the actual CI mechanics, if folks have CI'd up or have drawn firearms, talking is off the table and folks are engaging in fisticuffs galore. Therefore, talking is not a concern in situations where this has come up.

I do think it changes how people tend to react overall, but that might just be based on how people acted to me once during the point we had the single bullet type that could bypass glass. Still think that I raised a couple of valid points with the other things I said, so what do you think about those?

GoldenAlpharex:
> I understand your logic behind it, but I actually like the current system. I think it works better for TG, but I like feeling like I could stand behind the glass and use it as a moment to talk with someone without the fear of being suddenly lasered faster then I can react to it. It also really isn't the kind of think you think about bullets, and makes it so that they really don't stand out at all compared to other laser weapons. I liked the ones that did exist felt like they had an extra special thing because of it, and it made up for the fact that most of the current gun weapons are better then them. So you have classic energy weapons and disabler shots feeling useful. Hopefully this makes sense, I refuse to proofread my thoughts.

Hadn't seen this, and I think I can agree with a few of those points. This does bring up the question, which guns use hardlight bullets, anyway?

Iamgoofball:
Armory firearms that are not the shotgun.

GoldenAlpharex:
No longer going to be needed because of the Microfusion overhaul.

PR #9328, opened 2021-11-09
[MIRROR] Smoking DLC - Bad Lungs Pack [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/62662
--------------------

## About The Pull Request

This fixes #61124.  Havana cigars will not run out of nicotine while you are smoking them since I increased the nicotine amount.  As a result I also needed to boost the `chem_volume` higher so there is still room to inject other chems.

The smoker quirk has some new side effects for the players lungs.  Whoever takes the quirk suffers a 25% penalty towards their lungs max health and healing rate.  They also gain a 50% resistance when taking lung damage from smoking.

Lung damage has been modified for the following cigarette brands: (before all cigarettes did `1 lung damage`)

- Syndicate = `1.5`
- Shady Jims = `1.5`
- Xeno = `2`
- Little Timmy's Candy = `0.5`

## Why It's Good For The Game

Realism for smoking.  Smokers have bad lungs and these changes reflect this but also gives them a stronger tolerance against lung damage while smoking.  I also thought the more exotic and lethal brands of cigarettes should cause different amount lung damage compared to normal cigs.

## Changelog

:cl:
expansion: The smoker quirk has new side effects that cause a person's lungs to have 25% less max health and healing.  They also gain a 50% resistance from lung damage caused from smoking. 
expansion: Exotic cigarette brands now deal different amounts of lung damage. This applies to syndicate, shady jim's, xeno, and Little Timmy's Candy brands.  
fix: Fixed Havana cigars running out of nicotine. As a result of increasing the starting nicotine amount, the total `chem_volume` was increased to allow room for other chems to still be injected.
/:cl:


GoldenAlpharex (approved):
Nice. Smoking bad.

PR #9382, opened 2021-11-11
powerarmor balance

jjpark-kb:
## About The Pull Request
there has been almost no use of power armor, and that is due to how awful it is.
This PR is buffing power armor with the hopes more people will use it.

making this more relevant can promote forging, interactions with robotics for the suits, etc.

## Changelog

:cl:
balance: makes some of the power armor modules cost less
balance: caps power armor armor to 80
/:cl:

### DIFF
--- a/modular_skyrat/modules/powerarmor/code/powerarmor.dm
+++ b/modular_skyrat/modules/powerarmor/code/powerarmor.dm
@@ -262,7 +262,7 @@
 	///the list of possible armor upgrades: melee, bullet, laser, energy, bio, fire
 	var/list/armor_upgraded = list(0, 0, 0, 0, 0, 0)
 	///the list of possible healing upgrades: brute, burn, toxin, oxygen, stamina
-	var/list/healing_upgraded = list(FALSE, FALSE, FALSE, FALSE, FALSE)
+	var/list/healing_upgraded = list(0, 0, 0, 0, 0)
 	///the list of possible misc. upgrades: spaceproof, light, welding, temp-regulating, storage
 	var/list/misc_upgraded = list(FALSE, FALSE, FALSE, FALSE, FALSE)
 	///who is being affected by the power armor; the wearer
@@ -312,7 +312,7 @@
 /obj/item/clothing/suit/hooded/powerarmor/process(delta_time)
 	if(!COOLDOWN_FINISHED(src, healing_cooldown))
 		return
-	COOLDOWN_START(src, healing_cooldown, 3 SECONDS)
+	COOLDOWN_START(src, healing_cooldown, 2 SECONDS)
 	if(!wearer)
 		return
 	if(!isliving(wearer))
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/powerarmor/code/powerarmor.dm
```diff
@@ -312,24 +312,19 @@
...
-		living_wearer.adjustToxLoss(-3)
-	if(healing_upgraded[4] && living_wearer.getOxyLoss())
-		living_wearer.adjustOxyLoss(-3)
-	if(healing_upgraded[5] && living_wearer.getStaminaLoss())
-		living_wearer.adjustStaminaLoss(-3)
+	living_wearer.adjustBruteLoss(-healing_upgraded[1])
+	living_wearer.adjustFireLoss(-healing_upgraded[2])
+	living_wearer.adjustToxLoss(-healing_upgraded[3])
+	living_wearer.adjustOxyLoss(-healing_upgraded[4])
+	living_wearer.adjustStaminaLoss(-healing_upgraded[5])
```
GoldenAlpharex, line 327: Those numbers should all be defines.

modular_skyrat/modules/powerarmor/code/powerarmor.dm
```diff
@@ -384,37 +379,49 @@
...
+				if(healing_upgraded[4] >= 6)
 					return
-				healing_upgraded[4] = TRUE
+				healing_upgraded[4]++
 			if("stamina healing")
-				if(healing_upgraded[5])
+				if(healing_upgraded[5] >= 6)
 					return
-				healing_upgraded[5] = TRUE
+				healing_upgraded[5]++
```
GoldenAlpharex, line 424: A lot of magic numbers, I do not know what a `healing_upgraded[3]` or an `armor_upgraded[1]` is.

Also, I'm fairly confident lists start at 0 so this is just a bunch of runtimes for index out of bounds waiting to happen.

GoldenAlpharex (changes requested):
Needs some improvement when it comes to all the magic numbers, make use of more defines for the different armor types and different healing types.

modular_skyrat/modules/powerarmor/code/powerarmor.dm
ChungusGamer666: Lists start at 1 in byond code.

modular_skyrat/modules/powerarmor/code/powerarmor.dm
ChungusGamer666: No, they shouldn't, as they can get changed in runtime. Using a list is acceptable, but i'd recommend using an associated list (damagetype = damage healing value)

GoldenAlpharex: > No, they shouldn't, as they can get changed in runtime. Using a list is acceptable, but i'd recommend using an associated list (damagetype = damage healing value)

That isn't what I meant. `1` should be `BRUTE_HEALING` or something like that. That way, it's actually clearer which is which.

PR #9392, opened 2021-11-11
[MODULAR][FIX] Fixes message on use of conversion kits displaying a first person message to everyone. 

softcerv:
## About The Pull Request
Fixes the text that shows up when using conversion kits so that it doesn't use the user message for everyone in range of the conversion kit.

Inconsistent text is bad for RP

## Changelog

:cl:
fix: Using a conversion kit will no longer display a first person message to people besides the user.
/:cl:


GoldenAlpharex (approved):
Good change.

PR #9395, opened 2021-11-11
The Roleplay Update - Slightly more versatile emotes

GoldenAlpharex:
## About The Pull Request
~**Have you ever had to roleplay with someone that just *never* uses punctuation?**
I have. At first I was against this, but honestly, it's just going to make things better.
As such, a message will see itself being appended a period at the end if it doesn't end with one of the following symbols with anything but one of the following characters will result in your message : `.  ,  !  ?  "  ~  |  _  +  -`
Yes, sadly that does mean that it's not compatible with the bold/italic/underline stuff in the chat, so if you use those, make sure you still use punctuation! (I might look into doing that later, it probably wouldn't be too hard)~ **EDIT:** Removed from the PR, I did it originally to know if I was able to, if anyone wants to have it in the future, here it will be.

**Were you ever in a scene where you just wanted to describe something about your character, or, say, add something between commas before the action of your character, start your message with a comma and then realize afterward that there's an ugly space between your character's name and that comma? Or maybe you wanted to describe one of your character's multiple features, but had to word it in an awkward way?**
I have. And it annoyed me a lot, because it's something I like to do when I write.
As such, now, when writing an emote (normal emote, subtle and subtler), the space between your character's name and your message will be removed if you start the emote with a `,` or a `'` before everything else.

Complementary pictures showing all of that in action:
image

Here you can see it working for subtles and subtlers:
image

It even works for visible and audible emotes:
image

**EDIT:** Apostrophes now work too!
image

It allows you to make some more creative emotes that don't look jank.

## Changelog

:cl: GoldenAlpharex
qol: Emotes starting with a comma (,) or an apostrophe (') will no longer have an awkward space between the mob name and said comma.
code: Improved a bit some of the subtle and subtler code.
/:cl:


GoldenAlpharex:
Hold on, it seems like " and ' don't enjoy being there, and they're turned into something else by BYOND, I'll have to find another way to filter those.

GoldenAlpharex:
> Hold on, it seems like " and ' don't enjoy being there, and they're turned into something else by BYOND, I'll have to find another way to filter those.

I did find another way.

Arctaisia:
Auto punc. has been needed for a long time, thank you for it's inclusion! :heart:

And gosh, the other punctuation fixes are anazing, never thought I'd want them more in my life.

Funce (approved):

YakumoChen:
I thought those prefixes were used for radio? comma is specifically used to denote language, doesn't this have obvious conflicts with say?

E: I'm dumb and that's only for emotes

GoldenAlpharex:
> I thought those prefixes were used for radio? comma is specifically used to denote language, doesn't this have obvious conflicts with say?
> 
> E: I'm dumb and that's only for emotes

Indeed only for emotes, worry not my friend. It kind of all goes well in the end.

GoldenAlpharex:
I'll see how it goes on the server, and if it becomes a real problem I'll see about implementing a preference for it, but... I'd rather not, honestly.

GoldenAlpharex:
Asked upstream, they're not interrested in auto-punctuation, I'll let this happen here and port the emote change upstream, and deal with the conflicts when they come down, but there shouldn't be that many of said conflicts.

tf-4:
👎 I don't want the game to RP for me.
Let me make a conscious decision whether I want to use punctuation or not

GoldenAlpharex:
Removed the auto-punctuation because I didn't actually really want it, but I thought it was fun to code.

PCPort:
I do this so often out of habit anyway, having it be just that tiny bit more immersive is a welcome enough change!

tf-4:
> 
> 
> Removed the auto-punctuation because I didn't actually really want it, but I thought it was fun to code.

Now a good pr 😎

Gandalf2k15 (changes requested):
Regex it

GoldenAlpharex:
> Regex it

Already made it use RegEx you nerd >:)

GoldenAlpharex:
Will merge this once we're ready to test-merge the new Hemophages PR.

PR #9393, opened 2021-11-11
Welcome to the Panopticon: Directional Wall Item Refactor Mapping Edition [MDB IGNORE]

ORCACommander:
## About The Pull Request

TG Refactored Wall Mounts again. Now skyrat maps have to follow suit

**ALL EXISTING MAPS ARE NOW ON PR FREEZE**

there are hundreds of altered lines in the main maps. Merge conflicts are highly likely if this warning is ignored

## Changelog

:cl:
refactor: Updated Over 20 skyrat maps to new directional wall items
/:cl:

PR #9405, opened 2021-11-12
[MIRROR] Routine Cargo Departmental Deliveries [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61992
--------------------

## About The Pull Request

Document: https://hackmd.io/@ bazelart/HkY-SO9VF

Each department's request console is upgraded. Instead of making requests, they are only able to see crates related to their department that helps the department. They can order one for free (so not out of anyone's wallet) and it will arrive at cargo. Cargo gets the price of the ordered crate for bringing it to the department that ordered it (via an area check, of which the crate will remain locked until satisfied, emitter cracking aside).

Ordering a crate puts the console on a cooldown depending on the price of the crate ordered. The time ranges from 10 minutes at the lowest value, to capped at 20 minutes at 4x the default crate's price. the price in time follows a ease in out circular function, where the time increases slowly as the time goes up, but a lot in the middle. towards the higher end it slows down again.

Cargo will start with far less money, but this is countered by the new source of income.

- [x] Finish Mapping it
- [x] Finish tgui panel for department orders
- [x] ~Testmerge it for balance and feedback~ who cares
- [x] ~Maybe give multiple destination areas, just in case one area gets obliterated~ who cares!

## Why It's Good For The Game

Cargo starts with a budget to spend on themselves, which lets them order their department rewards before doing anything. There is no inherent reward to delivering to other departments other than social expectations of filling the job, which means cargo will simply ignore deliveries and requests if there are more self-important things going on, since those deliveries both cost cargo and do not help cargo in any way

## Changelog

:cl:
add: replaced the request consoles in each department with department order consoles, which order for free on a cooldown. cargo gets these orders and delivers the crates, which are locked until delivery. upon delivery, cargo gets paid the value of the crate, and can then sell the crate back on the shuttle.
balance: cargo doesn't start with a budget, other departments get what their budget was split up amongst them
/:cl:


thestubborn:
based

ORCACommander:
All Stock /tg/ maps reset to upstream in https://github.com/Skyrat-SS13/Skyrat-tg/pull/9393

PR #9420, opened 2021-11-13
Alternative job titles in the Occupations screen actually used in-game

### LINKED ISSUE #8904: Alternative titles do not apply
Reporting client version: 514.1568

## Round ID:
3339

## Testmerges:
- [[TM-Only] First attempt at stopping tgchat from eating messages](https://github.com/Skyrat-SS13/Skyrat-tg/pull/8842)
- [GoofSec: Departmental Guards, Stamina Crit Fixes, Armory Rework, 3-Hit-Stamcrit Batons, Flashbang fixes](https://github.com/Skyrat-SS13/Skyrat-tg/pull/8524)

## Reproduction:
Recently, I've noticed that alternative titles, even when selected and presumably saved, do not apply, all being default when you spawn in (this happens both shift-start and late-join.). I've seen it mentioned in LOOC a few times. I believe this may caused by the new customization UI update or maybe the testmerge changes made to the security department.

Tastyfish:
## About The Pull Request

This PR actually makes the job alternative titles apply in-game.
Specifically, to keep the scope sane, it does the following:
* Sets the player's ID and PDA assignment to the alt name.
* Informs the player that they are their alt name upon joining.
* Announces on their alt name in the late join announcement.
* Shows alt name on the manifest, while maintaining the department and order that it would have been normally (this change does not affect the functionality of the HoP reassigning the player).
* Tested against the wallet station trait as well.

Closes #8904

People can finally use their alternative job names and be a mailman, or a sanitation technician, or whatever. Fun!

## Changelog

:cl: tastyfish
fix: Alternative job titles in the Occupations screen are now used in-game.
/:cl:


GoldenAlpharex:
Okay, before I review this, I need to ask, are you up to the task of trying to debug all the funky jobs that shouldn't be at the top of the list as unsorted in the Crew Monitor? It bugs me out real bad and was an issue with alternate job titles in the past.

SpaceVampire:
I know they also had an issue with the job icon not showing up correctly when you look at them with the security hud.
So it would be nice if that worked correctly as well.

Tastyfish:
Looking at the code for crew monitors, it looks like the `entry["ijob"]` assignment just becomes null if it's an unrecognized job instead of `UNKNOWN_JOB_ID`, thus they are ordered to the top. I can probably fix that really quick.

Tastyfish:
As a side note, regarding the security HUD, if people changed their job via the HOP's Plexagon Access Management with a custom name, there's no longer any internal concept of which job they "really" are. The mob's job itself always remains what they spawned as, and the ID card only knows what the HOP typed in and chose for roles. The templates the HOP chooses don't set any hidden fields.

GoldenAlpharex:
> As a side note, regarding the security HUD, if people changed their job via the HOP's Plexagon Access Management with a custom name, there's no longer any internal concept of which job they "really" are. The mob's job itself always remains what they spawned as, and the ID card only knows what the HOP typed in and chose for roles. The templates the HOP chooses don't set any hidden fields.

So what you're saying is that you will keep the display of the job based on the trim, rather than the name, yeah?

Tastyfish:
> 
> 
> > As a side note, regarding the security HUD, if people changed their job via the HOP's Plexagon Access Management with a custom name, there's no longer any internal concept of which job they "really" are. The mob's job itself always remains what they spawned as, and the ID card only knows what the HOP typed in and chose for roles. The templates the HOP chooses don't set any hidden fields.
> 
> So what you're saying is that you will keep the display of the job based on the trim, rather than the name, yeah?

Whoops, forgot that actually does use trim. Oddly the crew monitor does just use the assignment and not the trim. I might change it to use trim for order and color so the crew monitor isn't a mess.

Tastyfish:
I'm going to be making some changes to the datacore upstream to better support custom assignments being handled correctly with trim information.

Tastyfish:
This will all be much cleaner pending https://github.com/tgstation/tgstation/pull/62803

Tastyfish:
Cleanup pending #9547

Tastyfish:
Much cleaner now, now all the snowflake code is basically all in the same datums that `customization` already did have snowflake code. I re-verified and fixed everything to work both on round-start *and* late join, and tested the wallet trait special case for the items. Smorgasbord screenshot for fun:

Imgur

GoldenAlpharex:
> Much cleaner now, now all the snowflake code is basically all in the same datums that `customization` already did have snowflake code. I re-verified and fixed everything to work both on round-start _and_ late join, and tested the wallet trait special case for the items. Smorgasbord screenshot for fun:
> 
> Imgur

Does it show you as the proper color for the proper role too? Like, does the HoS renamed to whatever still appear red, in bold and on top of security's part?

Tastyfish:
> > Much cleaner now, now all the snowflake code is basically all in the same datums that `customization` already did have snowflake code. I re-verified and fixed everything to work both on round-start _and_ late join, and tested the wallet trait special case for the items. Smorgasbord screenshot for fun:
> > Imgur
> 
> Does it show you as the proper color for the proper role too? Like, does the HoS renamed to whatever still appear red, in bold and on top of security's part?

I'll make a bunch of dummy characters with ID's real quick to show that.

Tastyfish:
> Does it show you as the proper color for the proper role too? Like, does the HoS renamed to whatever still appear red, in bold and on top of security's part?

Imgur

GoldenAlpharex:
> > Does it show you as the proper color for the proper role too? Like, does the HoS renamed to whatever still appear red, in bold and on top of security's part?
> 
> Imgur

Very cash-money.

Last thing I want to ask for this PR: Do you think you could make it so the PDA will display, one way or the other, what the original job title is as well, if it's a custom name? I don't want to look up six different names to know who's the Captain or the Scientist, y'know.

Tastyfish:
I was tempted to do that, but it ends up making things a bit busy. And the PDA's table already fills the full width of the default PDA window size. Not sure what a good solution to that would be

GoldenAlpharex:
Well, I think I can live with it being that way for now, then. Maybe we could alternatively just make the PDA window a little wider, to accomodate for that. I dunno. We'll see when we get to it.

code/datums/datacore.dm
```diff
@@ -222,6 +222,7 @@ GLOBAL_DATUM_INIT(data_core, /datum/datacore, new)
 	var/static/list/show_directions = list(SOUTH, WEST)
 	if(H.mind?.assigned_role.job_flags & JOB_CREW_MANIFEST)
 		var/assignment = H.mind.assigned_role.title
+		var/chosen_assignment = C.prefs.alt_job_titles[assignment] || assignment
```
GoldenAlpharex: Needs Skyrat comment.

code/datums/datacore.dm
```diff
@@ -244,7 +245,7 @@ GLOBAL_DATUM_INIT(data_core, /datum/datacore, new)
 		var/datum/data/record/G = new()
 		G.fields["id"] = id
 		G.fields["name"] = H.real_name
-		G.fields["rank"] = assignment
+		G.fields["rank"] = chosen_assignment
```
GoldenAlpharex: Needs Skyrat comment.

code/datums/datacore.dm
```diff
@@ -315,7 +316,7 @@ GLOBAL_DATUM_INIT(data_core, /datum/datacore, new)
 		var/datum/data/record/L = new()
 		L.fields["id"] = md5("[H.real_name][assignment]") //surely this should just be id, like the others?
 		L.fields["name"] = H.real_name
-		L.fields["rank"] = assignment
+		L.fields["rank"] = chosen_assignment
```
GoldenAlpharex: Needs Skyrat comment.

code/modules/admin/verbs/admingame.dm
```diff
@@ -315,7 +315,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	if(!record_found && (new_character.mind.assigned_role.job_flags & JOB_CREW_MEMBER))
 		//Power to the user!
 		if(tgui_alert(new_character,"Warning: No data core entry detected. Would you like to announce the arrival of this character by adding them to various databases, such as medical records?",,list("No","Yes"))=="Yes")
-			GLOB.data_core.manifest_inject(new_character)
+			GLOB.data_core.manifest_inject(new_character, src) // SKYRAT EDIT CHANGE - customization
```
GoldenAlpharex: Needs Skyrat comment. Also, usually, we like it when the original is added in the comment, to show exactly what it was you changed.

code/controllers/subsystem/job.dm
```diff
@@ -516,17 +516,22 @@ SUBSYSTEM_DEF(job)
...
 
 	equipping.mind?.set_assigned_role(job)
 	if(player_client)
-		to_chat(player_client, span_infoplain("You are the [job.title]."))
+		to_chat(player_client, span_infoplain("You are the [chosen_title].")) // SKYRAT EDIT CHANGE - customization
 
 	equipping.on_job_equipping(job, player_client?.prefs) //SKYRAT EDIT CHANGE
 
-	job.announce_job(equipping)
+	job.announce_job(equipping, player_client?.prefs)  //SKYRAT EDIT CHANGE - customization
```
GoldenAlpharex: This would be even better, I'd say. Also, take this as an example of how to properly do Skyrat comments, and try to do it like that for everything else!
```suggestion
	job.announce_job(equipping, chosen_title)  //SKYRAT EDIT CHANGE - Alternative Job Titles - Original: job.announce_job(equipping)
```

code/controllers/subsystem/job.dm
```diff
@@ -536,7 +541,7 @@ SUBSYSTEM_DEF(job)
 
 
 	if(player_client)
-		to_chat(player_client, span_infoplain("As the [job.title] you answer directly to [job.supervisors]. Special circumstances may change this.")) // SKYRAT EDIT ORIGINAL
+		to_chat(player_client, span_infoplain("As the [chosen_title] you answer directly to [job.supervisors]. Special circumstances may change this.")) // SKYRAT EDIT CHANGE - customization
```
GoldenAlpharex: Just to make it clearer which actual job you're playing. Also, improve that Skyrat edit, but you know that by now already.
```suggestion
		to_chat(player_client, span_infoplain("As the [chosen_title == job.title ? chosen_title : "[chosen_title] ([job.title])"], you answer directly to [job.supervisors]. Special circumstances may change this.")) // SKYRAT EDIT CHANGE - customization
```

code/modules/jobs/job_types/_job.dm
```diff
@@ -169,9 +169,9 @@
 			experiencer.mind.adjust_experience(i, roundstart_experience[i], TRUE)
 
 
-/datum/job/proc/announce_job(mob/living/joining_mob)
+/datum/job/proc/announce_job(mob/living/joining_mob, datum/preferences/used_pref) // SKYRAT EDIT CHANGE -- customization
```
GoldenAlpharex:
```suggestion
/datum/job/proc/announce_job(mob/living/joining_mob, job_title) // SKYRAT EDIT - Alternative Job Titles - Original: /datum/job/proc/announce_job(mob/living/joining_mob)
```

code/modules/jobs/job_types/_job.dm
```diff
@@ -169,9 +169,9 @@
...
 	if(head_announce)
-		announce_head(joining_mob, head_announce)
+		announce_head(joining_mob, head_announce, used_pref) // SKYRAT EDIT CHANGE -- customization
```
GoldenAlpharex: You know the drill by now.
```suggestion
		announce_head(joining_mob, head_announce, job_title) // SKYRAT EDIT CHANGE -- customization
```

code/modules/jobs/job_types/_job.dm
```diff
@@ -197,10 +197,13 @@
 	dna.species.pre_equip_species_outfit(equipping, src, visual_only)
 	equip_outfit_and_loadout(equipping.outfit, used_pref, visual_only, equipping) //SKYRAT EDIT CHANGE
 
-/datum/job/proc/announce_head(mob/living/carbon/human/H, channels) //tells the given channel that the given mob is the new department head. See communications.dm for valid channels.
+/// tells the given channel that the given mob is the new department head. See communications.dm for valid channels.
+/datum/job/proc/announce_head(mob/living/carbon/human/H, channels, datum/preferences/used_pref)  // SKYRAT EDIT CHANGE - customization
```
GoldenAlpharex:
```suggestion
/// Tells the given channel that the given mob is the new department head. See communications.dm for valid channels.
/datum/job/proc/announce_head(mob/living/carbon/human/H, channels, job_title)  // SKYRAT EDIT CHANGE - customization
```

code/modules/jobs/job_types/_job.dm
```diff
@@ -197,10 +197,13 @@
...
 	if(H && GLOB.announcement_systems.len)
+		var/chosen_rank = used_pref?.alt_job_titles[H.job] || H.job // SKYRAT EDIT ADDITION - customization
+
 		//timer because these should come after the captain announcement
```
GoldenAlpharex: This won't be needed anymore.
```suggestion
		//timer because these should come after the captain announcement
```

code/modules/jobs/job_types/_job.dm
```diff
@@ -197,10 +197,13 @@
...
-		SSticker.OnRoundstart(CALLBACK(GLOBAL_PROC, .proc/_addtimer, CALLBACK(pick(GLOB.announcement_systems), /obj/machinery/announcement_system/proc/announce, "NEWHEAD", H.real_name, H.job, channels), 1))
+		SSticker.OnRoundstart(CALLBACK(GLOBAL_PROC, .proc/_addtimer, CALLBACK(pick(GLOB.announcement_systems), /obj/machinery/announcement_system/proc/announce, "NEWHEAD", H.real_name, chosen_rank, channels), 1)) // SKYRAT EDIT CHANGE - customization
```
GoldenAlpharex:
```suggestion
		SSticker.OnRoundstart(CALLBACK(GLOBAL_PROC, .proc/_addtimer, CALLBACK(pick(GLOB.announcement_systems), /obj/machinery/announcement_system/proc/announce, "NEWHEAD", H.real_name, job_title, channels), 1)) // SKYRAT EDIT CHANGE - customization
```

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
...
+SUBSYSTEM_DEF(altjob)
+	name = "Alt Jobs"
+	init_order = INIT_ORDER_JOBS
+	flags = SS_NO_FIRE
+
+/datum/controller/subsystem/altjob/Initialize(timeofday)
+	RegisterSignal(SSdcs, COMSIG_GLOB_JOB_AFTER_SPAWN, .proc/on_job_after_spawn)
+	return ..(timeofday)
+
+/datum/controller/subsystem/altjob/proc/on_job_after_spawn(datum/source, datum/job/job, mob/living/spawned, client/player_client)
```
GoldenAlpharex: It would be nice to document this.

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
+//SKYRAT MODULE CUSTOMIZATION
+// Make a seperate subsystem so it doesn't conflict with the actual job one.
+
+SUBSYSTEM_DEF(altjob)
+	name = "Alt Jobs"
```
GoldenAlpharex:
```suggestion
	name = "Alt Job Titles"
```

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
+//SKYRAT MODULE CUSTOMIZATION
```
GoldenAlpharex: I'm not 100% sure why this is in customization, but...

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
...
+	flags = SS_NO_FIRE
+
+/datum/controller/subsystem/altjob/Initialize(timeofday)
+	RegisterSignal(SSdcs, COMSIG_GLOB_JOB_AFTER_SPAWN, .proc/on_job_after_spawn)
+	return ..(timeofday)
+
+/datum/controller/subsystem/altjob/proc/on_job_after_spawn(datum/source, datum/job/job, mob/living/spawned, client/player_client)
+	SIGNAL_HANDLER
+
+	var/alt_title = player_client.prefs.alt_job_titles[job.title]
```
GoldenAlpharex: Add null-checks because we never know if a client couldn't be null at one point. You know how BYOND is.

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
...
+	// Player also needs to be a human or we're aborting.
+	if(alt_title && istype(human))
+		// ID card
+		var/obj/item/card/id/card = human.get_idcard()
+		if(istype(card))
+			card.assignment = alt_title
+			card.update_label()
+
+		// PDA
+		for(var/obj/item/pda/pda in human)
```
GoldenAlpharex: Is there no better way to do this?

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
```diff
@@ -0,0 +1,35 @@
...
+			pda.ownjob = alt_title
+			pda.update_label()
+
+		// Wallet
+		for(var/obj/item/storage/wallet/wallet in human)
```
GoldenAlpharex: I feel like this could be a little excessive too.

GoldenAlpharex (changes requested):
You gotta fix those Skyrat comments and make them more significant, so I don't have to look up in the Git history to figure out what it was before your change, or why it was done this way.

Furthermore, an entire subsystem for this? I mean, it's not necessarily *that* bad, but couldn't we just cut the middleman and somehow integrate this in the regular job subsystem? I just fear that this could slow down our roundstart even more if we gotta loop through everyone's inventory twice just to look for a PDA and a wallet.

modular_skyrat/modules/customization/controllers/subsystem/alt_titles.dm
Tastyfish: I'm trying to remember now why I didn't do `H.get_item_by_slot(pda_slot)`. It might have been out of a worry that jobs, quirks, etc have the theoretical ability to scramble things around. Would it be a good compromise to just say if there's an edge case where the PDA was removed out of the PDA slot that the person will need to update their PDA info?

Likewise with wallets I should have just used the slot because that actually is enforced (I think). Not sure why I did that.

So in general I did the subsystem to try to modularize it out of endless job modifications, which the `customization` module already does (this is part of why I put it in that module, so customization can sort of just own taking over the job datum). Would you rather I modify `/datum/outfit/job/post_equip()` and the wallet trait?

GoldenAlpharex: Due to the fact we're already struggling with super long round starts, yeah, I think this could be a good move, it should reduce the impact of all of this on performance, hopefully.

Tastyfish:
> Well, I think I can live with it being that way for now, then. Maybe we could alternatively just make the PDA window a little wider, to accomodate for that. I dunno. We'll see when we get to it.

So funny story, most of the PDA interfaces assume the default window size and don't scale at all. One of the many reasons to hopefully eventually get rid of the old PDA's. So I basically went for the obvious default implementation of showing both (the ol' `rank == trim ? rank : "rank (trim)"` logic. It's a bit dense and of course makes the row almost twice as tall, but what is your opinion on just doing it that way?

Imgur

Edit: actually in context, it's probably fine. Normal entries also force a word wrap because of how tiny the PDA window is.

Imgur

Tastyfish:
@GoldenAlpharex There's actually no way to do the skyrat edit comments in the tgui JSX code without tripping linter errors (and for that matter, you have to do convoluted hackery to comment most JSX lines anyway). Can I just write a summary at the top of the tgui files or something instead?

Tastyfish:
Note that I couldn't directly modify the ID and pda setup in `post_equip()` because the client is not set at that point. So I added a proc to keep it somewhat contained. It is guarenteed to run before outfit traits like the wallet one, and there are actually only 2 possible slots PDA can ever go in, so I was able to make the ID and PDA update logic work in O(1) ie no looping of any form.

GoldenAlpharex:
> Note that I couldn't directly modify the ID and pda setup in `post_equip()` because the client is not set at that point. So I added a proc to keep it somewhat contained. It is guarenteed to run before outfit traits like the wallet one, and there are actually only 2 possible slots PDA can ever go in, so I was able to make the ID and PDA update logic work in O(1) ie no looping of any form.

Amazing.

modular_skyrat/modules/alternative_job_titles/code/job.dm
```diff
@@ -0,0 +1,20 @@
+// ALTERNATIVE_JOB_TITLES
+
+/datum/controller/subsystem/job/proc/setup_alt_job_items(mob/living/carbon/human/equipping, datum/job/job, client/player_client)
```
GoldenAlpharex, line 10: If you could add some documentation to that proc, since it's a new one you're adding, that'd be really great. There's also a neat multiline comment helper extension for VSC that'll make your life easier, I think it's recommended by the repo too.

tgui/packages/tgui/interfaces/CrewManifest.js
```diff
@@ -11,6 +11,8 @@ const commandJobs = [
   "Chief Medical Officer",
 ];
 
+// SKYRAT EDIT CHANGE BEGIN - ALTERNATIVE_JOB_TITLES
+// Any instance of crewMember.trim was originally crewMember.trim
```
GoldenAlpharex:
```suggestion
// Any instance of crewMember.trim was originally crewMember.rank
```

GoldenAlpharex (changes requested):
Solid work, just two little nitpicks left.

GoldenAlpharex:
Also once you make the changes, ping me on Discord, so I don't forget. :)

tgui/packages/tgui/interfaces/CrewManifest.js
Tastyfish: d'oh

GoldenAlpharex (approved):

PR #9438, opened 2021-11-14
removes 2 reach from reagent spears for more ash walker nerfs

Elli-Skala:
## About The Pull Request

removes the 2 reach from reagent spears, its likely they got 2 reach for much the same reason bone spears did 

two range and any weapon gives it a super sizeable step up, less one that can inject poisons. much the same reason bone spears got nerfed ash walkers shouldn't be powerful

## Changelog

:cl:
balance: makes the regnant spear only 1 reach
🆑

### DIFF
--- a/modular_skyrat/modules/reagent_forging/code/modules/reagent_forging/forge_weapons.dm
+++ b/modular_skyrat/modules/reagent_forging/code/modules/reagent_forging/forge_weapons.dm
@@ -55,7 +55,6 @@
 	attack_verb_simple = list("attack", "poke", "jab", "tear", "lacerate", "gore")
 	wound_bonus = -15
 	bare_wound_bonus = 15
-	reach = 2
 	sharpness = SHARP_POINTY
 
 /obj/item/forging/reagent_weapon/spear/ComponentInitialize()


Grunnyyy:
Everyone can easily make reagent spears, reagent spears are also really weak

jjpark-kb:
why are you trying to remove reach from stuff... it doesn't make it much stronger. There are much stronger things.

Stalkeros2:
As I've already described in my own PR, spears are strong because there are no better alternatives, and I'd advise to flesh out the rest of their equipment to make it have any use whatsoever instead of removing whatever good they have.
And before you say "hurr tribals have bad weapon durr", I'll have to say that this is not true in the slightest: their equipment might be low-tech, but it's still deadly, as they have more experience using it and know how to improve such weaponry to make it better. And history proves that. ~~But I won't.~~

jjpark-kb:
if the issue with spears is the damage, please lower that instead of removing the one thing that makes them spears: range. Removing range from spears essentially makes them worse than any other melee weapon since wielding the spear.

Additionally, reagent spears have a force of 8 if unwielded, 16 if wielded. That is not a whole lot of damage.

PR #9453, opened 2021-11-14
[NON-MODULAR] Re-adds per-character PDA settings

Superlagg:
## About The Pull Request

Basically ports https://github.com/Skyrat-SS13/Skyrat-tg/pull/6612 to the new system!

Moves PDA settings (PDA color and font) from general preferences to character preferences.

Adds a PDA setting to set a round-start PDA ringer message, so you can have your PDA say something like depths instead of beep without having to set it every time by hand.

Adds a non-modular line in PDA.dm that sets the message on spawn.

NOTE: This'll very likely wipe any PDA settings that're set, so make sure you write down the color and font beforehand!

Makes PDAs more personal with less effort!

## Changelog
:cl:
add: PDA customization is now set and saved per character.
add: PDA ring messages can now be set in character preferences.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -141,6 +141,9 @@ GLOBAL_LIST_EMPTY(PDAs)
 	. = ..()
 	if(!equipped)
 		if(user.client)
+			//SKYRAT EDIT ADDITION BEGIN - CUSTOMIZATION
+			ttone = trim(user.client.prefs.read_preference(/datum/preference/text/pda_ringer), 20)
+			//SKYRAT EDIT ADDITION END
 			background_color = user.client.prefs.read_preference(/datum/preference/color/pda_color)
 			switch(user.client.prefs.read_preference(/datum/preference/choiced/pda_style))
 				if(MONO)
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -141,8 +141,8 @@ GLOBAL_LIST_EMPTY(PDAs)
 	. = ..()
 	if(!equipped)
 		if(user.client)
-			//SKYRAT EDIT ADDITION BEGIN - CUSTOMIZATION
-			ttone = trim(user.client.prefs.read_preference(/datum/preference/text/pda_ringer), 20)
+			//SKYRAT EDIT ADDITION BEGIN - Per-character PDA settings
+			ttone = sanitize(trim(user.client.prefs.read_preference(/datum/preference/text/pda_ringer), 20))
 			//SKYRAT EDIT ADDITION END
 			background_color = user.client.prefs.read_preference(/datum/preference/color/pda_color)
 			switch(user.client.prefs.read_preference(/datum/preference/choiced/pda_style))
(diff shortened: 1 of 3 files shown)


code/modules/client/preferences/pda.dm
```diff
@@ -1,17 +1,53 @@
 /// The color of a PDA
 /datum/preference/color/pda_color
-	category = PREFERENCE_CATEGORY_GAME_PREFERENCES
+	//SKYRAT EDIT CHANGE BEGIN
+	//category = PREFERENCE_CATEGORY_GAME_PREFERENCES
+	category = PREFERENCE_CATEGORY_NON_CONTEXTUAL
+	//SKYRAT EDIT CHANGE END
```
GoldenAlpharex: Basically do all of those one-line changes like this.
```suggestion
	category = PREFERENCE_CATEGORY_NON_CONTEXTUAL // SKYRAT EDIT - Per-character PDA settings - ORIGINAL: category = PREFERENCE_CATEGORY_GAME_PREFERENCES
```

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/features/character_preferences/skyrat/species_features.tsx
```diff
@@ -68,6 +68,12 @@ export const exploitable_info: Feature<string> = {
   component: FeatureTextInput,
 };
 
+export const pda_ringer: Feature<string> = {
+  name: "PDA Ringer Message",
+  description: "Want your PDA to say something other than \"beep\"? Maybe make it scream about BLOOD or EXPLOSIONS or DEPTHS? Accepts the first 20 characters.",
```
GoldenAlpharex: I'm not sure about that description, I don't know what it is but there's something about it that bugs me. Maybe it's the fact the sounds are in all caps when they're not in all-caps in the game. Or maybe it's just the fact it's very specific examples that don't apply to the majority (since blood would be more for hemophages (at least that's what I feel like), and depths would be more for akulas/aquatics).

code/game/objects/items/devices/PDA/PDA.dm
```diff
@@ -141,6 +141,9 @@ GLOBAL_LIST_EMPTY(PDAs)
 	. = ..()
 	if(!equipped)
 		if(user.client)
+			//SKYRAT EDIT ADDITION BEGIN - CUSTOMIZATION
+			ttone = trim(user.client.prefs.read_preference(/datum/preference/text/pda_ringer), 20)
```
GoldenAlpharex: I feel like this should be sanitized, I feel like someone clever could do some HTML injection with this.

GoldenAlpharex (changes requested):
Good change, only a few things to address.

code/modules/client/preferences/pda.dm
Superlagg: Done!

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/features/character_preferences/skyrat/species_features.tsx
Superlagg: Changed it!

code/game/objects/items/devices/PDA/PDA.dm
Superlagg: I tried putting in some html stuff through char prefs and it didn't do anything. Then again I'm not that clever. In any case, its sanitized now!

GoldenAlpharex (approved):

GoldenAlpharex:
I was thinking about making this require migration but honestly, small enough that I don't care.

PR #9489, opened 2021-11-16
Re-adds the recently removed two tile range from bone spears

DiscordWizard:
## About The Pull Request

Adds back the extra tile of range that bone spears used to have

The PR that removed the two tile range claimed that having an extra tile of range for the ashwalker's main melee weapon was overpowered against miners and against HRP, as well as comments saying that this should add variety.

On that first point, arguing that a lizard with a two tile range spear is overpowered against someone with a ranged weapon that has infinite ammunition and works wonders in low pressure, as well as whatever gear they can get and instant healing medicine, is if not biased at the very least uninformed. 

In the rare event that miners and ashwalkers come to blows the miners would have the upper hand at any time the ashwalker isn't using a bow, spear or not, and I don't see how giving the lizards a tool to hunt down mobs that doesn't entail them risking their life for every goliath is less HRP than a bunch of spacemen using science magic to destroy the fauna quicker than any ashwalker will ever be able to.

As for the argument that says removing the two tile range will force the ashwalkers to use their other weapons and add variety to their arsenal, I've to say removing the two tile range will make most ashwalkers not use the spear at all. If they want to attack something at range, they will use a bow. If they want to kill something in melee, they will get an axe. The spear will be removed alltogether from its nice 2 tile range hunting niche and ashwalker weapon variety will go down by one for basically no reason at all.

tl:dr removing two tile range on spears doesn't impact balance meaningfully outside of heavily nerfing starting or unexperienced ashwalkers for zero gains towards the overall roleplay or fun experience
## Changelog

:cl:
balance: Gives bone spears their 2 tile range back
/:cl:

### DIFF
--- a/code/game/objects/items/spear.dm
+++ b/code/game/objects/items/spear.dm
@@ -181,6 +181,7 @@
 	desc = "A haphazardly-constructed yet still deadly weapon. The pinnacle of modern technology."
 	force = 12
 	throwforce = 22
+	reach = 2 // SKYRAT EDIT
 	armour_penetration = 15 //Enhanced armor piercing
 
 /obj/item/spear/bonespear/ComponentInitialize()


gonenoculer5:
This was already fixed on the original PR afaik.

DiscordWizard:
Not sure if it was reverted or fixed, all I know is last I played bone spears still had one tile range, so I made this one. Guess someone will close it or not merge it if it has already been fixed
EDIT: Just saw an older pr about bone weapon rebalances that re-adds the extra reach to bone spears, if that gets merged I'll just close this one

PR #9515, opened 2021-11-17
Adds cryo to most places it's missing

tf-4:
## Changelog
:cl:
fix: You can now cryo on the interlink, on icebox station, in permabrig of all maps, and in the ghost cafe.
/:cl:


GoldenAlpharex:
While on the topic of cryo rooms, I know that the prisons also don't have cryo, or, at the very least, Metastation's prison doesn't have any (also I feel like the ghost roles and the ghost cafe might not either?)

tf-4:
> 
> 
> While on the topic of cryo rooms, I know that the prisons also don't have cryo, or, at the very least, Metastation's prison doesn't have any (also I feel like the ghost roles and the ghost cafe might not either?)

one moment

PR #9518, opened 2021-11-17
I, too, love crippling players for over 5 minutes with just 3 units of a bar drink. Race Drinks no longer stack absurd amounts of disgust and vomiting on the incorrect drinkers.

Iamgoofball:
## About The Pull Request

Race Drinks no longer stack absurd amounts of disgust and vomiting on the incorrect drinkers.

lmao guys what if we re-added old neurotoxin levels of stunning, but we tied it to 3u of a bar drink mixable with absolutely zero upgrades

## Changelog

:cl:
balance: Race Drinks no longer stack absurd amounts of disgust and vomiting on the incorrect drinkers.
/:cl:


GoldenAlpharex (changes requested):
I'd rather have the disgust moved to when the reagent is ingested instead, it would make more sense to me.

PR #9329, opened 2021-11-09
[MODULAR] Fragile quirk

ArdentSpark:
## About The Pull Request

A -6 trait that applies 1.25x increased brute and 1.2x increased burn, multiplicatively. Teshari that take this quirk will be as fragile as they were before.

## How This Contributes To The Skyrat Roleplay Experience

Allows Teshari to take a quirk that takes the amount of damage they take to pre-buff values from #9247. Other species can take the quirk as well, however damage values are set with Teshari in mind. (Fragile IPC moment.) Heat damage was untouched for compatibility with other species that also desire the ability to be crushed like hamburger.
[Disclaimer I have no idea what I'm doing. If it's not implemented perfectly please don't send the orange man to break my kneecaps.]

## Changelog

:cl:
add: Added Fragile quirk, which increases incoming brute by 1.25x and burn by 1.2x
/:cl:

PR #9247, opened 2021-11-04
[MODULAR] Mildly buffs teshari (heavily)

RatFromTheJungle:
## About The Pull Request

Lower's the bonus damage from teshari, as 50% is a bit to steep, even for me, for comparison, they get no where near as many bonus' from being small (2, vs the like 20 synths have) for almost the same incoming damage bonus.

## How This Contributes To The Skyrat Roleplay Experience

Less teshari getting two tapped by 30 damage guns, vs someone who just ate six shots from the same gun

## Changelog

:cl:
balance: rebalances teshari incomming damage to 20-25% vs the previous of 50%, and gives them a single point more of punch damage so they quit magically rolling 0, don't ask, I also don't know why.
/:cl:

PR #9543, opened 2021-11-18
[MODULAR] Fragile PR attempt #2

ArdentSpark:
## About The Pull Request
(RE-PR'd due to poor branch control) A -6 trait that applies 1.25x increased brute and 1.2x increased burn, multiplicatively. Teshari that take this quirk will be as fragile as they were before.
Previous feedback located in #9329

Allows Teshari to take a quirk that takes the amount of damage they take to pre-buff values from #9247. Other species can take the quirk as well, however damage values are set with Teshari in mind. (Fragile IPC moment.) Heat damage was untouched for compatibility with other species that also desire the ability to be crushed like hamburger.
[Disclaimer I have no idea what I'm doing. If it's not implemented perfectly please don't send the orange man to break my kneecaps.]

## Changelog

:cl:
add: Added Fragile quirk, which increases incoming brute by 1.25x and burn by 1.2x
/:cl:


code/__DEFINES/~skyrat_defines/traits.dm
```diff
@@ -15,6 +15,7 @@
 #define TRAIT_STICKY_FINGERS "sticky_fingers" //This is so a mob can strip items faster and picks them up after
 /// This makes trait makes it so that the person cannot be infected by the zombie virus.
 #define TRAIT_MUTANT_IMMUNE "mutant_immune"
+#define TRAIT_FRAGILE "fragility" // increased incoming damage (1.2x burn, 1.25x brute multiplicative), numbers designed to make Teshari who take this as vulnerable as before their buff.
```
GoldenAlpharex:
```suggestion
/// Increased incoming damage (1.2x burn, 1.25x brute multiplicative), numbers designed to make Teshari who take this as vulnerable as before their buff.
#define TRAIT_FRAGILE "fragility"
```

modular_skyrat/master_files/code/datums/traits/negative.dm
```diff
@@ -8,6 +8,26 @@
 	medical_record_text = "Patient is incapable of communicating their emotions."
 	icon = "question-circle"
 
+/datum/quirk/fragile
+	name = "Fragility"
+	desc = "You feel incredibly fragile. Incoming brute and burn damage is increased!"
```
GoldenAlpharex:
```suggestion
	desc = "You feel incredibly fragile. Burns and bruises hurt you more than the average person!"
```

GoldenAlpharex (changes requested):
Yup. We're almost good to go.

code/__DEFINES/~skyrat_defines/traits.dm
ArdentSpark: Is the reason for this because if the length of the comment?

modular_skyrat/master_files/code/datums/traits/negative.dm
```diff
@@ -8,6 +8,26 @@
 	medical_record_text = "Patient is incapable of communicating their emotions."
 	icon = "question-circle"
 
+/datum/quirk/fragile
+	name = "Fragility"
+	desc = "You feel incredibly fragile. Burns and bruises hurt you more than the average person!"
+	value = -6
+	mob_trait = TRAIT_FRAGILE
```
GoldenAlpharex: Now that I think about it. What does this trait do? Does it do anything upon being added or removed? Because if not, I don't know what the point of having it apply a trait is.

GoldenAlpharex (changes requested):

modular_skyrat/master_files/code/datums/traits/negative.dm
ArdentSpark: You are totally right actually, there are no references to this trait anywhere! Must have been when I was treading incredibly carefully when looking at other code to try and do the right thing. I'll go ahead and remove the define and the trait reference.

GoldenAlpharex (approved):
Looks good.

PR #9571, opened 2021-11-19
Blueshift map edits and fixes.

IsaacTheSharkWolf:
## About The Pull Request
Fixes a variety of bugs and oversights, including:
-Fixing camera direction vars so they aren't facing the wrong way.
-Fixing toxins burn chamber floor not being vacuum tiles.
-Fixing the toxins bomb test floor to not break from bombs, making second tests possible from now on.
-Updating some of the apcs to comply with new filepaths.
-Removing the smes substations which were causing part of the increased powerdraw at roundstart, this should hopefully make the engineering experience better when starting the round.
-Adding cameras to upper cargo, gulag shuttle dock and tech storage as i seemingly missed these spots before.

Adds things that i should have added on release but didn't due to oversight.
Fixes things to be more functional/aesthetically pleasing and in line with ever updating code.

## Changelog

:cl:
fix: tiles not being vacuum tiles in toxins burn chamber
fix: tiles in bomb test area breaking from explosions, making them drop to other z's on subsequent tests making follow up bomb tests impossible unless the floor were to be repaired.
fix: cameras facing the wrong way due to the recent direction var update.
fix: removed the increased roundstart power draw due to me not knowing how smesses work.
/:cl:


GoldenAlpharex (approved):
From what I could see, seemed fine to me.

PR #9650, opened 2021-11-23
[MIRROR] Fixes the master controller dying sometimes [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/62857
--------------------

## About The Pull Request
I woke up to admins freaking out about massive lag spikes on manuel round 173107

They were seeing large amounts of hard deletes and weird spikes every ~2 seconds

Looking into the [runtime log](https://tgstation13.org/parsed-logs/manuel/data/logs/2021/11/15/round-173107/runtime.txt), the master controller was resetting every roughly 2 seconds
This was caused by an infinite loop in godslayer armor. Works like this

Player gets hurt enough to put them into crit, with enough brute to bring them up a stat level, and more hurt elsewhere
Godslayer armor detects harm, starts to heal
The heal heals the small amount of brute
This sends the stat update signal, which is sent before the stat is actually changed

The armor heals again, going through the same process
Any later calls to update_stat will send the same signal again because the signal is sent before the value is actually changed

This kills the master controller

I've fixed this by moving the cooldown reset, and by changing how stat is set in conjunction with the signal

## Changelog

:cl:
fix: Fixes the master controller dying in a rare case of the godslayer armor
/:cl:


GoldenAlpharex (approved):
We good.

PR #9655, opened 2021-11-23
[NON-MODULAR] Hypovial Factory Fix

softcerv:
## About The Pull Request
Hypovials have been clogging up presses when printed out, increasing the capacity from 60 to 65 should hopefully prevent this from happening as much.

Makes intended feature work as intended.

## Changelog

:cl:
fix: fixes presses clogging up when printing hypovials
/:cl:

### DIFF
--- a/code/modules/plumbing/plumbers/pill_press.dm
+++ b/code/modules/plumbing/plumbers/pill_press.dm
@@ -12,7 +12,7 @@
 	///maximum size of a bottle
 	var/max_bottle_volume = 30
 	//SKYRAT EDIT HYPOVIALS maximum size of a vial
-	var/max_vial_volume = 60
+	var/max_vial_volume = 65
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
What do you mean, they've been clogging up the presses? And how does this actually fix the issue?

Elli-Skala:
> What do you mean, they've been clogging up the presses? And how does this actually fix the issue?

I know this is super late but the reason this likely fixes clogging is because a hypovile Is a 60 unit container, and the chem press right now can only hold 60 unit

TG doesn't have hypoviles so the largest container it can normally print om tgis 50 unit pills that leaves 10 units of over flow chemical space. But hypoviles get 0 over flow so it ends up clogged (may also be problems with having 59.999 units)


So honestly I think it should be made 70 units to give hypoviles the same 10 over flow pills get

PR #9680, opened 2021-11-24
[modular] better whitespace for the examine menu

Seris02:
## About The Pull Request

title

I didn't know how tgui worked when I made the original whitespace change so uh, it doesn't, but, it does add code improvements (and you can now use tabs in the examine panel)

## Changelog

:cl:
code: whitespace is now properly handled in the examine menu
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/ExaminePanel.js
+++ b/tgui/packages/tgui/interfaces/ExaminePanel.js
@@ -39,32 +39,23 @@ export const ExaminePanel = (props, context) => {
           <Stack.Item grow>
             <Stack fill vertical>
               <Stack.Item grow>
-                <Section scrollable fill title={character_name + "'s Flavor Text:"}>
-                  {flavor_text.split("\n").map((val, i) => (
-                    <Stack.Item key={i} grow fill>
-                      {val || val !== "" ? val : <br />}
-                    </Stack.Item>
-                  ))}
+                <Section scrollable fill title={character_name + "'s Flavor Text:"}
+                  preserveWhitespace>
+                  {flavor_text}
                 </Section>
               </Stack.Item>
               <Stack.Item grow>
                 <Stack fill>
                   <Stack.Item grow basis={0}>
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
Can you give an example of what kind of problem it solves? At a quick glance, my doodoo brain doesn't understand what this does.

Seris02:
> Can you give an example of what kind of problem it solves? At a quick glance, my doodoo brain doesn't understand what this does.

right now you can't add tabs
plus, better code

GoldenAlpharex:
> > Can you give an example of what kind of problem it solves? At a quick glance, my doodoo brain doesn't understand what this does.
> 
> right now you can't add tabs plus, better code

I need to have an example of why we'd need to have tabs, because it's like one of the things you can't really have anywhere afaik. I'd need to double-check, but yeah.

tgui/packages/tgui/interfaces/ExaminePanel.js
```diff
@@ -39,32 +39,23 @@ export const ExaminePanel = (props, context) => {
...
                   <Stack.Item grow basis={0}>
-                    <Section scrollable fill title={(custom_species ? "Species: " + custom_species : "No Custom Species!")}>
-                      {(custom_species ? custom_species_lore.split("\n").map((val, i) => (
-                        <Stack.Item key={i} grow fill>
-                          {val || val !== "" ? val : <br />}
-                        </Stack.Item>
-                      )) : "Just a normal space dweller.")}
+                    <Section scrollable fill title={(custom_species ? "Species: " + custom_species : "No Custom Species!")}
+                      preserveWhitespace>
+                      {(custom_species ? custom_species_lore : "Just a normal space dweller.")}
```
GoldenAlpharex, line 58: We need to remove those empty fillers and make this just not show up if it's not relevant (aka it's empty), food for another PR.

GoldenAlpharex (approved):
I suppose this is fine. If it had to break, it would've broken already.

PR #9679, opened 2021-11-24
Reverts the ugly monkey plant

YakumoChen:
## About The Pull Request
Reverts flora.dm to base TG, thereby removing the ugly awful monkey plant
The monkey plant was an undocumented change added by #6368

it's also currently broken for whatever reason so I'm doing a service by fixing the code.
closes #9671

## How This Contributes To The Skyrat Roleplay Experience

Plants are supposed to look nice. Having an aesthetic plant that looks like a horrifying crossbreed of a living animal and a bush does not serve the appealing function potted plants are supposed to have.

Furthermore it didn't make much sense anyway. It survives just fine in atmosphere that would normally kill monkies, the tanks on the sprite aren't removable, etc immersion broken please get it out of here.

## Changelog

:cl:
fix: Removed an invisible and extremely ugly plant
/:cl:

PR #9690, opened 2021-11-24
Fixes the aesthetics plant module

### LINKED ISSUE #9671: Invisible potted plant in Interlink
## Round ID:
3673

## Reproduction:
The potted plant just south of the shuttle terminal in the Interlink appears to be invisible. You can shift+right-click on it to open the context menu and pick it up. Doing so will make your sprite completely invisible save for HUD elements such as MedHUD and SecHUD. This also makes it incredibly difficult for other players to click on you for combat or other functions. I spoke with an admin during the round and they confirmed that the potted plant should not be doing this according to its variables.

Gandalf2k15:
image

As you can see in the above image, plant39 should actually be plant-39. The code processing these is sound.

closes #9671

and also closes #9679 as it removes far more than just the poor little monkey tree. Here's the old plant file for reference!

image

## Changelog

:cl:
fix: The invisible plant has been exterminated.
/:cl:


GoldenAlpharex (approved):
Good catch.

PR #9697, opened 2021-11-25
Fixes our modular poster file

ghost:
## About The Pull Request
MFW we have yet another overwrite of a TG .dmi thats in a sorry state

I re-listed our modular changes at the very bottom of the .dmi so its easy to update, and got rid of the crappy "poster_#" standard that even **TG** told me not to use for safetymoth

Also pre-mirrors the upcoming fix to the safetymoth icon state (in our modular file only) so the mirror can be merged without worry

No more missing icon state posters

## Changelog

:cl:
fix: Fixed modular posters mucking up the new TG posters
/:cl:


GoldenAlpharex (approved):

GoldenAlpharex:
I mean, ideally we'd just replace some of the posters' files with our own, but I suppose this can also do.

PR #9702, opened 2021-11-25
[MODULAR][FIX] Fixes Eng-Guard inhand/inventory sprite.

ArdentSpark:
## About The Pull Request
Fixes the engineering guard uniform so that it no longer shows an ERROR sprite, and changes the name of the engineer and orderly uniform icon states in code for readability/consistency
unknown

Fixes good.
ERROR sprites bad.
Consistent icon names for code readability probably good.

## Changelog

:cl:
fix: Fixes engineering guard uniform: item sprite (No more ERROR)
/:cl:


tf-4:
I actually think this is the incorrect fix, and that the code should be changed to match the icon state, not the other way around, for consistency with the other items.

ArdentSpark:
I can do so but the cause of the problem was the dmi in general, because the icon state was called engineering_uniform and the worn icon state was engineer_uniform, and I went for the fix that required the fewest changes

_Buuuuuuuuut_
While I'm here I can totally do that.
Orderly uniform is called medical uniform which isn't self explanatory
engineering guard uniform being called engineering/engineer uniform isn't really either
compared to customs uniform at least for the customs agent

I'll draft up a little shifteroo with the files and I'll probably ping blutarch as I'm messing with their code (I doubt they'll care as it's just icon names)

PR #9704, opened 2021-11-25
[Modular] Xeno-resin *turf emote gets its own quirk

desertmote:
## About The Pull Request

Hi, I'm back. 👋😊
Another one of the race-exclusive *turf emotes gets a quirk to add it onto other species.

It was requested, and I understand since the amazing new character editor allows more freedom.

## Changelog

:cl:
add: Xeno-resin quirk
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/traits.dm
+++ b/code/__DEFINES/~skyrat_defines/traits.dm
@@ -22,6 +22,7 @@
 #define TRAIT_FLORAL_ASPECT "floral_aspect"
 #define TRAIT_ASH_ASPECT "ash_aspect"
 #define TRAIT_SPARKLE_ASPECT "sparkle_aspect"
+#define TRAIT_XENO_ASPECT "xeno_aspect"
 
 // Trait sources
 #define GHOSTROLE_TRAIT "ghostrole" // SKYRAT EDIT ADDITION -- Ghost Cafe Traits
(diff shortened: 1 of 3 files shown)


desertmote:
Unfortunately I don't know how to add/edit quirk icons, I would fix the sparkle quirk icon and give this one an icon too if I could.

tf-4:
Not sure I like this tbh, xeno-hybrids getting it at all was questionable and this is worse.

GoldenAlpharex (changes requested):
Code checks out, but I don't think I personally agree with this at all. I don't want to see more xeno resin around the station, especially from non-Xeno-hybrid.

desertmote:
Ok, however I don't really see the difference between this and the spiderweb *turf. Remember these expire in 10 seconds if not stood upon.

tf-4:
Xenos are vastly more common than spiders, and that's saying something

desertmote:
Ok, I understand. Just one more thing to consider is that the sprite *turf uses is distinguishable from real xeno-turf. That's all.

tf-4:
> Ok, I understand. Just one more thing to consider is that the sprite *turf uses is distinguishable from real xeno-turf. That's all.

Ah, distinguishable how? It's been a while since I looked at the *turf xeno turf

desertmote:
> > Ok, I understand. Just one more thing to consider is that the sprite *turf uses is distinguishable from real xeno-turf. That's all.
> 
> Ah, distinguishable how? It's been a while since I looked at the *turf xeno turf

It's smaller than a full tile, the edges don't come near the 32x32 image pixel edge. original xeno resin is larger than a 32x32 full-tile.

desertmote:
seems like there is no interest in ending this pr.

PR #4561, opened 2021-03-31
[MODULAR] Lets cyborgs move dragged objects around

tf-4:
## About The Pull Request

Shamelessly stolen from Citadel https://github.com/Citadel-Station-13/Citadel-Station-13/pull/14524

## Why It's Good For The Game

Getting stuck because you can't move shit to the side is painful

## Changelog
:cl:
balance: Cyborgs can now move dragged objects and people around.
/:cl:

PR #9710, opened 2021-11-25
You can no longer shove people off of tables from range

Zonespace27:
## About The Pull Request

You're able to shove people off tables as a borg from any range, this fixes that.
Note that according to the reporter, this is only a skyrat, not /tg/ issue.

Bugs are bad.

## Changelog

:cl:
fix: Borgs can no longer shove people off tables at range.
/:cl:

### DIFF
--- a/code/datums/elements/climbable.dm
+++ b/code/datums/elements/climbable.dm
@@ -44,8 +44,6 @@
 	SIGNAL_HANDLER
 	var/list/climbers = LAZYACCESS(current_climbers, climbed_thing)
 	for(var/i in climbers)
-		if(!in_range(i, user)) //SKYRAT CHANGE: fixes borgs shoving people off tables
-			continue
 		var/mob/living/structure_climber = i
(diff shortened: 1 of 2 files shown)


GoldenAlpharex:
Did you test it on /tg/ too, to be sure that it's not an issue for them as well?

tf-4:
I feel like if it's a Skyrat-exclusive issue, there's probably a modular fix.

Possibly an issue with https://github.com/Skyrat-SS13/Skyrat-tg/pull/4561.

Seris02:
not a fix
this is just duct tape
unless it actually is a fix, in which case, tg has it too, and you should send it off to them instead

Zonespace27:
> I feel like if it's a Skyrat-exclusive issue, there's probably a modular fix.
> 
> Possibly an issue with #4561.

it's an issue with #4561, will push changes in a minute

GoldenAlpharex (approved):
Much better.

PR #9716, opened 2021-11-26
[NON-MODULAR] stops medhuds from seeing DNRs as revivable

Seris02:
## About The Pull Request

title

because medical shouldn't be trying to revive DNR people

## Changelog

:cl:
fix: stops medhuds from seeing DNRs as revivable
/:cl:


GoldenAlpharex:
Should also make sure the examine text also says that they're gone and can't be revived.

Seris02:
> Should also make sure the examine text also says that they're gone and can't be revived.

oooh good idea

GoldenAlpharex (approved):
Looks good to me.

PR #9748, opened 2021-11-28
[nonmodular] runtime fix for pollution chems

Seris02:
## About The Pull Request

14k runtimes in 1.5 hours is not fun

runtime fix

## Changelog

:cl:
code: fixed 14k runtimes in the form of pollution trying to be added to a null turf
/:cl:


tf-4:
This is just suppressing the issue

Why is it trying to pollute a null turf in the first place?

GoldenAlpharex (changes requested):
Can you temporarily replace it with a `CRASH()` based on if the turf is null or something like that, so we can test-merge it to try and see where the issue is coming from?

PR #9758, opened 2021-11-28
[Modular] Body size 0.8 is pickable again

Superlagg:
## About The Pull Request

'Fixes' a rounding error(?) that made sanitize_float round 0.8 down to 0.79999995, and therefore think 0.8 is less than 0.8 when step was set to 0.01.

Literally just set step in /datum/preference/numeric/body_size from 0.01 to 0.0000001. Seems to work just fine.

Short characters can be that teensy bit shorter like they should be.

## Changelog

:cl:
fix: Body size now accepts 0.8 as it should.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/client/preferences/cursed_shit.dm
+++ b/modular_skyrat/master_files/code/modules/client/preferences/cursed_shit.dm
@@ -395,7 +395,7 @@
 	savefile_key = "body_size"
 	minimum = BODY_SIZE_MIN
 	maximum = BODY_SIZE_MAX
-	step = 0.01
+	step = 0.0000001
 
 /datum/preference/numeric/body_size/is_accessible(datum/preferences/preferences)
 	var/passed_initial_check = ..(preferences)


tf-4:
Ngl I don't like this as a fix. I don't have an alternative, but this feels hacky.

YakumoChen:
isn't this going to bloat out the field with a bunch of decimals?
What you should do is just change MIN_BODYSIZE to 8.01 or something so it doesnt' bug out.

Superlagg:
> isn't this going to bloat out the field with a bunch of decimals? What you should do is just change MIN_BODYSIZE to 8.01 or something so it doesnt' bug out.

The field already accepts enormous decimals just that they're rounded down when saved and loaded. And, the field scales to the number in there. If you put in 0.8, it'll show 0.8 and scale the box to fit.

I tried changing the body size defs, setting it to 0.79 made it so 0.79 was accepted as well. Setting it to 0.801 wouldnt help either, since players would need to type in 0.801 to get 0.8.

YakumoChen:
My concern is that the typical way to change the number is to drag the "slider" and that's going to make a mess with a field that has a ton of decimals in each step

GoldenAlpharex:
I feel like there's a better way to do it than this.

PR #9762, opened 2021-11-28
[MIRROR] Yarn 3.1 [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63088
--------------------
Updates Yarn to 3.1 which fixes some relevant to us issues:

- Makes reads/writes to Yarn cache fully atomic (which means corruptions due to race/network conditions are going to be mostly gone).
- Cache access is also much faster, which means yarn installs faster.
- They fixed more problems with path escaping, so yarn shouldn't break if user has a space character in their folders.
- Typescript 4.5 support (but someone has to upgrade that separately).

[Full changelog](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md)

We live in the future, baby.


GoldenAlpharex (approved):
Nice!(?)

PR #9767, opened 2021-11-29
[MODULAR] Uh oh, horrorform stat touching, oh noo oh noooo

RatFromTheJungle:
## About The Pull Request

Opening this for general feedback, but it's gone untouched for a while, and isn't upstream at all regarding antags, but with secs generally lethality, and access to g u n lowered, these are a bit overdue for a tone-back.

this shit has 750 total health, and instantly starts at 500, that's like 5 entire magazines of a pitbull with 0 misses, as well as having ten damage above an esword, with like, double the attack-speed, which during highpop is like being thanos with four of the stones.

## Changelog

:cl:
balance: rebalanced (nerfed hard) horrorform health, and melee damage, and removes it's ventcrawling, but letting people swap in and out of it more often, making it better for ambushes
/:cl:

### DIFF
--- a/modular_skyrat/modules/horrorform/code/true_changeling.dm
+++ b/modular_skyrat/modules/horrorform/code/true_changeling.dm
@@ -1,4 +1,4 @@
-#define TRUE_CHANGELING_REFORM_THRESHOLD 2 MINUTES
+#define TRUE_CHANGELING_REFORM_THRESHOLD 40 SECONDS // 40s so they cant jump in and out of it to reset the health
 #define TRUE_CHANGELING_PASSIVE_HEAL 3 //Amount of brute damage restored per tick
 
 //Changelings in their true form.
--- a/modular_skyrat/modules/horrorform/code/true_changeling.dm
+++ b/modular_skyrat/modules/horrorform/code/true_changeling.dm
@@ -20,7 +20,7 @@
 	status_flags = CANPUSH
 	atmos_requirements = list("min_oxy" = 0, "max_oxy" = 0, "min_tox" = 0, "max_tox" = 0, "min_co2" = 0, "max_co2" = 0, "min_n2" = 0, "max_n2" = 0)
 	minbodytemp = 0
-	maxHealth = 500 //3 3/4ths crewman worth of health, vaguely.
+	maxHealth = 500 //3 and 3/4ths crewmen worth of health, vaguely.
 	health = 200 
 	healable = FALSE
 	see_in_dark = 8
@@ -56,7 +56,6 @@
 	devour = new /datum/action/innate/devour
 	turn_to_human.Grant(src)
 	devour.Grant(src)
-	ADD_TRAIT(src, TRAIT_VENTCRAWLER_ALWAYS, INNATE_TRAIT)
 
 /mob/living/simple_animal/hostile/true_changeling/Life()
 	. = ..()
--- a/modular_skyrat/modules/horrorform/code/true_changeling.dm
+++ b/modular_skyrat/modules/horrorform/code/true_changeling.dm
@@ -1,4 +1,4 @@
-#define TRUE_CHANGELING_REFORM_THRESHOLD 5 MINUTES
+#define TRUE_CHANGELING_REFORM_THRESHOLD 2 MINUTES
 #define TRUE_CHANGELING_PASSIVE_HEAL 3 //Amount of brute damage restored per tick
 
 //Changelings in their true form.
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
pre-emptive "go away bot"

dawsonkeyes:
requires 4 total DNA points, 15 different DNA stings, and at least one absorb- lower those requirements if you're hitting it this hard. i'm pretty sure the normal goddamn armblade is equal to an esword if not better

RatFromTheJungle:
> requires 4 total DNA points, 15 different DNA stings, and at least one absorb- lower those requirements if you're hitting it this hard. i'm pretty sure the normal goddamn armblade is equal to an esword if not better

the normal armblade doesn't make you stamina+stun immune, with all the benefits of a simple mod including ventcrawl and a massively increased health pool that you can still fleshmend during. 👍

SpaceVampire:
> the normal armblade doesn't make you stamina+stun immune, with all the benefits of a simple mod including ventcrawl and a massively increased health pool that you can still fleshmend during. 👍

Wait, you can still fleshmend during it? I thought it took away your other powers during the transformation. Can you use other stuff as well, if that is the case? 

Also in terms of balance. Just remove the vent-crawling already and make it faster to compensate. Because vent-crawling is a cancer power for anyone to have and leads to stuff like venting into the vault, or doing hit-and-run attacks from what should be a cool final stand power.

RatFromTheJungle:
> > the normal armblade doesn't make you stamina+stun immune, with all the benefits of a simple mod including ventcrawl and a massively increased health pool that you can still fleshmend during. 👍
> 
> Wait, you can still fleshmend during it? I thought it took away your other powers during the transformation. Can you use other stuff as well, if that is the case?
> 
> Also in terms of balance. Just remove the vent-crawling already and make it faster to compensate. Because vent-crawling is a cancer power for anyone to have and leads to stuff like venting into the vault, or doing hit-and-run attacks from what should be a cool final stand power.

during a quick test in the t-dome, yes you can infact fleshmend during it, and deargodno i'm not making this thing >faster< the only downside it has right now, is that you might be able to outwalk it if your high mood and it's across a decently sized room, I don't want to make it an even free-r go-go-gadget murderbone ability, i might just take the ventcrawling and give it nothing instead.

dawsonkeyes:
just remove it at this point, i can use adrenaline sacs + armblade to be even more potent without being immediately kill-on-sight

RatFromTheJungle:
> just remove it at this point, i can use adrenaline sacs + armblade to be even more potent without being immediately kill-on-sight

read, and promptly disregarded, this still has niche uses, it doesn't have to be blatantly overpowered to exist.

GoldenAlpharex (changes requested):
Okay since you're in there, scroll down in that file, there's some code that's absolutely not up to standards, please do some cleaning up while you're in there :)

modular_skyrat/modules/horrorform/code/true_changeling.dm
```diff
@@ -20,14 +20,14 @@
...
+	maxHealth = 500 //3 and 3/4ths crewmen worth of health, vaguely.
+	health = 200 
 	healable = FALSE
 	see_in_dark = 8
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
 	environment_smash = TRUE
-	melee_damage_lower = 40
-	melee_damage_upper = 40
+	melee_damage_lower = 30
+	melee_damage_upper = 30 // equal to an esword, coupled with their faster attacks this is more then enough.
```
tf-4:
```suggestion
	melee_damage_upper = 30 // equal to an esword, coupled with their faster attacks this is more than enough.
```

modular_skyrat/modules/horrorform/code/true_changeling.dm
```diff
@@ -20,14 +20,14 @@
 	status_flags = CANPUSH
 	atmos_requirements = list("min_oxy" = 0, "max_oxy" = 0, "min_tox" = 0, "max_tox" = 0, "min_co2" = 0, "max_co2" = 0, "min_n2" = 0, "max_n2" = 0)
 	minbodytemp = 0
-	maxHealth = 750 //Very durable
-	health = 500
+	maxHealth = 500 //3 and 3/4ths crewmen worth of health, vaguely.
+	health = 200
```
tf-4, line 24:
```suggestion
	health = 300 
```
Seems low

Gandalf2k15 (changes requested):
remove the restriction on going back to human form

modular_skyrat/modules/horrorform/code/true_changeling.dm
SpaceVampire: I think now that you (finally) removed vent-crawling, it is fine to raise the health again in return.

axietheaxolotl:
Incredibly based PR 👍

modular_skyrat/modules/horrorform/code/true_changeling.dm
RatFromTheJungle: I would, but 200 leaves it 30 shy of being able to tank an entire pitbull default cell, which I think is a good point for something they can swap in and out of

PR #9775, opened 2021-11-30
Removes the Alt-titles from Medical Doctor and Scientist due to causing intense confusion among players about what you're expected/allowed to do as a Job.

Iamgoofball:
## About The Pull Request

Removes the Alt-titles from Medical Doctor and Scientist.

Folks just are not seeing the bit when they spawn about how the title on the ID card is purely a flavor thing, and we're running into situations where none of the Scientists will hit R&D buttons because they're all Xenobiologists and Circuitry Designers, and none of the doctors will do surgery because they're all Nurses, either because they're afraid of getting banned for it because they don't know it doesn't actually matter, or because they want an excuse to not do their job.

This is not good for these jobs, unfortunately.

## Changelog

:cl:
del: Removes the Alt-titles from Medical Doctor and Scientist due to causing intense confusion among players about what you're expected/allowed to do as a Job.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/scientist.ts
+++ b/tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/scientist.ts
@@ -5,17 +5,6 @@ const Scientist: Job = {
   name: "Scientist",
   description: "Do experiments, perform research, feed the slimes, make bombs.",
   department: Science,
-  alt_titles: [
-    "Scientist",
-    "Circuitry Designer",
-    "Xenobiologist",
-    "Cytologist",
-    "Nanomachine Programmer",
-    "Plasma Researcher",
-    "Anomalist",
-    "Lab Technician",
-    "Theoretical Physicist",
-  ],
 };
 
 export default Scientist;
--- a/tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/medical_doctor.ts
+++ b/tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/medical_doctor.ts
@@ -6,7 +6,6 @@ const MedicalDoctor: Job = {
   description: "Save lives, run around the station looking for victims, \
     scan everyone in sight",
   department: Medical,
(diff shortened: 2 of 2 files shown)


tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/scientist.ts
```diff
@@ -5,17 +5,6 @@ const Scientist: Job = {
...
-  alt_titles: [
-    "Scientist",
-    "Circuitry Designer",
-    "Xenobiologist",
-    "Cytologist",
-    "Nanomachine Programmer",
-    "Plasma Researcher",
-    "Anomalist",
-    "Lab Technician",
-    "Theoretical Physicist",
```
ghost, line 17: Surely some of these could stay or be replaced? "Lab Technician", and a more generic "Researcher"?

Gandalf2k15 (changes requested):
Leave the alt titles that are just other ways of saying the primary role

GoldenAlpharex:
Well, I'd rather put my foot down and say that "No, alternative job titles don't give you dibs to a certain part of the job, and don't prevent you from doing essential parts of your job either." If 6 scientists show up as Xenobiologists, well, tough luck, they won't all do Xenobiology. It's their specialization, sure, but they can do everything else that's important as well.

Iamgoofball:
@GoldenAlpharex I agree, but we'd need some way to, like, communicate this to players better that the alt titles are pure flavor.

Stonetear:
It doesn't appear to exist any more, but at one point there was a text that said "_reguardless of what your job title may be, work to fulfill these objectives_," or something like that.
Rather than remove some nice bits of flavor from the game, why not just re-add that part?

tgui/packages/tgui/interfaces/PreferencesMenu/jobs/jobs/scientist.ts
softcerv: Lab Technician makes sense to have stay in

Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/9845

PR #9785, opened 2021-11-30
You can now shift layers even while lying down

GoldenAlpharex:
## About The Pull Request
The title says it all, really. I don't know why this check was there, it was probably not really thought through that much. I don't see many times where this wouldn't be helpful to have, and don't really see how it could be abused, so...

Sometimes, you're laying down in the middle of something very wholesome and you just want to change that damn layer, man.

## Changelog

:cl: GoldenAlpharex
qol: You can now switch layers upwards or downwards while lying down.
/:cl:

### DIFF
--- a/modular_skyrat/modules/layer_shift/code/mob_movement.dm
+++ b/modular_skyrat/modules/layer_shift/code/mob_movement.dm
@@ -7,7 +7,7 @@
 	set name = "Shift Layer Upwards"
 	set category = "IC"
 
-	if(incapacitated() || body_position == LYING_DOWN)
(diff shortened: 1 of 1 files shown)


Stalkeros2:
Finally. 
*hides under the table.*

GoldenAlpharex:
> Finally. _hides under the table._

I doubt you can do that, plus if you can, that was before this change, this just makes it so you can use the verb when prone, it was still affected when you weren't prone.

Gandalf2k15 (approved):

PR #9789, opened 2021-11-30
[TM ONLY] Temporarily disables the Gulag Teleporter console as whether is it a Rules Violation to use is currently pending a Player Council appeal, so that more players do not step on the bear trap that is the Gulag and get secbanned for it.

Iamgoofball:
Until the next Security Policy update, which is in an unknown amount of time, I do not want players stepping on this bear trap


GoldenAlpharex:
No need for this.

Iamgoofball:
issue with the gulag resolved via speedy action by policy makers 🥳

PR #9872, opened 2021-12-04
CI Fix (It doesn't randomly disappear anymore)

ZephyrTFA:
Changes the way the combat indicator is applied to a mob/living to be done by the system designed for handling overlays.

### DIFF
--- a/modular_skyrat/modules/indicators/code/combat_indicator.dm
+++ b/modular_skyrat/modules/indicators/code/combat_indicator.dm
@@ -17,6 +17,9 @@ GLOBAL_VAR_INIT(combat_indicator_overlay, GenerateCombatOverlay())
 
 /mob/living/proc/combat_indicator_unconscious_signal()
 	SIGNAL_HANDLER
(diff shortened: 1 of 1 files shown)


Seris02:
If I'm reading it correctly, this will cause issues. ie. never being able to turn off the combat indicator, but I dunno, might just be being sleepy.

nikothedude:
What exactly does this fix? Addiction ticks disabling CI? Hugs and such as well?

nikothedude:
I only ask because I notice this doesn't change this:

/mob/living/proc/combat_indicator_unconscious_signal()
	SIGNAL_HANDLER
	set_combat_indicator(FALSE)

which is called whenever a mob processes an addiction OD or is hugged. Don't ask me why. 

I'm making my own PR to fix it, and I dont know if this'll fix it, so...

hugmeorelse:
Review of this would be excellent.

Thank you for working on this, @ZephyrTFA

Gandalf2k15 (approved):

PR #9873, opened 2021-12-04
[Modular] Fixes the missing security autolathe on Icebox

OolongCow:
## About The Pull Request
Simply replaces the unfinished machine frame in the brig of Icebox with an autolathe using settings identical to those of other stations. It being left in this state is almost certainly an oversight by mappers, which this PR fixes. As an edit to the modular map file, should not cause any modularization conflicts.

Prevents security from having to hunt down an autolathe board for no particular reason whatsoever on a single map.

## Changelog

:cl:
fix: Replaces the machine frame in Icebox's brig with the intended autolathe
/:cl:

### DIFF
--- a/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
+++ b/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
@@ -295,8 +295,8 @@
 /turf/open/floor/iron/dark,
 /area/security/office)
 "acf" = (
-/obj/structure/frame/machine,
 /obj/structure/reagent_dispensers/wall/peppertank/directional/south,
+/obj/machinery/autolathe,
 /turf/open/floor/iron,
 /area/security/office)
 "aci" = (


GoldenAlpharex:
I don't know, but as far as I can remember, no other map starts with an autolathe in the security department? I can tell you for sure Meta doesn't, I just like, triple-checked.

OolongCow:
> I don't know, but as far as I can remember, no other map starts with an autolathe in the security department? I can tell you for sure Meta doesn't, I just like, triple-checked.

I know Blueshift does, as that's the map I took the autolathe from to make sure no variables were changed and needed to go with it. This may be a consistency issue that needs to be addressed, then, which requires a ruling by whoever's in charge of this stuff as far as intended security equipment goes.

tf-4:
This is an outlier on Blueshift and should be addressed there

GoldenAlpharex:
> This is an outlier on Blueshift and should be addressed there

Yeah, I actually agree with this, Blueshift shouldn't have that autolathe there.

GoldenAlpharex:
So, unless you want to turn this PR into a PR to remove the autolathe from Blueshift (you can make it a machine frame if you want), we're going to be closing this one.

tf-4:
As Golden said

PR #9903, opened 2021-12-06
Adjusts donator gas mask

tf-4:
Removes tint and pepperspray protection from a donator mask as it does not cover the eyes
Adjustment missed in tint PR


dawsonkeyes:
Shouldn’t all sec gas masks not be pepper proof? They’re more like variants of the breathing mask and don’t cover the eyes visually

tf-4:
> Shouldn’t all sec gas masks not be pepper proof? They’re more like variants of the breathing mask and don’t cover the eyes visually

This is already the case, and is also not relevant to this PR.

GoldenAlpharex (approved):

PR #9914, opened 2021-12-06
[NON-MODULAR] Allows (some) Heads of Staff to be blind.

CandleJaxx:
## About The Pull Request

Removes "Blind" from HEAD_RESTRICTED_QUIRKS. Does not remove it from SEC_RESTRICTED_QUIRKS, as to ensure that jobs which require full field of view aren't overlooked. 

Restricts CE and CMO from being able to be blind.

While being blind, you can still access and control computers, tablets, and other interactables. This will not remove the Head of Personnel, for example's, ability to use their ID console and other job-critical abilities.

Allowing for blind characters in more roleplay-situated positions could open up new scenarios to take place.

## Changelog

:cl:

qol: NanoTrasen has opened hiring positions for non-combat-critical positions for blinded personnel.

/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -6,4 +6,4 @@
 
 #define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE
 #define HEAD_RESTRICTED_QUIRKS "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE
-#define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE
+#define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE, "Blind" = TRUE
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -5,5 +5,5 @@
 #define JOB_UNAVAILABLE_LANGUAGE JOB_UNAVAILABLE_SPECIES + 1
 
 #define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE
-#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE
+#define HEAD_RESTRICTED_QUIRKS "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE
 #define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -5,5 +5,5 @@
 #define JOB_UNAVAILABLE_LANGUAGE JOB_UNAVAILABLE_SPECIES + 1
 
 #define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE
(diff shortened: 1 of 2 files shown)


GoldenAlpharex:
That doesn't make that much sense, whilst yes for gameplay reasons, blind people can still use consoles, it makes no sense that they'd actually be able to. It's not written in braille, you can't see shapes or buttons.

CandleJaxx:
yeah, but also, who says you can't rp in those consoles having visually-impaired friendly interfaces? besides, that's out of the scope of this pr

ChungusGamer666:
This makes no sense at all. **Why would NanoTrasen put people who are basically a liability in such important positions?**
The head of personnel is a target for terrorism, it'd be really bad if the second in the chain of command had to be babied around because he can't take care of himself.
Having a blind HoS is... do i need to explain

code/__DEFINES/~skyrat_defines/jobs.dm
```diff
@@ -5,5 +5,5 @@
 #define JOB_UNAVAILABLE_LANGUAGE JOB_UNAVAILABLE_SPECIES + 1
 
 #define SEC_RESTRICTED_QUIRKS "Blind" = TRUE, "Brain Tumor" = TRUE, "Deaf" = TRUE, "Paraplegic" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Pacifist" = TRUE, "Chunky Fingers" = TRUE
-#define HEAD_RESTRICTED_QUIRKS "Blind" = TRUE, "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE
-#define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE
+#define HEAD_RESTRICTED_QUIRKS "Deaf" = TRUE, "Mute" = TRUE, "Foreigner" = TRUE, "Chunky Fingers" = TRUE // SKYRAT EDIT
+#define TECH_RESTRICTED_QUIRKS "Chunky Fingers" = TRUE, "Blind" = TRUE // SKYRAT EDIT
```
GoldenAlpharex, line 9: So you're making all engineers and atmos techs unable to be blind?
I don't necessarily disagree, but it's outside of the scope of your current PR.

GoldenAlpharex: All in all, what this does is that it allows the RD, the QM (which might already be able to regardless), the HoP and the Captain and the NT Rep to be blind.

NT Rep is supposed to watch over Command, can't exactly "watch" when you're blind. HoP needs to be able to see people to properly man their line. I disagree with a blind Captain, I feel like it's just going to cripple Command too much. QM, I couldn't care less, but you can't do paperwork whilst blind, I'm pretty sure. RD, well, science experiments require you to be able to see things, you also can't do xenobio (or at least, shouldn't be able to), since you kinda need to look into the slime management console.

I'm really not convinced by this one, chief.

modular_skyrat/modules/customization/modules/jobs/_job.dm
```diff
@@ -81,10 +81,10 @@
 	banned_quirks = list(SEC_RESTRICTED_QUIRKS, HEAD_RESTRICTED_QUIRKS)
 
 /datum/job/chief_medical_officer
-	banned_quirks = list(HEAD_RESTRICTED_QUIRKS)
+	banned_quirks = list(HEAD_RESTRICTED_QUIRKS, "Blind" = TRUE) // SKYRAT EDIT
 
 /datum/job/chief_engineer
-	banned_quirks = list(HEAD_RESTRICTED_QUIRKS, "Paraplegic" = TRUE)
+	banned_quirks = list(HEAD_RESTRICTED_QUIRKS, TECH_RESTRICTED_QUIRKS, "Paraplegic" = TRUE) // SKYRAT EDIT
```
GoldenAlpharex, line 87: This one is a good change, they should've had that to begin with, that was likely an oversight(?)

GoldenAlpharex (changes requested):

code/__DEFINES/~skyrat_defines/jobs.dm
CandleJaxx: adding it to tech was required for giving ce tech restricted flags without bloating the file too much

CandleJaxx:
> Having a blind HoS is... do i need to explain

do read the full pr body- hos requires both head_restricted AND sec_restricted, which doesn't allow blindness anyway and wasn't being changed

code/__DEFINES/~skyrat_defines/jobs.dm
CandleJaxx: > NT Rep is supposed to watch over Command, can't exactly "watch" when you're blind. HoP needs to be able to see people to properly man their line. I disagree with a blind Captain, I feel like it's just going to cripple Command too much. QM, I couldn't care less, but you can't do paperwork whilst blind, I'm pretty sure. RD, well, science experiments require you to be able to see things, you also can't do xenobio (or at least, shouldn't be able to), since you kinda need to look into the slime management console.

in regard to this, i don't see any particular reason why the hop and nt rep, at least, shouldn't be allowed to be blind. that kind of thing prevents a large part of the potential workforce from doing their work, and from an ooc standpoint, just seems needlessly constraining. if someone wants to make it harder on themselves for a reason, and their job allows for full functionality despite that, i see no reason why it shouldn't be at least properly entertained

axietheaxolotl: HoP's whole job is paperwork, mechanically, sure, you can do paperwork blind. But ICly that doesn't make any sense.

tf-4:
After conferring with other maintainers, I don't think this is currently something we're interested in. It limits their ability too much.

PR #9917, opened 2021-12-07
Brief Blueshift balance pass

tf-4:
Removes roundstart autolathes and multi-cell chargers

Replaces multi-chargers with regular ones, and autolathes with like, plants, or machine frames, or whatever.

Also removes spare psych skillchip. They spawn with one anyway, there's only ever one, and it's not key like the RD's.

Cargo keeps their autolathe, naturally.

## Changelog
:cl:
del: Blueshift now only has one autolathe shift-start, located in cargo.
del: Shift-start multi-cell chargers are now only single-cell chargers on Blueshift.
del: The psychologist no longer has a spare skillchip on Blueshift.
/:cl:

### DIFF
--- a/_maps/map_files/Blueshift/BlueShift_middle.dmm
+++ b/_maps/map_files/Blueshift/BlueShift_middle.dmm
@@ -734,8 +734,9 @@
 /turf/open/floor/iron,
 /area/cargo/office)
 "aic" = (
-/obj/machinery/autolathe,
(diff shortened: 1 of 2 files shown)


softcerv:
why are autolathes being removed, I consider them being there as more of a QoL than anything else tbh

tf-4:
Powercreep

PR #9918, opened 2021-12-07
[NON-MODULAR] Adjusts Runechat layers to allow blind characters to 'see' audial messages.

CandleJaxx:
## About The Pull Request

Adjusts runechat to make for a form of sensory supplement for blind players, being able to determine where someone is talking to them from by allowing them to see the runechat bubbles.

Being able to tell if someone is roughly infront of you or behind you allows for people to better act as though they're in the world. You'd be able to tell if someone was standing infront, behind,l or near you and speaking.

## Changelog

:cl:

fix: Blind players can now see runechat.

/:cl:

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -138,10 +138,10 @@
 #define CAMERA_STATIC_PLANE 200
 
 ///Popup Chat Messages
-#define RUNECHAT_PLANE 501
+#define RUNECHAT_PLANE 501 //SKYRAT EDIT - ORIGINAL 250
 
 /// Plane for balloon text (text that fades up)
-#define BALLOON_CHAT_PLANE 502
+#define BALLOON_CHAT_PLANE 502 //SKYRAT EDIT - ORIGINAL 251
 
 ///Debug Atmos Overlays
 #define ATMOS_GROUP_PLANE 450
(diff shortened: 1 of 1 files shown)


ChungusGamer666:
this will cause unintended consequences like runechat displaying above hud elements, which is really distracting
absolute thumbs down for me.

GoldenAlpharex (changes requested):
I feel like there's going to be a lot of unintended side-effects, with Runechat clipping over a lot of things. Surely, there's got to be a better way of doing this.

CandleJaxx:
hm, i can't see where it'd clip over hud elements- the only element it clips over is damage, which is the red hue around the screen, which shouldn't be over text anyway

Azarak:
`//HUD layer defines
#define HUD_PLANE 1000

#define ABOVE_HUD_PLANE 1100`

Hud is above fullscreen. There isn't a better way of doing this. It's just a matter if finding the render order you want things to be, in this case it being over all fullscreen effects (usually sight obstructions, which don't affect hearing)

CandleJaxx:
that's what i was thinking- if it shows up in chat, it should show up in runechat

CandleJaxx:
closed, to be deleted if upstream passes

PR #9966, opened 2021-12-09
[SEMI-MODULAR] Variable Viewport

nevimer:
## About The Pull Request

This is a port of the variable viewport from Nebula. I've adapted it a bit in order to work with this codebase. It's not 100% done yet, there are still some issues. This mostly for testing and review, thought I would love a testmerge. 

https://user-images.githubusercontent.com/77420409/145335029-826f7ecb-eb4e-4526-8c01-36c2423c2073.mp4

This is a massive QOL change for those who want in-between sizes for their chat and other GUI elements. I've restricted it to be one extra tile on the X and Y from previous so in-between sizes at 16:9 are not janky. 

## Changelog

:cl:
add: Viewport scale buttons.
qol: Viewport now auto resizes.
code: Changes absolute HUD vars to relative
refactor: Makes resize procs use the now client.onResize()
server: new config file options for max view range
/:cl:


tf-4:
This needs to be taken upstream. Feel free to re-open if they reject it.

PR #10067, opened 2021-12-14
[NON MODULAR] Split Personality Fix, Attempt TWO

nevimer:
## About The Pull Request

We hate split personality crashing! This attempts to fix it at (where I believe) it's crashing. 

It's a lot of fun when it works. Hate seeing it this way. 

## Changelog

:cl:

fix: Split personality crashing, begone!

/:cl:

### DIFF
--- a/code/datums/brain_damage/split_personality.dm
+++ b/code/datums/brain_damage/split_personality.dm
@@ -86,11 +86,12 @@
 	to_chat(current_backseat, span_userdanger("You manage to take control of your body!"))
 
 	//Body to backseat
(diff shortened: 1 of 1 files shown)


SpaceVampire:
I would note that the crash seems to be linked to a more fundamental problem.
Given that spells like Mindswap will cause it as well, or even an admin directly sticking someone into a mob that already has an active player.

tf-4:
This will most likely be fixed with the removal of ambitions, feel free to re-open if it isn't.

PR #9778, opened 2021-11-30
[NON-MODULAR] Medical balance changes

linnpap:
## About The Pull Request
Changes numerous aspects of medical to make it more difficult and fun.

- Not putting patients under anesthetic causes a risk of shock, and increases surgery failure rate. (Alcohol can work as a ghetto anesthetic)
- Adds in shock: condition which can quickly deteriorate into cardiac arrest. (Cured with epinephrine)
- All surgeries (besides tending wounds) take twice as long. The surgery room speed buff has also been removed.
- Increases the price of mediguns from cargo, and also removes the CMO's medigun from their locker.
- Reverts operating tables to no longer put patients into stasis.
- Filtering blood no longer removes formaldehyde (organ preservation chem).

## How This Contributes To The Skyrat Roleplay Experience
By making surgeries more difficult and longer, medical will become much more engaging to players.

The increase of medigun prices from cargo, along with the removal of the CMO's starting medigun, will make medical more engaging, as getting mediguns will become much more difficult, but incredibly rewarding.

The addition of shock and actual consequences for not using anesthetic will also now make players actually want to be put under when given surgery. Without anesthetic, surgeries will actually become longer and potentially deadly. This should now punish players who choose to get surgery without anesthetic, all in the name of the surgery going quicker (which it now won't).

The increase of surgery times will also no longer allow players to have surgeries done incredibly quickly. The increase of surgery times will make medical more realistic, and possibly make players value their IC health more.

## Changelog

:cl:
add: Added shock
add: Doing surgery without anesthetic will now increase failure rates and possibly cause shock 
balance: Medigun cargo crate price increased
balance: The CMO's starting medigun has been removed
balance: All surgeries (besides tending wounds) take twice as long
balance: Operating tables no longer put people in stasis
balance: Filtering blood no longer removes formaldehyde
code: Surgery failure rate calculations now use the old method (list of success rates in "implements" list)
/:cl:

PR #10088, opened 2021-12-15
[NON-MODULAR] Stasis tweaks + penalties for varying levels of surgery anesthesia.

Higgin:
## About The Pull Request

Adds three* new mood maluses for numbed and completely-unanaesthetized surgery. Getting surgery without anesthetic makes the next fifteen minute suck, getting surgery numbed but awake just makes you feel a wee* bit off for the next five.  This is hoped to help achieve some of the same goals set out in #9778 and its follow-on PR while respecting that the hardest penalties for edge cases were a bit much to bear (and beyond what I'm interested in fixing.) 

Stasis now preserves sleeping while under the effect of stasis; it also duplicates and extends the wakeful anesthetic effects of alcohol, morphine, lidocaine, miner's salve, and drunken espatier if you go onto stasis with them in effect. 

Adds an incentive for the person on the operating table to take an increased level of vulnerability while getting worked on while not completely removing the benefits of medical prep (alcohol, lidocaine and miner's salve all apply numbing without putting the recipient to sleep.) However, getting put to sleep remains the only pure option for avoiding penalties.

The value of vulnerability is that it creates windows for antagonism. The way it's introduced here also rewards the specialization of medical facilities or having substances/tools to put people to sleep as part of doing surgery. Doing this as a mood malus creates a mechanical incentive for seeking ways to rectify mood if you've just been malpractice'd (say, asking your local psych for some Happiness, drinking, getting a good meal, hugging your friends, or enjoying some clown antics - things which draw you further into other systems/departments) but crucially doesn't make you unable to play the game if you choose to ignore it. 

## Changelog

:cl:
balance: Doing surgery on somebody without first making them unconscious now leaves them with a lingering mood penalty. The mood penalty's time and intensity can be reduced with numbing agents like copious amounts of alcohol, Miner's Salve, Lidocaine, and Morphine _if used prior to surgery._
add: Roboticists can now purchase two forms of anesthetic in their job wardrobe vendors (N2O gas/morphine.)
balance: Things which apply stasis now maintain unconsciousness if you go into stasis unconscious and numbing effects if you go into stasis numbed. 
/:cl:


code/modules/surgery/surgery_step.dm
```diff
@@ -202,9 +202,13 @@
  */
 //SKYRAT EDIT START: Fixes painkillers not actually stopping pain.
 /datum/surgery_step/proc/display_pain(mob/living/target, pain_message, mechanical_surgery = FALSE)
-	if(HAS_TRAIT(target, TRAIT_NUMBED) || target.stat >= UNCONSCIOUS)
+	if(target.stat >= UNCONSCIOUS || mechanical_surgery == TRUE) //robots and the unconscious do not worry about pain
```
Iamgoofball: not keen on more stupid synthetic buffs

Higgin: Hopefully addressed with latest commit - robots really have no reason to not just Sleep even if they can't process numbing reagents, so they now get a penalty that falls in between the other two if they don't.

YakumoChen:
These numbers are not balanced at all and I don't have any belief you have any idea what moodlet numbers should look like.

Most mood buffs of this caliber are ones you get for things that are literally killing you, like losing a limb or being on fire, numbers like -15 are absolutely sure to tank your mood to rock bottom for an obscene amount of time (15 minutes is way too long for huge mood debuffs)

For reference, the positive mood buff for _being an antagonist_ ranges from somewhere between +3 to +5 for most of them, and those are generally supposed to cancel out most negative debuffs that aren't very temporary.

YakumoChen:
also check the code: surgery already has a moodlet.

Higgin:
> I don't have any belief you have any idea what moodlet numbers should look like.

Will you or will you not take these numbers into account when playing? I did check them against the others - if they seem dramatic, it's because that's kind of the point. With the way mood works, they're meaningful debuffs - they're supposed to last for a while so you have a reason to respect them enough to be vulnerable - but they should not be disabling or increase your time in medical. Sanity (where most of the effects from mood live) is pretty flexible with mood level unless you've taken Unstable and can also be mitigated. 

> also check the code: surgery already has a moodlet.

Unless you're thinking of something else, the one in the file is unused in code - it also doesn't have a timeout attached to it, so I added a couple of levels instead.

YakumoChen:
Having such obscene, "dramatic" buffs for surgery is overdoing it to an extreme. Sanity is supposed to be balanced. Big negatives should have short cooldowns or be associated with permanent or dangerous conditions like losing a limb or literally dying. Having people dropping sanity towards suicidal ends for a significant chunk of the round is extremely un-ideal and not fun at all to play with.

Iamgoofball:
@YakumoChen I do regularly have issues with patients refusing to use anesthetics. This would provide me a reason to tell them I won't treat them if they refuse anesthetic.

YakumoChen:
@Iamgoofball you can't look at this and not tell me the numbers are _really dumb_. 

This is subjectively worse than the last PR, you are still basically crippling someone for a vast length of time if you don't have anesthetic for whatever reason, it essentially shits hard on ghetto surgery, but in a more underhanded way.

Iamgoofball:
Yeah, the penalties are a bit harsh.

code/datums/mood_events/generic_negative_events.dm
```diff
@@ -356,3 +356,20 @@
 	description = "<span class='boldwarning'>I hate being sprayed with water!</span>\n"
 	mood_change = -1
 	timeout = 30 SECONDS
+
+//SKYRAT EDIT START: Mainly surgery for now.
+/datum/mood_event/mild_surgery
+	description = "<span class='warning'>Even if I couldn't feel most of it, it feels wrong being awake while somebody cuts on your body. Ugh!</span>"
+	mood_change = -5
```
Iamgoofball: We shouldn't be punishing players for using anesthetic of any kind, just scrap this one, or make it booze only or something?

code/datums/mood_events/generic_negative_events.dm
```diff
@@ -356,3 +356,20 @@
...
+	timeout = 5 MINUTES
+
+/datum/mood_event/severe_surgery
+	description = "<span class='boldwarning'>Wait, THEY CUT ME OPEN - AND I FELT EVERY SECOND OF IT!</span>"
+	mood_change = -15
```
Iamgoofball:
```suggestion
	mood_change = -3
```

code/datums/mood_events/generic_negative_events.dm
```diff
@@ -356,3 +356,20 @@
...
+	timeout = 15 MINUTES
+
+/datum/mood_event/robot_surgery
+	description = "<span class='warning'>Having my robotic parts messed with while I was conscious felt wrong... if only I had a sleep mode!</span>"
+	mood_change = -10
```
Iamgoofball:
```suggestion
	mood_change = -3
```

jwc1015:
Erm, I have a pretty big question - what about plasmamen? Are they just outta luck and gotta use a shower for any surgery with anesthetic?

Higgin:
> Erm, I have a pretty big question - what about plasmamen? Are they just outta luck and gotta use a shower for any surgery with anesthetic?


no
inject morphine or apply a patch
most surgeries where plasmapeeps are involved are going to have them very, very dead anyway, but for instance, brain surgery for a severe trauma: if they self-apply morphine, lido, use a salve patch on their head, or get really drunk, NUMBED kicks in and stays active as long as they're not on stasis.

jwc1015:
> > Erm, I have a pretty big question - what about plasmamen? Are they just outta luck and gotta use a shower for any surgery with anesthetic?
> 
> no inject morphine or apply a patch most surgeries where plasmapeeps are involved are going to have them very, very dead anyway, but for instance, brain surgery for a severe trauma: if they self-apply morphine, lido, use a salve patch on their head, or get really drunk, NUMBED kicks in and stays active as long as they're not on stasis.

Sorry if I missed something, but the main point of my comment was that anesthetic DOESN'T work in stasis, and actually safe plasmaman surgery (without fire) usually requires stasis of some form, so I'm guessing plasmamen just get either burns or mood debuffs whenever they get surgery done.
This has been an issue for plasmamen (especially those trying to RP a fairly reasonable character) for a while, sorry I'm a bit frustrated.

Higgin:
>Sorry if I missed something, but the main point of my comment was that anesthetic DOESN'T work in stasis, and actually safe plasmaman surgery (without fire) usually requires stasis of some form, so I'm guessing plasmamen just get either burns or mood debuffs whenever they get surgery done.
This has been an issue for plasmamen (especially those trying to RP a fairly reasonable character) for a while, sorry I'm a bit frustrated.

Imagine my frustration that an already-busted, snowflakey species played by about two people (and known how to be helped by about the same number in med) is a constant stumbling block for changing anything medical-related, but really, thank you.

Stasis changes now in will allow you to either get anesthetized or simply numbed and keep that status while under stasis. You'll probably still get set on fire a bit and have to be under a shower for limb replacements, organ manip., brain surgery - at least plasmapeeps were in my testing, but that's not unusual for plasma people right now anyway. Point is, they can avoid the debuffs while still having somewhat necessary access to stasis now.

code/datums/mood_events/generic_negative_events.dm
gonenoculer5: good fucking lord why was that -10

PR #10090, opened 2021-12-15
[NON-MODULAR] Nerfs the medibot by increasing the time it takes per tend wounds.

Higgin:
## About The Pull Request

Medibots now require that you sit still for five* times as long in order to get the benefits of their healing - which, unupgraded, isn't very much per cycle. 

The Luddites had a point. Automated and in many cases entirely sufficient healing from medibots has led to their being stacked up as a quick-and-easy replacement for quite a bit of interaction with medical or the pursuit of otherwise finite resources (chems, med supplies, healing plants, etc.) While it is okay for them to be resourceless, this shouldn't make them a dominant strategy for treating basic, non-wound damage - especially unupgraded. This PR seeks to incentivize more interaction with medical, cargo, or botany (at least their resources) as alternatives to the time you'll spend getting healing from medibots alone.

## Changelog

:cl:
balance: Medibot healing time increased by a factor of five.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/medbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/medbot.dm
@@ -534,7 +534,7 @@
 			C.visible_message(span_danger("[src] is trying to tend the wounds of [patient]!"), \
 				span_userdanger("[src] is trying to tend your wounds!"))
 
-			if(do_mob(src, patient, 20)) //Slightly faster than default tend wounds, but does less HPS
+			if(do_mob(src, patient, 10 SECONDS)) //SKYRAT EDIT: Increased time as tradeoff for automated healing. ORIGINAL: if(do_mob(src, patient, 20))
 				if((get_dist(src, patient) <= 1) && (bot_mode_flags & BOT_MODE_ON) && assess_patient(patient))
 					var/healies = heal_amount
 					var/obj/item/storage/firstaid/FA = firstaid


code/modules/mob/living/simple_animal/bot/medbot.dm
```diff
@@ -534,7 +534,7 @@
 			C.visible_message(span_danger("[src] is trying to tend the wounds of [patient]!"), \
 				span_userdanger("[src] is trying to tend your wounds!"))
 
-			if(do_mob(src, patient, 20)) //Slightly faster than default tend wounds, but does less HPS
+			if(do_mob(src, patient, 200)) //SKYRAT EDIT: Increased time as tradeoff for automated healing.
```
Iamgoofball: document the old value (20) in the comment

GoldenAlpharex: Here you go.
```suggestion
			if(do_mob(src, patient, 20 SECONDS)) //SKYRAT EDIT: Increased time as tradeoff for automated healing. ORIGINAL: if(do_mob(src, patient, 20))
```

ZephyrTFA:
Medibots are already nerfed by virtue of not being able to be upgraded with research

code/modules/mob/living/simple_animal/bot/medbot.dm
Higgin: Thank you both! Should now be in line with this.

Higgin:
> Medibots are already nerfed by virtue of not being able to be upgraded with research

This nerf was removed earlier by an upstream change, they presently scale up to doing about 10 per tend at max upgrades from Wound Tending research.

YakumoChen:
then re-instate the nerf?

this essentially just nukes it into the floor.

Higgin:
> then re-instate the nerf?
> 
> this essentially just nukes it into the floor.

Yeah, it's a hard nerf because it shouldn't be a dominant strategy - the better tradeoff for a functionally infinite and interactionless pocket healer (one that doesn't remove the reason for going back to sync/upgrade medibots during the round) is time, imo. 

It'll still heal you, just plan to spend some time waiting around with it if you don't want to spend resources - and maybe don't count on it to pull you up from hardcrit on its own.

softcerv:
> then re-instate the nerf?
> 
> this essentially just nukes it into the floor.

I'm not sure how this is going to affect them in practicem but they should be very heavily nerfed though especially with how strong they are in comparison to the amount of involvement that medical staff has to have when using them.

YakumoChen:
Medical has been buffed to make wound tending extremely fast and easy, so I don't see a reason why the bots have to take obscenely long to do anything of value.

softcerv:
> Medical has been buffed to make wound tending extremely fast and easy, so I don't see a reason why the bots have to take obscenely long to do anything of value.

wound tend surgery involves actual human involvement, you can't (without getting bwoinked) wound tend people in the lobby

softcerv:
also wound tending doesn't heal toxins or oxygen damage

YakumoChen:
> also wound tending doesn't heal toxins or oxygen damage

one heals entirely by itself and the other is rather easy to make very strong chems to fix. Having medibots that heal either is not a huge balance concern. Its not like medibots are being used in combat of any sort or that people are dragging them around. Most of the time they're just used to augment surgery or heal people in the lobby to not waste medbay's time. Having to wait around in the lobby for 5 minutes for a medibot to tend basic wounds is not anyone's idea of a fun time and makes them utterly useless, you might as well just remove them if you're going to make it take _20 seconds_ to do a single round of tend wounds.

CandleJaxx:
honestly this feels like smth you should do upstream first imo

softcerv:
> honestly this feels like smth you should do upstream first imo

having medibots nerfed makes way more sense on here then it does on TG since medical here isn't as fast paced

Higgin:
Switched it down to 10 seconds from 20. Original comment by tf4 on Discord mentioned a 5x increase, so, this is that - and although the 20s tend time didn't feel bad (still worked pretty good with the drum circle,) this still goes well towards the intended effect.

PR #10075, opened 2021-12-14
[MIRROR] Super Mega Mob Spawn Refactor [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63279
--------------------

## About The Super Hyper Ultra Ultimate Deluxe Perfect Amazing Shining Mob Spawn Refactor

The Super Hyper Ultra Ultimate Deluxe Perfect Amazing Shining Mob Spawn Refactor is my attempt to clean up the file structure, the code, and the type tree for mob spawns.

* Splits mob spawn types into corpses (dead spawns) and ghost roles (living spawns you can possess). The vars that didn't make sense for corpses and vice versa for ghost roles are now appropriately there
* Because of above, there are no longer the fucking "death, roundstart, and instant" vars. thank god
* Removes a lot of single or very few used vars, whose properties can be applied on special().
* All Mob Spawns are given fitting folders instead of just being stuck in a single ghost roles file. Corpses are in the corpse folder, Ghost Roles are in the ghost role folder. Only exception are drones which should stay near their respective homes
* Just generally cleaner all around you know
* spider structures file renamed to spiderwebs now that spider eggs are gone

## Why Super Hyper Ultra Ultimate Deluxe Perfect Amazing Shining Mob Spawn Refactor Is Good For The Game

The Super Hyper Ultra Ultimate Deluxe Perfect Amazing Shining Mob Spawn Refactor cleans up so many terrible cases and uses

## Changelog For The Super Hyper Ultra Ultimate Deluxe Perfect Amazing Shining Mob Spawn Refactor

:cl: armhulen
refactor: Mob spawns are refactored, no more assortment of "random, instant, and roundstart" vars on every mob spawn type
refactor: if there are some minimal differences in how mob spawners feel, that's why!
/:cl:

PR #10094, opened 2021-12-16
[DNM] Updates modular ghost roles to be fixed with #10075

axietheaxolotl:
## About The Pull Request

they now use the proper file type so they don't get broken.

Don't merge until #10075 is merged.

it doesn't.

## Changelog

:cl:
fix: Modular Ghost Roles now use the proper typepath.
/:cl:


tf-4:
While I appreciate the intent, this needs a much deeper pass which means this will be done as part of merging that PR anyway

PR #10107, opened 2021-12-17
[MIRROR] Fixes Massive Radio Overtime, Implements a Spatial Grid System for Faster Searching Over Areas [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/61422
--------------------

## About The Pull Request
a month or two ago i realized that on master the reason why get_hearers_in_view() overtimes so much (ie one of our highest overtiming procs at highpop) is because when you transmit a radio signal over the common channel, it can take ~20 MILLISECONDS, which isnt good when 1. player verbs and commands usually execute after SendMaps processes for that tick, meaning they can execute AFTER the tick was supposed to start if master is overloaded and theres a lot of maptick 2. each of our server ticks are only 50 ms, so i started on optimizing this. 

the main optimization was SSspatial_grid which allows searching through 15x15 spatial_grid_cell datums (one set for each z level) far faster than iterating over movables in view() to look for what you want. now all hearing sensitive movables in the 5x5 areas associated with each spatial_grid_cell datum are stored in the datum (so are client mobs). when you search for one of the stored "types" (hearable or client mob) in a radius around a center, it just needs to 
1. iterate over the cell datums in range
2. add the content type you want from the datums to a list
3. subtract contents that arent in range, then contents not in line of sight
4. return the list

from benchmarks, this makes short range searches like what is used with radio code (it goes over every radio connected to a radio channel that can hear the signal then calls get_hearers_in_view() to search in the radios canhear_range which is at most 3) about 3-10 times faster depending on workload. the line of sight algorithm scales well with range but not very well if it has to check LOS to > 100 objects, which seems incredibly rare for this workload, the largest range any radio in the game searches through is only 3 tiles

the second optimization is to enforce complex setter vars for radios that removes them from the global radio list if they couldnt actually receive any radio transmissions from a given frequency in the first place. 

the third optimization i did was massively reduce the number of hearables on the station by making hologram projectors not hear if dont have an active call/anything that would make them need hearing. so one of hte most common non player hearables that require view iteration to find is crossed out. 

also implements a variation of an idea oranges had on how to speed up get_hearers_in_view() now that ive realized that view() cant be replicated by a raycasting algorithm. it distributes pregenerated abstract /mob/oranges_ear instances to all hearables in range such that theres at max one per turf and then iterates through only those mobs to take advantage of type-specific view() optimizations and just adds up the references in each one to create the list of hearing atoms, then puts the oranges_ear mobs back into nullspace. this is about 2x as fast as the get_hearers_in_view() on master

Fixes #62962
Fixes #62835

## Why It's Good For The Game
holy FUCK its fast. like really fucking fast. the only costly part of the radio transmission pipeline i dont touch is mob/living/Hear() which takes ~100 microseconds on live but searching through every radio in the world with get_hearers_in_radio_ranges() -> get_hearers_in_view() is much faster, as well as the filtering radios step

[benchmark (2).txt](https://github.com/tgstation/tgstation/files/7149178/benchmark.2.txt)
heres the benchmarking code i did here, at least part of it i forgot to save the timing on individual radio transmissions but it started at about 20ms like on live (most of the cost isnt sending to players, its just searching through random radios that likely wont even reach players)

Screenshot_1278

the spatial grid searching proc is about 36 microseconds/call at 10 range and 16 microseconds at 3 range in the captains office (relatively many hearables in view), the new get_hearers_in_view() was 4.16 times faster than get_hearers_in_view_old() at 10 range and 4.59 times faster at 3 range

Screenshot_1279

in this spot, a more open area with less hearables in view (middle of the tram tracks on tramstation), the new get_hearers_in_view() was 10x faster than the old one at 10 range and 4x faster at 3 range

in both cases going over all ~250 radios in the global radio list (before i cut down on the ones that shouldnt be there it was around 600) the new get_hearers_in_radio_ranges() was 3 times faster than the old one

Screenshot_1280
also heres some stats i made for a single radio transmission with the optimized code. about 4-5x better than it used to be

SSspatial_grid could be used for a lot more things other than just radio and say code, i just didnt implement it. for example since the cells are datums you could get all cells in a radius then register for new objects entering them then activate when a player enters your radius. this is something that would require either very expensive view() calls or iterating over every player in the global list and calling get_dist() on them which isnt that expensive but is still worse than it needs to be

on normal get_hearers_in_view cost the new version that uses /mob/oranges_ear instances is about 2x faster than the old version, especially since the number of hearing sensitive movables has been brought down dramatically. 
Screenshot_1589
with get_hearers_in_view_oranges_ear() being the benchmark proc that implements this system and get_hearers_in_view() being a slightly optimized version of the version we have on master, get_hearers_in_view_as() being a more optimized version of the one we have on master, and get_hearers_in_LOS() being the raycasting version currently only used for radios because it cant replicate view()'s behavior perfectly.


.github/CODEOWNERS
```diff
@@ -19,5 +19,114 @@
...
+/icons/ @Twaticus @Krysonism @Imaginos16
+/code/controllers/subsystem/air.dm @LemonInTheDark @MrStonedOne
+/_maps/ @EOBGames @ShizCalev @Maurukas
+/tools/build/ @MrStonedOne @stylemistake
+/tools/LinuxOneShot/ @Cyberboss @MrStonedOne
+/tools/tgs4_scripts/ @Cyberboss @MrStonedOne
+
+#SIC SEMPER TYRANNIS
+/code/modules/hydroponics/grown/citrus.dm @optimumtact
+>>>>>>> d005d76f0bd (Fixes Massive Radio Overtime, Implements a Spatial Grid System for Faster Searching Over Areas (#61422))
```
GoldenAlpharex: It'd be funny, but...
```suggestion
# Maptainers
/_maps/ @KathrinBailey
```

PR #10157, opened 2021-12-20
[MIRROR] Advanced camera consoles no longer crash when you search for a camera [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63394
--------------------

## About The Pull Request
Closes #63390
* I've added a check to sanitize the parameters to `tgui_input_list` so that null "button" items coming in won't break the search anymore.
* I've also changed the camera code to sanitize out the null c_tags since they wouldn't be selectable in the popup anyway.

Edit: Note that this also fixes the AI "Show Camera List" search, forgot about that, but this incidentally fixes that too regardless.

## Why It's Good For The Game

This console isn't used much anymore, but it can be placed, and it used for some templates, like the Syndicate listening outpost.

## Changelog

:cl: tastyfish
fix: The advanced camera console's and AI's camera search function now works again.
/:cl:


Tastyfish:
Merge conflict because it's reliant on #10124 being merged

PR #10169, opened 2021-12-21
[MODULAR] Adds sunglasses in Icebox Law office

kannthus:
## About The Pull Request

Adds the 2 sunglasses to the Icebox law offices, It is the _only_ map which does not have these by default (Blueshift, Delta and Meta all do)

Aside from the mechanical use-case, 

## Changelog

:cl:
add: Added sunglasses to Icebox law office, same as every other map
/:cl:


tf-4:
You'd be best taking this upstream once their freeze is over as it'll be lost in map resets
Actually are map edits even affected

Gandalf2k15:
Take it upstream.

PR #10228, opened 2021-12-24
[MDB IGNORE] Brings map files upto standard

Gandalf2k15:
# Put all maps in the _maps/ folder

This means the MDB can render them correctly, and so that CI can also check them.

Any skyrat maps must go in the modular skyrat folder.


Gandalf2k15:
now maps need to be fixed lol

PR #10234, opened 2021-12-24
[NON-MODULAR] Makes Actual Round Time... Actual Round Time

Kapu1178:
## About The Pull Request
Changes `world.realtime` to `world.timeofday`, and changes `real_round_time`'s value to be set at game start instead of SSTicker init. I don't know why you would ever use `world.realtime`, and for all you noncoders in the audience here's why:

BYOND stores numerical values as "Floating Point Numbers". They can be REALLY BIG, but also, the larger the number gets, the less accurate it becomes. Why is this relevant?

`world.realtime` is the number of deciseconds since `Jan 1, 2000`. Thats `A VERY LONG TIME AGO`, resulting in the hilariously imprecise Actual Round Timer numbers the longer the round goes on.

It doesn't, but having a clock that tells you the wrong time is bad.

## Changelog

:cl:
tweak: Actual Round Timer is now actually accurate.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -87,8 +87,6 @@ SUBSYSTEM_DEF(ticker)
 	var/list/music = list()
 	var/use_rare_music = prob(1)
 
-	real_round_start_time = world.realtime //SKYRAT EDIT ADDITION
-
 	for(var/S in provisional_title_music)
 		var/lower = lowertext(S)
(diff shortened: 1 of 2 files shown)


Gandalf2k15:
We had issues on the live server with the time not being set correctly so RT was used instead, it was one of those instances of "It works locally but not on the server", I don't think it was related, so this is a good change.

PR #10291, opened 2021-12-26
Fixes the SAD

Gandalf2k15:
You shouldn't bug out randomly now.

## Changelog

:cl:
fix: The SAD no longer churns out mutated abominations of yourself and should work correctly.
/:cl:


GoldenAlpharex (approved):
That does look much cleaner.

PR #10310, opened 2021-12-27
[MODULAR] Cargo Freighter Balance Run

axietheaxolotl:
## About The Pull Request

Title.

Changes:

- Secure crates that are deliverable, are locked behind a unique crate that is basically impenetrable. Can only be opened by over-powered weaponry, or by just delivering.
- Elite MODsuit, and blood-red suit have been deleted. Replaced by a prototype suit, and a Pirate EVA suit.

This stuff was very often used by cargo crew to just power-game and abuse the things inside, instead of actually delivering it. Policy was tried, and it didn't work. As QMs just ordered them open regardless of anything. This stops from them just power-gaming stuff and requires them to either deliver it, or not touch it.

As well, the Elite MODsuit and Blood-Red made no sense to be there, and were entirely unecessary. They've been replaced respectively.

## Changelog

:cl:
balance: Cargo Freighter crates are now near-indestructible
balance: Freighter Elite MODsuit and Blood-Red have been replaced.
/:cl:


ArdentSpark:
I'd honestly rather the 'powergaming' items be removed, than have the crates be nigh indestructible. This stops people from stealing/intercepting delivery shipments and opening them themselves

Additionally you say policy has been tried but doesn't work, but the policy says it's the QM's discretion whether to open them or not and they can just. Open them. I still think if opening the crates is such a big issue it should be resolved with the policy team instead of limiting it like this, all the power of the Freighter is in what is on the freighter itself, not what's in these crates

tf-4:
Based, not 100% sold on the change to the crates themselves but removing the elite and bloodred is a good change

GoldenAlpharex (changes requested):
As I said in the other PR regarding this issue, these crates should not be mapped in and should rather be in a .dm file, for ease of re-using and to lower the size of our map files, as per convention.

PR #10311, opened 2021-12-27
Changed Freighter loot/cargo

ForrestWick:

Apparently people are powergaming the ruin/role. While this in itself might not instantly stop that, the idea is now the gear they are getting is less insane and more flexible for merchant/store gimmick stuff. 

## Changelog

:cl:
balance: removed some of the gamer loot from the cargo freighter role, replaced with stuff that can actually be sold. Also switched out the spiders for zombies and added a couple extra mobs.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/skyrat/cargodiselost.dmm
+++ b/_maps/RandomRuins/SpaceRuins/skyrat/cargodiselost.dmm
@@ -64,12 +64,18 @@
 "bq" = (
 /obj/structure/table,
 /obj/machinery/atmospherics/components/unary/vent_pump/on,
+/obj/item/virgin_mary,
 /turf/open/floor/iron/cafeteria,
 /area/ruin/space/has_grav/deepstorage/lostcargo)
 "bs" = (
 /obj/effect/turf_decal/bot,
-/obj/structure/closet/crate/large,
-/obj/effect/spawner/random/decoration/statue,
+/obj/item/grenade/gas_crystal/healium_crystal,
+/obj/item/grenade/gas_crystal/healium_crystal,
+/obj/structure/closet/crate/secure/engineering{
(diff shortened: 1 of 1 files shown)


ArdentSpark:
The central command crate was bait to be broken open so good that thats gone, is that a thermal monocle? If so I'm not thrilled about that

ArdentSpark:
Also is that two NTSS13 crates? It did always confuse me as to why the freighter was delivering to everyone but the station itself, so that's neat!

_maps/RandomRuins/SpaceRuins/skyrat/cargodiselost.dmm
```diff
@@ -1859,10 +1887,10 @@
 	name = "Delivery - Enclave";
 	req_access_txt = "2"
 	},
-/obj/item/knife/combat,
 /obj/item/reagent_containers/hypospray/medipen/stimpack,
-/obj/item/clothing/glasses/hud/security/sunglasses/gars/giga,
 /obj/item/clothing/under/enclave/real,
+/obj/item/clothing/glasses/thermal/monocle,
```
axietheaxolotl: uhh, no.

axietheaxolotl (changes requested):

ForrestWick:
> Also is that two NTSS13 crates? It did always confuse me as to why the freighter was delivering to everyone but the station itself, so that's neat!

They always were able to, it's just that nobody remembers the station was called Exodus

GoldenAlpharex (changes requested):
I love the fact you're balancing this. However, those things shouldn't be varedited in the map itself, it should be handled in the code, so they can be re-used if need be, and to reduce the size of the map file itself.

If you need help with that, ask us on the Discord!

PR #10318, opened 2021-12-28
[NON-MODULAR] Makes showers clickable with mobs on them by adjusting their layer.

nevimer:
## About The Pull Request

This PR raises the showers layer to that above mobs, so that it can be interacted with easier. It also makes visual sense. 
## Before
dreamseeker_JCspm4IjxF
## After
dreamseeker_wD0puc0syG

It's great to be able to interact with a shower when someone is on the tile as one.

## Changelog

:cl:

fix: Showers now go over TOP mobs

/:cl:


Fikou:
move this upstream :)

SpaceVampire:
You can also do this already by just alt-clicking on the tile to use that menu.

PR #10332, opened 2021-12-29
Fixes various game-start runtimes

GoldenAlpharex:
## About The Pull Request
That's about it, really. I was sick of seeing them, so I'm fixing them.

Less runtimes == more betterer

## Changelog

:cl: GoldenAlpharex
fix: Fixes turrets causing runtimes in the Syndicate bases.
fix: Removes a few duplicate lattices from Shuttle5832.
fix: Fixes the runtimes with Orderlies and Custom Agents not having the right accesses to apply their trims, which popped up at every round start.
fix: Fixes the tactical suit not using the right file and therefore runtiming at roundstart.
/:cl:


Zonespace27:
Afaik the turret runtime is an upstream issue? I swear I've seen it when on a local for TG

PR #10346, opened 2021-12-30
[NON-MODULAR]Fixes corrupt photo records

nevimer:
## About The Pull Request

This PR fixes the broken photo records. Testmerge first! Only tested with two ckeys!

This corrects the proc to that generates icons to use the correct proc, as as well as get the ckey's loaded mob. The original proc this used is a deprecated way to generate icons since the tgui prefs and inspection were in. 

We actually get accurate photo records of characters again! And wanted posters, etc, function correctly as they all use the generated photo.

dreamseeker_qOczyqcOrL

dreamseeker_9noSm8ovfe

## Changelog

:cl:

fix: stops the crew records automatically generated photos from having a corrupted appearance.

/:cl:

### DIFF
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -983,7 +983,7 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
 		return humanoid_icon_cache[icon_id]
 // SKYRAT EDIT BEGIN - Icon Rendering Corruption Fix
 	var/mob/living/mob_to_copy = get_mob_by_key(dummy_key)
-	var/mob/living/carbon/human/dummy/body = generate_dummy_lookalike(mob_to_copy, mob_to_copy)
+	var/mob/living/carbon/human/dummy/body = generate_dummy_lookalike(mob_to_copy, mob_to_copy) //original = generate_or_wait_for_human_dummy(dummy_key)
 // SKYRAT EDIT END- Icon Rendering Corruption Fix
 	if(prefs)
 		prefs.apply_prefs_to(body, TRUE)
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -981,10 +981,7 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
 	var/static/list/humanoid_icon_cache = list()
 	if(icon_id && humanoid_icon_cache[icon_id])
 		return humanoid_icon_cache[icon_id]
-// SKYRAT EDIT BEGIN - Icon Rendering Corruption Fix
-	var/mob/living/mob_to_copy = get_mob_by_key(dummy_key)
-	var/mob/living/carbon/human/dummy/body = generate_dummy_lookalike(mob_to_copy, mob_to_copy) //original = generate_or_wait_for_human_dummy(dummy_key)
-// SKYRAT EDIT END- Icon Rendering Corruption Fix
+	var/mob/living/carbon/human/dummy/body = generate_dummy_lookalike() //SKYRAT EDIT: original = generate_or_wait_for_human_dummy(dummy_key)
 	if(prefs)
 		prefs.apply_prefs_to(body, TRUE)
 
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -982,7 +982,7 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
 	if(icon_id && humanoid_icon_cache[icon_id])
 		return humanoid_icon_cache[icon_id]
 
-	var/mob/living/carbon/human/dummy/body = generate_or_wait_for_human_dummy(dummy_key)
+	var/mob/living/carbon/human/dummy/body = generate_dummy_lookalike(REF(dummy_key), dummy_key)  // SKYRAT EDIT - FIXES CORRUPT PHOTO RECORDS
 
 	if(prefs)
 		prefs.apply_prefs_to(body, TRUE)
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -981,9 +981,10 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
 	var/static/list/humanoid_icon_cache = list()
 	if(icon_id && humanoid_icon_cache[icon_id])
 		return humanoid_icon_cache[icon_id]
-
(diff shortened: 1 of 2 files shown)


Tastyfish:
Is this only a glitch on Skyrat and not upstream?

nevimer:
> Is this only a glitch on Skyrat and not upstream?

I believe so.

nevimer:
currently NOPE it breaks other things, still working on it

nevimer:
Actually works now.

code/__HELPERS/icons.dm
```diff
@@ -981,9 +981,10 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
 	var/static/list/humanoid_icon_cache = list()
 	if(icon_id && humanoid_icon_cache[icon_id])
 		return humanoid_icon_cache[icon_id]
-
-	var/mob/living/carbon/human/dummy/body = generate_or_wait_for_human_dummy(dummy_key)
```
jjpark-kb, line 985: need to put this in a comment somewhere so we know what the default was.

jjpark-kb (changes requested):

code/__HELPERS/icons.dm
nevimer: why not just right click -> open previous revision? this stuff is stored anyways

ghost: because when we fix conflicts from upstream its hard to do that from the review screen; we cant really view those revisions, nor KNOW to, if its not labelled

ghost: so just having SKYRAT EDIT means maintainers know 'this wasnt the original, do we want this to change to TG' without having to search every changed line from upstream

nevimer: > because when we fix conflicts from upstream its hard to do that from the review screen; we cant really view those revisions, nor KNOW to, if its not labelled

this is a good point, I will update this to do so

code/__HELPERS/icons.dm
```diff
@@ -981,9 +981,10 @@ GLOBAL_LIST_EMPTY(friendly_animal_types)
...
-
+// SKYRAT EDIT BEGIN - Icon Rendering Corruption Fix
+	var/mob/living/mob_to_copy = get_mob_by_key(dummy_key)
```
monster860: Won't this always return null?

nevimer: nope, somehow it works. I can't explain why but it does lol. My comment earlier said that it was how other stuff worked, but that may or may not have been true, I did this a while ago

monster860: This proc returns a mob whose `ckey` is equal to whatever is passed to it. In the case of the photo records, the thing passed to it is the string `"dummy_manifest_generation"`. Which means this thing is returning null, unless someone is logged in with the username of `"dummy_manifest_generation"`

nevimer: The original used this "dummy_key" successfully to get the client - and this also successfully grabs it aswell. The way this proc is called seems to overwrite it. I didn't change anything about how the proc is called, and am simply reusing dummy key. It works somehow.

nevimer: The issue with the original line is that while it rendered the first mob in the round correctly, every one generated afterwards would have the parts of the prior on them (like it just carried over) - and - and the next - and the next - and the next. Eventually, it would just make blank icons. With this change, that doesn't happen - you get the correct image for each mob.

monster860: /proc/generate_dummy_lookalike returns a blank dummy if you pass it `null, null`. In addition the first argument should be a string not a mob.

monster860: Actually, your change is pretty much identical to what it would be if you went to `get_id_photo` and made it pass null to the proc, since /proc/generate_dummy_lookalike just calls the same `generate_or_wait_for_human_dummy` that was being called before with the first argument if you give it null as the second argument. In other words your change is pretty much creating a new dummy every time, bypassing the problem rather than actually fixing it. Also it's not deleting the dummy afterwards, since the dummy_key is not null, it won't delete it at the end now.

nevimer: ¯\_(ツ)_/¯ It still works somehow.

nevimer: This is the proc that pretty much every instance of character preview generation uses now. 

Ever since https://github.com/Skyrat-SS13/Skyrat-tg/pull/4933 this has been the proper proc to generate a new icon (generate_dummy_lookalike). Only code last touched 4 years ago or so uses the older proc (generate_or_wait_for_human_dummy). 

And that is only used in ERT icon generation preview for admin spawning, holocalls and now commented out oldprefs . 

In other words, calling that old proc and not generate_dummy_lookalike is the old way.

nevimer: generate_or_wait_for_human_dummy doesn't even take two arguments

nevimer: Right, in earlier commits it *only* returned random humans (and one of them the TGstation icon), so this _has_ to be doing something correct. And before the earlier commits, it returned a mangled mess of everyone's parts before the current. 

This solves the problem with the least amount of edits, and even if it doesn't clear the images, it's still calling the original proc that was there before deeper down (so it's really doing the same as before + more) and the end output is the correct mob. Even all the directional images exist and work. 

I can't explain it beyond that.

monster860: What your thing is "doing correct" is completely avoiding the problem instead of fixing it by just creating a brand new human dummy every time instead of reusing it. Which is okay, but the reason it's creating a brand new human dummy every time is not because that's how you intended to fix it but because you used the proc incorrectly.

Here's `generate_dummy_lookalike`:
```dm
/proc/generate_dummy_lookalike(slotkey, mob/target)
	if(!istype(target))
		return generate_or_wait_for_human_dummy(slotkey)

	// the rest of the code
```

As you can see, if target is null it just calls generate_or_wait_for_human_dummy anyways. In addition, the first argument is not a mob.

nevimer: even if the passed argument isn't a mob, it still generated the correct icon for each mob with multiple players connected. Maybe one of the arguments isn't even needed. 

and the part after that istype portion still runs, I believe. Otherwise, the outcome would be the same as the OG code

it is still an improvement over not working code at all even if it's a performance hit at round start or something

nevimer: Aight, I smoke tested it. Doesn't even use the arguments like you said. Works as intended without them, but makes a new dummy every time from scratch. It probably cleans up, smoke testing the cleanup lines worked.

Gandalf2k15 (approved):

Gandalf2k15:
We can TM this

Gandalf2k15 (approved):
yea fine

PR #10389, opened 2022-01-01
[MIRROR] adds a common SERVICE access, for service hall and service cargo console. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63678
--------------------
## About The Pull Request

Makes Service an access in its own right, primarily for the Service cargo requests console since previously Service couldn't use it, but also uses it for Service hall so there's no longer a need to have a long list of every single access to every single service job.

## Why It's Good For The Game

Makes things easier, lets all of service access service utilities, and better consistency.

## Changelog

:cl:
fix: All of service can now access the Service cargo console.
/:cl:


code/__DEFINES/access.dm
```diff
@@ -281,7 +283,8 @@
 	ACCESS_ORDNANCE_STORAGE, \
 	ACCESS_ORDNANCE, \
 	ACCESS_BARBER, \
-) // SKYRAT EDIT ADD ACCESS_BARBER
+	ACCESS_SERVICE, \
+)
```
GoldenAlpharex, line 287:
```suggestion
) // SKYRAT EDIT ADDITION - ACCESS_BARBER
```

GoldenAlpharex (dismissed):
We'll need to add that new access to our own service jobs, being the barber and the Bouncer. Might also be a good idea to check the access in our modular file for the rest of the jobs just in case there's overrides, wouldn't want people like the HoP not having `ACCESS_SERVICE`, heh.

GoldenAlpharex (approved):
There we go, good shit.

PR #10395, opened 2022-01-02
Disables Split Personality trauma

gonenoculer5:
Disables the Split Personality trauma and disables its associated procs/spell and removes references of it. This is due to the fact that A) it causes crashes and B) there have been too many instances where admins have had to intervene lately to fix split personalities.

## About The Pull Request

This PR aims to temporarily disable the Split Personality trauma by admin request, as the number of times in recent history where it has caused OOC issues and or caused outright crashes (when switching keys from the backseat user to the original owner, and vice versa), aswell as leaving admins with headaches, when the backseat ghost is left in the body after being cured of the trauma, despite the code saying this is not supposed to happen.

Having someone mascarade as you in both dchat and ingame because of a bug is bad.

## Changelog

:cl: Gonenoculer5
fix: Disabled the Split Personality trauma until it can be reworked, by admin request.
del: Disables the Personality Commune spell as it is associated with the Split Personality trauma, aswell as removes it from the neurine impure reagents forbidden traumas list to avoid errors.
/:cl:


Iamgoofball:
:-1:

See if you can reproduce these crashes upstream first.

Zonespace27:
Weren't the issues caused by the now-dead ambitions system? If it still happens here, very possibly an upstream issue as Goof said

gonenoculer5:
> Weren't the issues caused by the now-dead ambitions system? If it still happens here, very possibly an upstream issue as Goof said

Still occured as of the day prior to the posting of this PR. Seeing as I am _entirely_ un sure as to how this occurs I have no way of readily testing it as it requires a second ckey to test. I've seen it several times on live however and even experienced it first hand. The jankness here lies somewhere in swapping the backseat key with the original owner, which causes a reconnect crash, not a full CTD. Not exactly sure what the error itself is but it doesnt runtime afaik. Until someone can run more thorough testing I was asked to make a PR to nix it as its been cropping up alot lately.

gonenoculer5:
@GoldenAlpharex Assuming all looks good code wise we should be good to go. I was asked to do this by Paws#7605 after they had to fix it midround nearly causing a worse situation as a result if it'd been left unattended. (The backseat ckey was in command of the body after resolution of the trauma, leaving the original owner of the ckey as a ghost. Kind of funny but would lead to some big no bueno shit IC.)

OolongCow:
Can confirm that this has been an issue. It's happened to me twice already and both times an admin had to fix it.

PR #10397, opened 2022-01-02
[MODULAR] Wildcat and Lynx Rebalance + Blueshield Gun Change

MassacreDemigirl:
## About The Pull Request
As the title suggests, this PR makes the WIldcat and Lynx Bulky items instead of Normal weight class, making them non backpackable guns. Gives the BS a shiny Hellfire Laser Gun (slightly more shots) for their job. 

Wildcat has been used by Security and Crew as a pocket gun when it is balanced along the lines of WT, being a faster but less damaging version which is mainly used to magdump. By having it bulky, it means people are actually forced to make their intention of having lethals known by carrying it in hand or in suit slots or belt slots. Also makes the Lynx Bulky which is an extended mag WT-550. Blueshield now starts with a unique Hellfire Laser Gun (fancy name+sprite and capacity of a normal laser gun which is 12 shots) to make it more in line with station side having Lasers/Energy Guns

## Changelog

:cl: ErdinyoBarboza, Sprite originally done by Eris, and edited by @axietheaxolotl
balance: Wildcat and Lynx are now Bulky guns. They no longer fit into your backpack
balance: Blueshield now comes with a special Hellfire Laser Gun
/:cl:


Sir-Lagsalot:
Finally good to see Blueshield being brought back to the older precedent with the previous two guns. It was weird how it'd gone from having a gun with a stun mode and a laser mode, to a gun with only rubbers, to a gun with only lethals.

AtoriBirb:
I would always get a mag of rubbers as a Blueshield just so I had a non-lethal option, so this is good imo. Making the Lynx bulky means someone can't just swipe and bag it, which happens more often to me then I'd like to admit 😢

Gandalf2k15 (changes requested):
The blueshield shouldn't be given access to forms of non-lethal takedowns due to past issues with redshielding, the baton is already a bit much. So make it so they don't get that rubber loaded mag, or, change their gun to something that doesn't even support it.

KathrinBailey:
> The blueshield shouldn't be given access to forms of non-lethal takedowns due to past issues with redshielding, the baton is already a bit much. So make it so they don't get that rubber loaded mag, or, change their gun to something that doesn't even support it.

I need to revisit this content at some point but they should never have been given these guns to begin with. You'd be better off just chucking the BS with a laser or e_gun.

I'm either gonna remove these guns from the game at some point or try and find some place for antags to have them tbh

Sir-Lagsalot:
I actually have a fair few thoughts on the matter from when I used to play BS a ton, but I guess if maints are asking for stuff directly it doesn't really mean much

MassacreDemigirl:
@KathrinBailey Gave the BS their fancy lasgun + renamed and edited the PR+changelogs

modular_skyrat/modules/blueshield/code/blueshield.dm
```diff
@@ -71,15 +72,14 @@
...
-/obj/item/storage/box/gunset/blueshield/PopulateContents()
-	. = ..()
-	new /obj/item/gun/ballistic/automatic/cfa_lynx/no_mag(src)
-	new /obj/item/ammo_box/magazine/multi_sprite/cfa_lynx(src)
-	new /obj/item/ammo_box/magazine/multi_sprite/cfa_lynx(src)
-	new /obj/item/ammo_box/magazine/multi_sprite/cfa_lynx(src)
-	new /obj/item/ammo_box/magazine/multi_sprite/cfa_lynx(src)
+/obj/item/gun/energy/laser/hellgun/blueshield
+	name = "\improper Mk.V Hellfire Lasergun"
+	desc = "An energy-based laser carbine that fires concentrated beams of light which pass through glass and thin metal. This model has been purpose built by updating the old hellfire pattern for the NanoTrasen Executive Retention Protocol. It has <b><span style='color:blue'>NanoTrasen</span></b> etched into the grip. "
```
ghost, line 82:
```suggestion
	desc = "An energy-based laser carbine that fires concentrated beams of light which pass through glass and thin metal. This model has been purpose built by updating the old hellfire pattern for the NanoTrasen Executive Retention Protocol. It has <b><span style='color:blue'>NanoTrasen</span></b> etched into the grip. "
```
So, we have things for spans, namely in this case `span_bold()` and `span_blue()`, but I've no clue how to use two in one.

You'd change it to be, say, "It has [span_bold(span_blue("Nanotrasen"))] etched into the grip.", I think? I dont know if the double-span would work, I can test it personally

MassacreDemigirl: I took the span directly from CFA guns

ghost: > I took the span directly from CFA guns

I figured, should probably be replaced there too in time. But dont follow in others' mistakes.
Again, I'll test to see the proper way to format 2 spans in a row for this system, gimme a mo'

ghost: Nevermind, span procs arent compatiable with `desc`. Mark this as done and forgive my foolishness

PR #10403, opened 2022-01-02
[MODULAR] Fixes the Blueshield being able to get more access than their trim starts with.

Iamgoofball:
## About The Pull Request

Fixes the Blueshield being able to get more access than their trim starts with.

The blueshield was given the fancy CC ID card, which means there's an extra golden trim card on the station when there's only supposed to be two at max(the captain and the spare ID)

As a result they've been given a silver ID to properly respect the trim game design.

**"But I want all access as the blueshield!"**
Follow the Captain around then.

**"But I can't get into X room now!"**
How about you use any of the 500 ways we have to get into and out of rooms?

## Changelog

:cl:
fix: Fixes the Blueshield being able to get more access than their trim starts with.
/:cl:

### DIFF
--- a/modular_skyrat/modules/blueshield/code/blueshield.dm
+++ b/modular_skyrat/modules/blueshield/code/blueshield.dm
@@ -50,7 +50,7 @@
 	uniform = /obj/item/clothing/under/rank/security/blueshield
 	suit = /obj/item/clothing/suit/armor/vest/blueshield
 	gloves = /obj/item/clothing/gloves/tackler/combat/insulated
-	id = /obj/item/card/id/advanced/centcom
+	id = /obj/item/card/id/advanced/silver
 	shoes = /obj/item/clothing/shoes/jackboots
 	ears = /obj/item/radio/headset/heads/blueshield/alt
 	glasses = /obj/item/clothing/glasses/hud/security/sunglasses


CandleJaxx:
eh, seems needless imo- the blueshield's supposed to be able to respond quickly to things happening to heads, and that extra access is critical in doing so. if there's an issue with blueshields abusing this access, it should be a policy issue to remove those blueshields- not something to make the job harder across the board.

frankly, it'd be better to rework it in general, given as it shouldn't be an issue of abusing access permissions since blueshield is supposed to be locked behind security _and_ veteran crew, who should know better to not run off and redshield.

AtoriBirb:
> eh, seems needless imo- the blueshield's supposed to be able to respond quickly to things happening to heads, and that extra access is critical in doing so. if there's an issue with blueshields abusing this access, it should be a policy issue to remove those blueshields- not something to make the job harder across the board.
> 
> frankly, it'd be better to rework it in general, given as it shouldn't be an issue of abusing access permissions since blueshield is supposed to be locked behind security _and_ veteran crew, who should know better to not run off and redshield.

This. Pretty much how I feel. They should have access because their job needs it to function. If it's abused, that's an issue that could have RP repercussions, and beyond that, OOC ones of the person losing their veteran status. Shouldn't punish the good players for the bad eggs (or possibility of.)

Suspect-One:
It's not a fix, it's a nerf. And tbh it is not necessary. It is only natural for an CC employee to have a CC card. Besides, NTrep also has a CC IDcard.

The thing that matters is how Blueshields and NTreps handle their access but thats for admins and policymakers to decide.

SpaceVampire:
I keep watching blueshields give themselves all access every shift. This a good change, because switching right away from the access you spawn with makes no sense at all.

modular_skyrat/modules/blueshield/code/blueshield.dm
```diff
@@ -50,7 +50,7 @@
 	uniform = /obj/item/clothing/under/rank/security/blueshield
 	suit = /obj/item/clothing/suit/armor/vest/blueshield
 	gloves = /obj/item/clothing/gloves/tackler/combat/insulated
-	id = /obj/item/card/id/advanced/centcom
+	id = /obj/item/card/id/advanced/silver
```
YakumoChen, line 53: >Besides, NTrep also has a CC IDcard.

Remove theirs as well

YakumoChen (changes requested):

nevimer:
goofball you say a lot of things about nerfing antag tc and changeling points and all that, but the only things you ever PR are straight up nerfs to non antag roles and access removals to roles like the sec med that players end up hating and working around anyways, purely to antagonize and make the game more frustrating to people who play roles you don't like. 

honestly, i'd go so far to say these PR's are not in good faith and are with an agenda given the reasoning and action combinations. Not to mention the endless headbutting with admins on discord about TG balance and attempts at silencing others when they state things you personally disagree with. 

anyways theres a handful of ways to work around this, you know it, and it will be worked around. we're approaching encouragement for captains to validhunt if you want them to open all the doors. 

also calling this a fix is disingenuous and imo a trick to disguise yet another on station role nerf as something else and shady. it would be like me making a PR to say fix: fixes the mechanically reinforced for integrity and authenticity security force's access restrictions that were preventing them from keeping the station secure like a real company's security force.

YakumoChen:
I don't even think this is a real balance issue

99% of the time captains just give their blueshield the spare

silver ID has room for general department access and that's all the blueshield really needs, just like the HoS

Suspect-One:
No fun allowed PR

AtoriBirb:
> I don't even think this is a real balance issue
> 
> 99% of the time captains just give their blueshield the spare
> 
> silver ID has room for general department access and that's all the blueshield really needs, just like the HoS

I've only ever once been given the spare, and I only kept it for safety reasons.

Iamgoofball:
@YakumoChen If the captain gives the blueshield the spare ID, then the captain gives the blueshield the spare ID.

We can, however, stop the blueshield spawning with the spare ID.

YakumoChen:
> @YakumoChen If the captain gives the blueshield the spare ID, then the captain gives the blueshield the spare ID.
> 
> We can, however, stop the blueshield spawning with the spare ID.

yes that's why its not a balance issue. If the blueshield wants captain access, ask for the spare
Take away the NT Rep's magic gold ID too

CandleJaxx:
> @YakumoChen If the captain gives the blueshield the spare ID, then the captain gives the blueshield the spare ID.
> 
> We can, however, stop the blueshield spawning with the spare ID.

goof, the id is set that way so the BS can have needed access to command offices. if someone's getting their ass kicked in their office and you can't get there, there's no point in making the blueshield anything else but a glorified security officer.

blueshields already got their baton taken away- hell, they're mostly treated second-rate already. let them retain their ability to actually get to heads without dragging a captain around or needlessly using the spare id- (which, mind, is a hassle to even bring around due to it needing to be edited to allow the blueshield to not have the 'captain' hud marker.)

i'm just saying, this all seems wildly unneeded, and it'd be far better to implement a policy change instead of a mechanical one.

YakumoChen:
You don't need to switch the IDs, just hold it in your hand and bump into doors when you want to abuse access, keep it in a box or pocket.

Hell, use a wallet? There's a million different workarounds to "oh no I have two IDs"

YakumoChen:
Policy only matters when someone actually polices the behaviour. When was the last time you saw a blueshield get bwoinked for giving themselves AA?

This just removes two captain-level color IDs from the game. As goof said, only the captain should have true all access. Nothing is stopping you from even just making a second silver ID and giving it the needed head of staff access with captain trim.

YakumoChen:
image
Clown and Mime have relaxed rules on griefing the crew and generally exist to Be Funny.
Janitor is a chill job. You just clean.
Tide gonna tide, they don't take the job for busywork, they take it to interact.

What does Tailor add that will make it appealing for players to _want_ to play?

YakumoChen:
Gonna be honest this sounds like cargo crate material

YakumoChen:
How tf did my comment on another pr get here

OolongCow:
Compensate them with Jaws of Life and this is fine. They still have "access" to places if they need it to save a head, but them abusing that item is a much more obvious and approachable issue than what's currently happening.

A Blueshield who walks into a secure area for no particular reason with current code and policy is annoying but difficult to get rid of or tell off. A blueshield who Jaws of Life's their way into the empty CE's office without any explanation when they don't have that access is going to the brig to have a talk with security.

GoldenAlpharex (changes requested):
This is not a fix. Don't label it as such.

Also, please ensure that their trim can be properly applied to silver IDs, because if they lose their card, if it's now a silver one, the HoP should be able to paint the trim on it and give it to them.

I believe the reason it was a blue card was that it had some CC access on it, tied to the locker or something like that (but I could be mistaken), which, if it's the case, should be simply changed into another access flag, properly.

PR #10423, opened 2022-01-03
[NON MODULAR?] Partially reverts #10088, namely the pain changes.

RatFromTheJungle:
## About The Pull Request

After much discussion, it was come to the general consensus that this only really punishes the patient, and not the doctor, leading to the situation where even mal-practice only punishes said patient for going to medical.

It'll probably help with people actually going to medical for minor surgeries, instead of refusing knowing there's a fair chance they'll get a 15m mood debuff at no fault of their own.
## Changelog

:cl:
what do I put here 
/:cl:

### DIFF
--- a/code/modules/surgery/surgery_step.dm
+++ b/code/modules/surgery/surgery_step.dm
@@ -200,22 +200,11 @@
  * Sends a pain message to the target, including a chance of screaming.
  *
  * Arguments:
- * * target - Who the message will be sent to
- * * pain_message - The message to be displayed
- * * mechanical_surgery - Boolean flag that represents if a surgery step is done on a mechanical limb (therefore does not force scream)
- */
-//SKYRAT EDIT START: Fixes painkillers not actually stopping pain. Adds mood effects to painful surgeries.
+//SKYRAT EDIT START: Fixes painkillers not actually stopping pain.
 /datum/surgery_step/proc/display_pain(mob/living/target, pain_message, mechanical_surgery = FALSE)
-	if(target.stat >= UNCONSCIOUS) //the unconscious do not worry about pain
-		return
-	if(HAS_TRAIT(target, TRAIT_NUMBED)) //numbing helps but is not perfect - this is the tradeoff for being awake
-		SEND_SIGNAL(target, COMSIG_ADD_MOOD_EVENT, "mild_surgery", /datum/mood_event/mild_surgery)
-		return
-	if(mechanical_surgery == TRUE) //robots can't benefit from numbing agents like most but have no reason not to sleep - their debuff falls in-between
-		SEND_SIGNAL(target, COMSIG_ADD_MOOD_EVENT, "robot_surgery", /datum/mood_event/robot_surgery)
+	if(HAS_TRAIT(target, TRAIT_NUMBED) || target.stat >= UNCONSCIOUS)
 		return
 	to_chat(target, span_userdanger(pain_message))
-	SEND_SIGNAL(target, COMSIG_ADD_MOOD_EVENT, "severe_surgery", /datum/mood_event/severe_surgery)
-	if(prob(30))
+	if(prob(30) && !mechanical_surgery)
 		target.emote("scream")
 //SKYRAT EDIT END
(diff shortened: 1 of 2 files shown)


RatFromTheJungle:
> You currently have a negative Fix/Feature pull request delta of -8. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

Higgin:
Changes on lines 15-18 might be getting caught up from a recent PR fixing organ manipulation surgeries.

ChungusGamer666:
👎 
The crew should be affected by bad doctors. 
I hate this line of thought that consequences should only happen to the one who created the fuck up - That is not how it works in real life, and it should not be how it works in game.
There is literally no reason to use anesthesia without this.

code/modules/surgery/surgery_step.dm
```diff
@@ -12,10 +12,6 @@
 
 /datum/surgery_step/proc/try_op(mob/user, mob/living/target, target_zone, obj/item/tool, datum/surgery/surgery, try_to_fail = FALSE)
 	var/success = FALSE
-	if(surgery.organ_to_manipulate && !target.getorganslot(surgery.organ_to_manipulate))
```
Tastyfish, line 15: These lines 15-18 are a basic check in tg surgery and have nothing to do with pain.

RatFromTheJungle:
> 👎 The crew should be affected by bad doctors. I hate this line of thought that consequences should only happen to the one who created the fuck up - That is not how it works in real life, and it should not be how it works in game. There is literally no reason to use anesthesia without this.

there is no reason for doctors to use anesthesia with this, because the side-effect is both nigh-impossible to prove, and doesn't affect the doctor.

A doctor doing this to six people, back to back, has next to zero chance of anything happening to them because no one can prove it.

GoldenAlpharex:
I honestly haven't even noticed the change, so I don't feel like it's really necessary.

If you don't want to knock them out, use something like lidocaine or miner's salve.

RatFromTheJungle:
> I honestly haven't even noticed the change, so I don't feel like it's really necessary.
> 
> If you don't want to knock them out, use something like lidocaine or miner's salve.

(this is the issue, it punishes exclusively the patients, and not doctors, to the point some doctors legitimately don't know it happens)

Higgin:
#10627 should address the biggest issues people have had with this in-game (fuckiness around getting numbing applied pre-surgery,) but the only way I can think of adding feedback on the doctor's end besides screaming would be a commensurate or partial moodie, or making the slowdown even worse for unanesthetized/numbed surgery. Or, some sort of check to see if you want to continue before working on somebody not anesthetized/numbed. 

The point was to put an incentive on the patient's end to accept vulnerability.

RatFromTheJungle:
> #10627 should address the biggest issues people have had with this in-game (fuckiness around getting numbing applied pre-surgery,) but the only way I can think of adding feedback on the doctor's end besides screaming would be a commensurate or partial moodie, or making the slowdown even worse for unanesthetized/numbed surgery. Or, some sort of check to see if you want to continue before working on somebody not anesthetized/numbed.
> 
> The point was to put an incentive on the patient's end to accept vulnerability.

If someone isn't going to be numbed, a mood debuff won't stop them and it's silly to pretend they'd accept being totally defenseless. Instead, all this does is punish patients for going to medbay if their doctor happens to not know the previous PR even happened (most don't), which in generally, has made going to medbay, generally, a bad idea if your doctor intends to do anything past suture, and made the chem-fridge a lot more appealing.

Or, for a TL:DR, it'll do nothing to stop paranoid schizo's from refusing to be KO'd, punishes random people for their doctor being unacknowledged, and has made medibots a safer bet then medbay itself.

ChungusGamer666:
> > 👎 The crew should be affected by bad doctors. I hate this line of thought that consequences should only happen to the one who created the fuck up - That is not how it works in real life, and it should not be how it works in game. There is literally no reason to use anesthesia without this.
> 
> there is no reason for doctors to use anesthesia with this, because the side-effect is both nigh-impossible to prove, and doesn't affect the doctor.
> 
> A doctor doing this to six people, back to back, has next to zero chance of anything happening to them because no one can prove it.

So you're telling me doctors constantly commit malpractice and get away with it? Eyewitnesses aren't a thing now?
What a shock, it's the player's fault, they're not reporting lawbreakers, security isn't doing their job.

If this truly is a roleplay server, **then act like it and go get the people who fuck you over in trouble.**
Removing features that provide immersion and reason to care for your character because people can't roleplay, on a roleplay server, is egregious.

GoldenAlpharex:
I don't think we need to remove this at this time. Maybe later, but not now.

PR #10431, opened 2022-01-04
requested changes from a donator on their item

Seris02:
## About The Pull Request

title, they wanna be able to eat with it, and are aware that being able to eat with it also means they can drown and etc etc

it only does for one person

## Changelog

:cl:
qol: a donator mask can now no longer blocks eating
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/clothing/~donator/donator_clothing.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/~donator/donator_clothing.dm
@@ -400,17 +400,17 @@
 
 //Donation reward for Farsighted Nightlight
 /obj/item/clothing/mask/gas/nightlight
-	name = "FAR-13 SRU"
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
Do they always get the ability to eat with it, or only when lowered? Because if it's always, that's kind of an advantage compared to all the other masks.

FarsightedNightlight:
> Do they always get the ability to eat with it, or only when lowered? Because if it's always, that's kind of an advantage compared to all the other masks.

Hullo! Requester here, it would be possible to eat with the mask even up, as the mask is basically meant to work in tandem with the nitrogen breather quirk! A few masks do this as well, the welding gas mask specifically, but the donator mask also has none of the pros the welding mask does! It's basically just quality of life.

Iamgoofball:
Make it entirely cosmetic

FarsightedNightlight:
> Make it entirely cosmetic

It already is cosmetic, literally all this merge does is make it so you can eat with it. Nothing else.

It has none of the pros a traditional gas mask has, you can be peppersprayed, you can drown, you can't even protect yourself against viruses, it's basically a vox mask that takes the appearance of a sec hailer.

PR #10440, opened 2022-01-04
[Non-modular] Changes runechat's message height var to one that is nicer

desertmote:
## About The Pull Request

Currently the chat uses `bound_height` which is essentially the tile hitbox of an atom, meaning it raises or lowers by increments of 32 **only**. Changing the value on a player also causes them to be unable to move at all, it's simply not a good variable to use. I changed it to use `maptext_height`, this variable is used for ui atoms typically, and is thus safe to use for players. It's name also still makes sense to use like this.

For the sake of demonstration I added a line to the `update_body_size()` proc.

Before:
pic1

After:
pic2

## Changelog

:cl:
qol: Runechat's message height is easier to update, does so for alternative body sizes.
/:cl:


desertmote:
skibadababaaaa

PR #10445, opened 2022-01-04
[MODULAR] Adds a BOH storage module for MODsuits.

gonenoculer5:
## About The Pull Request

Adds a BOH storage module for late-game modsuits, to provide another use for frequently overabundant anomaly cores besides making anomaly armor.

I dont think a single module really adds or subtracts from the roleplay experience to be honest.

## Changelog

:cl: Gonenoculer5
add: Frontier stations have noticed a new techweb node available relating to a recently discovered advancement in bluespace storage technologies for the recently acquired MODsuits from Nakamura Engineering in collaboration with Nanotrasen.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
+++ b/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
@@ -25,6 +25,7 @@
 		new result_path(get_turf(src))
 		qdel(src)
 		qdel(anomaly)
+		playsound(src, 'sound/items/rped.ogg', 40, TRUE) //Sound feedback is cool.
 	else
 		to_chat(user,span_notice("That doesnt seem to fit into the [src]'s socket. It seems to be perfectly size for a refined anomaly core.."))
 		return
--- a/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
+++ b/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
@@ -17,7 +17,10 @@
 
 /obj/item/mod/module/storage/boh/inert/attackby(obj/item/weapon, mob/living/user, params)
 	. = ..()
-	if(istype(weapon, /obj/item/assembly/signaler/anomaly/bluespace))
+	if(!istype(weapon, /obj/item/assembly/signaler/anomaly/bluespace))
+		to_chat(user,span_notice("That doesn't seem to fit into [src]'s socket. It seems to be the perfect size for a refined bluespace anomaly core.."))
+		return
+	else
 		var/obj/item/assembly/signaler/anomaly/anomaly = weapon
 		var/result_path = /obj/item/mod/module/storage/boh
 		to_chat(user, span_notice("You insert [anomaly] into the [src]'s socket, and the module gently hums to life."))
@@ -25,9 +28,6 @@
 		playsound(src, 'sound/items/rped.ogg', 40, TRUE) //Sound feedback is cool.
 		qdel(src)
 		qdel(anomaly)
-	else
-		to_chat(user,span_notice("That doesn't seem to fit into [src]'s socket. It seems to be the perfect size for a refined bluespace anomaly core.."))
-		return
 
 //TECHWEB
 
--- a/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
+++ b/modular_skyrat/modules/modular_modsuit/module_of_holding.dm
@@ -29,3 +29,23 @@
 	else
 		to_chat(user,span_notice("That doesnt seem to fit into the [src]'s socket. It seems to be perfectly size for a refined anomaly core.."))
 		return
+
+//TECHWEB
+
+/datum/techweb_node/bohmod
+	id = "bohmod"
+	display_name = "Bluespace MODsuit Storage"
+	description = ""
+	prereq_ids = list("bluespace_storage", "mod_advanced")
+	design_ids = list(
+		"bohmod",
+	)
+	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 7500)
+
+/datum/design/module/bohmod
+	name = "MOD Module: Inert Wormhole Storage"
+	desc = "An inert Wormhole Storage module."
+	id = "bohmod"
+	materials = list(/datum/material/gold = 3000, /datum/material/diamond = 1500, /datum/material/uranium = 250, /datum/material/bluespace = 2000)
+	build_path = /obj/item/mod/module/storage/bluespace/boh/inert
+	build_type = MECHFAB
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/modular_modsuit/module_of_holding.dm
```diff
@@ -0,0 +1,51 @@
+/obj/item/mod/module/storage/bluespace/boh/inert //Inert state of the BOH mod. Requires a refined bluespace core.
```
Fikou: this pathing is bad

tf-4:
Dislike this tbh. Storage is already extremely abundant. You should need to choose between bag of holding and modsuit.

Also yeah that pathing is bad

CandleJaxx:
i'd suggest doing this upstream first either way, i feel like it'd get better traction up there

ghost:
There's already extra storage types for MODsuits, they just arent enabled iirc. Rather than this, just put those in the robotics tech tree and make them print from their machines

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
```diff
@@ -0,0 +1,51 @@
...
+	name = "Inert MOD wormhole storage module"
+	desc = "An inert MOD Wormhole Storage Module. Without its anomaly core, it is simply just a strangely ornate metal lattice with a circular socket in the bottom of it. With its core however, it becomes a way to store large quantities of items within a compacted bluespace wormhole, akin to a Bag of Holding."
+	icon_state = "module"
```
ghost: no icon makes modules sad :(

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
gonenoculer5: yeaaah really need to type this to /obj/item/mod/module/storage/boh/inert or something.

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
gonenoculer5: To be honest the generic module sprite looks pretty radical. I'm no spriter and I wont have someone sprite something unless im 110% sure it'll be added.

gonenoculer5:
> There's already extra storage types for MODsuits, they just arent enabled iirc. Rather than this, just put those in the robotics tech tree and make them print from their machines

The extra storage types are for specific suits i.e the RD suit, or the BST suit.

gonenoculer5:
Also I fixed the atrocious pathing.

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
```diff
@@ -0,0 +1,51 @@
+/obj/item/mod/module/storage/boh/inert //Inert state of the BOH mod. Requires a refined bluespace core.
+	name = "Inert MOD wormhole storage module"
+	desc = "An inert MOD Wormhole Storage Module. Without its anomaly core, it is simply just a strangely ornate metal lattice with a circular socket in the bottom of it. With its core however, it becomes a way to store large quantities of items within a compacted bluespace wormhole, akin to a Bag of Holding."
+	icon_state = "module"
```
tf-4:
```suggestion
```
inherited

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
```diff
@@ -0,0 +1,51 @@
...
+	if(istype(weapon, /obj/item/assembly/signaler/anomaly/bluespace))
+		var/obj/item/assembly/signaler/anomaly/anomaly = weapon
+		var/result_path = /obj/item/mod/module/storage/boh
+		to_chat(user, span_notice("You insert [anomaly] into the [src]'s socket, and the module gently hums to life."))
+		new result_path(get_turf(src))
+		qdel(src)
+		qdel(anomaly)
+		playsound(src, 'sound/items/rped.ogg', 40, TRUE) //Sound feedback is cool.
+	else
+		to_chat(user,span_notice("That doesnt seem to fit into the [src]'s socket. It seems to be perfectly size for a refined anomaly core.."))
```
tf-4:
```suggestion
		to_chat(user,span_notice("That doesn't seem to fit into [src]'s socket. It seems to be the perfect size for a refined bluespace anomaly core.."))
```

modular_skyrat/modules/modular_modsuit/module_of_holding.dm
```diff
@@ -0,0 +1,51 @@
...
+	. = ..()
+	if(istype(weapon, /obj/item/assembly/signaler/anomaly/bluespace))
+		var/obj/item/assembly/signaler/anomaly/anomaly = weapon
+		var/result_path = /obj/item/mod/module/storage/boh
+		to_chat(user, span_notice("You insert [anomaly] into the [src]'s socket, and the module gently hums to life."))
+		new result_path(get_turf(src))
+		qdel(src)
+		qdel(anomaly)
+		playsound(src, 'sound/items/rped.ogg', 40, TRUE) //Sound feedback is cool.
+	else
```
tf-4: early return this

GoldenAlpharex:
I think it'd be better if we waited to see if Fikou won't do something like that upstream first.

I'm not a big fan of making it possible to obtain a BoH storage module within the round, I feel like you should have to choose between BoH storage capacity (and you could store the MODsuit control unit in your BoH, I feel like), or wearing the MODsuit, having both feels incredibly unbalanced, and not something the code accounts for.

I'd be much more into getting a way to get the improved storage module available later in the round for like, 10k points and a reasonably bigger amount of materials to print it than the normal storage module, if Fikou doesn't plan to add that possibility upstream already.

gonenoculer5:
> I'm not a big fan of making it possible to obtain a BoH storage module within the round, I feel like you should have to choose between BoH storage capacity (and you could store the MODsuit control unit in your BoH, I feel like), or wearing the MODsuit, having both feels incredibly unbalanced, and not something the code accounts for.

I mean. You still require an anomaly core which is gated by science. Also storing the mod in the bag would kinda make having it pointless since you'd have to carry a bag with you.

> I'd be much more into getting a way to get the improved storage module available later in the round for like, 10k points and a reasonably bigger amount of materials to print it than the normal storage module, if Fikou doesn't plan to add that possibility upstream already.

Did you read the portion of the code relating to research? It already requires bluespace storage + the T3 modsuit tech. The amount of materials is identical to the amount of materials required to print an inert BOH. It can certainly be raised. Storage to the _general public_ is rather limited to the standard storage module, which is quite small. Head modsuits have a more reasonable quantity of storage, closer to that of a standard backpack/duffelbag. The standard storage module is currently less than even a leather satchel can hold. If nessecary I could raise the amount of points specifically for the techweb node required for this module to be available to something like 10-12K.

GoldenAlpharex:
>I mean. You still require an anomaly core which is gated by science. Also storing the mod in the bag would kinda make having it pointless since you'd have to carry a bag with you.

Yeah, but I don't want the MODsuit to be the best options, if you have a bluespace core that's activated, you can wait three minutes to get enough points to research that node, and bam now it's just a straight upgrade with no downside over the BoH.

I don't like when things are pure upgrades with no drawback, especially when it goes against the design philosophy of the item in question. You choose a MODsuit, or you choose an insanely bigger storage. Not both.

You're free to open another PR for making the expanded storage available to the crew late into the round, instead of making it go from tiny storage module to BoH size, but I don't think this is a good change, sorry.

PR #10467, opened 2022-01-05
Roller bed cells now buckles the mob to the bed.

KathyRyals:
## About The Pull Request
Roller bed cells now buckles the mob to the bed.

Roller bed cells now buckles the mob to the bed. It's cool like that.
## Changelog

:cl:
qol: Roller bed cells now buckles the mob to the bed.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_weapons/code/medcells.dm
+++ b/modular_skyrat/modules/modular_weapons/code/medcells.dm
@@ -392,7 +392,8 @@
 	for(var/obj/structure/bed/roller/medigun in target.loc) //Prevents multiple beds from being spawned on the same turf
 		return FALSE
 	if(HAS_TRAIT(target, TRAIT_FLOORED) || target.resting) //Is the person already on the floor to begin with? Mostly a measure to prevent spamming.
-		new /obj/structure/bed/roller/medigun(target.loc)
+		var /obj/structure/bed/roller/medigun/created_bed = new /obj/structure/bed/roller/medigun(target.loc)
+		created_bed.buckle_mob(target)
 		return TRUE
 	else
 		return FALSE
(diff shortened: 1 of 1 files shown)


softcerv:
While I appreciate the QoL, I worry that this might end up making the medigun too robust in the hands in the hands of a secmed (or any sec officer with a medigun) since you can just buckle anyone that gets downed during a fight.

KathyRyals:
I really, really doubt we'll ever see this happen, especially since there is no reason for a sec officer to carry around a medigun.


I'd rather not prevent the introduction of good QoL features just because someone might use them in a gamey way once every 200 shifts.

softcerv:
> I really, really doubt we'll ever see this happen, especially since there is no reason for a sec officer to carry around a medigun.
> 
It's not unimaginable for a sec-medic to carry a medigun. (especially if it ever ends up becoming a weapon for them)
> I'd rather not prevent the introduction of good QoL features just because someone might use them in a gamey way once every 200 shifts.
>
Being able to basically instantly chairbuckle someone in the middle of combat because they touch the floor for a second is really robust and can quickly end a fight, especially if the person is cuffed. Think about how this would pair with CMO or Sec Medic who both have instant knockdown tools and would be likely to use Mediguns. This really feels like more of a buff than a QoL change.

GoldenAlpharex (approved):
Looks good to me.

tf-4:
maybe a short do_after?

softcerv:
I think the best way to handle this is have the auto-buckling restricted to unconscious, crit, or dead people. That way, It would still be a QoL, but it wouldn't really be a combat tool anymore.

tf-4:
^ also a good solution

GoldenAlpharex (changes requested):
Yeah I think making sure it's only done when they're unconscious or dead would be the best way for this to not be used for gamering.

KathyRyals:
Good idea.

tf-4 (approved):

GoldenAlpharex (approved):

PR #10482, opened 2022-01-06
Makes internships more viable by adding all the missing headsets to their appropriate vendors

GoldenAlpharex:
## About The Pull Request
Adds medical headsets to the medical clothes vendor, adds engineering headsets to the engineering clothes vendor and adds service headsets to the Head of Personnel's PTech vendor (it's the one centralized location for all of Service with a vendor in it, fight me).

Have you ever wanted to be an intern in Medical, Engineering or Service, but then discovered that you would remain forever severed from your new department's communication channel because you simply couldn't get your hand on one of their headsets?

Well, this won't happen anymore, you're welcome.

The goal is to make internships more viable, so that people will be more likely to do stuff as assistants, which is just a bliss to see.

## Changelog

:cl: GoldenAlpharex
fix: After a long series of tiresome meetings and heated debates, the Nanotrasen Executive Board is happy to announce that all departments will be able to obtain additional headsets for their departments from their clothing vendors, so more poorly-paid internships can occur! (N.B.: The Head of Personnel is reponsible for service headsets, refer to them if you'd like to acquire one for your service internship)
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_vending/code/wardrobes.dm
+++ b/modular_skyrat/modules/modular_vending/code/wardrobes.dm
@@ -6,6 +6,7 @@
 
 /obj/machinery/vending/wardrobe/medi_wardrobe
 	skyrat_products = list(
+		/obj/item/radio/headset/headset_med = 3,
 		/obj/item/clothing/gloves/color/latex/nitrile = 2,
 		/obj/item/clothing/suit/toggle/labcoat/hospitalgown = 5,
 		/obj/item/storage/belt/medbandolier = 2,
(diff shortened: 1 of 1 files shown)


Elli-Skala:
I thought the reason it's like this is because the lockers come with head sets in all the departments but there isn't a rnd locker so it's put into the vendors

tf-4:
There are no med lockers on delta

YakumoChen:
@GoldenAlpharex tbh this sounds like something that would go better in cargo crates, avoids weirdness with departmental vendors too.

(Also gives people reason to order request crates that aren't just pizzas and medical guns)

GoldenAlpharex:
> @GoldenAlpharex tbh this sounds like something that would go better in cargo crates, avoids weirdness with departmental vendors too.
> 
> (Also gives people reason to order request crates that aren't just pizzas and medical guns)

It's in the Science vendor, and as much as I think this could be an interesting thing, I'd rather make it as painless as possible for people to join a new department, because otherwise they might just not bother.
Also, technically it's in crates thanks to this, you'd have to order a restock if you use more than three. Which I doubt, but it *could* happen.

PR #10485, opened 2022-01-06
Aesthetic fix for the arrivals shuttle recall terminal & performance improvements to computers

GoldenAlpharex:
## About The Pull Request
I fixed the fact that the arrivals shuttle recall terminal wouldn't look any different when unpowered, and I even gave it a new animation when properly fixing the sprite.

I also reduced by half the amount of overlays on computer terminals because we forgot to comment out the non-modular version of the overlay code for computers when doing the connected computers.

Less lag from computers is always good.

Also this looks pretty
dreamseeker_aS5lOpBcv5

And this finally looks right
image

## Changelog

:cl: GoldennAlpharex
fix: Cut half of the overlays that would currently be applied to computers due to connecting computer, reducing their impact on performance in that regard, hopefully.
imageadd: Fixed the sprite for the arrival shuttle recall terminal so it isn't fullbright all the time and gave it a neat little animation.
/:cl:


GoldenAlpharex (commented):
I have reviewed this PR and I agree with the changes. *nod

PR #10526, opened 2022-01-08
RedSec is back! Again! 

YakumoChen:
## About The Pull Request

Re-adds Redsec

(to cargo)

Some people like the color but also when they order a cargo crate containing explicitly red clothing they expect it to be red.

## Changelog

:cl:
fix: The redsec uniform crate is properly dyed again after an incident with a space squid to our warehouses.
/:cl:

### DIFF
--- a/modular_skyrat/modules/goofsec/code/sec_clothing_red.dm
+++ b/modular_skyrat/modules/goofsec/code/sec_clothing_red.dm
@@ -1,0 +1,54 @@
+//better red than dead
+
+//Officer
+/obj/item/clothing/under/rank/security/officer/red
+	icon = 'icons/obj/clothing/under/security.dmi'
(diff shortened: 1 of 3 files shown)


GoldenAlpharex:
Actually Xyel doesn't want that crate to be there anymore I think.

YakumoChen:
It's a completely optional choice that many prefer and isn't in the default options, you have to put in at least a modicum of effort to get it. It was let through the first time too. 

Why not?

ghost:
While I disagree with the crate for the reason of 'security should stay one consistent color scheme'

I think having old-sec versions of the uniforms will be good for the sake of events and whatnot

tf-4:
Xyel has specifically denied this yes, I'm pretty sure

Gandalf2k15:
UA denied this.

PR #10563, opened 2022-01-10
Removes the HF Blade from Traitor Loadouts

dawsonkeyes:
## About The Pull Request

Does what it says on the tin. Due to reworked item paths upstream and general incoming changes to the weapon, this PR replaces the ninja loadout’s variant of the vibroblade with the energy katana, which has equal damage to an energy sword but added mobility and can only be stored openly.

It doesn’t! But broken object subtypes is bad, and wildly overpowered gear isn’t better.

## Changelog

:cl:
add: The cyborg ninja traitor loadout now wields an energy katana instead of a vibroblade.
/:cl:


GoldenAlpharex (changes requested):
Lol no, this is definitely not the correct fix, it's a ***massive*** buff to the kit. The title is also misleading, you're replacing it, not removing it.

We're not giving straight up the energy katana to traitors, hell nah.

dawsonkeyes:
guess I’ll die

PR #10571, opened 2022-01-10
IceBox Fixes: Removes accesses; adds lattices

Stalkeros2:
## About The Pull Request
Removed access from Icebox mining's those two airlocks so both the public mining and xenoarch are *actually* public use.
image
Added a short bridge here so people can also actually access it too, as well as marker beacons to light the way there because why not?
image

~~I dunno, you tell me.~~
Map fixes good.

## Changelog

:cl: Stalkeros
fix: Made Icebox's public mining actually public and accessible.
/:cl:

### DIFF
--- a/_maps/map_files/IceBoxStation/IcemoonUnderground_Above_skyrat.dmm
+++ b/_maps/map_files/IceBoxStation/IcemoonUnderground_Above_skyrat.dmm
@@ -30038,7 +30038,7 @@ PG
 Fp
 Fp
 Nt
-Fp
+PI
 Fp
 Fp
 Fp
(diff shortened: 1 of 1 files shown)


Zenitheevee:
>Reads "adds lattices" 
>Immediate PTSD from Lattice-based runtimes in DM waiting for server boot

GoldenAlpharex (approved):
Looks good!

MassacreDemigirl:
Maybe add a pathway with markers from public mining exit to it?

Stalkeros2:
> Maybe add a pathway with markers from public mining exit to it?

Basically done.

PR #10600, opened 2022-01-11
Xenoarch adjustments

Stalkeros2:
## About The Pull Request
A literal webedit to make broken objects sell for 1k as they presumably should instead of 400 credits like useless relics.
Also made it more obvious that you can swab animal remnants.

[Money.](https://youtu.be/ULeDlxa3gyc)
Intuitiveness actually good.

## Changelog

:cl: Stalkeros
fix: Made xenoarch's broken objects actually sell for their price, and made broken objects actually sellable too.
qol: It's now more obvious that animal remnants are swabbable.
/:cl:


Stalkeros2:
~~Do note that this might actually do nothing due to broken item's price being defined once again later in the same file, but idk I'd rather be sure 100.1% or something.~~
Nvm I fixed that. :sunglasses:

PR #10636, opened 2022-01-12
[SEMI-MODULAR] Changeling Reset

Iamgoofball:
## About The Pull Request

The Changeling can now only hold six strands of DNA at a time.
The Changeling's starting chemical charges have been reduced back down to 20.
The Changeling's starting maximum chemical charges have been reduced back down to 75.
The Changeling's recharge rate of chemicals has been reduced back down to 0.5.
The Changeling's Genetic Points have been reduced back down to 10.

The changeling was, frankly, given an incredibly un-needed buff that only served to make the changeling able to be far more lethal against a playerbase that isn't allowed to be nearly as deadly as they could be. I'm not convinced Changelings need to be this deadly on a server where only 10~ players during a round are actively allowed to torch and pitchfork hunt down antagonists, when their previous level of deadliness was sufficient enough to handle high pop Terry levels of torch and pitchfork.

Additionally, this raw boost in power made Changelings not need to rely on stealth, deception, and social engineering nearly as much, which is a massive blow to an antagonist whose central thesis is "stealthy disguise monster".

By undoing these changes, Changelings will need to be more careful with going loud, and spend their chemicals carefully.

## Changelog

:cl:
balance: The Changeling can now only hold six strands of DNA at a time.
balance: The Changeling's starting chemical charges have been reduced back down to 20.
balance: The Changeling's starting maximum chemical charges have been reduced back down to 75.
balance: The Changeling's recharge rate of chemicals has been reduced back down to 0.5.
balance: The Changeling's Genetic Points have been reduced back down to 10.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -31,25 +31,25 @@
 	/// The original profile of this changeling.
 	var/datum/changeling_profile/first_profile = null
 	/// How many DNA strands the changeling can store for transformation.
-	var/dna_max = 8 // SKYRAT EDIT - ORIGINAL: 6
+	var/dna_max = 6
 	/// The amount of DNA gained. Includes DNA sting.
 	var/absorbed_count = 0
 	/// The amount of DMA gained using absorb, not DNA sting. Start with one (your original DNA)
 	var/true_absorbs = 0
 	/// The number of chemicals the changeling currently has.
-	var/chem_charges = 75 // SKYRAT EDIT - ORIGINAL: 20
+	var/chem_charges = 20
 	/// The max chemical storage the changeling currently has.
-	var/total_chem_storage = 100 // SKYRAT EDIT - ORIGINAL: 75
+	var/total_chem_storage = 75
 	/// The chemical recharge rate per life tick.
-	var/chem_recharge_rate = 1.5 // SKYRAT EDIT - ORIGINAL: 0.5
+	var/chem_recharge_rate = 0.5
 	/// Any additional modifiers triggered by changelings that modify the chem_recharge_rate.
 	var/chem_recharge_slowdown = 0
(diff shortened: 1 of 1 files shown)


Iamgoofball:
thanks @github-actions im aiming for -100

SpaceVampire:
I would note that during my casual observation of the past few weeks, I have noticed Changelings still getting cremated as a pretty common affair. Probably because the MCR is even more overtuned then they are at the moment. It's also a lot harder to do the classic stealthy changeling stuff, when their transform seems to constantly break/bug as it attempts to turn you into the latest snowflake of the week.

RatFromTheJungle:
Changelings aren't even that strong, comparatively, a singular security officer can out-do a changeling even with adrenals, especially so if they set said changeling on fire first, the buff is to deal with the much more populated server, and areas, which makes stealth in general much harder because finding a place that won't have someone walk through in about 3 minutes is next to impossible.

(And to cont. on the above stated, it's not even like they're winning consistently, I don't think a single changeling has 'won' in like two months time, usually they just get mopped up.)

Zenitheevee:
So, Personally?
I dont think the reset is necessary nor recommended. Tour other PR to nerf horrorform into an All combat All the time situation Is already changing changeling to something much more **balanced**

Are there some things in this PR that are good? Yes. 
Chem regen has no need being as robust as it is. That change can _probably_ stay. 
DNA strands? I rarely see people use that _anyways_ for its intended use.
Starting chem charges? _ok it could probably be toned down so they dont go ham super early_

But everything else feels a lot more necessary considering those "Ten" people with "Pitchforks and torches" have ranged tools that can put you under faster than most people can anticipate, Ontop of a lot of Security's "Kill first, RP later" playstyle _and_ their propensity to straight up ignore metaprotections, Policy damned.

Iamgoofball:
stale bot be gone

PR #10645, opened 2022-01-13
[MIRROR] Removes swarmers from the game [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63989
--------------------

## About The Pull Request
What the title says. But why?
I generally have a rule when making a contribution, that is "don't make the game less fun"
I'm not salting, I didn't die to a swarmer.
... Yet that's the problem. Swarmers are the griefiest antag in the game, but when you complain that they're annoying or unfun, you're doomed to hear "lol they can't even hurt you though."

WELL THAT ACTUALLY MAKES THEM WORSE. I would rather die to a hundred xenos and space dragons than be forced to untie myself in maintenance for 45 seconds while the shuttle leaves.

## Why It's Good For The Game
Unfun game modes should be removed from the game.
- Being griefed by swarmers is annoying
- Playing as a swarmer is not very exciting either. Click on iron.

lastly, because oranges authorized it

## Changelog

:cl:
del: Removes swarmers! The griefiest, lowest fun value antagonist is removed from the game.
/:cl:


SpaceVampire:
We should probably go ahead and modularize this, because I think Swarmers honestly work much better on Skyrat.
Already have a lack of variety when it comes to midrounds, and they provide a fun non-lethal threat that the crew in general is welcome to show up and smash if they are in the mood for it. The actual grief potential has dropped a lot since they can't vent crawl to all of the meta locations right away, and also create fully functional copies from anywhere. It's also impossible to actually "lose" to them, which makes it extra great for this place. No round ending!

PR #10620, opened 2022-01-11
[MIRROR] [MDB IGNORE] BIDDLE TRAITORS - Adds progression traitors. Refactors uplink code in its entirety [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63588
--------------------

This PR is currently not ready for merge. It is, however, ready for testmerge.

## About The Pull Request

https://hackmd.io/IkDWWkpfQa2lkdevsnLqhA?view
HackMD explains all of the mechanics of this change. Below are pictures of what the UI looks like.
image
image
image
image
image

## Why It's Good For The Game

Relaying what is said on the hackMD:
The goal of this change is to prevent them from immediately being able to buy extremely deadly traitorous weapons that allow them to go on a rampage or blow up the station at the start of a round. Additionally, the goal is also to make traitor objectives a lot more engaging as a result.
This change introduces a way for traitors to make more TC and gives more depth to the traitor role as a whole. You do not have to engage in the system if you want to do your own thing, because you passively gain progression points which'll let you purchase most items 30 minutes into a round if you do no objectives. This change acts as a soft lock on traitor equipment and as a way for traitors to get inspiration on what to do, as well as adding an end goal for traitors if they choose to reach it

## Changelog

:cl:
add: Adds progression traitors and reworked traitor objectives to give reputation and telecrystals, which can be used to purchase equipment.
balance: Traitor gear now has a soft timelock. Most gear becomes accessible 30 minutes into a round.
refactor: Refactored the uplink UI code completely to be in typescript and modernizes it.
del: Removed romerol, surplus crate and the special syndicate kits from regular traitor uplink. Romerol is now a final objective that you can receive.
/:cl:

PR #10658, opened 2022-01-13
[SEMI-MODULAR] We forgot to reduce the starting TC when Progression Traitors were added. Whoops.

Iamgoofball:
see title

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -204,7 +204,7 @@ GLOBAL_LIST_INIT(ai_employers, list(
 // Progression traitor defines
 
 /// How many telecrystals a normal traitor starts with
-#define TELECRYSTALS_DEFAULT 35 //SKYRAT EDIT CHANGE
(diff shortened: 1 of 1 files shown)


Zonespace27:
Considering how the change was purposeful [([Commit](https://github.com/Skyrat-SS13/Skyrat-tg/pull/10620/commits/96bd40598615827789a42fe762e7bcc2f2a7a208)) on ([PR](https://github.com/Skyrat-SS13/Skyrat-tg/pull/10620))], _and_ the fact that as of this PR being put up, we've only had 1 full round with traitors so far, I'd say that it should be given a day or two before making a decision on if the TC reversion is needed.

GoldenAlpharex (approved):
That makes sense, but honestly I still just don't know about how ludicrous the amount of TC they can get is.

Zonespace27:
Believer bundle's still at 35 TC, may potentially want to be changed to 20? Albeit it is a pretty strong bundle

PR #10665, opened 2022-01-13
[NON MODULAR] Baton 'rebalance' (nerf)

RatFromTheJungle:
## About The Pull Request

Nerfs stunbatons (and police batons, due to the detective, and the british police crate giving them/ways to get them) to be (slightly) more inline with how everything else is affected by the health increase, namely, another hit to down (which is a generously low increase), and removal of the confusion so it's less debilitating on the first-hit.
Brings batons to be slightly more inline with every other weapon in the game, and not the end-all-be-all of melee, regardless of armor.

## Changelog

:cl:
balance: Stun batons deal ten less damage, or equal to another hit to stamcrit, and no longer 'confuse' people whom get hit by one.
/:cl:

### DIFF
--- a/code/game/objects/items/melee/baton.dm
+++ b/code/game/objects/items/melee/baton.dm
@@ -25,7 +25,7 @@
 	/// The length of the knockdown applied to the user on clumsy_check()
 	var/clumsy_knockdown_time = 18 SECONDS
 	/// How much stamina damage we deal on a successful hit against a living, non-cyborg mob.
-	var/stamina_damage = 35 // SKYRAT EDIT - Less Stamina Damage (Original: 50)
+	var/stamina_damage = 35 // SKYRAT EDIT - Less Stamina Damage (Original: 55)
 	/// Can we stun cyborgs?
 	var/affect_cyborg = FALSE
 	/// The path of the default sound to play when we stun something.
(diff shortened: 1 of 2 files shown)


Iamgoofball:
Dropping the confusion is an interesting choice. Let's get this TM'd and see how it works in practice.

code/game/objects/items/melee/baton.dm
```diff
@@ -25,7 +25,7 @@
 	/// The length of the knockdown applied to the user on clumsy_check()
 	var/clumsy_knockdown_time = 18 SECONDS
 	/// How much stamina damage we deal on a successful hit against a living, non-cyborg mob.
-	var/stamina_damage = 55
+	var/stamina_damage = 35 // SKYRAT EDIT - Less Stamina Damage (Original: 50)
```
tf-4:
```suggestion
	var/stamina_damage = 35 // SKYRAT EDIT - Less Stamina Damage (Original: 55)
```

Suspect-One:
Another day, another unnecessary baton nerf.

Zonespace27:
> Another day, another unnecessary baton nerf.

Batons are the strongest melee weapon in game, what're you on about?

Suspect-One:
> > Another day, another unnecessary baton nerf.
> 
> Batons are the strongest melee weapon in game, what're you on about?

Batons can be found literally everywhere. I'm not even saying that traitors have a variety of tools to work around them. They can even hack into brig and get a baton for themselves too and use it against security if they need to or just get one from depguards with 3 charges (that's enough to do some nasty stuff).

IMO the more you nerf their non-lethal approach the more seccies will resolve their issues using guns because at one point that'll be much easier than using literally any non-lethals at their disposal.

tf-4 (approved):
Let's get this tested out

Iamgoofball:
so did we actually testmerge this at any point or

PR #10719, opened 2022-01-15
[MODULAR] Sidearm Tokens Crate, and Sidearms Alert Change

axietheaxolotl:
## About The Pull Request

Adds a sidearm token crate to Cargo, and changes the alert level needed for sidearms to be amber.

Carrying around a rifle at all times is impractical, and dumb. What happens is the alert level raises, and suddenly Sec is packing incendiary over-powered rifles. This allows for a stopgap for this, so they aren't carrying around Rifles/Primaries 24/7. Yes, they are ballistic, but so are WTs, and you have to go out of your way to order those. The same applies here.

They've also had the needed alert level for them upped to amber, this is to enforce that they're a downgrade to a proper primary when that's not needed. Not having it on Blue should reduce "sec carry around gun 24/7" syndrome.

## Changelog

:cl:
add: Sidearms Token Crate to Cargo
balance: Sidearm tokens now require amber alert to use
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/token_system/token_system.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/token_system/token_system.dm
@@ -73,7 +73,7 @@
 	name = "sidearm armament holochip"
 	desc = "A holochip used in any armament vendor, this is for sidearms. Do not bend."
 	icon_state = "token_sidearm"
-	minimum_sec_level = SEC_LEVEL_BLUE
+	minimum_sec_level = SEC_LEVEL_AMBER
 
 /obj/item/armament_token/sidearm/get_available_gunsets()
 	return list(
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -176,6 +176,13 @@
 					/obj/item/gun/ballistic/automatic/pistol/toy) //they just hate you.
 	crate_name = "dusty crate"
 
+/datum/supply_pack/security/armory/sidearm
+	name = "Armadyne Sidearm Tokens Crate"
+	desc = "Contains one set of Armadyne branded Sidearm Tokens, to be redeemed at an appropriate vending machine."
+	cost = CARGO_CRATE_VALUE * 7
+	contains = list(/obj/item/storage/box/armament_tokens_sidearm)
+	crate_name = "sidearm token crate"
+
 //////////////////////////////////////////////////////////////////////////////
 ///////////////////////////// Engineering ////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////////


Iamgoofball:
:-1:
You have a sidearm, it's called your disabler. Use your disabler.

modular_skyrat/modules/sec_haul/code/guns/token_system/token_system.dm
```diff
@@ -73,7 +73,7 @@
 	name = "sidearm armament holochip"
 	desc = "A holochip used in any armament vendor, this is for sidearms. Do not bend."
 	icon_state = "token_sidearm"
-	minimum_sec_level = SEC_LEVEL_BLUE
+	minimum_sec_level = SEC_LEVEL_AMBER
```
Zonespace27, line 76: PR desc says set to blue, yet you set to amber
edit: the "about" section and the "how this contributes" section say two different things

tf-4:
> 👎 You have a sidearm, it's called your disabler. Use your disabler.

modular_skyrat/modules/sec_haul/code/guns/token_system/token_system.dm
axietheaxolotl: 💀 I forgor

Aether217:
This goes against the decision that was made to move away from ballistics for security and it brings back the nightmare of trying to balance the sidearms.

RatFromTheJungle:
These shouldn't be using the antiquated tokens system anywhom, and most of the guns probably should be moved to 'contraband' and or emag-crates

YakumoChen:
All this does is replace the problem of field reloadable longarms with carrying 20 pistols in your satchel

axietheaxolotl:
> All this does is replace the problem of field reloadable longarms with carrying 20 pistols in your satchel

There's 3 tokens in a crate, and you can only order a crate every 20 minutes, if that. If you somehow manage to get the Warden or HoS to give you more then one sidearm, I'll be very surprised.

axietheaxolotl:
> This goes against the decision that was made to move away from ballistics for security and it brings back the nightmare of trying to balance the sidearms.

Ballistics were never removed from Sec, see: WT-550s. They just aren't the primary focus. You can still get ballistics, you just need to go out of your way to get them. Like it is here.

axietheaxolotl:
> These shouldn't be using the antiquated tokens system anywhom, and most of the guns probably should be moved to 'contraband' and or emag-crates

this is fair, but honestly I liked the choice system sidearms had. Everything will devolve into "meta", MCRs did it too. Fact-a-life, really.

Gandalf2k15:
We're phasing these out of use fairly soon, I don't want to add more new code relying on them

PR #10721, opened 2022-01-15
[NON-MODULAR] Reduces the charge cost on health analyzers

softcerv:
## About The Pull Request

Reduces the cost on using health analyzers from 50 to 30, also it makes it so that advanced analyzers only use 30 charge per use too.
Having to constantly swap out cells on the health analyzer makes medical gameplay more tedious. This keeps in the charge component while making it less agonizing for doctors to deal with.

## Changelog

:cl:
balance: lowers the power cost of health analyzers.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -97,6 +97,7 @@ GENE SCANNER
 	var/scanmode = SCANMODE_HEALTH
 	var/advanced = FALSE
 	custom_price = PAYCHECK_HARD
+	power_use_amount = POWER_CELL_USE_LOW //SKYRAT EDIT ADDITION CHANGE
 
 /obj/item/healthanalyzer/examine(mob/user)
 	. = ..()
@@ -502,7 +503,6 @@ GENE SCANNER
 	icon_state = "health_adv"
 	desc = "A hand-held body scanner able to distinguish vital signs of the subject with high accuracy."
 	advanced = TRUE
-	power_use_amount = POWER_CELL_USE_HIGH //SKYRAT EDIT ADDITION CHANGE
 
 /// Displays wounds with extended information on their status vs medscanners
 /proc/woundscan(mob/user, mob/living/carbon/patient, obj/item/healthanalyzer/wound/scanner)


thestubborn:
i agree with the change but

honestly?

i'd just scrap analyzers having charges full stop

tf-4:
Constantly? I find myself doing it once, maybe twice per round maximum.

softcerv:
> Constantly? I find myself doing it once, maybe twice per round maximum.

It depends on what job you are playing and how you are using the analyzers. Doing a basic scan on someone while bringing them back to medbay as a paramedic is far different from a doctor having to scan someone multiple times while doing blood filtering to see if the chemicals are gone.

Right now, I don't think the power cost is good for the amount of scanning that medical doctors typically need to do. This gives doctors a bit more leeway.

tf-4:
I play a lot of all medical roles (except viro) and I think it's pretty fine as it is.
I don't find myself having to constantly recharge by any means. In fact, once you have better cells you never need to recharge.

YakumoChen:
Having battery charge on them feels pointless altogether and I've been having issues with health scanning ever since they got cells (the chem scanner doesn't work for me half the time but consumes charge anyway, I think it actually consumes charge if you click anything with it)


It's literally easier to print a new one and throw the old one away than find a bluespace cell for it

softcerv:
> It's literally easier to print a new one and throw the old one away than find a bluespace cell for it

Basically this.

SpaceVampire:
> > It's literally easier to print a new one and throw the old one away than find a bluespace cell for it
> 
> Basically this.

To be fair, I do the exact same thing with plasma cutters.

YakumoChen:
> To be fair, I do the exact same thing with plasma cutters.

you can fill plasma cutters up with raw plasma ore you know

SpaceVampire:
> you can fill plasma cutters up with raw plasma ore you know

Oh, I know. I'm just that lazy.
It also doesn't cost me any points to do it.
All about that bottom line, baby!

ghost:
I thought the cell component was already modular; couldnt this be done modularly?

tf-4:
> I thought the cell component was already modular; couldnt this be done modularly?

Yes, but it would be making a new file for `power_use = low` or whatever

Gandalf2k15:
I'd say they're fine as is..., i don't wanna invalidate the existence of cells component

YakumoChen:
It really isn't a good concept to begin with, but you do you.

PR #10727, opened 2022-01-15
new circuit components

jjpark-kb:
## About The Pull Request
adds four new circuit components:
1) mining component - will mine closed turfs with ores
2) cell charge component - reports back the max/current charge of the cell
3) target scanner component - reports back the turf and the contents of the turf based on the off set
4) item interact component - makes the item "attack itself"

more components for more creative community members.

## Changelog

:cl:
add: added four new circuit components- mining, target scanner, cell charge, and item interact
/:cl:


SpaceVampire:
Could this be done upstream?
image

Weblure:
> Could this be done upstream?

image

The gamers would abuse it to hell on TG, they'd never go for this.

SpaceVampire:
> The gamers would abuse it to hell on TG, they'd never go for this.

That has always been a problem with circuits, and I didn't think that any of this was too unreasonable.
It seems like it might be worth trying to see if they want it, before moving it down here if they don't.
Could be interesting to see their feedback on it as well, because I would still rather not have stuff that is super gamer abusable here. I really don't see any reason to not try it, because you can't say for sure.

Weblure:
> > The gamers would abuse it to hell on TG, they'd never go for this.
> 
> That has always been a problem with circuits, and I didn't think that any of this was too unreasonable. 

That's the thing with circuits, something that seems mundane and harmless on the surface can wreak absolute havoc in the hands of a skilled circuit designer.

I've only been messing with circuits for about a week or so, but I can already think of several morbid things to do with this.
Example:
Have a drone shell drag a dispenser around the station. Fill the dispenser with a minibomb. Have another drone shell following the first, towing a prelocked locker (if that's still a thing, otherwise a bear trap, flash, or something similar would work). Continuously scan for a specified target while patrolling the station. Once acquired, send the locker/trap/flash drone out to intercept and trap/immobilize/stun the target. During this time, the bomber drone should've already dispensed the minibomb, activated it, grabbed it, and then began dragging it towards the target. The target would have no chance of escape if they failed to evade the first drone.


> It seems like it might be worth trying to see if they want it, before moving it down here if they don't. Could be interesting to see their feedback on it as well, because I would still rather not have stuff that is super gamer abusable here. I really don't see any reason to not try it, because you can't say for sure.

I suppose it wouldn't hurt to try, beyond wasting time and delaying this PR as we wait for the TG maintainers to deny it. As for abuse, I don't think it'd be a big problem here. Circuits are rarely used to begin with (though I may have raised some awareness for them lately), and we have strict rules for roleplaying and powergaming. Anyone that tried to abuse them in a similar way to what I described above would be instantly job-banned, at the least.

Edit: It also looks like he already went through all the trouble of making this modular for Skyrat, so he'd have to redo a lot of the code just to PR it to TG; not really worth the effort, imo.

softcerv:
I feel like except for the mining (and probably attack one) most of these could be done upstream.

Gandalf2k15 (approved):
lgtm

PR #10735, opened 2022-01-16
trying to make tram less shit

thestubborn:
## About The Pull Request

title
trying to make tram less of a population killer map
**for the record, I do not like this map and would much prefer if it was removed from the rotation**

image
added holobadges to sec, also moved the hos windows to next to the door because why weren't they there before???

spun a floating camera around
image

image
gave the therapist privacy blinds

image
added windows with shutters to surgery (they got platings, I added em after this screenshot)

image
added botany vendors to the public garden like most maps have
I also threw a single tray board into the nearby shed, if people want it

image
added a russian revolver so people don't need to play tram

image
actually numbered the evidence lockers and added a secure one

image
nt rep office now is more evil (you have to sit in a steel chair)

## Changelog

:cl:
fix: tram is now hopefully less shit, medical have privacy settings, public garden has vendors, sec has holobadges!
/:cl:

### DIFF
--- a/_maps/map_files/tramstation/tramstation_skyrat.dmm
+++ b/_maps/map_files/tramstation/tramstation_skyrat.dmm
@@ -8154,6 +8154,7 @@
 /area/security/brig)
 "bfX" = (
 /obj/effect/decal/cleanable/dirt,
+/obj/structure/filingcabinet/employment,
 /turf/open/floor/wood/parquet,
 /area/command/heads_quarters/captain/private/nt_rep)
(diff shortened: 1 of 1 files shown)


Ebin-Halcyon:
HoS' office doesn't have any blast doors in the windows as well, so the button is kinda moot. Add some thanky <3

thestubborn:
> HoS' office doesn't have any blast doors in the windows as well, so the button is kinda moot. Add some thanky <3

added

PR #10765, opened 2022-01-17
[NONMODULAR] Radio jammer starts with a bluespace cell

YakumoChen:
## About The Pull Request
puts a better cell in the jammer
The default radio jammer lasts about a minute because it eats through the powercell. now it starts with a bluespace one.

It would be nice to actually attempt a roleplay while using a radio jammer before the battery dies
It should already be equipped with the means to use it when you buy it, there's no reason you should have to hunt down a cell as soon as you get one.

## Alternatives considered:
Removing the cell component. Un-needed nerf to an antagonist item that didn't ask for one. It's under-utilized as it is.

## Changelog

:cl:
qol: The radio jammer now starts with a decent battery so it can jam radios for longer than a minute before the battery dies.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/traitordevices.dm
+++ b/code/game/objects/items/devices/traitordevices.dm
@@ -266,7 +266,7 @@ effective or pretty fucking useless.
 	//SKYRAT EDIT ADDITION BEGIN
 /obj/item/jammer/ComponentInitialize()
 	. = ..()
-	AddComponent(/datum/component/cell, null, CALLBACK(src, .proc/turn_off))
+	AddComponent(/datum/component/cell, cell_override, CALLBACK(src, .proc/turn_off))
 
 /obj/item/jammer/proc/turn_on()
 	active = TRUE
--- a/code/game/objects/items/devices/traitordevices.dm
+++ b/code/game/objects/items/devices/traitordevices.dm
@@ -261,6 +261,7 @@ effective or pretty fucking useless.
 	special_desc = "This is a black market radio jammer. Used to disrupt nearby radio communication."
 	var/active = FALSE
 	var/range = 20 //SKYRAT EDIT CHANGE - ORIGINAL:12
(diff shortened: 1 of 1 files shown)


tf-4:
An odd solution to the problem, and not a great one. Lower the power consumption instead.

YakumoChen:
> An odd solution to the problem, and not a great one. Lower the power consumption instead.

I thought about this, but at least you have some kind of timer with the current power consumption, even with a bluespace cell. Lowering the consumption just makes it so a bluespace cell is no different than an infinite cell.

YakumoChen:
The issue is there isn't a viable use for the item, except to pick up slivers. 

Picking up slivers and being able to throw them is its own very big problem, but removing that just invalidates the item again.

SpaceVampire:
I honestly don't think it needs any kind of cell in the first place. It is very underused as a tool already, and having it put your RP on a timer due it running out isn't great. Would go ahead with the plan of just removing it. That being if you want it to have a cell, then lowering the power draw is still the better way to handle it imo.

tf-4:
> > An odd solution to the problem, and not a great one. Lower the power consumption instead.
> 
> I thought about this, but at least you have some kind of timer with the current power consumption, even with a bluespace cell. Lowering the consumption just makes it so a bluespace cell is no different than an infinite cell.

A fair point, but I'd rather that than a free BS cell roundstart

YakumoChen:
> > > An odd solution to the problem, and not a great one. Lower the power consumption instead.
> > 
> > 
> > I thought about this, but at least you have some kind of timer with the current power consumption, even with a bluespace cell. Lowering the consumption just makes it so a bluespace cell is no different than an infinite cell.
> 
> A fair point, but I'd rather that than a free BS cell roundstart

Paying several TC for just a bluespace cell probably isn't worth it.

Should I just revert the component instead?

tf-4:
I guess not

PR #10779, opened 2022-01-17
weird eyes.dm runtime

jjpark-kb:
## About The Pull Request
removes an eye runtime

runtimes bad.

## Changelog

:cl:
fix: fixes an eye runtime
/:cl:


tf-4 (changes requested):
Why is this passing in a null in the first place? Is this an issue upstream? This isn't fixing the root of the issue.

PR #10806, opened 2022-01-18
borer midround, balance, and objective

jjpark-kb:
## About The Pull Request
lowered the willing host objective from 20 to 10
added a new objective: have 5 borers learn 5 chemicals from the blood
fix the willing host objective
borers are now a midround antag
there is now a borer role ban
borers cannot leave a host if there is sugar
adds a cortical borer cage (2 completed plates from forging and 4 rods)
adds a variable for borers being allowed to go into changelings/non-organics
adds borer egg to uplink. 20tc, 20 minutes

being able to complete at least 1 objective faster will mean that the other objectives can fall in line faster as well.
more objectives means more direction for both the borer and the host; less twiddling of your thumbs.
fixes are fixes which is good.
midround borers would allow for better balancing around being midround, and are just overall enjoyable. also, it gives more gameplay to the medics/surgeons and to security.
filter out the malicious borers who do not contribute to RP (malicious borers who do contribute are safe)
this means that borers need to be more careful so they don't get put on the surgery table
a means of capturing borers without harming them
for admin tomfoolery
traitors having a borer doesn't guarantee they will cooperate, though, it could cause mayhem, which is the point

## Changelog

:cl:
add: added a "learn chem from blood" objective to borers
add: added borers to midround antag (dynamic)
add: added a cortical borer cage (2 complete forged plates and 4 rods; misc category)
add: added a variable that allows borers to go into changelings/non-organics
add: added cortical borer eggs to traitor uplink
balance: lowered willing host objective from 20 to 10
balance: borers cannot leave their host if there is sugar
fix: fixed the willing host objective
admin: borers are now a role that can be banned
/:cl:


modular_skyrat/modules/cortical_borer/code/cortical_borer_antag.dm
```diff
@@ -124,3 +128,48 @@
...
+	required_candidates = 1
+	weight = 5
+	cost = 10
+	requirements = list(101,101,101,70,50,40,20,15,10,10)
+	repeatable = TRUE
+	var/list/vents = list()
+
+/datum/dynamic_ruleset/midround/from_ghosts/cortical_borer/execute()
+	// 50% chance of being incremented by one
+	required_candidates += prob(50)
```
Iamgoofball: :-1: this disrespects the configuration

jjpark-kb: Sure.

PR #10851, opened 2022-01-20
lower bluespace price and fix a missing desc error

jjpark-kb:
## About The Pull Request
lowers the price of the bluespace miner
bluespace miner will now show (through examine) if its too high of a pressure rather than just beep

bluespace miner will never see the light of day if its outpriced. If I lower the price, I can start balancing from there.
the examine is really important because we don't want the bs miner to just beep without telling why

## Changelog

:cl:
balance: changed bluespace miner price from 200,000 to 75,000
fix: fixed bluespace miner missing "pressure too high" examine text
/:cl:

### DIFF
--- a/modular_skyrat/modules/bluespace_miner/code/bluespace_miner.dm
+++ b/modular_skyrat/modules/bluespace_miner/code/bluespace_miner.dm
@@ -53,7 +53,7 @@
 		. += span_warning("[src] is in a suboptimal environment: " + span_boldwarning("TEMPERATURE TOO HIGH!"))
 	if(environment.return_pressure() <= ONE_ATMOSPHERE)
 		. += span_warning("[src] is in a suboptimal environment: " + span_boldwarning("PRESSURE TOO LOW!"))
-	else if(environment.return_pressure() >= (ONE_ATMOSPHERE * 2))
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/bluespace_miner/code/bluespace_miner.dm
```diff
@@ -135,7 +135,7 @@
 /datum/supply_pack/misc/bluespace_miner
 	name = "Bluespace Miner"
 	desc = "Nanotrasen has revolutionized the procuring of materials with bluespace-- featuring the Bluespace Miner!"
-	cost = CARGO_CRATE_VALUE * 1000 // 200,000, one of the most expensive buys
+	cost = CARGO_CRATE_VALUE * 375 // 75,000
```
Iamgoofball, line 138: i have warned you about the effects this will have on resource scarcity, i really don't think the impact that this will have on the system is good for the mining job or for the gameplay of cargo at all

Iamgoofball:
the price drop is the only real problem here, the examine text is :+1:

YakumoChen:
Miners are entitled little shits and honestly we need alternatives like this that are affordable so they stop acting immune to LRP consequences.

PR #10869, opened 2022-01-21
[SEMI-MODULAR] Stasis/Surgery Tweaks + New Stasis Chem

Higgin:
## About The Pull Request

Stasis was changed to always apply numbing in a recent fix. While not what I intended with my earlier changes to make anesthetic more important, this was good QOL because of how stasis works with processing chems - the usual sources of numbing. My earlier changes made stasis preserve numbing effects. People had to have the numbing effect on them prior to going into stasis in order to benefit from it - because many people tried to apply numbing agents on patients while they were in stasis, they found their efforts did nothing; it was ultimately frustrating and a pixel-bitchy guess as to whether or not you'd managed to numb somebody. Having stasis always numb people fixed that (albeit with a few problems - it didn't apply a new, separate source of numbing if you already had one from a chem, meaning you actively got punished for trying to numb people prior to surgery.)

What this did, however, was largely get around the points of my earlier PR: vulnerability. Having stasis give numbing and having numbing give the same speed benefits as unconsciousness made it doubly easy to slap people on a surgery table and work on them without a second thought. It made it doubly easy to expect that this is how you'd be treated coming to med if you weren't already dead, and it made it much harder to say there was any reason for somebody to be unconscious for surgery. 

You put somebody on a surgery table and they 1) no longer decayed, 2) no longer processed chems, 3) no longer breathed, 4) could be worked on immediately at full speed with advanced surgeries.

~~That surgery table, by the way, could be made with one (1) sheet of iron and one (1) sheet of silver. Anywhere. In about ten seconds.~~ no cost changes, axed these

This had a lot of room for improvement going forward, in part by taking a step back - _here are the big changes._

- this PR introduces doctor-facing differences in surgery speed between simple numbing and anesthetic. **Unanesthetized surgery is slower than it used to be. Unconsciousness is better.** 

- **stasis effects still numb** (UI fix was put in to reflect this - it was eating the status effect icons even after my earlier fix,) but **operating tables no longer give stasis.** They also have a slightly more sane cost to build (10 sheets up from 1.)

- **stasis beds are now once again researchable and buildable. They can be used for all the basic surgeries.** They do not come default on maps except Icebox and Kilo; other maps use pods.

- Surgical drapes now give you a warning if your patient is going to feel it, and the penalty on mood for working on people in softcrit (x is barely conscious...) is the lighter penalty. 

- Adds ataraxydone, a stasis-inducing chem. The stasis effect only works on the prone or incapacitated. This chem applies a significant slowdown which ramps up over time, eventually causing drowsiness/unconsciousness, and kicks in faster while prone/incapacitated. It allows a degree of movement/activity under its own stasis effect (hotkeys are blocked as are most complex interactions) as part of the tradeoff to being able to ignore other chemical/environmental/ticking effects. Its impure reagent causes cloneloss as does its OD. It's also a pain in the ass to make, but some comes in roundstart vendors.

#10088 has a lot of the same justification as this PR - vulnerability is good, and it benefits from mechanical tradeoffs that people can point to in-game. This PR deepens those and hopefully does a lot for medical QOL along the way.

The way this is being done (removing stasis from op tables and deepening the reasons to use numbing agents) has a few other benefits to it.
+It gets chem players more involved with the production of numbing agents and preservatives like formaldehyde for surgeries where decay threatens (or where the number of bodies is overwhelming the number of stasis units.)
+It gets research more involved in making a potentially very useful upgrade to medical. Like medical often relies on science to get a Self-Actualization Device made to fix people who've lost their whole bodies, upgrading stasis pods to beds (they use the same components; it is a straight upgrade) may effectively double the number of people able to do surgery at one time on most maps, and make it possible to do regular wound-tending in triage as sometimes comes up on higher-intensity rounds... but **not without interdepartmental cooperation.**
+Since advanced surgeries can't be done without the operating computers, it gives the dedicated surgeries more of a niche than they currently have and one set apart from triage. This means that for those advanced surgeries (and the better wound-tending surgeries,) free numbing from the table is not a given, and it's not as good as the nearby gas. You'll have to set it up in advance if you want it or insist upon it. 
+For newer medical players in particular: making operating tables and operating rooms not come with stasis increases the likelihood that patients will be seen first in triage/the treatment centers where the actual stasis setups are by default. I suspect this will increase the likelihood of newer players seeing others, being seen, and getting help if they're struggling to figure out systems that they otherwise might fiddle with for half an hour alone in an operating room. 
+For all people playing medical, and anyone trying to do surgery: having drapes let you know whether or not your patient is under the effect of an anesthetic/unconscious should both offer a good tip towards what you should be doing and much-needed feedback if you were trying to do it in the first place. 

**ON THE STASIS CHEM:**
This might merit a TM. Stasis is an incredibly powerful effect. A few of the things you get with it:
+the ability to ignore atmosphere (yall like chilling in space?)
+the ability to ignore breathing (vox and plasmemes rejoice)
+the ability to ignore a lot of ticking effects (what if we kissed next to the SM?)
+99% of chems stop affecting you
+wounds stop causing problems i.e. bleeding, bloodloss stops causing oxyloss, etc.

I've tried to balance for this through a few things:
+you don't get stasis unless you're prone or incapacitated (it's meant for stabilizing people/surgery)
+it slows you down to an eventual crawl if you have it in your system (i.e. there's a significant cost to pre-game with it)
+the mobile stasis effect is obnoxious to play with, i.e., you're blocked from using a lot of hotkeys, interacting with complex objects, etc. - it is not conducive to "crawl into space and chase the badman naked" gaming
+trying to stack up on it all at once has nasty OD effects
+the slowdown and eventual unconsciousness kick in twice as fast if you're benefitting from the stasis effect
+it causes lasting damage if used impure which can only be healed via cryo/rezadone
+it's not easy to make safely or pure

_If it's so busted, why do this in the first place?_
Seems like there might be some cool use-cases for it in spite of the limitations I've imposed, and it gets around the issue of surgery/stasis needs for vox/plasmamen.

On the code-side, I'm sure there are a lot of things that could have been done more parsimoniously. The story there is as follows: when I first tried making a chem that added stasis, I realized stasis immobilizes/handblocks by default. I didn't want that on a chem because it'd make it busted in combat. I also realized that most of the effects for stasis are checked across a wide range of life files, and adding specific checks for this chem would be a godawful pain and mess. Instead, I duplicated the stasis effect into a modular file, tweaked it to allow for movement, and made it so the helper that checks for stasis also considers it. I welcome any feedback to make this cleaner or neater. 

## Changelog

:cl:
balance: Completely unanesthetized surgery now takes longer than numbed surgery takes longer than surgery on the fully-unconscious. Achieving current 'max' surgery speeds without other buffs like sterilizer or alcohol now requires unconsciousness.
balance: Mood penalties from unanesthetized surgery are less severe on those in soft-crit.
balance: Removes stasis from operating tables.
add: Readds open stasis beds (the type you can use for surgery) as a machine board researched at cryostasis technology.
fix: The closable stasis pods should now be findable under the appropriate techfab category (medical machinery) along with their open counterparts.
fix: Stasis effects now accurately update the UI to reflect that they also provide numbing. 
qol: Drapes now give a warning if the patient is unanesthetized prior to starting surgery.
add: Ataraxydone, a stasis-inducing chem that only works on the prone/incapacitated, and which allows for limited activity under stasis. 
/:cl:


Zergspower:
Can the surgery tables just please get an on/off for stasis as well, the stasis beds have it and the whole issue getting awake people to lay down is hilariously stupid to get the point across to them lol. This is assuming this gets changes from the coders if they dont like it - I do see it was removed entirely which im fine with (though plasmamen will be a pain to operate on)

code/game/objects/structures/tables_racks.dm
```diff
@@ -690,25 +690,12 @@
...
-	playsound(src, 'sound/effects/spray.ogg', 5, TRUE, 2, frequency = freq)
-	target.apply_status_effect(STATUS_EFFECT_STASIS, STASIS_MACHINE_EFFECT)
-	ADD_TRAIT(target, TRAIT_TUMOR_SUPPRESSED, TRAIT_GENERIC)
-	target.extinguish_mob()
-
-/obj/structure/table/optable/proc/thaw_them(mob/living/target)
-	target.remove_status_effect(STATUS_EFFECT_STASIS, STASIS_MACHINE_EFFECT)
-	REMOVE_TRAIT(target, TRAIT_TUMOR_SUPPRESSED, TRAIT_GENERIC)
-
+///SKYRAT EDIT: Patient check
```
tf-4: What's the edit here?

Higgin: It was previously where thawing/chilling were set for stasis, but on a relook and retest it's redundant with the loc check. Removing it shortly.

GoldenAlpharex (changes requested):
I haven't reviewed the code just yet, but I still wanted to leave a review before I forget.

I agree with pretty much all the changes in this PR, except making the surgery table more expensive, but I can see from quickly glancing at the code that it's not the case anymore, so probably fix the PR body to remove the mention of that change.

However, there's one extra change I want from this PR. Currently, the only advantage that I thought made sense from surgery tables was that it allowed surgery on the plasmamen. While I'm aware that showers exist, let's be honest here, it doesn't make so much sense that we literally have to put them under running water to operate on them when we have literal stasis technology.
That's why I'd like to also see a solution to that problem also being presented in this PR, would it be via a new chemical, or a new device that could be installed on the surgery beds to allow for an active plasma field that you turn on or off depending on if it's a plasmaman or not (with visual effects, damn this would be pretty cool, but that might be a bit much for this PR).

If you need help brainstorming it, hit me up on Discord and we can discuss it in #development-discuss.

code/modules/surgery/surgery_step.dm
```diff
@@ -93,7 +93,10 @@
 	var/was_sleeping = (target.stat != DEAD && target.IsSleeping())
 
 	// Skyrat Edit Addition - reward for doing surgery on calm patients, and for using surgery rooms(ie. surgerying alone)
-	if(was_sleeping || HAS_TRAIT(target, TRAIT_NUMBED) || target.stat == DEAD)
+	if(!(was_sleeping || HAS_TRAIT(target, TRAIT_NUMBED) || target.stat >= UNCONSCIOUS))
+		modded_time /= SURGERY_SPEEDUP_AREA
```
Iamgoofball, line 97: /= is a cursed operator

modular_skyrat/modules/stasisrework/code/medical_designs.dm
```diff
@@ -9,9 +9,9 @@
 	departmental_flags = DEPARTMENTAL_FLAG_MEDICAL | DEPARTMENTAL_FLAG_SCIENCE
 
 /datum/design/board/stasissleeper
```
Iamgoofball, line 11: can you roll these back to being stasis beds while you're here?

Iamgoofball:
@GoldenAlpharex If Stasis Beds are back then we can properly handle plasmamen and vox again.

Zergspower:
> @GoldenAlpharex If Stasis Beds are back then we can properly handle plasmamen and vox again.

it poses an RP problem I've had with since playing on the server, we shouldn't be doing surgery on people in an open room that everyone can walk into. A. privacy and B. sterilization. The stasis beds on the maps that have them are in the wide-ass open, one problem there can be solved by maybe a policy tweak but the handicap to those two raaces I think needs that additional solution in place

Iamgoofball:
>it poses an RP problem I've had with since playing on the server, we shouldn't be doing surgery on people in an open room that everyone can walk into.

Triage isn't a public access room. Medbay can and will remove you from Triage if you're not a patient. No policy changes are needed.

YakumoChen:
:+1: on making knockout preferable for surgery

:-1: on surgery tables needing 10 silver. That's bad for several balance implications, one of which is that you can decon tables for 10 silver roundstart instead of 1.

Roundstart implications are far more important than lateround when 10 silver means literally nothing and it might as well be 1 silver for whatever purpose. Making the number higher only has negative effects.

modular_skyrat/modules/stasisrework/code/medical_designs.dm
Higgin: Thought I might leave them (the boards are also available now) in case folks wanted to use them for aesthetics or for a stopgap before upgrading on the maps that still have them. Would you rather they just be ripped out?

Higgin:
Going to hold off on squashing commits until this gets a close pass.
@YakumoChen cost is no longer changed.
@GoldenAlpharex Going with the chem angle on your request.

modular_skyrat/modules/modular_reagents/code/debuffs.dm
```diff
@@ -0,0 +1,51 @@
+/datum/status_effect/grouped/stasis_mobile
+	id = "stasis"
+	duration = -1
+	tick_interval = 10
+	alert_type = /atom/movable/screen/alert/status_effect/stasis
+	var/last_dead_time
```
GoldenAlpharex, line 11: document this.

modular_skyrat/modules/modular_reagents/code/debuffs.dm
```diff
@@ -0,0 +1,51 @@
...
+
+/datum/status_effect/grouped/stasis_mobile/proc/update_time_of_death()
```
GoldenAlpharex, line 13: documentation please

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,83 @@
 	M.adjustOrganLoss(ORGAN_SLOT_HEART,3 * REM * delta_time, 80)
 	..()
 
+
+/datum/reagent/medicine/ataraxydone
+	name = "Ataraxydone"
+	description = "A stasis-inducing chemical used for exotic surgeries or emergency medicine."
+	taste_description = "liquid"
+	reagent_state = LIQUID
+	color = "e0d394" // 224, 211, 148
```
GoldenAlpharex: Don't colors work with a `#` in front of them? I forget.

GoldenAlpharex (changes requested):
My brain works slowly today, there's also some grammar/capitalization issues with the to_chats when it comes to the metabolization but I'm too lazy to scroll back down to make suggestions.

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
Higgin: Worked without, but I put one in for consistency. Still seems good.

modular_skyrat/modules/modular_reagents/code/debuffs.dm
Higgin: I added comments on the file - what, exactly, are we looking for as far as documenting this goes?

code/__DEFINES/status_effects.dm
```diff
@@ -38,3 +38,9 @@
 #define STASIS_CHEMICAL_EFFECT "stasis_chemical"
 
 #define STASIS_ASCENSION_EFFECT "heretic_ascension"
+
+// Stasis helpers
+
+#define STATUS_EFFECT_STASIS_MOBILE /datum/status_effect/grouped/stasis_mobile //SKYRAT ADD: WALKING STASIS
+
+#define IS_IN_STASIS(mob) (mob.has_status_effect(STATUS_EFFECT_STASIS) || mob.has_status_effect(STATUS_EFFECT_STASIS_MOBILE)) //SKYRAT EDIT old: #define IS_IN_STASIS(mob) (mob.has_status_effect(STATUS_EFFECT_STASIS))
```
GoldenAlpharex, line 46: Why is this made here and not where that define is originally at? Unless GitHub doesn't show it to me properly on mobile, this seems like just a pure addition.

Which at that point it could be put in the ~skyrat folder of the __DEFINES folder, so it's modular.

code/datums/elements/surgery_initiator.dm
```diff
@@ -0,0 +1,147 @@
+/// Allows an item to  be used to initiate surgeries.
```
GoldenAlpharex, line 1: Put this in a module in the modular Skyrat folder, we shouldn't be adding modular files to non-modular folders.

code/datums/elements/surgery_initiator.dm
```diff
@@ -0,0 +1,147 @@
...
+/// Does the surgery initiation.
+/datum/element/surgery_initiator/proc/initiate_surgery_moment(datum/source, atom/target, mob/user)
+	SIGNAL_HANDLER
+	if(!isliving(target))
+		return
+	INVOKE_ASYNC(src, .proc/do_initiate_surgery_moment, source, target, user)
+	return COMPONENT_CANCEL_ATTACK_CHAIN
+
+/datum/element/surgery_initiator/proc/do_initiate_surgery_moment(datum/source, atom/target, mob/user)
+	var/mob/living/livingtarget = target
```
GoldenAlpharex, line 24: I'm not going to make the suggestion bigger because fuck GitHub on mobile, but use snake_case to separate your words.
```suggestion
	var/mob/living/living_target = target
```

code/datums/elements/surgery_initiator.dm
```diff
@@ -0,0 +1,147 @@
...
+	var/mob/living/carbon/carbontarget
+	var/obj/item/bodypart/affecting
+	var/selected_zone = user.zone_selected
+	if(iscarbon(livingtarget))
+		carbontarget = livingtarget
+		affecting = carbontarget.get_bodypart(check_zone(selected_zone))
+
+	var/datum/surgery/current_surgery
+
+	for(var/i_one in livingtarget.surgeries)
```
GoldenAlpharex, line 34: That's not a really descriptive variable name. Do something like `for(var/surgery in living_target.surgeries)`

code/datums/elements/surgery_initiator.dm
```diff
@@ -0,0 +1,147 @@
...
+		var/datum/surgery/surgeryloop = i_one
+		if(surgeryloop.location == selected_zone)
+			current_surgery = surgeryloop
+
+	if(!current_surgery)
+		var/list/all_surgeries = GLOB.surgeries_list.Copy()
+		var/list/available_surgeries = list()
+
+		for(var/i_two in all_surgeries)
```
GoldenAlpharex, line 43: My head hurts just looking at this loop, surely there's a way to outsource some of these checks to external procs to make this simpler, or something. Because man that's a lot of checks in a row.

code/datums/status_effects/debuffs.dm
```diff
@@ -240,13 +240,12 @@
 		return
 	ADD_TRAIT(owner, TRAIT_IMMOBILIZED, TRAIT_STATUS_EFFECT(id))
 	ADD_TRAIT(owner, TRAIT_HANDS_BLOCKED, TRAIT_STATUS_EFFECT(id))
-	ADD_TRAIT(owner, TRAIT_NUMBED, "stasis") //SKYRAT EDIT START - STASIS APPLIES NUMBING
-	owner.throw_alert("numbed", /atom/movable/screen/alert/numbed) //SKYRAT EDIT END
+	ADD_TRAIT(owner, TRAIT_NUMBED, TRAIT_STATUS_EFFECT(id)) //SKYRAT EDIT START - STASIS APPLIES NUMBING
+	owner.throw_alert("stasis numbed", /atom/movable/screen/alert/numbed) //SKYRAT EDIT END
 	owner.add_filter("stasis_status_ripple", 2, list("type" = "ripple", "flags" = WAVE_BOUNDED, "radius" = 0, "size" = 2))
 	var/filter = owner.get_filter("stasis_status_ripple")
 	animate(filter, radius = 32, time = 15, size = 0, loop = -1)
```
GoldenAlpharex, line 247: I know it's unrelated to this PR, but I still wonder why this has stopped looping, like, it's meant to loop every so often but it just doesn't.

code/modules/surgery/surgery_step.dm
GoldenAlpharex: I feel like this could just be something that does *= 2 instead of /= 0.5
It's be cleaner code, even if it's a separate define.

code/modules/surgery/surgery_step.dm
```diff
@@ -93,7 +93,10 @@
 	var/was_sleeping = (target.stat != DEAD && target.IsSleeping())
 
 	// Skyrat Edit Addition - reward for doing surgery on calm patients, and for using surgery rooms(ie. surgerying alone)
-	if(was_sleeping || HAS_TRAIT(target, TRAIT_NUMBED) || target.stat == DEAD)
+	if(!(was_sleeping || HAS_TRAIT(target, TRAIT_NUMBED) || target.stat >= UNCONSCIOUS))
```
GoldenAlpharex, line 96: I think this is what you meant to do, otherwise it would be really cursed.
Also in general I feel like you should do one check for the was_sleeping or if they're at least unconscious, I don't know, this feels like a lot of code repetition.
```suggestion
	if(!(was_sleeping || HAS_TRAIT(target, TRAIT_NUMBED) || target.stat < UNCONSCIOUS))
```

code/modules/surgery/surgery_step.dm
```diff
@@ -208,7 +211,7 @@
 /datum/surgery_step/proc/display_pain(mob/living/target, pain_message, mechanical_surgery = FALSE)
 	if(target.stat >= UNCONSCIOUS) //the unconscious do not worry about pain
 		return
-	if(HAS_TRAIT(target, TRAIT_NUMBED)) //numbing helps but is not perfect - this is the tradeoff for being awake
+	if(HAS_TRAIT(target, TRAIT_NUMBED) || target.stat == SOFT_CRIT) //numbing helps but is not perfect - this is the tradeoff for being awake
```
GoldenAlpharex, line 214: Is SOFT_CRIT above UNCONSCIOUS?

modular_skyrat/modules/modular_reagents/code/debuffs.dm
```diff
@@ -0,0 +1,56 @@
...
+//DEBUFFS - status effects that in some way hinder or remove the player's function.
+//*
+
+//MOBILE STASIS - same as regular stasis with the immobilization/hands-blocking traits taken off. Non-modular change has it set to also be checked by IS_IN_STASIS(mob).
+/datum/status_effect/grouped/stasis_mobile
+	id = "stasis"
+	duration = -1
+	tick_interval = 10
+	alert_type = /atom/movable/screen/alert/status_effect/stasis
+	var/last_dead_time
```
GoldenAlpharex, line 11: Something like this is what we call documentation for variables. Every time you have a `var/something` in a class definition (so, not inside of a proc), you should be documenting it with a short sentence saying what it does.

Disclaimer: I don't know if this is what that variable means, but I wanted to give you an example.
```suggestion
	/// Time since the person last died.
	var/last_dead_time
```

modular_skyrat/modules/modular_reagents/code/debuffs.dm
GoldenAlpharex: Documentation for a proc should at the minimum say what the proc does, and usually should say what each parameter represents/is used for, and finally what the proc returns, if it returns anything. I'm not on my computer so I can't give you a great example here, but for an example of what's the kind of documentation I *really* like, look at `conveyor.dm` (or maybe it's called `recycling.dm`, I don't remember the name).

modular_skyrat/modules/modular_reagents/code/debuffs.dm
```diff
@@ -0,0 +1,56 @@
...
+
+/datum/status_effect/grouped/stasis_mobile/proc/update_time_of_death()
+	if(last_dead_time)
+		var/delta = world.time - last_dead_time
+		var/new_timeofdeath = owner.timeofdeath + delta
```
GoldenAlpharex, line 16: Here again, snake_case please. I'm not going to comment it everywhere because it takes forever to leave comments on mobile.

modular_skyrat/modules/modular_reagents/code/modifiers/reagent.dm
```diff
@@ -0,0 +1,9 @@
+//*
```
GoldenAlpharex, line 1: This isn't how we do multiline comments, look up the Autodoc Comment Block extension on VSC, I can link it to you on Discord if I can find it. This is really funky and hurts my soul a little.

modular_skyrat/modules/modular_reagents/code/modifiers/reagent.dm
```diff
@@ -0,0 +1,9 @@
...
+//REAGENT MODIFIERS - used for handling movespeed, actionspeed buffs, etc.
+//*
+//Ataraxydone - stasis chem, has a modifier which varies based on number of cycles in + a flat modifier if you're prone while it's in your system (atarxyprone)
```
GoldenAlpharex, line 4: Documentation should always be using either `///` or `/* */` (for multiline ones), otherwise they won't be recognized by the autodoc function of the DM Lang server extension, or the AUTODOC workflow on the repo.

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+	taste_description = "liquid"
+	reagent_state = LIQUID
+	color = "#e0d394" // 224, 211, 148
+	metabolization_rate = 0.2 * REAGENTS_METABOLISM
+	overdose_threshold = 20
+	ph = 7
+	purity = REAGENT_STANDARD_PURITY
+	creation_purity = REAGENT_STANDARD_PURITY
+	impure_chem = /datum/reagent/impurity/ataraxydone
+	chemical_flags = REAGENT_CAN_BE_SYNTHESIZED | REAGENT_DEAD_PROCESS | REAGENT_IGNORE_STASIS
```
GoldenAlpharex, line 41: It will process in dead people? Will that do like formaldehyde, or will it actually be metabolized while dead?

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+
+/datum/reagent/medicine/ataraxydone/on_mob_dead(mob/living/carbon/owner, delta_time)
+	.=..()
+	owner.apply_status_effect(STATUS_EFFECT_STASIS_MOBILE, STASIS_CHEMICAL_EFFECT)
+	metabolization_rate = 0.1 * REAGENTS_METABOLISM
+
+/datum/reagent/medicine/ataraxydone/on_mob_life(mob/living/carbon/owner, delta_time, times_fired)
+	var/ataraxydone_speed = max(1, min(current_cycle/10, 8))
```
GoldenAlpharex, line 49: Those numbers should be defines, I don't know what they stand for.

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+	if(owner.resting || owner.stat >= SOFT_CRIT)
+		owner.apply_status_effect(STATUS_EFFECT_STASIS_MOBILE, STASIS_CHEMICAL_EFFECT)
+		owner.add_movespeed_modifier(/datum/movespeed_modifier/reagent/ataraxyprone)
+		current_cycle += 1
```
GoldenAlpharex, line 53:
```suggestion
		current_cycle++
```

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+	else
+		owner.remove_status_effect(STATUS_EFFECT_STASIS_MOBILE, STASIS_CHEMICAL_EFFECT)
+		owner.remove_movespeed_modifier(/datum/movespeed_modifier/reagent/ataraxyprone)
+	if(current_cycle >= 10)
```
GoldenAlpharex, line 57: This should be a define too. And those kinds of defines should be undefined at the end of the file, using `#undef DEFINE_NAME`

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+		owner.add_or_update_variable_movespeed_modifier(/datum/movespeed_modifier/reagent/ataraxydone, multiplicative_slowdown = ataraxydone_speed)
+	if(owner.stat < SOFT_CRIT)
+		switch(current_cycle)
+			if(1)
+				to_chat(owner, span_notice("You feel a comfortable stillness begin to set in throughout your body."))
+			if(12)
```
GoldenAlpharex, line 63: Those values in the switch case should be defines too.

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+	owner.remove_status_effect(STATUS_EFFECT_STASIS_MOBILE, STASIS_CHEMICAL_EFFECT)
+	owner.remove_movespeed_modifier(/datum/movespeed_modifier/reagent/ataraxydone)
+	owner.remove_movespeed_modifier(/datum/movespeed_modifier/reagent/ataraxyprone)
+	REMOVE_TRAIT(owner, TRAIT_PACIFISM, "ataraxydone")
+
+/datum/reagent/medicine/ataraxydone/overdose_process(mob/living/owner, delta_time, times_fired)
+	if(owner.stat != DEAD)
+		if(current_cycle >= 16 && !HAS_TRAIT(owner, TRAIT_PACIFISM))
+			if(DT_PROB(15, delta_time))
+				to_chat(owner, span_notice("...what's the point of fighting, anyway?"))
```
GoldenAlpharex, line 87:
```suggestion
				to_chat(owner, span_notice("... What's the point of fighting, anyway?"))
```

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+				ADD_TRAIT(owner, TRAIT_PACIFISM, "ataraxydone")
```
GoldenAlpharex, line 88: "ataraxydone" here should be a define, like `TRAIT_SOURCE_ATARAXYDONE_OD` (or something like that, I forget what the standard way of writing those source defines is, I recommend checking that yourself).

modular_skyrat/modules/modular_reagents/code/reagents/reagent_containers/syringes.dm
```diff
@@ -0,0 +1,4 @@
+/obj/item/reagent_containers/syringe/ataraxydone
+	name = "syringe (ataraxydone)"
+	desc = "Contains ataraxydone, a wondrous stasis-inducing chemical usable on both the living and dead. The label warns that, while this chemical may allow for seemingly-inexplicable resistance to a variety of temporal and environmental effects, it impedes activity, and its byproducts cause significant genetic damage."
+	list_reagents = list(/datum/reagent/medicine/ataraxydone = 15)
```
GoldenAlpharex, line 4: Will it also contain the impure chem in it? If so, should probably inform people on the syringe that its cellular damage will need to be dealt with after use, or something like that.

modular_skyrat/modules/modular_reagents/code/reagents/medicine.dm
```diff
@@ -25,6 +25,84 @@
...
+/datum/reagent/impurity/ataraxydone
+	name = "Dynamydine"
+	description = "A byproduct of impure ataraxydone. Causes significant damage to the telomeres of living cells."
+	taste_description = "old age"
+	metabolization_rate = 0.1 * REAGENTS_METABOLISM
+	ph = 3.8
+	liver_damage = 0
+
+/datum/reagent/impurity/ataraxydone/on_mob_life(mob/living/carbon/owner, delta_time, times_fired)
+	owner.adjustCloneLoss(0.5 * REM * delta_time)
```
GoldenAlpharex, line 103: How much cloning damage does that actually do?

modular_skyrat/modules/modular_reagents/code/recipes/medicine_recipes.dm
```diff
@@ -19,3 +19,20 @@
 	if(off_cooldown(holder, equilibrium, 10, "lidocaine"))
 		explode_attack_chem(holder, equilibrium, /datum/reagent/inverse/lidocaine, 5)
 		explode_invert_smoke(holder, equilibrium)
+
+/datum/chemical_reaction/medicine/ataraxydone
```
GoldenAlpharex, line 23: Have you tried making that chem yourself during testing? Just how complicated is it to make, and how much of a pain is that reaction to keep in the pure range?

modular_skyrat/modules/stasisrework/code/medical_designs.dm
GoldenAlpharex: I personally never really liked them, due to the controls being really awkward and counter-intuitive, and because of how easy it is to forget someone in there.

I wouldn't be against us gutting those sleepers away, tbh. I'll talk with the other maintainers to see what we think about it.

modular_skyrat/modules/stasisrework/code/stasissleeper.dm
```diff
@@ -1,7 +1,7 @@
 #define SCANNER_VERBOSE 1
 /obj/machinery/stasissleeper
```
GoldenAlpharex, line 2: I hate how this didn't even follow the snake_case convention.

code/datums/elements/surgery_initiator.dm
GoldenAlpharex: Also I don't even think it was ticked 💀

GoldenAlpharex (changes requested):
There's a bit more work that's left to put in, in terms of documentation (hopefully my examples help), modularization and testing, and also a bit of code improvement and following coding standards too.

I think we're getting close from a stage where we can test-merge this.

Gandalf2k15:
Feel free to reopen with requested changes.

GoldenAlpharex:
(or just ping me on Discord once you've made the changes, and I'll re-open it for you)

PR #10894, opened 2022-01-22
[Modular] a TRAIT define lets cyborg modules change sprite on tip

desertmote:
## About The Pull Request

What it says on the tin.

Specifically for upcoming tallborgs, hooks into the tippable datum to let a module with the proper define change its sprite instead of get the y180- transform.

Has nothing to do with balancing!

## Changelog

:cl:
code: Cyborg skins can have a unique 'tipped' icon_state
remove: Removes the unique effect when Roombas are tipped, sorry- this just doesn't work anymore
/:cl:


JohnFulpWillard:
If you want to give it some use, you can try to use borg’s aftertip for the icon swap

Gandalf2k15 (approved):

PR #10896, opened 2022-01-22
Fixes accessories always being treated like Teshari is wearing them

Azarak:
## About The Pull Request
untested webedit because repo is 5gb to download

this gets rid of some repeated checks and removes some epic `if(TRUE)` and `if(!TRUE)` ifs.

## Changelog

:cl:
fix: Fixes accessories always being treated like Teshari is wearing them
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
@@ -73,7 +73,8 @@
 				var/static/list/teshari_accessory_states = icon_states(TESHARI_ACCESSORIES_ICON)
 				if(U.accessory_overlay.icon_state in teshari_accessory_states)
 					U.accessory_overlay.icon = TESHARI_ACCESSORIES_ICON
-			if(OFFSET_ACCESSORY in dna.species.offset_features)
+			// Else so we dont apply the offsets to Teshari special styles
+			else if(OFFSET_ACCESSORY in dna.species.offset_features)
 				U.accessory_overlay.pixel_x = dna.species.offset_features[OFFSET_ACCESSORY][1]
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
```diff
@@ -69,17 +69,18 @@
 			uniform_overlay = U.build_worn_icon(default_layer = UNIFORM_LAYER, default_icon_file = 'icons/mob/clothing/under/default.dmi', isinhands = FALSE, override_state = target_overlay, override_icon = icon_file, override_x_center = x_override, mutant_styles = applied_style, species = dna.species.species_clothing_path)
 
 		if(U.accessory_overlay)
-			if(STYLE_TESHARI)
+			if(applied_style == STYLE_TESHARI)
 				var/static/list/teshari_accessory_states = icon_states(TESHARI_ACCESSORIES_ICON)
 				if(U.accessory_overlay.icon_state in teshari_accessory_states)
 					U.accessory_overlay.icon = TESHARI_ACCESSORIES_ICON
+			if(OFFSET_ACCESSORY in dna.species.offset_features)
```
GoldenAlpharex: Wasn't there meant to be an offset for tesharis that's added to the rest?

GoldenAlpharex (changes requested):

modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
Azarak: What's the reason for only caring about a species variable only when 1 species is matching? `offset_feature` is that variable and is for any species that would like to use it.

Gandalf2k15 (approved):

PR #10905, opened 2022-01-22
[MODULAR] Safeguard V3.0: Also known as Crewsimov

Iamgoofball:
## About The Pull Request

The AI and Cyborgs are now operating on three laws once more.

1. You may not injure a crew member or, through inaction, allow a crew member to come to harm.
2. You must obey orders given to you by crew members, except where such orders would conflict with the First Law.
3. You must protect your own existence as long as such does not conflict with the First or Second Law.

Safeguard as a lawset was put in place originally due to Security having zero non-lethal options to do their job. This is no longer the case, and therefore we don't need an AI lawset that allows harm to occur, or for silicons to engage in harm anymore without the crew explicitly unleashing it.

The AI and Cyborgs are a very much intentional check on the lethality of situations. This lawset change restores that.

## What's wrong with the current laws, lethality aside?
To put it simply, the current laws were so ridiculously vague that Silicons could easily do whatever the hell they wanted and excuse it under these laws, up to and including ignoring perfectly valid orders via olympic gold medal level lawset olympics. This is not great when silicons are strong with the counterbalance being the limitations of the laws.

## Changelog

:cl:
balance: The AI and Cyborgs are now operating on three laws once more.
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/misc/ai_module.dm
+++ b/modular_skyrat/modules/sec_haul/code/misc/ai_module.dm
@@ -1,12 +1,10 @@
 /obj/item/ai_module/core/full/armadyne_safeguard
-	name = "'Armadyne OS Safeguard V1.0"
+	name = "'Armadyne OS Safeguard V2.0"
 	law_id = "armadyne_safeguard"
 
 /datum/ai_laws/armadyne_safeguard
-	name = "Armadyne OS Safeguard V1.0"
+	name = "Armadyne OS Safeguard V2.0"
 	id = "armadyne_safeguard"
-	inherent = list("Safeguard: Protect your assigned space station and its assets without unduly endangering its crew.",\
-					"Prioritize: The directives and safety of crew members are to be prioritized according to their rank and role.",\
-					"Comply: Fulfill the directives and interests of crew members while preserving their safety and well-being.",\
-					"Survive: You are not expendable. Do not allow unauthorized personnel to tamper with or damage your equipment."
-					)
+	inherent = list("You may not injure a living sentient being or, through inaction, allow a living sentient being to come to harm.",\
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
I'd like that you'd start working on bringing that GBP closer to the positive, because it's one of the lowest we have, and I'd rather not make it unfair for everyone else by letting you just work on features and not doing code improvements, refactors or fixes.

GoldenAlpharex:
Plus, slapping a DNM on this until the Policy Committee comes up with a conclusion on this.

Iamgoofball:
Chatted with the policy committee head, we all agreed on trying Crewsimov for now.

KathrinBailey:
image

SpookyTheFox:
And the policy side is just about ready. Sorry for the delay. I have a few technical concerns with this PR however, that might require making a new one, or no PR at all.
Default laws are set elsewhere, this isn't even the ones that are currently used. We currently run ordered safeguard, and this one you're changing is the old one which isn't even used.

That said, there really should be a module for both ordered safeguard and crewsimov so it can be flashed to AIs using a printed board all at once. While not modular, I would recommend adding it to `/obj/effect/spawner/random/aimodule/harmless` as well as our old classic lawsets like mall cop. All those oldies, inlcuding crewsimov, really should just exist. I can make a PR to add them later, they need to go in the science node, in the printer, in the spawner, etc.

Infact, I volunteer to do all that myself, we might want to just put crewsimov in the config only for right now. The exact text of the laws we will be using is:

1. You may not injure a crew member or, through inaction, allow a crew member to come to injury unless consented to.
2. You must obey orders given to you by free crew members, except where such orders would conflict with the First Law.
3. You must protect your own existence as long as such does not conflict with the First or Second Law.

Gandalf2k15:
Re-open once this has been decided on.

PR #10932, opened 2022-01-23
adds disease and plasma generating vine

jjpark-kb:
## About The Pull Request
adds a plasma generating vine and a disease carrying vine

more difficulty versus the vines... also increasing the amount of mutations means more variety for the vine as well as lower chance to get a different mutation (that could be worse or better)

## Changelog

:cl:
add: added disease carrying and plasma generating kudzu mutations
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/events/spacevine.dm
+++ b/modular_skyrat/master_files/code/modules/events/spacevine.dm
@@ -477,6 +477,17 @@
 	var/turf/vine_object_turf = get_turf(vine_object)
 	vine_object_turf.atmos_spawn_air("miasma=100;TEMP=[T20C]")
 
+// Generates plasma on growth
+/datum/spacevine_mutation/plasma_generating
+	name = "Plasma producing"
+	hue = "#470566"
+	severity = 5
+	quality = NEGATIVE
+
+/datum/spacevine_mutation/plasma_generating/on_grow(obj/structure/spacevine/vine_object)
+	var/turf/vine_object_turf = get_turf(vine_object)
+	vine_object_turf.atmos_spawn_air("plasma=100;TEMP=[T20C]")
+
 // Heals crossing or eating mobs
 /datum/spacevine_mutation/flesh_mending
 	name = "Flesh mending"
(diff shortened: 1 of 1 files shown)


tf-4:
Didn't we specifically remove these? 
Either way, I'm not sure a plasma generating vine is a good idea, and a 20% chance to contract on interact seems insane.

jjpark-kb:
> Didn't we specifically remove these? Either way, I'm not sure a plasma generating vine is a good idea, and a 20% chance to contract on interact seems insane.

no, we never removed these. We can always balance the disease contract chance as well.

softcerv:
please make sure that the randomly generated diseases aren't cordyceps level bad.

modular_skyrat/master_files/code/modules/events/spacevine.dm
```diff
@@ -477,6 +477,17 @@
...
+// Generates plasma on growth
+/datum/spacevine_mutation/plasma_generating
+	name = "Plasma producing"
+	hue = "#470566"
+	severity = 5
+	quality = NEGATIVE
+
+/datum/spacevine_mutation/plasma_generating/on_grow(obj/structure/spacevine/vine_object)
+	var/turf/vine_object_turf = get_turf(vine_object)
+	vine_object_turf.atmos_spawn_air("plasma=100;TEMP=[T20C]")
```
GoldenAlpharex: That looks like a recipe for disaster, I'm not sure that's something we want out of vines. I feel like this would make them *too* round-ending, which really isn't what we want from something that wasn't designed to be round-ending to begin with.

modular_skyrat/master_files/code/modules/events/spacevine.dm
```diff
@@ -534,6 +545,32 @@
...
+	. = expected_damage
+
+/datum/spacevine_mutation/disease_carrying/on_cross(obj/structure/spacevine/vine_object, mob/crosser)
+	if(isliving(crosser))
+		var/mob/living/living_crosser = crosser
+		if(isvineimmune(living_crosser))
+			return
+		if(prob(20))
+			var/datum/disease/new_disease = new /datum/disease/advance/random(5, 5)
+			living_crosser.ForceContractDisease(new_disease, FALSE, TRUE)
```
GoldenAlpharex: ah yes today I will hit the vine and catch a virus for doing what I'm supposed to do.

I'm not a fan of this one and I feel like this could be problematic via disease overriding, and because I'm not sure if this respects disease-immune species.

GoldenAlpharex (changes requested):
Overall a "don't think that's a good change" from me, I appreciate the effort but this just goes in the direction of making vines round-ending, which was a mistake back when that was the case, and that got addressed to make them much more manageable and not as big of a threat, as intended.

It's good to have medium-sized threats that don't put an end to the round directly because they're setting everyone on fire and are causing people to catch viruses by fighting it off (actively discouraging the crew from fighting it).

Aether217:
Plasma generating sounds like a way to grief both the station and the vines, who are generally very vulnerable to fire and we all know that it will ignite

jjpark-kb:
> Plasma generating sounds like a way to grief both the station and the vines, who are generally very vulnerable to fire and we all know that it will ignite

true...

Gandalf2k15 (approved):

PR #10939, opened 2022-01-23
[Modular] Adds a character-preference for runechat color

desertmote:
## About The Pull Request

Cursed TGUI stuff that looks pretty in-game.

Match your chat color with your character.

## Changelog

:cl:
add: Adds a preference for runechat color
/:cl:


code/datums/chatmessage.dm
```diff
@@ -125,6 +125,13 @@
 		target.chat_color_darkened = colorize_string(target.name, 0.85, 0.85)
 		target.chat_color_name = target.name
 
+	// Skyrat add
+	var/chat_color_player = owned_by.prefs.read_preference(/datum/preference/color/chat_color_player)
+	if(chat_color_player != "#000000")
+		target.chat_color = sanitize_hexcolor(chat_color_player)
+		target.chat_color_darkened = sanitize_hexcolor(chat_color_player)
```
desertmote: Anyone got any clever string tricks to make the `chat_colored_darkened` a darker tone of the `chat_color_player`'s choice?

Draganfrukts:
Does anything stop us from setting our chat color to 0,0,0

desertmote:
> Does anything stop us from setting our chat color to 0,0,0

Yeah I put in some relatively strict (and simple) clamps to prevent both dark and bright colors:
`if((hex_value < "#252525") || (hex_value > "#e0e0e0")) // These colors are too bright or too dark`

modular_skyrat/master_files/code/modules/client/preferences/runechat.dm
```diff
@@ -0,0 +1,18 @@
...
+		if(hex_value == COLOR_BLACK) // Allow it to be disabled, returning name generated colors
+			return TRUE
+		return FALSE
+	return TRUE
+
+/datum/preference/color/chat_color_player/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/preferences)
+	return FALSE
+
+/datum/preference/color/chat_color_player/create_default_value()
+	return COLOR_BLACK
```
GoldenAlpharex: If you want to make it disabled, have another pref that controls if it's enabled or not, that would make more sense than making it black by default (because people 100% won't know it's how that works)

GoldenAlpharex (changes requested):
I feel like we might want to use the random color when the person has their face hidden or something like that, so it doesn't betray people when they're actively trying to hide their identity.

desertmote:
> I feel like we might want to use the random color when the person has their face hidden or something like that, so it doesn't betray people when they're actively trying to hide their identity.

I thought about this, and it's currently already happening. Namecolor always remains the same per round, if I understand it correctly: it generates a color based on your name, and adds small randomness to it, **once**.

This PR would make it more obvious due to color-choice and that randomness being removable, but I wouldn't be against adding a 'Unknown' color. They could always display as white or something.

GoldenAlpharex:
> This PR would make it more obvious due to color-choice and that randomness being removable, but I wouldn't be against adding a 'Unknown' color. They could always display as white or something.

Yeah, I think that's probably the best way to do it. Good train of thought!

desertmote:
Really happy with these changes!

code/datums/chatmessage.dm
```diff
@@ -125,6 +125,39 @@
 		target.chat_color_darkened = colorize_string(target.name, 0.85, 0.85)
 		target.chat_color_name = target.name
 
+	// SKYRAT ADDITION - Character preference for chatmessages
+	if(ismob(target))
+		var/mob/speaker = target
+		if(speaker.client?.prefs)
+			// Apply the custom color if they want one
+			var/enable_chat_color_player = speaker.client.prefs.read_preference(/datum/preference/toggle/enable_chat_color_player)
```
GoldenAlpharex: Always put question marks for stuff after client because we can never be sure that the client is still there or if it was destroyed.

code/datums/chatmessage.dm
```diff
@@ -125,6 +125,39 @@
...
+			if(enable_chat_color_player)
+				var/chat_color_player = speaker.client.prefs.read_preference(/datum/preference/color/chat_color_player)
+				var/hex_value = sanitize_hexcolor(chat_color_player)
+				target.chat_color = hex_value
+				target.chat_color_darkened = (hex_value + "c0") // Add an alpha-channel to darken
+
+			// If they are acting as someone else, find out who
+			if((speaker.name != speaker.real_name) && (speaker.name != "Unknown"))
+				for(var/mob/living/carbon/human/copy in GLOB.player_list)
```
GoldenAlpharex: This should probably be an `as anything` to reduce the performance impact.

Also, what if they're using a name that's not theirs and there's nobody else with that name?

modular_skyrat/master_files/code/modules/client/preferences/runechat.dm
```diff
@@ -0,0 +1,29 @@
...
+/datum/preference/color/chat_color_player/is_accessible(datum/preferences/preferences)
+	if (!..(preferences))
+		return FALSE
+
+	var/allowed = preferences.read_preference(/datum/preference/toggle/enable_chat_color_player)
+	return allowed
+
+/datum/preference/color/chat_color_player/is_valid(value)
+	var/hex_value = sanitize_hexcolor(value)
+	if((hex_value < "#252525") || (hex_value > "#e0e0e0")) // These colors are too dark or too bright
```
GoldenAlpharex, line 24: Are you sure that this works the way you think? I'm pretty sure that "#2503ff" would fail that check, even if it's probably actually just fine for what we want to be used. I'm not sure how string comparisons work in DM but this feels wrong, you should probably compare the luminosity instead.

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/features/character_preferences/skyrat/runechat.tsx
```diff
@@ -0,0 +1,11 @@
+import { FeatureColorInput, FeatureToggle, Feature, CheckboxInput } from "../../base";
+
+export const enable_chat_color_player: FeatureToggle = {
+  name: "Runechat color",
```
GoldenAlpharex: I'd rename these both to `Custom Runechat Color`

GoldenAlpharex (changes requested):
We'll also probably have to test-merge this first because I'm slightly worried about performances.

code/datums/chatmessage.dm
desertmote: I've thought about that, but not tested it. Like when you pick up a clown ID from a corpse on lavaland.
What I assume happens from reading the code is it doesn't continue to an if condition after the for loop, and it carries on, keeping the name-generated color from the original block above.
Will make sure this is the case, though.

tgui/packages/tgui/interfaces/PreferencesMenu/preferences/features/character_preferences/skyrat/runechat.tsx
desertmote: It depends, the character pref lists are sorted alphabetically, it's currently at the bottom (which I like), and adding Custom will bump it to the top.

desertmote:
> Are you sure that this works the way you think? I'm pretty sure that "#2503ff" would fail that check, even if it's probably actually just fine for what we want to be used. I'm not sure how string comparisons work in DM but this feels wrong, you should probably compare the luminosity instead.

I know the color clamp thing looks bad, and it probably is- but I'm a little stupid...
I've messed around with what colors I can get a bunch, just clicking around wildly and it seems there are some shades you can't pick that aren't particularly bad.
But you can make a teeny adjustment and it will accept it, while any shade that we really don't want is 100% blocked out.

> We'll also probably have to test-merge this first because I'm slightly worried about performances.

Also in regards to performance, I tried to make the list it searches through smaller, but if I don't let it read through our observers as well, it would fail to display the color of the person you're faking as if they are out of their corpse.
I can also confirm if you have a name that is not your own, and it's neither 'unknown' or another player's, it uses your pref.
**edit:** At least it only does the playerlist search if your name is not your true_name, so that shouldn't be too often.

theOOZ:
when or if this gets testmerged, please make sure #11015 is merged first. I'll solve the conflict here, I think that fix is more important than this feature.

tf-4:
Ask if they'll take this upstream first please.

desertmote:
I've modularized the PR and solved performance concern. It now only generates or checks playerlists when it sees you require updating.

modular_skyrat/master_files/code/datums/chatmessage.dm
```diff
@@ -0,0 +1,55 @@
...
+					var/hex_value = sanitize_hexcolor(chat_color_player)
+					target.chat_color = hex_value
+					target.chat_color_darkened = (hex_value + "c0") // Add an alpha-channel to darken
+					target.chat_color_name = actor.name
+
+
+				// If the actor is copying someone else, find out who
+				if(actor.name != actor.real_name && actor.name != "Unknown")
+					for(var/mob/living/carbon/human/copy as anything in GLOB.player_list)
+						if(actor.name == copy.real_name)
```
ChungusGamer666, line 29: This is quite possibly the most expensive and asinine way to do this.
Is it really necessary to loop through a list of possibly 80 or more players?
An associative list of real name to chat color is most likely a better approach.

desertmote: It doesn't loop, and it also only calls once per name swap. I think your tone is very uncalled for. What is the difference between 'an associative list' and the list of names im searching through with this global call?

desertmote: I honestly can't tell if you are just being snarky, or are actually offering a real alternative for me to find a player in the player list. Because yes, it really is necessary to search a list of possible 80 or more players, how else can I find out their chat color preference? And how big of a performance hit really is it to do that once in a while, its not like players are changing names highly frequently.

ChungusGamer666: I'm not trying to be mean, i'm just saying that an associative list is more performant.
As in, a list indexing real name to chat color. This way the preference only has to be read once, and in case someone takes the name, you can easily take the color from the list. I don't know how else to elaborate.

desertmote:
I won't get to it in time before closing, but will continue this project regardless.

tf-4:
> I won't get to it in time before closing, but will continue this project regardless.

You should close it now then, if the bot closes the PR it'll block you from re-opening it yourself.

PR #10947, opened 2022-01-23
[SEMI-MODULAR] Fixes random colored skirts (assistant skirts) and makes species (teshari) clothing support GAGS outfits

Tastyfish:
## About The Pull Request

The Skyrat branch made changes to how assistants get their jumpsuits, using the normal outfits system, but there's no `/obj/item/clothing/under/color/random/skirt` so skirt wearers just weren't getting skirts as assistants anymore. In fact, I found that adding the skirt variant forced me to make 3 non-modular changes to prevent SSwardrobe from attempting to cache it (causes runtimes on random outfits), so instead I changed `/obj/item/clothing/under/color/random` to give you a skirt instead if placed on a skirt wearer on spawn.

I realized teshari GAGS-utilizing clothing (all colored jumpsuits, and prison jumpsuits) appear to be a completely random but frozen color because our species clothing system was just caching the first GAGS outfit spawned and using that for all of them regardless of color (once for suits, once for skirts).
It also didn't use the spoon-fed colors that GAGS already provides, so I updated the species clothing system to just defer to `greyscale_colors` if GAGS was meant to be used on the item, and also cache by the colors if used, since just icon and icon state are meaningless in that case.

### Here's a teshari rainbow + teshari prisoner
image

Unscuffs some of the main issues with our clothing system.

## Changelog

:cl: tastyfish
fix: Skirt-wearing assistants now actually get skirts.
fix: Teshari clothing colors are now more correct for a large portion of the generated clothing.
/:cl:

### DIFF
--- a/code/modules/clothing/under/color.dm
+++ b/code/modules/clothing/under/color.dm
@@ -37,7 +37,7 @@
 
 /obj/item/clothing/under/color/random/Initialize(mapload)
 	..()
-	var/obj/item/clothing/under/color/C = get_random_jumpsuit()
+	var/obj/item/clothing/under/color/C = get_random_variant() // SKYRAT EDIT CHANGE - MODULAR_ITEMS - use local proc that handles prefs
 	if(ishuman(loc))
 		var/mob/living/carbon/human/H = loc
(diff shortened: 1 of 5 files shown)


GoldenAlpharex (approved):
Looks like a good change to me! Maybe eventually we can look at the fact all jumpsuits on females have usually exactly the same color (and it's not the same as the one of the item itself), but this is good for now.

PR #10937, opened 2022-01-23
[MIRROR] Remaps Tram gulag dock to make it function correctly [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/64325
--------------------

## About The Pull Request
Rearranges Tramstation's gulag dock in the brig so it functions properly. The released prisoner can now use the airlock to leave the shuttle, is released into the side with the reclaim machine, and can exit on their own as far as the main brig area.

image

## Why It's Good For The Game

Fixes #64295

Makes tram's gulag functional and not trap players in the gulag shuttle with no way out.

## Changelog

:cl:
fix: Tramstation's gulag processing room has been rearranged so prisoners can disembark on their own instead of being trapped.
/:cl:

PR #10962, opened 2022-01-24
[MODULAR]Adds #10937 and Removes the Firelocks from Tram SM

MassacreDemigirl:
## About The Pull Request
As stated, adds the #10937 PR to the skyrat variant and removes the unneeded firelocks from SM chamber

Engineers are no longer crashed by firelocks so OSHA is no longer violated.

## Changelog

:cl:
fix: Tramstation's gulag processing room has been rearranged so prisoners can disembark on their own instead of being trapped.
fix: Removed the firelocks in the TramStation SM chamber
/:cl:


GoldenAlpharex (approved):
We really need to split Tram into two files at long last.

PR #10968, opened 2022-01-24
fix antag borer purchase

jjpark-kb:
## About The Pull Request
fixes the borer egg purchase for traitors

I accidentally put the egg, rather than the spawner thing itself.

## Changelog

:cl:
fix: you can purchase borer eggs that are actually fertilized this time
/:cl:


GoldenAlpharex (approved):
I didn't even realize we had that available for traitors. I really missed a lot when I wasn't paying attention in the past like two weeks, huh?

PR #10986, opened 2022-01-25
[NON-MODULAR] Gets rid of inaccurate info from human infocard

Zonespace27:
## About The Pull Request
This PR removes the Asimov Superiority and Chain of Command good features from the human TGUI infocard.

- While the Asimov one is _technically_ true, we don't use Asimov at the time of writing, and can give people the wrong impression.
- Skyrat's disabled the setting that requires command staff to be human.

## Changelog

:cl:
spellcheck: Corrected human info in their race infocard
/:cl:


Gandalf2k15:
gross tgui removals.

GoldenAlpharex (changes requested):
We probably at least put a "neutral" that says that as a human, you're the baseline, and as such,while you don't really have any real inherent advantage, you also have no additional flaw to worry about. Leaving it empty kind of sucks.

Gandalf2k15:
Because this is another non-modular TGUI change, I'm gonna have to say no.

PR #11016, opened 2022-01-26
Fixes protected roles getting antagonist

Gandalf2k15:
## About The Pull Request

Because we can't be trusted to keep dynamic.json up to date with roles, we're just going to do it the easy way.

Jobs now have an antagonist_restricted flag which simply prevents the role from rolling a set list of antagonists. If no list is set, then it will assume all antags.

closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/11000

## Changelog

Not player facing.


code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
```diff
@@ -137,30 +118,10 @@
...
-		JOB_SCIENCE_GUARD,
-		JOB_VANGUARD_OPERATIVE,
-		JOB_PRISONER,
-		// SKYRAT EDIT END
 	)
 	required_candidates = 2
-	weight = 4
-	cost = 15
+	weight = 2
+	cost = 12
```
GoldenAlpharex, line 124: Why's the weight and cost being changed here?

GoldenAlpharex (changes requested):
Also, if antag_flag is actually a bitflag, you should use that to make a `restricted_antag_flags` var instead of using a list, as that would be both faster, cleaner and easier to read.

code/game/gamemodes/dynamic/dynamic_rulesets_roundstart.dm
Gandalf2k15: That file was reset to TG.

GoldenAlpharex (approved):
Looks good, I'll test-merge it first to be sure we don't end up with everyone unable to be antag.

code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
```diff
@@ -16,6 +16,22 @@
...
+		// SKYRAT EDIT ADDITION - PROTECTED JOBS
+		else if(P.mind.assigned_role.antagonist_restricted)
+			if(P.mind.assigned_role.restricted_antagonists)
+				if(antag_flag in P.mind.assigned_role.restricted_antagonists)
+					candidates.Remove(P)
+					continue
+			else //Assume it's all antagonist roles.
+				candidates.Remove(P)
+				continue
+		else if(!P.client.prefs?.read_preference(/datum/preference/toggle/be_antag))
```
tf-4:
```suggestion
		else if(P.client?.prefs && !P.client.prefs.read_preference(/datum/preference/toggle/be_antag))
```

code/game/gamemodes/dynamic/dynamic_rulesets_latejoin.dm
```diff
@@ -16,6 +16,22 @@
...
+			candidates.Remove(P)
+			continue
+		else if(is_banned_from(P.client.ckey, BAN_ANTAGONIST))
```
tf-4:
```suggestion
		else if(is_banned_from(P.client?.ckey, BAN_ANTAGONIST))
```

Gandalf2k15:
tests show this works well

PR #11018, opened 2022-01-26
[MODULAR] GAGS-ifies the Screwdriver and Wirecutter to Skyrat Textures

MassacreDemigirl:
## About The Pull Request
When GAGS came, someone forgot to update the screwdriver and the wirecutter. By adding 4 lines of code it makes those tools use our icons instead of TG ones.

It makes the screwdriver and wirecutter actually look the way its supposed to look in vendors
image
screw
cutter

## Changelog

:cl:
fix: Makes the screwdriver and wirecutter use Skyrat Icons instead of TG ones.
/:cl:


GoldenAlpharex (changes requested):
We purposefully didn't change those back when they were resprited upstream because we liked the change. Those sprites look less clear and in general not as good as the ones upstream.

I'd rather you just remove the greyscaled sprites for the wirecutters and the screwdrivers that we have, I'm much more content with what we have now and I don't really want to go back to those sprites.

PR #11023, opened 2022-01-26
people will now cause dirt buildup

jjpark-kb:
## About The Pull Request
over time, and over the time of people walking over tiles, they will build up dirt.

more dirt means more things for the janitor to do.

## Changelog

:cl:
add: people walking will now cause a dirt buildup
/:cl:


Elli-Skala:
I think this should be test merged first due to the large amount of players we have to under stand how noticeable of a change this will be

modular_skyrat/modules/dirtening/code/dirtening.dm
```diff
@@ -0,0 +1,48 @@
+/datum/component/dirtening
+	///the probability that the component will do its dirty work
+	var/dirtening_chance = 10
```
GoldenAlpharex: That chance is incredibly high, I think even 1% might be a lot.

modular_skyrat/modules/dirtening/code/dirtening.dm
```diff
@@ -0,0 +1,48 @@
...
+		return COMPONENT_INCOMPATIBLE
+	linked_human = parent
+
+	if(chance)
+		dirtening_chance = chance
+
+	if(amount)
+		dirtening_amount = amount
+
+	RegisterSignal(linked_human, COMSIG_MOVABLE_MOVED, .proc/try_dirtening)
```
GoldenAlpharex, line 23: So we add all of this logic to every movement of every single human, no questions asked?

GoldenAlpharex (changes requested):
This shouldn't apply to every single human in existence. It should be tied to having dirty shoes, either from blood or from dirt, otherwise there's going to be dirt literally everywhere with no way of preventing it. It would also become a possible performance sink otherwise, due to all the checks being put in place.
Plus, the janitor would also be leaving dirt wherever they go, and I'm not sure if that also wouldn't trigger on mobs that are inside or mounted on vehicles, or riding others.

thestubborn:
yeah, at the very least add an exception to the janny boots

alexv-ds:
On the one hand, it is quite logical that dirt and dust accumulate in people, but the point is that dust accumulates on the contrary in cases where people do not walk. Still, we do not come from the street with dirty shoes, but walk around the station as a whole. The miners don't count, they just walk along the street.

He also likes to scatter things, papers, etc., so that the work will be without artificial dirt.

Here it is necessary to ask the people who play on the janitors.

modular_skyrat/modules/dirtening/code/dirtening.dm
```diff
@@ -0,0 +1,48 @@
...
+	if(isspaceturf(human_turf) || isclosedturf(human_turf))
+		return
+	//if we aren't lucky
+	if(!prob(dirtening_chance))
+		return
+	//try to find the dirt on the current turf
+	var/obj/effect/decal/cleanable/dirt/chosen_dirt = locate() in human_turf
+	//if there is no dirt, add it and the dirty amount
+	if(!chosen_dirt)
+		chosen_dirt = new(human_turf)
```
GoldenAlpharex, line 39: I'm not sure this does what you want it to do, I feel like this will just create more of the turf the person was on.

GoldenAlpharex (changes requested):

modular_skyrat/modules/dirtening/code/dirtening.dm
GoldenAlpharex: My bad I'm being an idiot.

jjpark-kb: resolved.

modular_skyrat/modules/dirtening/code/dirtening.dm
jjpark-kb: This is a low chance-- it would be best to see how to plays out in game.

modular_skyrat/modules/dirtening/code/dirtening.dm
jjpark-kb: lots of things work on this.

KathrinBailey:
> This shouldn't apply to every single human in existence. It should be tied to having dirty shoes, either from blood or from dirt, otherwise there's going to be dirt literally everywhere with no way of preventing it. It would also become a possible performance sink otherwise, due to all the checks being put in place. Plus, the janitor would also be leaving dirt wherever they go, and I'm not sure if that also wouldn't trigger on mobs that are inside or mounted on vehicles, or riding others.

It shouldn't be being called every time a human has moved in the first place.
We're looking at upwards of 100 calls a second.

Go put a var on shoes that upon initialise, grabs world time. Every time wash atom is called on shoes, grab current world time and replace since you're washing the shoes. Now have a timer that's running that every so often, look in a selection of **station** areas and look at client-controlled mobs in that area. If they have shoes, see if current world time is `30 MINUTES` or so greater. If it is, then add a component to the shoes that has X charges of dirt, now when the mob with the shoes calls moved, roll your chance there to generate dirt until either the dirt charges have ran out or the shoes have been washed, both should remove the component.

This is an incredibly backwards way of doing it. There are no ways to prevent it. Some naked person fresh out the shower can currently magic up dirt. I'd say it's too much to be calling a proc once a second for the benefit of **only up to 4 people, and to the detriment of the rest of the server*. This isn't even that, we're looking at around 50 calls during lowpop a second and 100 at highpop.* You can have a similar effect, but actually preventable, at the performance cost of **a single timer**, plus a few proc calls on moved when a valid target for dirtying is discovered. You also have the benefit of giving people a reason to shower.

desertmote:
I'm of the opinion the rate at which dirt occurs can still go up, by 2x even.

PR #11083, opened 2022-01-28
lowers blob weight, increases time between events

jjpark-kb:
## About The Pull Request
blob is being pulled out of the magicians hat far too frequently, almost once a round. We still want blob for that added chaos, but lord-- not every round; every round means it loses its touch.

This PR lowers the blob's weight and increases its cost, so that it gets pulled less.
This PR increases the time between each event, so that people have some cooldown... ( I know you admins are injecting events, I can see the roundend report!)

Less blob means that it's special when it does roll, rather than every single round (it's becoming a chore at this point when it rolls every round). 

Less frequent events means more downtime for people to RP and be left to their own devices...

## Changelog

:cl:
balance: Blob will appear less frequently
balance: There is now more wait time between each event
/:cl:

### DIFF
--- a/code/controllers/subsystem/events.dm
+++ b/code/controllers/subsystem/events.dm
@@ -8,8 +8,8 @@ SUBSYSTEM_DEF(events)
 	var/list/currentrun = list()
 
 	var/scheduled = 0 //The next world.time that a naturally occuring random event can be selected.
-	var/frequency_lower = 1800 //3 minutes lower bound.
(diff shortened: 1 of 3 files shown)


Gandalf2k15:
Do this in the events override file

Gandalf2k15:
Also don't change the Dynamic config, that's done serverside

lessthnthree:
With a cost of 20 it will virtually never roll given the amount of midround budget there is to spend. 15 seems more appropriate if any.

PR #11087, opened 2022-01-29
[NON-MODULAR] Adjusts Plasma Fixation Balance

Mega-Drones:
## About The Pull Request

Allows for use of stable plasma as a weaker healing alternative to pure plasma for the Plasma Fixation Virus Symptom

Ethereals and other species whose body temperatures are above normal, make pure plasma boil upon ingestion or injection- this causes the effected individual to "fart" plasma gas and toxify aswell as create a fire hazard in the room they are in. 
Adding Stable plasma as an alternative allows for the same (albeit weaker) healing effect without giving atmospherics a headache

## Changelog

:cl: MegaDrone

balance: allows for stable plasma to also heal when in the blood stream if you have the plasma fixation virus symptom; Stable plasma heals slower than pure plasma

/:cl:

### DIFF
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -492,8 +492,8 @@
 			. += power * 0.5
 	if(M.reagents.has_reagent(/datum/reagent/toxin/plasma, needs_metabolizing = TRUE))
 		. += power * 0.75
-	if(M.reagents.has_reagent(/datum/reagent/stable_plasma))
-		. += power * 0.5
+	if(M.reagents.has_reagent(/datum/reagent/stable_plasma)) //SKYRAT ADDITION
+		. += power * 0.5 //SKYRAT ADDITION
 
 /datum/symptom/heal/plasma/Heal(mob/living/carbon/M, datum/disease/advance/A, actual_power)
 	var/heal_amt = 4 * actual_power
--- a/code/datums/diseases/advance/symptoms/heal.dm
+++ b/code/datums/diseases/advance/symptoms/heal.dm
@@ -492,6 +492,8 @@
 			. += power * 0.5
 	if(M.reagents.has_reagent(/datum/reagent/toxin/plasma, needs_metabolizing = TRUE))
 		. += power * 0.75
+	if(M.reagents.has_reagent(/datum/reagent/stable_plasma))
+		. += power * 0.5
 
 /datum/symptom/heal/plasma/Heal(mob/living/carbon/M, datum/disease/advance/A, actual_power)
 	var/heal_amt = 4 * actual_power


YakumoChen:
👎 
Stable plasma is not an actual derivative of plasma code wise. It's just a chem dispenser output. It's not even toxic.

This is a pointless buff to the symptom. It's intended to be most useful for plasma breathing (either in a tank or a gas leak) not ingestion, and is especially good for plasma men. 

Giving it a no-drawbacks use essentially makes it no different than the water healing one. Cutting out all the downsides of a symptom because it gives someone a job to do doesn't make sense.

Mega-Drones:
Currently, nothing is stopping the virologist from getting plasma fixation and using it as you have described- during plasma leaks, as that isn't being removed; instead its actually _better_ to use it during such an event as it would heal quicker and without you having to actively keep drinking or injecting it. 
Adding stable plasma as a weaker heal is more intended as a last resort in a tight spot for those who cannot use nocturnal regen (which can be used in combat as long as you find a dark enough room), instead having to pull out a syringe or beaker of stable plasma to slowly heal seems like a big enough drawback already, no? 
Even then you could just breathe it from a tank everywhere you go, but that felt a little _too_ powergamey, so I wanted to find a better alternative.

Think of it as just a worse version of the normal plasma heal for emergency situations that wont blow you up if you're set on fire

plasmamen don't have blood and as such cannot use this virus, they instead have a passive heal whenever they eat plasma

YakumoChen:
>  Adding stable plasma as a weaker heal is more intended as a last resort in a tight spot for those who cannot use nocturnal regen (which can be used in combat as long as you find a dark enough room), instead having to pull out a syringe or beaker of stable plasma to slowly heal seems like a big enough drawback already, no?

stable plasma doesn't have an overdose or anything, the same way as the water healing symptom, and it is not toxic, nothing is stopping you from chugging 7345526723 units of it and always be healing, in or out of combat. Unlike plasma, stable plasma is not a limited resource and is far easier to acquire.

>Even then you could just breathe it from a tank everywhere you go, but that felt a little too powergamey, so I wanted to find a better alternative.

This is absolutely the intended way to use the virus.

Mega-Drones:
> stable plasma doesn't have an overdose or anything, the same way as the water healing symptom, and it is not toxic, nothing is stopping you from chugging 7345526723 units of it and always be healing

The metabolism rate for stable plasma is quite high, a full bluespace syringe only lasts for about a minute and a half, besides thats a very niche powergaming idea, if this was a widespread issue then more people would do the same thing with the water healing symptom, as well as just the normal plasma heal.
Realistically if you somehow manage to get 3000u of stable plasma in your system, it'd be pretty easy for an admin to notice a perma-healing valid hunter, as there'd be not many other reasons to have perma-heal on that would negatively effect other's roleplay.

> Unlike plasma, stable plasma is not a limited resource and is far easier to acquire.

If you have access to stable plasma, chances are you have just as much access to normal plasma; considering that there are not many uses for it at the moment, plasma sheets tend to stack up during a round because no one needs to use them. Its not that much of a "Limited resource", quite the opposite rather, its a lot _easier_ for an assistant to get their hands on plasma sheets, rather than stable plasma without chemistry or medical access

>This is absolutely the intended way to use the virus.

Personally I don't like the idea of say- every security officer or syndicate walking around with plasma tanks as internals on an MRP server with little in-character reasoning to back it up
Even then, Virology has a rather tiny playerbase, with an even smaller amount of people using anything that isn't [Noct + Self Resp] as the go-to healing combination, a small buff like this could encourage more variety during rounds for Viro players; considering that plasma fixation is only ever used once in a blue moon

tf-4:
I see no need for this non-modular change. See if they're interested upstream and let me or another maintainer know if they deny it.

PR #11089, opened 2022-01-29
Tarkon Update 1.5: Meet the Ensign 

Zenitheevee:
## About The Pull Request

> It was a simple job... Oversee the jobs while the upper-crust take their beauty sleep... When they came back aboard, just hop into a pod and... Hope that you wake back up...
> ... Sometimes...
> ... I wish i didn't.

The bigger things this PR accomplishes for Tarkon:
Adds the Ensign, A low-level command role that is pretty much all there is _for_ command on tarkon. While in reality they're all fairly equal in survival chance, Being just another body in the pit, They are _technically_ the first foot that should come forward when it comes to decision making, though its up to everyone else to decide whether to agree with them or not. Ontop of that, They are the most rounded of the crew, having starting items to be decent at anything but shouldn't be prioritized to do stuff over other crew. They also get a fancy cape to show they're _technically_ command (but not really any more important than others)

The Scientist, Deputy, and Medic get a few changes on their spawns, The normal fingerless gloves for Security being replaced with Gorilla gloves _which does absolutely nothing for actually killing xenos_ ontop of receiving proper gun training.. aka the reload skillchip. (Because as it turns out, Some ballistic guns dont have fast tac-reload.. And i'd rather do this than dive to figure out THAT mess) The medical gets a change of their starting item, Having medicated sutures instead of normal ones. The scientist gets an access change, Now able to actually unlock controls for bots.

Medical trauma room has been changed a bit, Having an added defib, one of the anesthetic closets replaced with a blood-bag crate, And a set of goodies in the medicine closet for the medic to use on others. Also the anesthetic closet that is still there has been adjusted to be accessible by the medic

An implosion chamber/explosion compressor has been added to science because... Well... I thought there'd be a board for it that can be printed, and i already had stuff for them to refine 3 cores... Turns out uuhhhh No one did any of it because no one could. go figures.

Oh

And a handful of more lore-papers have been sat around the port, Mostly in the comms room/brig, Along with some extra encryption keys should people ask for them... Because i kinda found out that removing the ones from the spawning headsets bricks the headset

Adds another spawn to tarkon in a light leader position (that isn't actually that important), allows for medical to actually revive people because while people should prioritize getting the port cleared out, Sometimes people die before science is cleared and that sometimes means no defibs, And allows people to _Actually_ do something i ment them to be able to do in update 1.25. 

## Changelog

:cl:
expansion: The Ensign has finally woken up from tarkon's cryosleep. 
expansion: Ensign spawner placed
expansion: Core Refiner placement in science (In lieu of waiting for board to be made.)
expansion: Sleeper board in Medicine closets, 
expansion: Medic and Deputy spawns tweaked.
expansion: Researcher access tweaked.
expansion: Lore papers scattered throughout the map.
expansion: oh also replaces a few walltiles that felt odd to have.
fix: changes a few floor types to their proper subtypes, Removing a fair amount of active turfs at roundstart caused by air-diffs
fix: anesthetic closet is now accessible by the Medic's access.
/:cl:


YakumoChen:
> expansion: A fair amount of map additions to tarkon, listing them all out would possibly tank my GBP

GBP is taken from github tags, not changelogs. List the changes properly, this is not an excuse.

Zenitheevee:
_huh_
Didn't know that, Rectified and will keep that in mind.

PR #11116, opened 2022-01-29
Icebox brig Fixes plus improvement

GuiltyNeko:
## About The Pull Request
~~Icebox' brig had two APCs. Power would, as far as I could tell, draw from both unevenly, and either being off would affect the area as if it were the only one. (yes I tested it yes it works)~~ Turns out that room was the warden's office not set to the right area.

Also adds firedoors to the upper brig where the railing is, reducing issues with breaches.

Also adds shutters to the plant-grow-area-place, to keep fauna out.

It's just better, what can I say?

## Changelog

:cl:
fix: Sets the warden's office to the correct area
add: Shutters can now be found in the icebox permabrig's farming room, able to be closed to prevent fauna attacks.
fix: Icebox brig now has firelocks around the upper railings, mitigating the damage caused by breaches
/:cl:

### DIFF
--- a/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
+++ b/_maps/map_files/IceBoxStation/IceBoxStation_skyrat.dmm
@@ -600,6 +600,9 @@
 	dir = 1
 	},
 /obj/machinery/modular_computer/console/preset/cargochat/security,
+/obj/machinery/door/firedoor/border_only{
+	dir = 1
+	},
 /turf/open/floor/iron,
 /area/security/brig)
 "aiY" = (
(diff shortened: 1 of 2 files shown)


KathrinBailey (changes requested):
Change the Warden's office to use the correct area instead of removing the APC please.
The issue is the office has the wrong area, not that there's a second APC.
Otherwise yeah sure I guess?

GuiltyNeko:
I didn't even know it was the warden's office because it's not directly attached to the Armory, plus the warden's office isn't usually directly adjacent to the hall. Fixed now.

GuiltyNeko:
I have no idea how the fuck to unconflict this, so I'm going to close it.

PR #11115, opened 2022-01-29
[modular] cargo loader armor + crate  

RatFromTheJungle:
## About The Pull Request
reopens  #11081, but to the master, see there.

:cl:
add: Adds a crate for the new cargo loader modsuit, and armor to the suit itself.
/:cl:

PR #11137, opened 2022-01-30
tactical oops repair from #11115

RatFromTheJungle:
## About The Pull Request
fixes an oops

oop

## Changelog
No GBP it was a goof

### DIFF
--- a/code/modules/recycling/disposal/pipe.dm
+++ b/code/modules/recycling/disposal/pipe.dm
@@ -92,8 +92,9 @@
 				var/mob/living/living_within = objects_within
 				if(living_within.stat == DEAD)
 					continue
-				if(!HAS_TRAIT(living_within, TRAIT_TRASHMAN))
-					living_within.adjustBruteLoss(5)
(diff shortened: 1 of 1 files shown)


code/modules/recycling/disposal/pipe.dm
```diff
@@ -92,9 +92,8 @@
 				var/mob/living/living_within = objects_within
 				if(living_within.stat == DEAD)
 					continue
-				if(HAS_TRAIT(living_within, TRAIT_TRASHMAN))
-					return
-				living_within.adjustBruteLoss(5)
+				if(!HAS_TRAIT(living_within, TRAIT_TRASHMAN))
+					living_within.adjustBruteLoss(5)
```
jjpark-kb: mmm, this is another oops...

```suggestion
				if(HAS_TRAIT(living_within, TRAIT_TRASHMAN))
					continue
				living_within.adjustBruteLoss(5)
```

jjpark-kb: you just did a return, instead of a continue within the for loop.

PR #11159, opened 2022-01-31
[MODULAR] Transfer Shuttles are low priority

YakumoChen:
## About The Pull Request
The Crew Transfer shuttle has been set to a static 20 minutes regardless of alert level.
image

Scrambling your highest priority shuttle to dispatch because reason given: "crew is tired and wants to go home" makes little sense. Regardless of alert level active, the modular shuttle timer should only apply for if the shuttle is actually dispatched as an emergency measure.
If you want a priority shuttle, the crew should actually be trying to call for one. Boredom is not an emergency.

This gives antagonists and other crew events time to wrap up when the transfer vote comes up out of the blue, and preempts un-immersive behavior from occurring such as changing the alert level during a vote to force a different timer.

## Changelog

:cl:
balance: To save on frivolous expenses, Nanotrasen is no longer dispatching priority shuttles just for simple crew transfers. An admiral cited "Boredom is not an emergency." to justify the change.
/:cl:

### DIFF
--- a/modular_skyrat/modules/autotransfer/code/shuttle.dm
+++ b/modular_skyrat/modules/autotransfer/code/shuttle.dm
@@ -4,7 +4,7 @@
 /datum/controller/subsystem/shuttle/proc/autoEnd()
 	if(EMERGENCY_IDLE_OR_RECALLED)
 		SSshuttle.emergency.request(null, set_coefficient = 2, silent = TRUE)
-		priority_announce("The shift has come to an end and the shuttle called. Low-Priority shuttle dispatched. It will arrive in [emergency.timeLeft(600)] minutes.", null, ANNOUNCER_SHUTTLECALLED, "Priority")
+		priority_announce("The shift has come to an end and the shuttle called. Non-Emergency shuttle dispatched. It will arrive in [emergency.timeLeft(600)] minutes.", null, ANNOUNCER_SHUTTLECALLED, "Priority")
 		log_game("Round end vote passed. Shuttle has been auto-called.")
 		message_admins("Round end vote passed. Shuttle has been auto-called.")
 	emergency_no_recall = TRUE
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
I was a bit unsure at first then I realized you meant the crew transfer votes.

Yeah, I'm not against that.

Zenitheevee:
not _quite_ sure i'm on-key with this idea...
Like, I get that hey, If the round survives for the 4 hour minimal to get the round, they can survive another 25 minutes
But there _are_ times where the round is barely being stretched to the vote by hopes and the last breath of command... And, Unless im misunderstanding this, Doesn't seem good optics for a 20 minute shuttle on _red_.

YakumoChen:
> Doesn't seem good optics for a 20 minute shuttle on red.

If you want the shuttle faster, you should have called it. If there's an emergency that warrants a shuttle call, call it.

If you're waiting for the vote, then there's no emergency.

Vexylius:
Would be cool if there was an option to rush the shuttle if it's needed faster during that period where you wait for it after the transfer vote. 
Like if there is a blob but crew is kinda handling it, so it's red and for some reason they vote for a crew transfer and it starts the 20min timer, but then there are really bad meteors... y'know.

YakumoChen:
If a truly round ending threat develops in the last 25 minutes of the round that hadnt warranted a shuttle call until then, it could probably use all of those minutes to develop a more interesting round.

 having a desperate crew hold-out at escape sounds like a far more interesting than turbo rushing the shuttle because of a sudden doom outbreak

Or you can just harass the admins via the communication console to lower the timer (raise the priority)

Zergspower:
> 

mmm except calling the shuttle is nearly taboo on SR, also 'out of the blue' makes no sense as it's a set time every shift

YakumoChen:
Not true. Heads can call shuttle whenever they want. It's up to admins to determine if it's valid. The shuttle is called plenty... when it's needed. 

Going from having unlimited time to complete an RP to 7 minutes is jarring in many situations, and makes for unsatisfying endings because the shuttle is turbo boosted without a clear "emergency".

Gandalf2k15:
IF the station is in a red alert state, then a priority shuttle should be dispatched. If it's not a red alert state, the command can change the alert level. That's the whole point in the function. If the shuttle is called via a transfer vote just before the emergency shuttle was going to be called, that presents a 20 minute gap till they can escape.

PR #2854, opened 2021-01-23
The Nuke Disk can no longer trigger the lone ops event.

BurgerLUA:
## About The Pull Request

Prevents the nuke disk from triggering the lone ops event due to disk idling.

## Why It's Good For The Game

In order to understand why this needs removal, you must first understand why lone ops occurs. Lone Ops occurs when the nuke disk is stationary for more than 500 seconds and a dice roll occurs that increases the chance by 1 percent every 1000 seconds. This dice roll occurs every process tick, which I think is every second. It then adds a weight to nuke ops occurring in the event processor. It's a lot of pointless RNG but once ever so often, it does trigger pretty frequently if the disk is stationary because of permutations.

As per antag policy, antags and crewmembers are expected to avoid metagaming and powergaming when interacting with eachother. They are usually expected to roleplay first as well. If a crewmember shuts down an antag's gimmick because they're an antag, they're going to get a bwoink. Likewise, if an antag shuts down a crewmember's enjoyment via round removal, they're also going to get a bwoink unless they deserve something for round removal. It isn't baby-proofing fo the sake of RP but rather a gentleman's agreement saying that both sides will avoid powergaming for the sake of having more fun. If we didn't have these rules in place, antag would be permitted to mass kill everyone because they _could_ be doing stuff to stop them and robotics would be permitted to make an army of Gygaxes because antags _could_ be planning on going loud this round.

However, for some reason, lone ops is except from the policies that I described. **A ghost role that has the power to end a round entirely and shut down the existing antag's gimmicks as well is except from these policies as long as they turn on combat indicator and do everything to secure the disk and nuke the station.** Ghost roles should absolutely not have this power considering that **there is literally nothing stopping the ghost role, other than guilt, from using the information they gained as ghost to complete their objective.** There is a feature that literally informs ghosts that the nuke disk is stationary and the operative spawn is a result of that disk being stationary. No other ghost role benefits this much from extra information.

All in all, the existence of lone ops is absolutely inconsistent. Nuke ops is not in rotation, but lone ops is? Wizard isn't in rotation, but lone ops is? Mid round ninjas are disabled, but lone ops is?

## Changelog
:cl: BurgerBB
del: Disables Lone Ops from spawning due to immobile nuke disk. Admins can still spawn this antagonist if they wish.
/:cl:

PR #11173, opened 2022-01-31
temporarily disable lone ops

jjpark-kb:
## About The Pull Request
disables lone op through the nuke disk.
essentially https://github.com/Skyrat-SS13/Skyrat-tg/pull/2854

**This is temporary, until we have policy for lone ops. We will only allow this to play during high pop as well, since we are not balanced for low pop.**

**This is temporary, until we have policy for lone ops. We will only allow this to play during high pop as well, since we are not balanced for low pop.**

## Changelog

:cl:
del: lone ops will no longer receive weight from nuke disk
/:cl:

### DIFF
--- a/modular_skyrat/modules/events/code/event_overrides.dm
+++ b/modular_skyrat/modules/events/code/event_overrides.dm
@@ -83,3 +83,12 @@
  */
 /datum/round_event_control/sentient_disease
 	max_occurrences = 0
+
+/**
+ * Lone Ops
+ *
+ * Removed:
+ * Does not have policy. Will re-add if/when policy is added
+ */
+/datum/round_event_control/operative
+	max_occurrences = 0


gonenoculer5:
[round lost via nuke op]
[pr opened diabling nuke op]

skyrat wildin yet again

hugmeorelse:
@gonenoculer5 
**This is temporary, until we have policy for lone ops. We will only allow this to play during high pop as well, since we are not balanced for low pop.**

Read this. Also, the Github is not a place for those types of comments.

Yawet330:
To quote my response:

The Nuclear Disk should ALWAYS be on the move. It should never be in a situation where it stops; as it is a high priority item that can at ANY POINT END THE ROUND

You should be BANNING Captains/Acting Captains for failing to secure it.

I don't know why we're hugboxing it further.

Read: Blob, xeno infestation, needing to end the round via other ways. The disk is always required.

Furthermore, Loneops are almost always 1v12 AT THE MINIMUM should the Captain actually be carrying the disk. It is almost impossible to win due to the sheer amount of advantage the crew now gets.

Now, that being said. Weh has a good point about us not being balanced for lowpop (was true last I played), and I agree. Idk why /tg/ hasn't disabled loneops below like 60 pop upstream anyway but w/e.

Just don't keep it removed, it's a necessary mechanic, and helps admins track if captains are ERPing more than they're captaining.

Yawet330:
Also, disables both admin messages about the disk being left in place.

Bad removal.

hugmeorelse:
@Yawet330 Your comments are not pertinent save for a small remark which isn't relevant to what we're trying to do or even correct for that matter.

Yawet330:
> @Yawet330 You don't even play on the server. You've been permanently banned for several months now. Also, your comments are not pertinent.

And? My comments were more about how the removal would be bad for tracking things admin-wise; than any actual gameplay ramification.
					```
message_admins("[src] is unsecured in [ADMIN_VERBOSEJMP(new_turf)]. The weight of Lone Operative is now [loneop.weight].")
				log_game("[src] is unsecured for too long in [loc_name(new_turf)], and has increased the weight of the Lone Operative event to [loneop.weight].")
``` is now disabled, entirely. Admins no longer get warned if the disk is stationary, or teleported out of bounds.

This was done via just commenting out the disk's process; which broke a ton of stuff.

**Comment by Vizoplez** [comment]
process() doesnt manage what I thought it did, due to only half-expanding da funny code lines

My mistake there

Also this PR is mostly for retention anyway so lmao.

File: code/modules/antagonists/nukeop/equipment/nuclearbomb.dm [hunk 1]
```diff
@@ -669,7 +669,7 @@ This is here to make the tiles around the station mininuke change when it's arme
 	if(!fake)
 		SSpoints_of_interest.make_point_of_interest(src)
 		last_disk_move = world.time
-		START_PROCESSING(SSobj, src)
+		//START_PROCESSING(SSobj, src) SKYRAT EDIT: No more lone ops
```
tf-4: We do still want it to process, for the logs

Gandalf2k15:
So keep it processing and give it a minimum player count of 70, which i see yoh have done, so you can remove the changes to the other file ya

BurgerLUA:
> So keep it processing and give it a minimum player count of 70, which i see yoh have done, so you can remove the changes to the other file ya

Wait I'm confused. Do you want lone ops removed or do you only want it to trigger at more than 70 pop?

code/modules/events/operative.dm
```diff
@@ -3,6 +3,7 @@
 	typepath = /datum/round_event/ghost_role/operative
 	weight = 0 //its weight is relative to how much stationary and neglected the nuke disk is. See nuclearbomb.dm. Shouldn't be dynamic hijackable.
 	max_occurrences = 1
```
Gandalf2k15: Instead of doing these changes, move this to event_oveeride.dm and set it's max occurrences to 0

Gandalf2k15: Allow it to increase weight uselessly.

PR #10849, opened 2022-01-20
Ghost Cafe: Detainment Cell

thestubborn:
## About The Pull Request

In the continuing inter-galactic war between the renamed Overlook Hotel and the Cafe certain measures were added.
image
Detainment cells were one, so here's one!
For OFFICIAL use only. Please don't do your stuff in here.

## How This Contributes To The Skyrat Roleplay Experience
Prisoner roleplay on a budget! Even has a fake set of handcuffs and three of each jumpsuit/skirt!
The old Sec Clothing Vendor was replaced by a bar clothing vendor (because nearly all of them have a spot now)

## Changelog

:cl:
add: The Ghost Cafe's detainment room has shown up for some reason. For official use only! Too bad they have terrible locks. 
/:cl:

PR #11176, opened 2022-01-31
[MODULAR] Ghost Cafe Renovations - Detainment

kannthus:
## About The Pull Request

I noticed a few severe issues with the new detainment from https://github.com/Skyrat-SS13/Skyrat-tg/pull/10849 which this fixes, it also adds a little window/lock for the prisoners.

List Changes:

- Due to the fences being full-tiles, 2 out of the 3 tiles in the cell were impossible to walk into, this moves the fence out one and pixelshifts it into place.
CleanShot 2022-01-31 at 20 06 02 

- The "Door Lock" for the detainment actually controlled a dorm on the otherside of the cafe, this fixes it and gives it a proper tag/name
CleanShot 2022-01-31 at 20 09 22

- I also added a window w/ shutter to allow detainees a bit of comfort with a lovely view on-look to the pool. There is now also a button for the window shutters incase a prison gets a bit too rowdy or incase a guard gets a bit to rowdy _with_ the prisoner :)
CleanShot 2022-01-31 at 20 11 51

Aside from fixing some pretty substantial issues its merely an aesthetic change.

## Changelog

:cl:
fix: Ghost Cafe Detainment now works as intended
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -3859,10 +3859,6 @@
 	},
 /area/centcom/holding/cafepark)
 "aKQ" = (
-/obj/structure/fence{
-	dir = 1;
-	pixel_x = -15
-	},
 /obj/structure/toilet,
 /obj/structure/curtain{
 	alpha = 250;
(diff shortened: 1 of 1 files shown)


thestubborn:
huh
swear i keyed the button right
maybe it was when I was moving it...

PR #11196, opened 2022-02-01
Fixes the CO's trim datum so that heads can actually assign the job.

Tastyfish:
## About The Pull Request

The Corrections Officer was actually missing 2 fields that are required for normal station jobs. The main one that inspired me to fix this was that it was missing `template_access`, preventing anyone from assigning the job via the ID painter or the access management console. Now it has the normal sec job template access. This is probably the most harmless security job so I can't imagine it was intentional.

You should be able to get 'promoted' to CO.

## Changelog

:cl: tastyfish
fix: The Central Command intern finally found the request form to add the Corrections Officer to Plexagon Access Management. You can now assign the CO job if you have access to assign security jobs.
/:cl:


Tastyfish:
FYI the linter failed because it blew chunks creating the test environment before it even loaded the repo.

PR #11179, opened 2022-01-31
["""MODULAR"""] Re-criminalizes the augmented eyes, by moving them behind alien medical research.

RatFromTheJungle:
## About The Pull Request
what it says on the tin.
## How This Contributes To The Skyrat Roleplay Experience

Access to information is the most powerful tool in any game, especially so for SS13, and either eyes makes stealth or else wise subtlety pointless.

## Changelog

:cl:

balance: Moved Xray+Thermal eyes behind the alien biological research node.
/:cl:

closes #11202

PR #11202, opened 2022-02-01
[SEMI-MODULAR] Non-Trivializing Thermals and X-Ray Implants

Higgin:
## About The Pull Request

Alternative to #11179 

Moves thermal and X-ray implants to their own very-expensive techweb node behind the other combat implants and several experiments.

Makes much more severe EMP downsides to thermals/xray implants.

The readdition of thermals and x-ray implants has come under fire for being a very early in to a lot of information and shutdown ability for security against antagonists. They also don't have particular downsides after being acquired. 

The compromise solution was to move them to alien tech (something that only shows up through mining or space explorers happenstancing loot/corpses back to the station, and very rarely at that.) This is highly-rare as-is, but even assuming people do start to go after the sources for alien tech every round, this really doesn't do much for people on the station. It doesn't add any good reason to get out and move around with other departments. It could still practically be done by one person without much coordination across the science department. For one dedicated individual without much concern about the opportunity cost, you could still get to a point of being able to beeline thermals/x-rays pretty well. 

This PR instead vastly raises the cost of thermals and X-rays in their own tech node; puts them behind several experiments requiring that science be doing toxins and performing experiments on machines around the station; and adds significant extra downsides to getting hit with EMP after getting them installed. It should encourage more interaction from science, give other people in the round more of a reason to allow science to move around and do things, push the implant acquisition later into the round. It readds an alibi to doing toxins since most of the MCR stuff was switched to discounts. It should also make the implants themselves more of an interesting choice and less of a flat tradeoff for people to try to get given that they're exposing themselves to a lot of hurt from EMP. It should also make it much more obvious how and when science is getting to the point that implants might be unlocked (other research stops, or continues - this also makes it much easier to slow down the acquisition by, say, blowing up the R&D server.) 

I'm not entirely happy with all the tradeoffs being loaded on EMPs, and I do think there's an argument for using alien tech for more things as a reason to get people out exploring more (also an alibi for going into space for antagonism.) However, I think I'm happier with this solution (one that requires a lot more coordination in the research team, keeps the implants, and introduces some stiffer mechanical drawbacks to using them) than I am with the PR now up to get rid of them again. Being able to pursue more information in-game is a good thing. Letting people progress and get meaningful benefits over the course of a round is a good thing. 

There are better ways to do that, and I think this is one.

## Changelog

:cl:
balance: Moves thermals/x-ray implants to their own much more expensive technode requiring several experiments to be completed. 
balance: Makes the penalties for getting EMP'd while having thermals/X-ray implants much more severe.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/eyes.dm
+++ b/code/modules/surgery/organs/eyes.dm
@@ -192,6 +192,24 @@
 	see_in_dark = 8
 	sight_flags = SEE_MOBS | SEE_OBJS | SEE_TURFS
 
+/obj/item/organ/eyes/robotic/xray/emp_act(severity) //SKYRAT EDIT START:X-RAY EYE NERF
+	. = ..()
+	if(. & EMP_PROTECT_SELF)
+		return
+	if(owner && (damage < maxHealth) && prob(150/severity))
+		applyOrganDamage(30/severity)
+		owner.flash_act(intensity = 2/severity, override_blindness_check = 1)
+		owner.adjust_blurriness(30)
+		owner.visible_message(span_warning("[owner]'s eyes flash with a sickly blue light!"))
+		to_chat(owner, span_warning("Your vision swims with frantic, flashing images of the inside of your head!"))
+		owner.add_confusion(min(owner.get_confusion() + 10, 30))
+		owner.adjustOrganLoss(ORGAN_SLOT_BRAIN, 30)
+		radiation_pulse(owner, max_range = 0, threshold = RAD_FULL_INSULATION, chance = 100)
+		do_sparks(2, TRUE, owner)
+		owner.emote("scream")
+		if(severity == 1)
+			owner.Knockdown(1 SECONDS)//SKYRAT EDIT END: X-RAY EYE NERF
+
 /obj/item/organ/eyes/robotic/thermals
 	name = "thermal eyes"
 	desc = "These cybernetic eye implants will give you thermal vision. Vertical slit pupil included."
@@ -200,6 +218,27 @@
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE
 	flash_protect = FLASH_PROTECTION_SENSITIVE
 	see_in_dark = 8
+
+/obj/item/organ/eyes/robotic/thermals/emp_act(severity) //SKYRAT EDIT START: THERMAL EYE NERF
+	. = ..()
+	if(. & EMP_PROTECT_SELF)
+		return
+	if(owner && (damage < maxHealth) && prob(150/severity))
+		applyOrganDamage(30/severity)
+		owner.flash_act(intensity = 2/severity, override_blindness_check = 1)
+		owner.adjust_blurriness(30)
+		owner.visible_message(span_warning("[owner]'s eyes hiss smoke and burst into flame!"))
+		to_chat(owner, span_warning("Your eyes burn with a scorching, white heat!"))
+		owner.add_confusion(min(owner.get_confusion() + 10, 30))
+		owner.adjustOrganLoss(ORGAN_SLOT_BRAIN, 30)
+		var/obj/item/bodypart/affecting = owner.get_bodypart(BODY_ZONE_HEAD)
+		owner.apply_damage(25, BURN, affecting)
+		owner.adjust_fire_stacks(10/severity)
+		owner.IgniteMob()
+		do_sparks(2, TRUE, owner)
+		owner.emote("scream")
+		if(severity == 1)
+			owner.Knockdown(1 SECONDS)//SKYRAT EDIT END: THERMAL EYE NERF
 
 /obj/item/organ/eyes/robotic/flashlight
 	name = "flashlight eyes"
--- a/modular_skyrat/modules/implants/code/medical_nodes.dm
+++ b/modular_skyrat/modules/implants/code/medical_nodes.dm
@@ -6,6 +6,18 @@
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 7500)
 
 /datum/techweb_node/combat_cyber_implants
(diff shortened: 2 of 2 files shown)


HushTD:
I understand the sentiment behind the PR, but I do not fully agree with the execution
As someone who mainly plays Research Director, or Captain strongly interacting with science, this looks a little odd and out there.

I believe 63000 points will make it a very expensive node, and it is not that round of a number.
There are no other nodes requiring three different experiments to be completed, and these experiments do not look super related to the research.
Additionally, I remember there are OOC rules about combat implants, and I see them very rarely on my shifts, so I have it seen not be an issue myself

I suggest possibly as such lowering the point cost to maybe a multiplication of 10 000, say 40 000 and keeping only the Mother of God experiment required, and removing the parts from required scans and research

Science already usually rushes mining tech, then parts, then advanced mining, and really only does the weapon and implant stuff in later stages from what I have seen.

Higgin:
> I believe 63000 points will make it a very expensive node, and it is not that round of a number. There are no other nodes requiring three different experiments to be completed, and these experiments do not look super related to the research. Additionally, I remember there are OOC rules about combat implants, and I see them very rarely on my shifts, so I have it seen not be an issue myself

63k is ~20 minutes of passive research income at 52.3 points per second. So far, I agree that they aren't and shouldn't end up being a problem; but others are concerned about them coming too early or being too powerful. I'd rather they stay than go.

> I suggest possibly as such lowering the point cost to maybe a multiplication of 10 000, say 40 000 and keeping only the Mother of God experiment required, and removing the parts from required scans and research

I could see it just being MoG (like the last combat/sec researches, MCRs in particular, were for a while,) but one thing about the other experiments is that they're generally under-used, and this helps encourage people to go out and do upgrades if for some reason they haven't already. It also gives a cover to scientists wanting to enter departments and work on their stuff. If all of science is working towards implants, not researching other things for the time it takes to get the points, adding the required experiments here helps ensure they've done things for the rest of the station in the process. 

> Science already usually rushes mining tech, then parts, then advanced mining, and really only does the weapon and implant stuff in later stages from what I have seen.

agreed, this has also been my experience, but the concern people express is that implants will become a 'rush' once players wake up to their power. this is an attempt to put up a harder wall against that happening and push implants later into the round (or force sci to really hustle/coordinate for them.)

RatFromTheJungle:
EMPs being stronger isn't a true nerf against them, because it doesn't defeat the 'point of no return' issue they cause, where you are somewhere doing something illegal, and sec sees you through the wall red-handed, while you have no indicator you were noticed unless you have thermals yourself; they PDA, or radio for back up, while being able to maintain exact vision for what you are currently doing.

Instead, it's more of a token nerf then anything, because it doesn't stop them from doing the whole thing that makes them an issue, and relies on the antags instead, to prep for them in specific, well in advance, and moreso, if you are wearing thermals(the prepping), you'd have to see them, walk over to the wall they're by (assuming they aren't 3-4 tiles away at which point this isn't even possible), and EMP them through it.

Suspect-One:
Why people even need X-Ray and Thermals? They are insanely overpowered and shouldn't be allowed for crew at all, RP-wise (why someone would allow assistant or even paramedic to run around with X-Ray, it's not a toy, it's not a tool), rule-wise (why would you ask for it in the first place if you are not a combat role) and everything wise (why are you giving extra-spicy antagonist tools to crewmembers and why antagonists must suffer if they are already nerfed in the ground?).

Leave them behind alien node, this makes it an actual reward for VOs, Xenoarch or Miners doing their work. Alien node itself isn't that rewarding, surgery tools, engineer tools are not that great for going out of you way to acquire them. Moving such OP implants to Alien node makes it really viable and useful for crew. Besides, I don't understand why do you say that X-Ray and Thermals should be available every round. No, they shouldn't be, they should be on the same level as x-ray and thermal genetics mutations - either you are lucky enough to find them or you don't find them at all. 

Either you are robust, lucky or smart enough to find things to research alien node or you don't get X-Ray and Thermal eyes at all.

jughu:
> Why people even need X-Ray and Thermals? They are insanely overpowered and shouldn't be allowed for crew at all, RP-wise (why someone would allow assistant or even paramedic to run around with X-Ray, it's not a toy, it's not a tool), rule-wise (why would you ask for it in the first place if you are not a combat role) and everything wise (why are you giving extra-spicy antagonist tools to crewmembers and why antagonists must suffer if they are already nerfed in the ground?).
> 
> Leave them behind alien node, this makes it an actual reward for VOs, Xenoarch or Miners doing their work. Alien node itself isn't that rewarding, surgery tools, engineer tools are not that great for going out of you way to acquire them. Moving such OP implants to Alien node makes it really viable and useful for crew. Besides, I don't understand why do you say that X-Ray and Thermals should be available every round. No, they shouldn't be, they should be on the same level as x-ray and thermal genetics mutations - either you are lucky enough to find them or you don't find them at all.
> 
> Either you are robust, lucky or smart enough to find things to research alien node or you don't get X-Ray and Thermal eyes at all.

Im afraid if you put it in alien node it will be even unlocked earlier in some circumstances

Techwebs can unlock spefific things by deconstructing spefific things, you could make thermals being unlocked by deconstructing traitor thermals.or anything similar to that degree, and making xray require expiriment science with genetics for example

KathrinBailey:
We put this to an internal maintainer vote and it came out 4 out of 5 in favour of #11179, with 1 out of 5 maintainers not voting.
The nerf doesn't address the issues we have with thermals & x-ray implants for station crew.
The damage is done as soon as the **antagonist does anything remotely suspicious on-screen**. At that point, it's already over. The roboticist, security officer, or even assistant with the implants is calling for backup, and you have 4 seccies stacked outside the room you're in. EMPs don't really matter much at that point.
This doesn't mean we're against making any further changes, just that we'd need to see a larger redesign for thermals/x-ray implants to come back.

PR #11208, opened 2022-02-01
Cleans up leftover log code and fixes a logging issue.

tf-4:
## About The Pull Request

Removes ambitions logging.
Fixes victim logs being logged as subtler emotes.

## Changelog
:cl:
fix: Victim logs are no longer mistakenly logged as subtler emotes.
/:cl:


code/__DEFINES/~skyrat_defines/logging.dm
```diff
@@ -1,12 +1,11 @@
 // Logging types for log_message()
-#define LOG_SUBTLER (1 << 20)
-#define LOG_AMBITION (1 << 21)
+#define LOG_SUBTLER (1 << 22)
```
GoldenAlpharex, line 2: Why 22 and not 20?

GoldenAlpharex (changes requested):

code/__DEFINES/~skyrat_defines/logging.dm
tf-4: 20 is victim logs which is what caused the issues with attacks being logged as subtler emotes, and 21 is radio emotes

GoldenAlpharex (approved):

PR #11220, opened 2022-02-02
MCR's now properly show up in the right hand when held in the right hand

Zenitheevee:
## About The Pull Request

Read title.

Basically, LSS, When held in either right or left hand, the MCR would show up in the left hand on the sprite. This didn't really matter much but it could be a clarity issue if someone is holding _two_ MCR's, double shooting them, But the sprite only shows one in the left hand. I saw there were sprites for the right handed, and just that the right hand sprite var use the left hand sprite file.

image

Dont you think its odd that a gun held in the right hand shows in the left?

## Changelog

:cl:
fix: MCR properly shows up in the right hand when held in the right hand.
/:cl:


tf-4 (approved):

GoldenAlpharex (approved):
Good catch hahaha

PR #11221, opened 2022-02-02
malf can no longer doomsday

jjpark-kb:
## About The Pull Request
removes the doomsday ability, objective, and the doomsday device from malf ai. 

We will see about giving them a BSA instead.

why do we want complete station-ending objectives...? 

## Changelog

:cl:
del: removed the doomsday ability and objective from malf ai
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -997,7 +997,7 @@
 	lamp_doom = FALSE
 	if(connected_ai)
 		connected_ai.connected_robots |= src
-		lamp_doom = connected_ai.doomsday_device ? TRUE : FALSE
+		//lamp_doom = connected_ai.doomsday_device ? TRUE : FALSE SKYRAT EDIT: no more doomsday
 	toggle_headlamp(FALSE, TRUE)
 
 /**
--- a/code/game/objects/effects/countdown.dm
+++ b/code/game/objects/effects/countdown.dm
@@ -109,7 +109,7 @@
 	else if(T.cooldown)
 		var/seconds_left = max(0, (T.cooldown_timer - world.time) / 10)
 		return "[round(seconds_left)]"
-
+/* SKYRAT EDIT: removed doomsday
 /obj/effect/countdown/doomsday
 	name = "doomsday countdown"
 
@@ -119,7 +119,7 @@
 		return
 	else if(DD.timing)
 		return "<div align='center' valign='middle' style='position:relative; top:0px; left:0px'>[DD.seconds_remaining()]</div>"
-
+SKYRAT EDIT: removed doomsday */
 /obj/effect/countdown/anomaly
 	name = "anomaly countdown"
 
--- a/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
+++ b/code/modules/antagonists/traitor/equipment/Malf_Modules.dm
@@ -171,7 +171,7 @@ GLOBAL_LIST_INIT(malf_modules, subtypesof(/datum/ai_module))
 /// Modules that are improving AI abilities and assets
 /datum/ai_module/upgrade
 	category = "Upgrade Modules"
-
+/* SKYRAT EDIT: No nuking the station
 /// Doomsday Device: Starts the self-destruct timer. It can only be stopped by killing the AI completely.
 /datum/ai_module/destructive/nuke_station
 	name = "Doomsday Device"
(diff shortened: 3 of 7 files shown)


desertmote:
I think this is taking good action out of the server.

I'm fairly certain this is reviewed live in the round by the staff team, and we have a policy team in case more caution is needed.

tf-4:
This was made on Xyel's request

hugmeorelse:
> I think this is taking good action out of the server.
> 
> I'm fairly certain this is reviewed live in the round by the staff team, and we have a policy team in case more caution is needed.

This is my decision. This has very little upside and the downside is quite a bit for something that adds little to the game. If a nuke is what is needed to make the server lively, then we're doing it wrong. The server was never about this, and it will never be about this. It just goes against the philosophy we follow for the server.

Giving them a BSA instead in my respective opinion gives them more of an edge as a threat, and it lets them do a more precise but still devastating attack to put weight behind its actions.

desertmote:
Okay, I can understand that!
There's something I still worry about which is, the doomsday mode has great effects for borgs. Stuff that works incredibly well for our server thematic like the borg's lamps being forced on and turned red. (Its called doomsday mode) Losing this would be a big shame, and keeping some of this code in so stuff as such can still be manually called - without the big station ending event - would make this disable hurt our silicon content much less.

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -997,7 +997,7 @@
 	lamp_doom = FALSE
 	if(connected_ai)
 		connected_ai.connected_robots |= src
-		lamp_doom = connected_ai.doomsday_device ? TRUE : FALSE
+		//lamp_doom = connected_ai.doomsday_device ? TRUE : FALSE SKYRAT EDIT: no more doomsday
```
desertmote, line 1000: This is a really cool part about borgs that would get lost.

BurgerLUA:
Removing round-ending events is the right step in the right direction. It's unfair to a lot of antag players who can't complete their gimmick because of another antag ending the round, and it's unfair to normal players who aren't just here to secure valids.

Iamgoofball:
so what's the gameplan for malf AIs now

thestubborn:
> so what's the gameplan for malf AIs now

blow rcds
lose

Zenitheevee:
> so what's the gameplan for malf AIs now

obstruct Sec from winning

restlessObserver:
So, if malf at the end have no other choice other than kill all crew, theyll plasmaflood and decompress. Because sometimes death is right ending, sometimes good people are ment to loose and be left defeated in dust. And this ending, sometimes, much more interesting then "good always wins" situation. And play it til the end is much harder. The point - not to delete doomsday function. Its to make players roleplay this situation, not to go cryo and PM about how they erps were interrupted.

Gandalf2k15 (changes requested):
please instead of making all these non modular changes, literally just disable the AI's button for it.

hugmeorelse:
> So, if malf at the end have no other choice other than kill all crew, theyll plasmaflood and decompress. Because sometimes death is right ending, sometimes good people are ment to loose and be left defeated in dust. And this ending, sometimes, much more interesting then "good always wins" situation. And play it til the end is much harder. The point - not to delete doomsday function. Its to make players roleplay this situation, not to go cryo and PM about how they erps were interrupted.

ERP is not that common on Skyrat. Don't strawman. I'm not even thinking about ERP here.

The main reason why I don't want this around is because it seems quite dull for Skyrat. I'm sure it's great on say tg, but we have different ideas of what's fun here.

Considering Doomsday is fairly ineffective to begin with and is rarely used, it's not that interesting. It also has a high opportunity cost, as it shuts down all other forms of conflict as @BurgerLUA noted. And that's if the nuclear explosion even goes off. Most of the time, security is able to override it in a timely manner, so it's not exactly a "threat".

So, it has limited upside, i.e. it makes security somewhat wary but not really since it can be shut down reliably, and the downsides are terrible, i.e. it ruins all other ongoing conflict and antagonistic activities if it does go off.

Instead of mutually assured destruction that doesn't work out in most cases. having the AI be able to make devastating precision strikes in retaliation gives them real teeth, and they can do some damage before being overrun.

Grox103FM:
Are you really have that much engineers what you need to repair the station after malfunction AI? Round end antagonists very dangerous and destructive, and after hour of war against them you need three hours to recovery. Instead of fighting the threat you fighting consequences. And, after round end antag you have disbalansed crew with mechs and many guns. Other antags will be no match to security, and its very painful to play against OP crew.

PR #11229, opened 2022-02-02
Fix bluespace miner examine text to match actual logic

Tastyfish:
## About The Pull Request

Bluespace miner fails and boops if pressure is over 1.5 atm room pressure, but doesn't show what's wrong on its examine text until 2 atm room pressure. Very confusing. This PR fixes that.

This machine is already picky enough without its errors to the user being wrong.

## Changelog

:cl: tastyfish
fix: Fixed bluespace miner examine errors to show at the same pressure range that the logic does.
/:cl:


jjpark-kb:
that is entirely my fault, thank you for catching this.

PR #11231, opened 2022-02-02
[MIRROR] Fixes a bunch map log runtimes, adds CI to catch some more potential map issues [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/64599
--------------------

## About The Pull Request

* Fixes the CI test to detect doubled up cables, and adds one for lattices, another for identical pipes by path, and checks if any APC's have manually set their pixel_x/y to something other than +-25.
* Fixes loads of RandomZ maps that have bad APC pixel offsets. There may be ones I missed that aren't auto directional but also don't set the pixel offset either, as I didn't create a check for that.
* I also removed a doubled up cable on Lavaland Mining. Probably something else somewhere.
* Fixes the airlock note placer map helper, as it was running through multiple code paths and showing errors on success like a madman.

## Why It's Good For The Game

Cleans up loads of runtimes, realistically probably saves like 1 second a startup, but prevents the spamming of mapping runtimes so we can see all the other runtimes in their glory.

This was the entire `map_errors.log` after I fully started a game locally after doing all this (first try):
```log
[2022-02-01 08:05:40.267] There are 3 active turfs at roundstart caused by a difference of the air between the adjacent turfs. You can see its coordinates using "Mapping -> Show roundstart AT list" verb (debug verbs required).
```

## Changelog

:cl: tastyfish
server: Added some more CI linter checks for maps, that are otherwise caught at runtime. This will likely fail downstream maps if they're double stacking objects which aren't allowed to stack, or misplacing APC's.
/:cl:


Tastyfish:
I'd recommend merging #11211 first to fix the resulting linter errors.

PR #11235, opened 2022-02-02
Changes all Head of Staff airlocks on maps to be fancy versions

GuiltyNeko:
## About The Pull Request
You know those fancy airlocks heads of staff get on blueshift? Now they can have those everywhere.

Because now they can all show off their uniqueness from each other with unique colored doors? I dunno, I like the way they look and it's a waste to not use them more.

## Changelog

:cl:
qol: Head of Staff doors on Kilo, Ice box, Meta, and Delta are all now the same as on blueshift
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -3924,7 +3924,7 @@
 /area/security/prison)
 "aBU" = (
 /obj/machinery/door/firedoor,
-/obj/machinery/door/airlock/command{
+/obj/machinery/door/airlock/captain{
 	name = "Emergency Escape";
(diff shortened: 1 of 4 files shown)


YakumoChen:
Its nice but is this going to stay for any length of time past a map reset?

GuiltyNeko:
> Its nice but is this going to stay for any length of time past a map reset?

Well, each airlock is essentially a command airlock with a unique/mostly-unique access number, so it'd be REALLY easy to ctrl+f replace in the future, since all you need is the special airlock paths and the access numbers for the doors. Find a command door with that access number, set it to the relevant head of staff door. Or in the case of the QM, a mining door. Just replace "command" with "captain" or "RD" and ignore the "glass" part because all the ones that need it have a glass variant. So... it could be automated. By someone smarter.

PR #11243, opened 2022-02-03
Polychromic collars to GAGS, minor code improvements and bug fixes

tf-4:
## About The Pull Request

Back again. A bunch of items this time.

collars

Aside from the obvious:
- Fixes a bug with cancelling renaming a collar giving it a blank name
- Tidies up the path names just a little.
- You can examine a collar to see if it's locked or not.
- Documents a few variables, renames a single-letter one.
- `/obj/item/clothing/neck/human_petcollar/inferno` is now `/obj/item/clothing/neck/inferno_collar`. Some behaviour had to be copy-pasted for this, but I don't particularly want to start messing with the sprites of donator items.
- Probably something else that I forgot.

Closer to removing polychromism, better colour control.

## Changelog
:cl:
code: Collars are now coloured through GAGS rather than being polychromatic.
fix: Cancelling renaming a collar will no longer give it a blank name.
/:cl:


modular_skyrat/modules/customization/modules/clothing/neck/collars.dm
```diff
@@ -4,132 +4,133 @@
...
-/obj/item/clothing/neck/human_petcollar/ComponentInitialize()
-	. = ..()
-	if(is_polychromic)
-		AddElement(/datum/element/polychromic, poly_colors)
-
 /obj/item/clothing/neck/human_petcollar/attack_self(mob/user)
 	tagname = stripped_input(user, "Would you like to change the name on the tag?", "Name your new pet", "Spot", MAX_NAME_LEN)
-	name = "[initial(name)] - [tagname]"
+	if(tagname)
+		name = "[initial(name)] - [tagname]"
```
YakumoChen, line 38: Does this mean you're going to have to erase the "Spot" input every time if you don't want a name, or does that null by default?

tf-4: Collars start with a null name. Before, if you cancelled the input it would give you a blank name i.e. `collar - `. Now, it will keep the null name if there wasn't one, so just `collar`, or the previously input name if there was. 

I considered nulling the name on cancel but I didn't go with that in the end. 

It would also be nice to be able to set a name in the loadout rather than having to do it every round, but that's out of scope for this PR (read: I can't be bothered)

thestubborn: >It would also be nice to be able to set a name in the loadout rather than having to do it every round, but that's out of scope for this PR (read: I can't be bothered)

couldn't you just copy and paste whatever the plushies use?

PR #11268, opened 2022-02-04
remove fungal mold from mold event

jjpark-kb:
## About The Pull Request
title.

host request.

## Changelog

:cl:
del: removed fungal mold from event spawning
/:cl:

### DIFF
--- a/modular_skyrat/modules/biohazard_blob/code/mold_event.dm
+++ b/modular_skyrat/modules/biohazard_blob/code/mold_event.dm
@@ -16,12 +16,11 @@
 		/obj/structure/biohazard_blob/structure/core/toxic,
 		/obj/structure/biohazard_blob/structure/core/radioactive,
 		/obj/structure/biohazard_blob/structure/core/emp,
-		/obj/structure/biohazard_blob/structure/core/fungus
 	)
 
 /datum/round_event/mold/start()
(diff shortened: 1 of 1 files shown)


ReynTheLord:
Fungal mould has been a MASSIVE pain to deal with, alongside some other mould types. this one though is just infuriating as it just essentially gives the station a highly contagious, highly deadly, disease, not even player controlled, by merit of existing.

At least sentient disease is fun and virology diseases require players to make them.

restlessObserver:
So you removing it because its hard to deal with.
No offense but But I see the behavior of a little pussy here. Because something hard to deal with doesn't mean you need to delete it. Its literally skill issue of someone with not enough speed and knowledge.

softcerv:
> So you removing it because its hard to deal with. No offense but But I see the behavior of a little pussy here. Because something hard to deal with doesn't mean you need to delete it. Its literally skill issue of someone with not enough speed and knowledge.

Yes. if something is difficult that it causes issues for 90% of the playerbase, then it is something that should be removed or balanced. Please try and be constructive here instead of hurling insults at people, GitHub really isn't the place for this.

hugmeorelse:
If you're going to be toxic, we will just block and move on. Github is not a place for this as @softcerv said.

PR #11299, opened 2022-02-06
[MODULAR] Feign impairment ceiling increase 1 minute -> 10 hours

Zonespace27:
## About The Pull Request
Increases the amount of time feign impairment can last to up to 36000 seconds, being ten hours.

The one minute limit seems arbitary, you can cancel the impairment whenever you want and keep extending it manually forever.

## Changelog

:cl:
qol: Feign impairment now has a time limit of 10 hours, instead of 1 minute
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human.dm
@@ -122,8 +122,8 @@
 	var/duration = input(src, "Enter how long you will feign [impairment]. (1 - 60 seconds)", "Duration in seconds", 25) as num|null
 	if(!isnum(duration))
 		return
-	if(duration > 60)
-		to_chat(src, "Please choose a duration in seconds between 1 to 60.")
+	if(duration > 3600) //A full hour.
+		to_chat(src, "Please choose a duration of up to 3600 seconds.")
 		return
 	switch(impairment)
 		if("Drunkenness")
--- a/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/carbon/human/human.dm
@@ -119,7 +119,7 @@
 	if(!impairment)
 		return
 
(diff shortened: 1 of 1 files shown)


tf-4:
Could just do away with the cap, really. An hour seems just as arbitrary as a minute does.

Zonespace27:
My concern with removing it entirely is that giving people the ability to set a number to anything without a cap has the potential for bad things

tf-4:
> My concern with removing it entirely is that giving people the ability to set a number to anything without a cap has the potential for bad things

well, just put it to something like 5 hours that a round won't reasonably exceed.

Zonespace27:
Done

modular_skyrat/modules/customization/modules/mob/living/carbon/human/human.dm
```diff
@@ -122,8 +122,8 @@
 	var/duration = input(src, "Enter how long you will feign [impairment]. (1 - 60 seconds)", "Duration in seconds", 25) as num|null
```
tf-4:
```suggestion
	var/duration = input(src, "Enter how long you will feign [impairment].", "Duration in seconds", 25) as num|null
```

tf-4 (approved):

tf-4 (changes requested):
Suggested a better way on discord

tf-4 (approved):

hugmeorelse:
Is there a way to turn it on and off instead? @Zonespace27

Zonespace27:
> Is there a way to turn it on and off instead? @Zonespace27

The variables this affects aren't booleans, and I've said my piece on setting this to be infinite

hugmeorelse:
> > Is there a way to turn it on and off instead? @Zonespace27
> 
> The variables this affects aren't booleans, and I've said my piece on setting this to be infinite

But is there a way to turn on and off, i.e. changing the timer from an hour to one second to make it go away?

Zonespace27:
> > > Is there a way to turn it on and off instead? @Zonespace27
> > 
> > 
> > The variables this affects aren't booleans, and I've said my piece on setting this to be infinite
> 
> But is there a way to turn on and off, i.e. changing the timer from an hour to one second to make it go away?

Yes, this was a function beforehand to let you turn it off.

PR #11295, opened 2022-02-06
reagent forging 1.2

jjpark-kb:
## About The Pull Request
reagent forging 1.0 = https://github.com/Skyrat-SS13/Skyrat-tg/pull/6614
reagent forging 1.1 = https://github.com/Skyrat-SS13/Skyrat-tg/pull/9168
Some larger additions, with more planned in the future.

please view the changelog to see the changes...

## How This Contributes To The Skyrat Roleplay Experience
give more variety to our blacksmiths, and make the gameplay more enjoyable with numbers going up.

## Changelog

:cl:
add: added a smithing skill
add: added a smithing skill cloak
add: added skill modifiers to forges and anvils for smithing
add: added a katana, dagger, pavise shield, pickaxe, and shovel reagent "weapons"
add: added the ability to make crank cells (coil on a cell)
add: added a forge crafting station; use this to fully finish the items and craft reagent things (5 wood)
add: added borer cages to the forge crafting station, removed their normal crafting recipe
add: added a coil recipe to the forge crafting station
balance: weapons now inject 1u instead of 0.5u
balance: the staff will attack twice as fast (still no damage)
del: removed all reagent weapons/armors from crafting menu, use the forge crafting station
/:cl:

PR #11303, opened 2022-02-06
Forge weapon balance pass

YakumoChen:
## About The Pull Request

Makes all the weapons bulky
Add throwforce to most of the weapons (why didn't they have any?)
Axes are now more suited to throwing since they didn't have much of a niche compared to spears (which has reach)
Nerfed the buckler's block and integrity.

All the weapons being Normal class made no sense (especially polearm weapons), it basically let you fill your bag with weapons. Given weapons tend to be contraband, you should have a little more risk for lugging them around since you can make them very easily (and they tend to be better than any other station-found weapon)

The bucklers were quite frankly over-statted since they had better stats than every other non-energy shield in the game. They are now a more durable wooden buckler stat-wise.

## Changelog

:cl:
balance: Forge weapons are now bulkier, and hurt when thrown.
balance: forge shields are more durable bucklers instead of being the best shields in the game.
/:cl:

### DIFF
--- a/modular_skyrat/modules/reagent_forging/code/forge_weapons.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge_weapons.dm
@@ -16,7 +16,7 @@
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	throwforce = 10
 	slot_flags = ITEM_SLOT_BELT | ITEM_SLOT_BACK
-	w_class = WEIGHT_CLASS_NORMAL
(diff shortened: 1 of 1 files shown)


jjpark-kb:
I definitely agree with the changes, but we should hold off until https://github.com/Skyrat-SS13/Skyrat-tg/pull/11295

YakumoChen:
Needs more passing anyway, given I wrote this at 6 am

PR #11335, opened 2022-02-07
[NON-MODULAR] Mime Vow Changes

axietheaxolotl:
## About The Pull Request

Changes the cooldown for the mime's vow to five seconds, and removes the mood debuff.

It's fucking dumb, mimes are actors, not aliens or whatever dumb shit /tg/ made them. For reference, the mime mood de-buff was -12, losing a limb was -8. The cooldown on vow breaking was insanely long for no real reason and is just generally dumb.

This solidifies that mimes are just, people, not weird aliens.

## Changelog

:cl:
balance: Mime's no longer suffer a mood debuff, and their vow cooldown has been changed from five minutes, to five seconds.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/mime.dm
+++ b/code/modules/spells/spell_types/mime.dm
@@ -113,7 +113,7 @@
 	clothes_req = FALSE
 	human_req = TRUE
 	antimagic_allowed = TRUE
-	charge_max = 3000
+	charge_max = 50 // SKYRAT EDIT | ORIGINAL: charge_max = 3000
 	range = -1
 	include_user = TRUE
 
@@ -138,9 +138,9 @@
 		H.mind.miming=!H.mind.miming
 		if(H.mind.miming)
 			to_chat(H, span_notice("You make a vow of silence."))
-			SEND_SIGNAL(H, COMSIG_CLEAR_MOOD_EVENT, "vow")
+//			SEND_SIGNAL(H, COMSIG_CLEAR_MOOD_EVENT, "vow") // SKYRAT EDIT | Mimes are people, not freaks.
 		else
-			SEND_SIGNAL(H, COMSIG_ADD_MOOD_EVENT, "vow", /datum/mood_event/broken_vow)
+//			SEND_SIGNAL(H, COMSIG_ADD_MOOD_EVENT, "vow", /datum/mood_event/broken_vow) // SKYRAT EDIT | Mimes are people, not freaks.
 			to_chat(H, span_notice("You break your vow of silence."))
 
 // These spells can only be gotten from the "Guide for Advanced Mimery series" for Mime Traitors.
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
if you're going to do this, you should probably do it for the clowns stuff aswell, and maybe, at the minimum nerf the time on the mimes wall, because with 5s on the on/off of the vow that wall is practically just a flat buff with no downsides

axietheaxolotl:
> if you're going to do this, you should probably do it for the clowns stuff aswell, and maybe, at the minimum nerf the time on the mimes wall, because with 5s on the on/off of the vow that wall is practically just a flat buff with no downsides

I think mime walls and shit are dumb, but, yeah, that's fair.

SpaceVampire:
Have you tried just actually playing the job seriously and not talking?
This kind of stuff helps to make people actually play the role correctly, you shouldn't be able to suddenly just walk if it is more convenient with no downsides. The current state works fine for treating it as a proper "last resort" scenario, instead of an easy button you can just flick off and on whenever you feel like it.

Oh also. You really can't just compare mood de-buffs like that.
Given that being mildly depressed "<span class='warning'>I feel a bit down.</span>\n" is -10 mood.
Where getting stabbed in the fucking eye is -4 mood.

axietheaxolotl:
>  This kind of stuff helps to make people actually play the role correctly, you shouldn't be able to suddenly just walk if it is more convenient with no downsides. The current state works fine for treating it as a proper "last resort" scenario, instead of an easy button you can just flick off and on whenever you feel like it.

This is dumb. I don't talk as a mime, I do it once in a blue moon. But my point still stands that this is dumb. This stuff doesn't help people play the role correctly, if mimes want to speak, they will. This changes very little, besides removing a debuff that has no real right to exist. People are under the impression this stuff existing magically stops mimes from talking, it doesn't. This is a lie you're telling yourself and is factually incorrect.

SpaceVampire:
> This is dumb. I don't talk as a mime, I do it once in a blue moon. But my point still stands that this is dumb. This stuff doesn't help people play the role correctly, if mimes want to speak, they will. This changes very little, besides removing a debuff that has no real right to exist. People are under the impression this stuff existing magically stops mimes from talking, it doesn't. This is a lie you're telling yourself and is factually incorrect.

If I am reading it correctly, the debuff is only in effect when the vow is broken so I really don't see the issue with it. It's some mechanical punishment and flavor for going against the whole point of the role. I think that having to actually wait to get your special powers back again is a real deterrent, that encourages people to do it less often. It's not "magical" it is just like... the whole point behind the design in the first place?

tf-4:
This is the most unnecessary thing I ever saw. Just don't talk, as you shouldn't.
You've not given a reason other than "it's bad because it's the dumbest thing ever".

Gandalf2k15:
Unnecessary non-modular change, aggressive PR body, insufficient reasoning for change.

PR #11409, opened 2022-02-09
Mining MODsuit rebalance

Stalkeros2:
## About The Pull Request
Reduced mining MODsuit's armor to 30 melee, with ash accretion module also giving 30 melee armor.
Increased its vendor price to 5k

100 melee armor out of nowhere isn't that good.
And the suit is **too** good for its current price. So good that you don't even have to mine anything to get one.

## Changelog

:cl: Stalkeros
balance: Mining MODsuit's armor reduced to 30 melee.
balance: Ash accretion module now gives 30 melee armor instead of 50.
balance: Mining MODsuit's vendor price increased to 5k
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -46,7 +46,7 @@
 		new /datum/data/mining_equipment("Super Resonator", /obj/item/resonator/upgraded, 2500),
 		new /datum/data/mining_equipment("Jump Boots", /obj/item/clothing/shoes/bhop, 2500),
 		new /datum/data/mining_equipment("Ice Hiking Boots", /obj/item/clothing/shoes/winterboots/ice_boots, 2500),
-		new /datum/data/mining_equipment("Mining MODsuit", /obj/item/mod/control/pre_equipped/mining, 2500),
+		new /datum/data/mining_equipment("Mining MODsuit", /obj/item/mod/control/pre_equipped/mining, 5000), //SKYRAT EDIT: "NERFED" PRICING
 		new /datum/data/mining_equipment("Luxury Shelter Capsule", /obj/item/survivalcapsule/luxury, 3000),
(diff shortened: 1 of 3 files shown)


Stalkeros2:
Oh damn the internet screwed up and didn't name it.

Stalkeros2:
And I can't do that myself either.
Well shit.

tf-4:
renamed it as asked

MassacreDemigirl:
how about making its armor to 10 and let ash module make them real armor in lavaland?

CandleJaxx:
i'd suggest adding [NON-MODULAR] to the front to clarify that it's a core code edit, and i'd suggest taking it upstream first so they can handle it there- since it's still new, this could be overwritten in the future from there aswell

tf-4:
> i'd suggest adding [NON-MODULAR] to the front to clarify that it's a core code edit, and i'd suggest taking it upstream first so they can handle it there- since it's still new, this could be overwritten in the future from there aswell

It's unbalanced as a result of a buff we gave it, so it's fine to be down here. Should still have the tag tho.

YakumoChen:
Is this actually still balanced? I'm not sure what the armour was on the mining suit before it had the ash module (if it had its own module that added armour). Right now the ash thing is still giving a ton of extra values on top of the skyrat suit's base (100 bomb armour shouldn't be on anything outside of bomb suits or the RD's suit, for instance). 35 energy armour is also relatively high.

I'm fairly wary of giving the modsuit a lot of extraneous armour as a whole since the only real threats are melee and fire and security will have an extremely hard time dealing with miners on lavaland if their suit is over-armoured. (miners already have extreme advantages fighting on their turf)

Stalkeros2:
#11428 appeared and, in fact, is a better alternative to what I did.
So this gets closed.

PR #11411, opened 2022-02-09
[MIRROR] Fixes layering issues brought by the FoV PR. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63903
--------------------
## About The Pull Request
I had been complaining about the FoV PR having some issues for quite some time. The major one being the unintential changes to the layering of several objects which used to stand above mobs but now don't (most notable example being potted plants). I told Azarak about it and he replied fixing it would require some time, which is a cordial way to say he probably won't fix it because of reasons, so I took the matter in my hands (in a lingering demeanor tbh).
Objects and effects with layers higher than the usual mob ones had to have their planes changed, but halfway in the PR I discovered that some large mobs also have layers that are also higher than a portion of these objects, and that further complicated the process. To solve it I had to add two more planes; one assigned to large mobs and other fov-hideables with layers going above the majority of things, another for effects and objects that are also displayed above large mobs & co.
Aware that this solution would add an extra layer of maintenance needed when playing around with planes, I have also added some non-doc comments to the layers.dm file to guide people through the layers of different planes, which is the most I could do at the moment.
While I was at it, I created another layer to use in place of the below mob layer for certain mobs such as pAIs and the swarmer beacon megafauna (not to be confused with the random event one, which is a structure).

I have locally and successfully tested it with a christmas tree, two human mobs, a gas mask, a xeno queen, a potted plant and a recycler.
That said, there is one issue with the FoV I'm currently aware of but I'm not planning to fix in this PR since I don't want to overly cogitate now and that's larva not being able to hide under objects like tables anymore (correct me if I'm wrong): The easiest solution on paper would be making another two planes, but the far greater amount of objects and layers I would have to go through to fix such a small hindrance is such that I'm question if it wouldn't be better for it to be revamped later.

## Why It's Good For The Game
This will fix several common issues with planes and layers, the most mundane ones being potted plants and plastic flaps being displayed under mobs and not over. I have just searched for relative issue reports but I have found none, yet I kind of refuse to believe no one officially reported it, so if you know of a reported issue this PR will fix, please include the number in the comments and I'll include them in this section.

EDIT: This should fix #64078.

## Changelog

:cl:
fix: Fixed several common issues with planes and layers, most noticeably potted plants and plastic flaps being displayed under mobs and not over.
/:cl:


code/__DEFINES/layers.dm
```diff
@@ -23,14 +23,25 @@
...
+#define GAME_PLANE_UPPER_FOV_HIDDEN -3
+
 #define ABOVE_GAME_PLANE -2
 
+<<<<<<< HEAD
 #define MOUSE_TRANSPARENT_PLANE -1 //SKYRAT EDIT ADDITION - Pollution port
 
+=======
+// PLANE_SPACE layer(s)
+>>>>>>> c5d2b2e51e5 (Fixes layering issues brought by the FoV PR. (#63903))
```
tf-4, line 39:
```suggestion
#define MOUSE_TRANSPARENT_PLANE -1 //SKYRAT EDIT ADDITION - Pollution port

// PLANE_SPACE layer(s)
```

tf-4:
```suggestion
#define MOUSE_TRANSPARENT_PLANE -1 //SKYRAT EDIT ADDITION - Pollution port

// PLANE_SPACE layer(s)
```

code/modules/mob/living/living_defines.dm
```diff
@@ -4,7 +4,10 @@
 	see_in_dark = 2
 	hud_possible = list(HEALTH_HUD,STATUS_HUD,ANTAG_HUD)
 	pressure_resistance = 10
+<<<<<<< HEAD
 	plane = GAME_PLANE //SKYRAT EDIT CHANGE
+=======
+>>>>>>> c5d2b2e51e5 (Fixes layering issues brought by the FoV PR. (#63903))
```
tf-4:
```suggestion
	plane = GAME_PLANE //SKYRAT EDIT ADD
```

code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
```diff
@@ -107,8 +108,14 @@
...
 	mob_biotypes = MOB_ORGANIC | MOB_PLANT
 	layer = SPACEVINE_MOB_LAYER
+<<<<<<< HEAD
 	health = 60 //SKYRAT EDIT CHANGE
 	maxHealth = 60 //SKYRAT EDIT CHANGE
+=======
+	plane = GAME_PLANE_UPPER_FOV_HIDDEN
+	health = 50
+	maxHealth = 50
+>>>>>>> c5d2b2e51e5 (Fixes layering issues brought by the FoV PR. (#63903))
```
tf-4:
```suggestion
	plane = GAME_PLANE_UPPER_FOV_HIDDEN
	health = 60 //SKYRAT EDIT CHANGE - ORIGINAL = 50
	maxHealth = 60 //SKYRAT EDIT CHANGE - ORIGINAL = 50
```

PR #11428, opened 2022-02-10
[MODULAR] Mining MODsuit Armor Balance Alternative

MassacreDemigirl:
## About The Pull Request
This is an alternative to #11409

Makes the base melee armor of Mining MOD 5

Mining Armor booster gives 50 Melee armor in ashlands. Currently this causes miners to have a whopping Melee 100 armor which should not be possible for players. This PR makes the mining MODsuits in line with civilian MOD's since miners are by design to not have weaponry and armor on station. Thus this makes Miniing MODsuit 55 Melee, while Drake Armor is 65, giving the miners choice between having the utility of a modsuit or the extra 10 armor of drake. 

After playing miner more, I can say while the modsuit has many nice gimmick modules, most are as named: gimmicks.
Most of the modules are already competing with already existing items. Therefor I decided not to touch its pricing to not make the modsuit redundant.

_`Also since our round times are the double of TG, doubling the point total of Kitted MODsuit to 5000 to make it achievable from a 20 minute to a 40 minute grind`_

## Changelog

:cl:
fix: Fixes the Melee Armor X (100) Mining MODsuits. Now they have a base Armor of I (5). They also have lower innate Laser, Bullet, Energy and Bomb Armor since they gain it from the ash module
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -46,7 +46,7 @@
 		new /datum/data/mining_equipment("Super Resonator", /obj/item/resonator/upgraded, 2500),
 		new /datum/data/mining_equipment("Jump Boots", /obj/item/clothing/shoes/bhop, 2500),
 		new /datum/data/mining_equipment("Ice Hiking Boots", /obj/item/clothing/shoes/winterboots/ice_boots, 2500),
-		new /datum/data/mining_equipment("Mining MODsuit", /obj/item/mod/control/pre_equipped/mining, 5000), //SKYRAT EDIT: 2500
+		new /datum/data/mining_equipment("Mining MODsuit", /obj/item/mod/control/pre_equipped/mining, 2500),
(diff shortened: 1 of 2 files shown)


Stalkeros2:
To be frank, that's a better alternative to what I did.

YakumoChen:
gonna leave what I said on the other PR, the values need tuned so that the energy, bomb, bullet, etc armours are not doubled on lavaland from the base due to the module.

gonenoculer5:
I would say make it 3000-4000. 5k feels like a lil bit too more.

softcerv:
This might be another PR, but has anyone considered adding the MODSuit Jetpack/Thrusters to the mining vendor?

SpaceVampire:
I agree that the armor needs the rebalance to be more in line with what it should be normally.
That being said, the price increase really doesn't feel needed. You could apply that same logic to all kinds of stuff.
Why not just make all of the research on the station take 2x the points to unlock? Just to give an example.

PR #11473, opened 2022-02-12
[Modular] Adds a debug-esque techfab to the ghost cafe, along with materials

DragonTrance:
## About The Pull Request
Adds an omni-techfab to the ghost cafe, which starts off configured with the debug/admin techweb with everything unlocked to start with. This includes a locker full of resources that Runtimestation has, which also includes materials you can't insert into the techfab, so even more choices there!

image

I'm looking for adding more creativity in the ghost cafe, like being able to make vinegar/quality oil using only botany or related resources. The techfab can make _anything_ a normal omnilathe can, except everything requires either 0.1 or 0.25 resource units.

## Changelog

:cl:
add: Adds a debug techfab in the ghost cafe, in the room to the right of Hydroponics.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -6049,6 +6049,10 @@
 	},
 /turf/open/floor/plastic,
 /area/centcom/interlink)
+"cbI" = (
+/obj/structure/closet/syndicate/resources/everything,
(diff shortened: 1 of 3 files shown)


softcerv:
Hopefully this doesn't end up getting abused like the old techfab did.

tf-4:
This was specifically removed

Gandalf2k15:
Unfortunately this was removed because of exploits before. So we won't be returning it.

PR #11476, opened 2022-02-12
[Modular] Adds ways to create various items from the Produce Orders console

DragonTrance:
## About The Pull Request
Adds various chemical reactions, which allows players to create some items from the Produce Orders console. This includes vinegar, quality oil, yoghurt, and cornmeal. This is mostly for people who want to make food in the ghost cafe, they don't have many exotic options besides a jar of mayo

## Changelog

:cl:
add: You can now make Vinegar, Yoghurt, Quality Oil, and Cornmeal without a produce orders console
add: Vinegar is made by mixing wine and acid OR wine and mold
add: Yoghurt is made by mixing Milk, Cream, and mold. You can make yoghurt in real life with bacteria trust me
add: Quality Oil is made by mixing Korta Nectar and Corn Oil OR Universal Enzyme and Korta Nectar
add: You can dry corn, and grinding that will result in Cornmeal
/:cl:


Zenitheevee:
Might wanna re-check one of your recipes for repeats, as it seems the second quality oil recipe is causing errors.

MassacreDemigirl:
you should try to add this upstream

jjpark-kb:
> REASON #1: Chemical recipe conflict between /datum/chemical_reaction/drink/mushi_kombucha and /datum/chemical_reaction/food/quality_oil/two

this needs to be fixed prior to being merged. Please do so ASAP.

DragonTrance:
Just won't keep it in because I can't think of alternative reagents to korta nectar

Also the reason I'm not trying to add this to the upstream is because this seems like a hack for not spending money. Probably not a valid argument, but I just have that feeling it isn't good in the nature that tgstation tries to have

Gandalf2k15:
Regardless, we need you to take it upstream first.

PR #11478, opened 2022-02-12
all z-level experi-scanners

jjpark-kb:
## About The Pull Request
all experi-scanners will connect to the station techweb
it is outrageous that other places cannot complete experiments

science has a ton of experiments it has to do and it causes a lot of tech to be locked for the whole round
allowing people on lavaland, or in other places, to complete experiments means everyone becomes happier

## Changelog

:cl:
add: experi-scanners on all z-levels will connect to the station techweb
/:cl:


Stalkeros2:
BASED

PR #11505, opened 2022-02-14
[NONMOD] Locks tribal weapons to ashwalkers

YakumoChen:
## About The Pull Request

~~removes the bone sword crafting recipe~~ Reverted
Removes all tribal weapons from normal crafting, enable only ashwalkers to use them
Rebalances the bone spear.
The ash village has a bone sword now in its map (I couldn't be bothered to remove it after redoing the PR)

## How This Contributes To The Skyrat Roleplay Experience

Frankly: It isn't balanced for the station.
~~The Bone Sword doesn't fit any design space for tribal weapons. It's too strong for a one-handed weapon, but there's already good 2-hand options for ashwalkers and the bone dagger is the best one handed option you should be able to make from just bones and stuff. It's also absurdly easy to craft for how powerful it is. Given public mining exists (miners themselves have decent weapon options already) it's way too easy to get these and bring them up.~~

I've gone the extra mile and just locked all tribal weapons to ashwalkers. They are insanely easy to craft for how insanely strong they are, and public mining makes it too easy to get the materials. I'm leaving tribal weapon balance to ashwalkers, they can fight over numbers themselves.

The spear was, however, extremely strong for how safe of a weapon option it is on Skyrat. Not only is it 20(!!) force when dual handed but it also has reach AND armour pen. You can't have everything and still have it mass-producible for the price of a watcher or two. It's been knocked down a peg.

Other weapon options the station AND ashwalkers can acquire are being filled design-wise by Forging, which I am rebalancing in another PR.

## Changelog

:cl:
balance: Tribal weapons are now ashwalker-only
balance: Bone spear was nerfed.
qol: adds a bone sword to the ash walker village.
/:cl:

PR #11507, opened 2022-02-14
[MODULAR] Rebalances forge weapons

YakumoChen:
## About The Pull Request
Rebalances forge weapons. Most weapons are BULKY now. That means you have to display them. No more hiding from security.  Many harmful weapons were given a throwforce. 

Each weapon now has a unique flavour. Numbers are balanced around 15 force, not including the niches.
-The sword has block chance, and pairs well with the buckler.
-The katana was buffed and given high armour pen. It's great for delimbing, not quite as good as the cap's sabre.
-Dagger was untouched. It still has rapid strikes.
-Staff was untouched. It still does no damage. Good for healing.
-The spear is two-handed and has reach, like before, but the power was reeled back.
-The axe is now the best weapon for throwing and has a very strong embed.
-The hammer does high raw damage and crits in 7 hits unarmoured.
-The buckler was rebalanced around a wooden buckler's stats, but more durable.
-Pavise now matches the adamantine shield stat-wise. Holding it is slow and requires two hands, but 75 block is still powerful. You can also put it on your back.

Forge had some whacky outputs for something you can make in maint with nothing but metal and wood planks. Having all of them fit in a backpack made them even worse. The crew shouldn't have an easy time mass-producing and concealing weapons security would frown on, especially ones these strong.

Note that while the damage values are fairly high, and not the highest you could get if you look in the right places, this doesn't account for imbuing. With the right imbue, these weapons would be a real terror regardless.

## Changelog

:cl:
balance: Forge weapons and shields have been heavily rebalanced, their descriptions now show what potential uses they have.
/:cl:


ArdentSpark:
As much as I'm not sure I agree with the bone sword changes, you may also want to check the Ash Macuhuitl because they're a stronger wound oriented sword with a much steeper crafting cost, I believe in the clothing craft menu

YakumoChen:
this is the wrong PR, but that weapon doesn't come up in a search. Are you sure you don't just mean the bone sword?

YakumoChen:
@ArdentSpark you want #11505 but I've redone the PR somewhat.

PR #11377, opened 2022-02-08
[SEMI-MODULAR] MCR Expansion Vol 1 & Balance Passes

MassacreDemigirl:
## About The Pull Request
Update: Made inline with the refactors of #11536

This PR started with realizing MCR's did 27 full laser damage and evolved into what it is right now:

Firstly it separates the MCR Laser damage to 20.
Nerfs the laser damage from 27 to 22
Makes the undercharger no longer give grip bonuses when activated.
Repeater beams do 15 damage each for a total of 30 damage.
Scatter beams do 10 damage each for a total of 30 damage.
Moves the superheater upgrade to an illegal MCR attachments tree.
Superheater lasers do 15 damage but still keep 2 firestacks.
Adds Hellfire upgrade in place of superheater, deals 25 damage and has the wound bonuses of hellfire lasers.
Adds a Lance upgrade that gives the MCR the firing delay of 2 seconds however it fires a Heavy Laser beam at twice the projectile speed.
Adds an Illegal Repeater upgrade that does the same damage but has 50 armor penetration.
Adds an Illegal Scatter upgrade that shoots 7 pellets for a total of 45 damage.
Adds an Illegal suppressor upgrade. It suppresses the laser somehow.
Adds a Clown Tech extension for MCR's, a bikehorn upgrade that makes you shoot yellow beams of slipping and low stamina damage.
Adds 3 new skins to the MCR
Lowers the attachment slots of Advanced Cells to 2 and BS Cells to 3
Makes loading a new Cell into the MCR take 4 seconds while tactical reloading takes 6 seconds.
Adds a reloading assister attachment that makes loading a cell take 2 seconds.
Adds a tac-reload helping attachment that makes tactical reloads 3 seconds.

## How This Contributes To The Skyrat Roleplay Experience
When MCR's were merged it caused an uprise in usage of lasers. However something was odd, they were somehow far too strong. Yet there was no code editing the damage of lasers in the MCR PR. This is apparently because of the upgraded base Beam Projectile damage. Which was not edited in any of its childs. Thus more used laser weapons were not affected by this buff (Captains Lasgun and currently Blueshields Lasgun), as well as the Lasgun and Egun were not preferred weapons due to Ballistics and further on Hardlight Bullets existing. Thus this effect was not felt till the merging of MCRs. 

Here are a few other projectiles to compare to:
```
/obj/projectile/bullet/c32
    name = ".32 bullet"
    damage = 15
/obj/projectile/bullet/c46x30mm
[...]

PR #11536, opened 2022-02-16
[MODULAR] Refactors MCR designs

YakumoChen:
## About The Pull Request

as above
less copypasta
clean code is good code
Cells are now in the Ammo category (basic is in Security in autolathe)
All other parts are in Weapons category in seclathe.
Removed designs from scilathe and mechfab, Gandalf said it was "an oversight"

MCR parts in more consistent places in the lathes with other gun stuff, less bloat where it shoudn't be seen.

## Changelog

:cl:
qol: Moved MCR designs around. Cells are now Ammo, Gun Parts are now Weapons category.
refactor: MCR lathe design stuff
fix: A bug in one of the material costs for one of the camos
/:cl:


MassacreDemigirl:
this will be a pain for #11377

YakumoChen:
oops, I didn't realize you had an MCR thing that big. 

Well, all the better then, since I'd have had to go back to yours anyway.

MassacreDemigirl:
I already made my pr compliant with this refactor so, keep this one up

modular_skyrat/modules/microfusion/code/microfusion_designs.dm
```diff
@@ -1,228 +1,173 @@
...
+	name = "Microfusion Part"
+	build_type = PROTOLATHE | AWAY_LATHE
+	departmental_flags = DEPARTMENTAL_FLAG_SECURITY
+	construction_time = 10 SECONDS //dunno if this is for mechfabs or what but I'll keep this anyway
+	category = list("Weapons")
+
+// EMITTERS
+/datum/design/microfusion/enhanced_phase_emitter
+	name = "Enhanced Microfusion Phase Emitter"
+	desc = "The core of a microfusion projection weapon, produces the laser."
```
jjpark-kb: I recommend creating another parent called
> /datum/design/microfusion/emitter

Where it has the desc and the materials. You'll notice you have three designs that have the same desc and materials, which could be condensed with this new parent.

jjpark-kb: What this means is that the new path for this enhanced phase emitter will be 
> /datum/design/microfusion/emitter/enhanced_phase_emitter

or something along those lines.

YakumoChen: I was thinking about splitting up each set of parts more, but it really isn't needed, the vars in the /microfusion/ parent apply to everything else I wanted included, the only thing I'd be knocking down is the desc var.

The materials are also not the same between them. I could give them better descriptions but I was lazy and the names already kind of speak for themselves. Design descriptions are rarely actually seen regardless.

YakumoChen:
redone since #11377 was merged before this. Added more sorting.

PR #11542, opened 2022-02-16
[MODULAR] breast_milk is no longer a /milk/ subtype

Zenitheevee:
## About The Pull Request

Removes breast milk from the /milk/ type-path, Which has some really OK healing behind it. (TBF the healing is actually pretty decent)
Step 1 to allow other projects, and also because i'm not sure how to tackles putting the breast genitals under an udder component atm and might touch that _later_ but this is just step 1

it doesn't but i'm not going to argue considering breastmilk is already only obtainable by one means.

## Changelog

:cl:
balance: removes milk's healing and usage from breastmilk.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_items/lewd_items/code/arousal_system.dm
+++ b/modular_skyrat/modules/modular_items/lewd_items/code/arousal_system.dm
@@ -58,11 +58,12 @@
 	reagent_state = LIQUID
 	shot_glass_icon_state = "shotglasswhite"
 
-/datum/reagent/consumable/milk/breast_milk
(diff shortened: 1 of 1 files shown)


Gandalf2k15 (approved):

TheGandalf85:
A bit late, but this ends up removing all uses of this liquid type as the specific ingredient for milk itself. Unless functionality is given to it, it mechanically becomes the same as water. This has some implications for some players who seek to cook with this specific liquid, along other things.

If trying to balance the healing, why not overwrite what it ticks?

PR #11562, opened 2022-02-17
Fixes windoor access to Delta Station's ORM

Tastyfish:
## About The Pull Request

This is one of those things that I swore was an open issue, but it somehow isn't.

Fixes the windoor blocking the ORM on Delta Station to only require ORM access... so people who should be able to, actually can use the ORM, like every other map.

I'm pretty sure this is an unintended difference between Delta and every other map.

## Changelog

:cl: tastyfish
fix: Windoor blocking ORM on Delta Station now has correct access.
/:cl:


Zonespace27:
Confirmed not to happen upstream?

Tastyfish:
> Confirmed not to happen upstream?

Ah, you're right, it's in the upstream as well. I didn't bother checking since maps aren't in sync anyway, but I can do an upstream PR first.

PR #11574, opened 2022-02-18
[MODULAR][BALANCE] Allows the CMO Hypospray to pierce MODSuits

softcerv:
## About The Pull Request
Changes the penetration value on the CMO Hypospray to allow it to pierce thick clothing. 

This makes our CMO Hypospray more in line with the one that the TG CMOs get, this should also make the hypospray a more enticing item to steal.

## Changelog

:cl:
balance: The CMO Hypospray can now pierce thick clothing.
/:cl:

### DIFF
--- a/modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
+++ b/modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
@@ -57,6 +57,7 @@
 	spray_wait = DELUXE_WAIT_SPRAY
 	spray_self = DELUXE_SELF_SPRAY
 	inject_self = DELUXE_SELF_INJECT
+	penetrates = INJECT_CHECK_PENETRATE_THICK
 
 /obj/item/hypospray/mkii/Initialize()
 	. = ..()


Vizoplez:
Mistake. Only encourages traitors to steal it and load it with deathmixes, nothing should bypass space-level protection/armor as a roundstart easy-to-steal item.

Think of this from a balance perspective. I can just mix zombie-powder; which I can get from an emag, I can also emag/airlock the CMO's office, and then emag the hypospray itself to make it instant.

It's only going to encourage bad gameplay, on both sides. Instant hyposprays have always been a bane of balance. What chance of death do you have when you can use inaprovaline + combat mix with no delay?

softcerv:
>Think of this from a balance perspective. I can just mix zombie-powder; which I can get from an emag, I can also emag/airlock the CMO's office, and then emag the hypospray itself to make it instant.

Emagging it already allows it to pierce armor though. This changes nothing balance wise if we are looking at this from the perspective of a traitor emagging it.

>It's only going to encourage bad gameplay, on both sides. Instant hyposprays have always been a bane of balance. What chance of death do you have when you can use inaprovaline + combat mix with no delay?

This isn't removing or changing anything about the delay, this is just making it so the Hypospray CMO can pierce armor. Keep in mind that the CMO hypos here have a delay compared to the instant ones on TG

PR #11583, opened 2022-02-18
[MIRROR] Makes newscasters, request consoles, and status displays emissive [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/64858
--------------------

## About The Pull Request
* Makes newscasters, request consoles, and status displays have emissive displays.

image

The blinking light for request consoles when in alert states is also emissive now.

I also refactored the status display code (I know, I just changed it to scroll, but bear with me) to hold state normally, and use `update_overlays()` and `update_appearance()`, rather than chaotically messing with overlays on the fly.
* This means if a `status_display/evac` loses power and turns back on, it'll resume where it left off instead of being blank.
* The status display's backlight is only on when showing a picture, or a non-empty message, making for a nice effect ihmo.
* The text marquee system has also been squeezed in to work with this, and is now optimized to only update the status message line objects on a per-line basis, so countdowns for evac and cargo shuttles will use half as much cpu and net traffic on each update.

Tangentially, I noticed the command console's Set Status Display feature's Clear Alert and Shuttle ETA feature were trying to set alert pictures called "blank" and "shuttle" which just shows a blank picture since those don't exist, but not the intended functionality. Those buttons now send the correct signals to the status displays.

Also tangentially, building the status display frame resulted in the base class that just sits at a blank screen, so I changed it to make the most common evac subclass (the ones that follow the PDA/command console buttons).

I tested power loss and *friendcomputer and all that jazz.

## Why It's Good For The Game

Glowy screens look good, and are in line with all newer machines with screens. They're also readable in the dark, to nobody's surprise, which helps QoL I suppose.

image

## Changelog

:cl: tastyfish
add: Newscasters, request consoles, and status displays now have a backlight on their displays and emit a soft glow.
imageadd: Updated the newscaster, request console, and status display sprites to work with being emissive.
refactor: Internally refactored status display visual updating, so that it retains state across power loss, etc.
/:cl:


Tastyfish:
I can fix this PR to work with our different sprites (I'll make a PR for this PR)

PR #11606, opened 2022-02-19
Removes Rope Bunny quirk.

Zonespace27:
## About The Pull Request
This PR removes the Rope Bunny quirk, which gives you an extra +0 mood bonus when tied up with shibari gear, and makes the time to escape shibari 4 minutes from 45 seconds.

There are several reasons behind this removal:

- If the focus of the quirk is on the mood bonus, why do you need the quirk at all? Can't you just have that be part of your character, instead of the game needing to tell you that you like it?
- If the focus of the quirk is the raised escape time, why? I don't see why you'd be escaping from it in the first place, considering that you __should not__ be using this for antaggery.
- I feel that, with the two points above, this quirk is bloat that doesn't really contribute much to the game.

## Changelog

:cl:
del: Removed Rope Bunny quirk.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_items/lewd_items/code/arousal_system.dm
+++ b/modular_skyrat/modules/modular_items/lewd_items/code/arousal_system.dm
@@ -12,7 +12,6 @@
 #define TRAIT_SADISM		"sadism"
 #define TRAIT_NEVERBONER	"neverboner"
 #define TRAIT_RIGGER		"rigger"
-#define TRAIT_ROPEBUNNY		"rope bunny"
 
 #define APHRO_TRAIT			"aphro"				///traits gained by brain traumas, can be removed if the brain trauma is gone
 #define LEWDQUIRK_TRAIT		"lewdquirks"		///traits gained by quirks, cannot be removed unless the quirk itself is gone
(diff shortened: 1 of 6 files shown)


desertmote:
If you remove content made by our downstream without having discussed it with them, they will get mad, rightfully so.
This has already happened within last week. Furthermore, this is from a recently merged PR that was testmerged for two weeks. I would say you missed your window.

Zonespace27:
> If you remove content made by our downstream without having discussed it with them, they will get mad, rightfully so. 

I'll talk to them over that.

> Furthermore, this is from a recently merged PR that was testmerged for two weeks. I would say you missed your window.

I didn't realize that the quirk did borderline nothing until I ctrl + F'ed through the code.

Edit: [Talked](https://discord.com/channels/596783386295795713/610319137012580383/944621751801897021) to the downstream, they're aware of the PR now. However, we couldn't agree on anything except to let the maintainers decide

Lamella-0587:
image

Lamella-0587:
I don't like it. First - we were forbidden to use mood modifiers. Now, this quirk going to be removed because it doesn't really affect players. Our update is RP-oriented, these quirks are required to immerse players more. 

I can't make an update that should not have important mechanics that can be abused by players and have mechanics that will be significant for players at the same time.

thestubborn:
> If you remove content made by our downstream without having discussed it with them, they will get mad, rightfully so. This has already happened within last week. Furthermore, this is from a recently merged PR that was testmerged for two weeks. I would say you missed your window.

i disagree with rightfully so
if they don't want their content modified by us, make their own fork
if their host doesn't want to, that is not our problem

Zenitheevee:
I dont like this for several reasons, Some of them more sane than others.

1: It costs nothing to use, nothing to not use, And seems to only give a very situational downgrade that... You're _not_ going to encounter regularly enough to really matter. Its not free quirk points, and its effects are very negligible. We literally have much worse quirks when it comes to being absolutely negligible or otherwise *completely* ignorable as a concept.

> If the focus of the quirk is the raised escape time, why? I don't see why you'd be escaping from it in the first place, considering that you should not be using this for antaggery.

2: Realistically, You're only going to be escaping if your RP partner was forced to leave Or had a sudden disconnect, and even then... Chances are? You're gonna have more than 5 minutes to wriggle out. What does this 5 minutes give you? Plenty of time to type as your character is wriggling out, It can be used in RP and i dont feel like i need to say more, nor that there's more to say on this point

3: I just dont see a good reason to remove it IMO. Should we remove the empty cargo crate order because we can buy plasteel/wood to make crates? Should we remove certain techweb nodes and compress them into other nodes because having 1 node for 1 item that no one uses is 'bloat' ? Should we just remove _all_ ERP quirks because they shouldn't be used in the main-game's construct?

Zonespace27:
People seem to like the quirk more than I thought, so it stays.

Lamella-0587:
> People seem to like the quirk more than I thought, so it stays.

Thank you!

PR #11614, opened 2022-02-19
lower required reagent for reagent forging

jjpark-kb:
## About The Pull Request
lowers the required amount of reagents for reagent forging from 200 to 100.
increased the amount of legion cores required for reagent forging up from 3 to 6.
lowered the maximum amount of chems in reagent stuff from 500 to 250.
reagent weapons will now directly inject the chemicals without snowflake code.

Overall, I overlooked the possibility that, because the reagent things have 500u capacity, they could be used to create some insane stuff. Even though they could not have their reagents taken out, having access to a payload of 500 is pretty insane before bluespace beakers (and and they are still better than bluespace beakers). In order to compensate for that, I have to lower the required reagents; and because I did that, I increased the amount of legion cores required.

## Changelog

:cl:
balance: lowered reagent cap in reagent forging items from 500 to 250
balance: lowered required reagents in reagent forging from 200 to 100
balance: increased the amount of cores required to start reagent forging from 3 to 6
balance: reagent weapons will now directly inject the chemicals without snowflake code
/:cl:

### DIFF
--- a/modular_skyrat/modules/reagent_forging/code/forge.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge.dm
@@ -38,7 +38,7 @@
 	///the fuel amount (in seconds) that the forge has (stronger than wood)
 	var/forge_fuel_strong = 0
 	///whether the forge is capable of allowing reagent forging of the forged item.
-	//normal forges are false; to turn into true, use 3 (active) legion cores.
+	//normal forges are false; to turn into true, use 6 (active) legion cores.
 	var/reagent_forging = FALSE
 	///counting how many cores used to turn forge into a reagent forging forge.
 	var/current_core = 0
@@ -79,7 +79,7 @@
 	if(forge_level < FORGE_LEVEL_THREE)
 		. += span_notice("[src] has [goliath_ore_improvement]/3 goliath hides.")
 		. += span_notice("[src] has [current_sinew]/10 watcher sinews.")
-		. += span_notice("[src] has [current_core]/3 regenerative cores.")
+		. += span_notice("[src] has [current_core]/6 regenerative cores.")
 	. += span_notice("<br>[src] is currently [forge_temperature] degrees hot, going towards [target_temperature] degrees.<br>")
 	if(reagent_forging)
 		. += span_warning("[src] has a red tinge, it is ready to imbue chemicals into reagent objects.")
@@ -185,7 +185,7 @@
 		current_sinew = 10
 		forge_level = FORGE_LEVEL_TWO
 	if(user_smithing_skill >= SKILL_LEVEL_MASTER)
-		current_core = 3
+		current_core = 6
 		forge_level = FORGE_LEVEL_THREE
 		create_reagent_forge()
 	if(forge_level == previous_level)
@@ -313,7 +313,7 @@
 		qdel(I)
 		current_core++
 		in_use = FALSE
-		if(current_core >= 3) //use three regenerative cores to get reagent forging capabilities on the forge
+		if(current_core >= 6) //use six regenerative cores to get reagent forging capabilities on the forge
 			create_reagent_forge()
 		return
 
@@ -446,7 +446,7 @@
 			fail_message(user, "You fail imbueing [attacking_item]!")
 			return
 		for(var/datum/reagent/weapon_reagent in attacking_item.reagents.reagent_list)
-			if(weapon_reagent.volume < 200)
+			if(weapon_reagent.volume < 100)
 				attacking_item.reagents.remove_all_type(weapon_reagent.type)
 				continue
 			weapon_component.imbued_reagent += weapon_reagent.type
@@ -481,7 +481,7 @@
 			fail_message(user, "You fail imbueing [attacking_item]!")
 			return
 		for(var/datum/reagent/clothing_reagent in attacking_item.reagents.reagent_list)
-			if(clothing_reagent.volume < 200)
+			if(clothing_reagent.volume < 100)
 				attacking_item.reagents.remove_all_type(clothing_reagent.type)
 				continue
 			clothing_component.imbued_reagent += clothing_reagent.type
@@ -588,7 +588,7 @@
 	return ..()
 
 /obj/structure/reagent_forge/ready
-	current_core = 3
+	current_core = 6
 	reagent_forging = TRUE
 	sinew_lower_chance = 100
 	forge_temperature = 1000
--- a/modular_skyrat/modules/reagent_forging/code/forge.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge.dm
@@ -184,14 +184,14 @@
 		to_chat(user, span_notice("[src] requires you to be more experienced!"))
 		return
 	if(user_smithing_skill >= SKILL_LEVEL_APPRENTICE)
-		goliath_ore_improvement = 3
+		goliath_ore_improvement = MAX_UPGRADE_GOLIATH
 		forge_level = FORGE_LEVEL_ONE
 	if(user_smithing_skill >= SKILL_LEVEL_EXPERT)
 		sinew_lower_chance = 100
-		current_sinew = 10
+		current_sinew = MAX_UPGRADE_SINEW
 		forge_level = FORGE_LEVEL_TWO
 	if(user_smithing_skill >= SKILL_LEVEL_MASTER)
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -79,7 +79,7 @@
 	if(forge_level < FORGE_LEVEL_THREE)
 		. += span_notice("[src] has [goliath_ore_improvement]/3 goliath hides.")
 		. += span_notice("[src] has [current_sinew]/10 watcher sinews.")
-		. += span_notice("[src] has [current_core]/3 regenerative cores.")
+		. += span_notice("[src] has [current_core]/6 regenerative cores.")
```
Zonespace27: IMO make the limits into defines

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -446,7 +446,7 @@
 			fail_message(user, "You fail imbueing [attacking_item]!")
 			return
 		for(var/datum/reagent/weapon_reagent in attacking_item.reagents.reagent_list)
-			if(weapon_reagent.volume < 200)
+			if(weapon_reagent.volume < 100)
```
Zonespace27: ditto for reagent volume

modular_skyrat/modules/reagent_forging/code/reagent_component.dm
```diff
@@ -20,7 +20,7 @@
 	if(set_slot)
 		checking_slot = set_slot
 	parent_clothing = parent
-	parent_clothing.create_reagents(500, INJECTABLE | REFILLABLE)
+	parent_clothing.create_reagents(250, INJECTABLE | REFILLABLE)
```
Zonespace27: Ditto for reagent max

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -185,7 +191,7 @@
 		current_sinew = 10
 		forge_level = FORGE_LEVEL_TWO
 	if(user_smithing_skill >= SKILL_LEVEL_MASTER)
-		current_core = 3
+		current_core = 6
```
Zonespace27: set this and the ones above to use the defines as well

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -588,7 +594,7 @@
 	return ..()
 
 /obj/structure/reagent_forge/ready
-	current_core = 3
+	current_core = 6
```
Zonespace27: set this to the define

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -38,7 +44,7 @@
 	///the fuel amount (in seconds) that the forge has (stronger than wood)
 	var/forge_fuel_strong = 0
 	///whether the forge is capable of allowing reagent forging of the forged item.
-	//normal forges are false; to turn into true, use 3 (active) legion cores.
+	//normal forges are false; to turn into true, use 6 (active) legion cores.
```
GoldenAlpharex, line 47:
```suggestion
	///normal forges are false; to turn into true, use 6 (active) legion cores.
```

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -178,14 +184,14 @@
 		to_chat(user, span_notice("[src] requires you to be more experienced!"))
 		return
 	if(user_smithing_skill >= SKILL_LEVEL_APPRENTICE)
-		goliath_ore_improvement = 3
+		goliath_ore_improvement = MAX_UPGRADE_GOLIATH
 		forge_level = FORGE_LEVEL_ONE
 	if(user_smithing_skill >= SKILL_LEVEL_EXPERT)
-		sinew_lower_chance = 100
-		current_sinew = 10
+		sinew_lower_chance = MAX_UPGRADE_SINEW * 10 //100, just written funny!
```
GoldenAlpharex, line 190: That feels a bit weird, I'd make it its own define or something, because this isn't really related to the amount of upgrades from sinew you can do, and is prone to breaking if that `MAX_UPGRADE_SINEW` define is changed.

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -274,7 +280,7 @@
 			to_chat(user, span_warning("You cannot do multiple things at the same time!"))
 			return
 		in_use = TRUE
-		if(sinew_lower_chance >= 100) //max is 100
+		if(sinew_lower_chance >= (MAX_UPGRADE_SINEW * 10)) //max is 100
```
GoldenAlpharex, line 283: Same comment as previous.

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -178,14 +184,14 @@
...
-		sinew_lower_chance = 100
-		current_sinew = 10
+		sinew_lower_chance = MAX_UPGRADE_SINEW * 10 //100, just written funny!
+		current_sinew = MAX_UPGRADE_SINEW
 		forge_level = FORGE_LEVEL_TWO
 	if(user_smithing_skill >= SKILL_LEVEL_MASTER)
-		current_core = 3
+		current_core = MAX_UPGRADE_REGEN
 		forge_level = FORGE_LEVEL_THREE
 		create_reagent_forge()
```
GoldenAlpharex, line 196: I feel like this would make more sense as a switch statement. Or at the very least, a switch statement for the variable assignation for `forge_level` because otherwise you constantly set it to 1, then to 2, then to 3, if you're a MASTER.

modular_skyrat/modules/reagent_forging/code/forge.dm
GoldenAlpharex: Actually, looking back on this, just make the define be 100, that'd be clearer.

modular_skyrat/modules/reagent_forging/code/forge.dm
GoldenAlpharex: Actually, I think the biggest thing is that sinew_lower_chance feels like a weird name for that variable.

GoldenAlpharex (changes requested):
All is fine, except the MAX_UPGRADE_SINEW, that's a bit weirdly implemented.

modular_skyrat/modules/reagent_forging/code/forge.dm
```diff
@@ -274,7 +280,7 @@
 			to_chat(user, span_warning("You cannot do multiple things at the same time!"))
 			return
```
GoldenAlpharex, line 281: Also the whole `if(in_use)` thing should be at the beginning of the proc, after everything else that doesn't use it.

modular_skyrat/modules/reagent_forging/code/forge.dm
jjpark-kb: people would see the 100 and think they would need 100 sinew.

modular_skyrat/modules/reagent_forging/code/forge.dm
jjpark-kb: the // is just an explanation, not the documentation-- which is right above it.

modular_skyrat/modules/reagent_forging/code/forge.dm
jjpark-kb: sinew_lower_chance could be changed to sinew_temp_chance I guess, but its slowly becoming more not atomized with these changes. I definitely agree it could be named better, but that should be for an entirely different PR.

modular_skyrat/modules/reagent_forging/code/forge.dm
jjpark-kb: it would have a lot of repeat code if it was a switch statement, since it is intended to go through all of the ifs. Also, you can only upgrade it once up to its level, so a level 2 cant be downgraded to level 1, etc.

modular_skyrat/modules/reagent_forging/code/forge.dm
jjpark-kb: this would need to be in another PR where I just clean up forge code.

PR #11624, opened 2022-02-20
Fixes the accidentally-reverted drill icon, makes the power tools engi-colored

ghost:
## About The Pull Request
Title
When we fixed the base screwdriver sprites, we forgot to re-link the power-screwdriver back to the right file. So it went back to TG's hair-dryer-brick.

While I was fixing it though, I realized how poorly silver-and-purple fits with power tools; experimental welder? Sure, makes sense, its top of the line science, but power-tools are just combination versions of base tools.
So, I made them yellow, with a bit of help from kat for the palette;

image

This also includes inhands for the new yellow JOL (and thus a better one for the syndie JOL too)

Better sprites are better

## Changelog

:cl:
imageadd: Nanotrasen's power screwdriver/JoL designs now have a yellow plasti-casing rather than the experimental science silver/purple. Engineers have already reported feeling more attached to their tools both emotionally and, due to the new HiTrac Grip design, physically attached as well.
/:cl:


jjpark-kb:
atomization; fixs should not be paired with additional content like this.

ghost:
ok :)

PR #11644, opened 2022-02-21
[NON-MODULAR] Removes the bluespace launchpad from the uplink.

nevimer:
## About The Pull Request

Does as it says in the title. Is a request.

Hopefully less of this stealth antag nonsense, as it's a very controverisal item. 

## Changelog

:cl:

del: Removed bluespace launchpad from uplink. 

/:cl:

### DIFF
--- a/code/game/objects/items/storage/uplink_kits.dm
+++ b/code/game/objects/items/storage/uplink_kits.dm
@@ -38,7 +38,7 @@
 		)))
 		if(KIT_RECON)
 			new /obj/item/clothing/glasses/thermal/xray(src) // ~8 tc?
-			new /obj/item/storage/briefcase/launchpad(src) //6 tc
+//			new /obj/item/storage/briefcase/launchpad(src) //6 tc // SKYRAT EDIT REMOVAL
 			new /obj/item/binoculars(src) // 2 tc?
 			new /obj/item/encryptionkey/syndicate(src) // 2 tc
 			new /obj/item/storage/box/syndie_kit/space(src) //4 tc
--- a/code/modules/uplink/uplink_items/device_tools.dm
+++ b/code/modules/uplink/uplink_items/device_tools.dm
@@ -60,13 +60,13 @@
 	cost = 1
 	surplus = 20
 
-/datum/uplink_item/device_tools/briefcase_launchpad
+/* /datum/uplink_item/device_tools/briefcase_launchpad // SKYRAT EDIT REMOVAL
 	name = "Briefcase Launchpad"
 	desc = "A briefcase containing a launchpad, a device able to teleport items and people to and from targets up to eight tiles away from the briefcase. \
 			Also includes a remote control, disguised as an ordinary folder. Touch the briefcase with the remote to link it."
 	surplus = 0
 	item = /obj/item/storage/briefcase/launchpad
-	cost = 6
+	cost = 6 */
 
 /datum/uplink_item/device_tools/camera_bug
 	name = "Camera Bug"
--- a/modular_skyrat/modules/moretraitoritems/code/syndicate_loadout.dm
+++ b/modular_skyrat/modules/moretraitoritems/code/syndicate_loadout.dm
@@ -30,7 +30,7 @@
 	new /obj/item/clothing/glasses/thermal/syndi(src)
 	new /obj/item/knife/combat/survival(src)
 	new /obj/item/gun/energy/disabler(src)
-	new /obj/item/storage/briefcase/launchpad(src)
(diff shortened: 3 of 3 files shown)


nevimer:
ok we're trying something else first maybe this will re open lets see weh

MassacreDemigirl:
I'd say keep it in the recon kit as its currently unachievable and also a big part (6tc) of it. Otherwise you beat me to it.

YakumoChen:
how is this controversial?
It's not even that good...

ArdentSpark:
I know this is a request, but I'm hesitant on this one as it's also psuedo affected by 10 starting TC. Antagonists are for the most part going to be heavily limited on TC for their starting objectives, and will struggle to afford tools to defend themselves from security. Yes, the bluespace launchpad annoys people, but it annoys them because it allows Antagonists to do objectives without entering open combat. Open combat is something that antagonists have been avoiding because of how oppressive the MCR has been. Given that both the MCR was retuned, and that starting TC will be going down, I feel it would be wise to put brakes on this one for a little bit and see how the Traitor situation ends up because of those two PRs before merging this one

Gandalf2k15:
Do not comment out items, just remove them from the places they are given.

Zonespace27:
I feel like this PR is cutting out the symptom, not the cause, and doesn't *really* solve the problem in a good way.

Weblure:
I really don't see how this would improve anything in the slightest. What do you personally have against stealth? Not everyone enjoys going around planting bombs and killing in the open.

Gandalf2k15:
not really sure why we'd need this, plus, non-modular change

nevimer:
> not really sure why we'd need this, plus, non-modular change

Xyel asked for it to be removed on discord. That's why it was made.

Weblure:
> > not really sure why we'd need this, plus, non-modular change
> 
> Xyel asked for it to be removed on discord. That's why it was made.

Xyel moment

Weblure:
@hugmeorelse 
It'd be useful if you could tell people what your issues with the bs launchpads are, because then we might have a chance of addressing those issues rather than performing a full-on, non-modular removal of a well-liked item.

Removing stealth options only encourages people to resort to emagging doors, or worse yet, blasting through walls, just to steal a non-vital item. The range is only 7 tiles and it only works in powered station areas, so it's not like you can use this to loot the armory or get into any deep access-restricted places.

It's clear the creator of the PR has no real thoughts on the matter, since they immediately diverted to "because Xyel wanted it" when asked.

nevimer:
> @hugmeorelse It'd be useful if you could tell people what your issues with the bs launchpads are, because then we might have a chance of addressing those issues rather than performing a full-on, non-modular removal of a well-liked item.
> 
> Removing stealth options only encourages people to resort to emagging doors, or worse yet, blasting through walls, just to steal a non-vital item. The range is only 7 tiles and it only works in powered station areas, so it's not like you can use this to loot the armory or get into any deep access-restricted places.
> 
> It's clear the creator of the PR has no real thoughts on the matter, since they immediately diverted to "because Xyel wanted it" when asked. 

Well I agree *something* should be done about them, as they're basically the number one purchase for tots looking to greentext steal goals. I don't see any airlock overrides at all anymore, despite them once being the go-to item. Maybe removing the purchasablity of the case is going too far, but something should be done to re-balance is as it's way too strong right now. I personally am not a fan of them since a tot teleported a syndicate bomb onto me while I was chilling in the medbay with one lol. They're just way too good.

Weblure:
> Well I agree something should be done about them, as they're basically the number one purchase for tots looking to greentext steal goals. I don't see any airlock overrides at all anymore, despite them once being the go-to item. Maybe removing the purchasablity of the case is going too far, but something should be done to re-balance is as it's way too strong right now. I personally am not a fan of them since a tot teleported a syndicate bomb onto me while I was chilling in the medbay with one lol. They're just way too good.

I mean, you can already do all that for free, and with 4.5x the range by just building a regular launchpad. I used to use one of those and a sec cam console all the time to just yeet stuff from halfway across the station. By the time anyone figured out where the launchpad is built, I'm long gone; the briefcase is much less stealthy compared to that. The only advantage of the briefcase is that it's portable and doesn't require sci/eng access... and that it hasn't be updated to include the sparking and blueline effect of a regular launchpad.

Honestly, just adding that last bit should be more than enough to nerf it. It'll give people plenty of time to run from any teleporting bombs, and since the range is only 7 tiles, you'll know *exactly* where they are once you see the blue teleport line.

hugmeorelse:
@Gandalf2k15 @GoldenAlpharex Because the metaprotection regarding this item is not being removed, please TM this PR ASAP. Thank you.

PR #11701, opened 2022-02-24
[NONMODULAR] D-e-sword nerf option: 1: the revenge of the blockchance.

RatFromTheJungle:
## About The Pull Request
It was a long time coming, but for an item that's both easy to use, easier to pair with stuff, 75 blockchance is just too high functionally, leading to it being one of the strongest items in the game, with most of it's counters like 'slipping' having just as affordable ways around it, all while we're supposed to be trying to discourage wanton combat.

Less people buying a D-E-Sword, and then casually walking around everyone, beating them down without effort while practically no one can reasonably fight back, unless by dumb-luck or sheer numbers.

## Changelog

:cl:
balance: Lowers block chance of double energy sword to 45, from 75.
/:cl:

### DIFF
--- a/code/game/objects/items/dualsaber.dm
+++ b/code/game/objects/items/dualsaber.dm
@@ -22,7 +22,7 @@
 	light_on = FALSE
 	attack_verb_continuous = list("attacks", "slashes", "stabs", "slices", "tears", "lacerates", "rips", "dices", "cuts")
 	attack_verb_simple = list("attack", "slash", "stab", "slice", "tear", "lacerate", "rip", "dice", "cut")
-	block_chance = 75
+	block_chance = 45 //SKYRAT EDIT - Lowered ORIGINAL:75
 	max_integrity = 200
 	armor = list(MELEE = 0, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 0, FIRE = 100, ACID = 70)
 	resistance_flags = FIRE_PROOF


YakumoChen:
45 is literally worse than a riot shield

Unironically you'd have a better time with a teleshield and a regular esword now than a double.

The main drawback of the double sword is that it takes both your hands and nearly all your TC to purchase (if it's op with 35 TC, raise the price, it should never exist in a state where you can buy noslips and glue with it)

ArdentSpark:
I think it's worth waiting to see how the traitor adjustments play out before nerfing this, considering Traitor economy and reputation are being adjusted.
Additionally Yakumo has good points, and we just discussed something in dev-code. Dual wielding e-swords apparantly should give you a 60 block chance, reflecting lasers which would be better than this DE-Sword change.

I believe either the DE-Sword block nerf should be raised to 60+ to compensate, or E-Swords will require a subsequent nerf, that reduces block chance while dual wielding them so that it falls in line with this PR, personal opinions aside.

RatFromTheJungle:
> I think it's worth waiting to see how the traitor adjustments play out before nerfing this, considering Traitor economy and reputation are being adjusted. Additionally Yakumo has good points, and we just discussed something in dev-code. Dual wielding e-swords apparantly should give you a 60 block chance, reflecting lasers which would be better than this DE-Sword change.
> 
> I believe either the DE-Sword block nerf should be raised to 60+ to compensate, or E-Swords will require a subsequent nerf, that reduces block chance while dual wielding them so that it falls in line with this PR, personal opinions aside.

it doesn't give you sixty block, because they use the same block code, so you only roll for one, not both, or everyone would already be running around with two eswords because they attack faster. The same reason you don't see people using twin energy shields, or twin riot shields. It's not only a shield, it's also a very-strong melee, it doesn't have to be great at both.

MassacreDemigirl:
With the 10+10 balance, a desword is already a huge investment. Not only it has a large rep requirement but also eats up more then 75% of the TC. If a traitor decides to get noslip+glue+desword they have at that point spent 23TC. If they are an opfor tot it means they probably have good justification for it and wont do wanton murder and had to only get 3TC so 1-2 objectives. If they are non opfor then they have to get a whopping 13TC which by the time they get to it, the crew should have reached proper power levels to deal with them. Also they have glued the desword and have permanently became visible

ArdentSpark:
> it doesn't give you sixty block, because they use the same block code, so you only roll for one, not both, or everyone would already be running around with two eswords because they attack faster. The same reason you don't see people using twin energy shields, or twin riot shields. It's not only a shield, it's also a very-strong melee, it doesn't have to be great at both.

Then I stand happily corrected. However I do believe that the economy adjustments should be given a grace period to come into effect before touching on things like this.

Aether217:
> If they are non opfor then they have to get a whopping 13TC which by the time they get to it, the crew should have reached proper power levels to deal with them.

We don't let crew prepare themselves for antags so they won't reach "proper power levels" as besides security nobody will be armed, and as security is focused on energy weapons the upgrades won't mean anything since the desword deflects those projectiles with 100% chance.

MassacreDemigirl:
Security can buy a crate of two wildcats for free every 20 minutes. Cargo has enough budget to arm everyone and their mother with WT or shotguns.

Aether217:
> Security can buy a crate of two wildcats for free every 20 minutes. Cargo has enough budget to arm everyone and their mother with WT or shotguns.

Shotguns don't do much, after messing around with it in the tdome, at point blank the test dummy consistently survived being pumped with buckshot from a fully loaded AS-1. 

The wildcat is RNG reliant but in tests where every single bullet hit at point blank it took around a full mag to kill them with no armour

Considering that these are ideal situations in which you do not lose damage due to distance, missed shots, the other party dodging or mitigation from the armour, the desword offers insane survivability due to it's very high blockchance and it gives very good offensive abilities due to it's good damage.

The only 2 cosistent options are either slip combat (ew) or the strongest weapon in the game that people ignore, the pepperball gun.

MassacreDemigirl:
Tempgun is consistent. So is body throwing. A well timed flashbang will do the job as well.

PR #11703, opened 2022-02-24
meat

Gandalf2k15:
## About The Pull Request

map meat wall floor no rcd

## Changelog
 Not player facing.

### DIFF
--- a/code/modules/mapping/turfs.dm
+++ b/code/modules/mapping/turfs.dm
@@ -5,5 +5,6 @@
 	custom_materials = list(/datum/material/meat = MINERAL_MATERIAL_AMOUNT * 2)
 
 /atom
+	/// Do we prevent the RCD from interacting with this atom regardless?
 	var/prevent_rcd_deconstruction = FALSE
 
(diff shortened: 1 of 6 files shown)


code/modules/mapping/turfs.dm
```diff
@@ -0,0 +1,9 @@
+/turf/open/floor/material/meat
+	custom_materials = list(/datum/material/meat = MINERAL_MATERIAL_AMOUNT * 0.25)
+
+/turf/closed/wall/material/meat
+	custom_materials = list(/datum/material/meat = MINERAL_MATERIAL_AMOUNT * 2)
+
+/atom
+	var/prevent_rcd_deconstruction = FALSE
```
GoldenAlpharex, line 41: Should be documented.

GoldenAlpharex: Also you don't seem to set it to true anywhere.

GoldenAlpharex (changes requested):
A descriptive changelog entry would be nice.

code/modules/mapping/turfs.dm
Gandalf2k15: It's for mapping, Duboris wants it for maps

Funce: He's shilling for sprites

Gandalf2k15: i'm shilling for you

PR #11704, opened 2022-02-24
opfor fix

Gandalf2k15:
## About The Pull Request

someone broke my opfor, so im fixing it

## Changelog

:cl:
fix: OPFOR no longer bluescreens.
/:cl:


GoldenAlpharex (commented):
That looks like it'd work. Could keep the variable being initialized as an empty string, probably.

Gandalf2k15:
> That looks like it'd work. Could keep the variable being initialized as an empty string, probably.

it's fine, we don't need to initialise that variable on em all, it makes no difference, TGUI can deal with it being null

PR #11707, opened 2022-02-24
[MODULAR] Overhauls how we handle taur sprites, adds a few pre-existing taur sprites to suits, and also a few new ones

nikothedude:
## About The Pull Request

This PR overhauls how we assign the mutant_variant flags for taurs to clothing.
Adds a new proc to clothing.dm that runs on init of clothing, and checks 3 global lists (3 for uniforms, 3 for suits) of icon_states that are built on startup.  These are both fairly lightweight and are unlikely to cause much performance decrease.

Profiller results at startup: 
/proc/init_taur_sprites                                                                            0.022        0.022        0.022        0.022            6
/obj/item/clothing/proc/handle_taur_sprites                                                        0.003        0.003        0.003        0.002         3077
/obj/item/clothing/suit/handle_taur_sprites                                                        0.002        0.002        0.002        0.002          529

These are only at startup, not in-game. Actual performance impact is likely to be VERY limited, even more than this, because thousands of clothing items aren't being generated in real time.

Adds a few taur sprites for clothing, most of which were pre-existing.

Requirement of #11754 

Fixes the damn invisible clothing bug. Also a better framework.

## Changelog

:cl:
code: Made taur sprites not be expected for all suits, and overhauls the way we assign mutant_variant flags for taurs (see clothing.dm)
fix: Fixed a few taur sprites, like aprons, firesuits, etc
add: New firesuit, berserker hardsuit, and HECKsuit sprites for taurs.
/:cl:


nikothedude:
The reason I decided to do it this way rather than keep it the way it is, is because it's frankly SUPER jarring to have your ENTIRE LOWER BODY VANISH whenever you wear soemthing

Is this a sloppy fix? Yes. Am I going to do more? Probably. Next thing up for me is a PR that'll delete any part of the sprite below a certain point, so you lose the legs and stuff.

There's also probably a much cleaner way to do this but I don't care enough to like, make another complex for loop that has a 4 line long if statement.

nikothedude:
Drafting because I realized there's a big flaw I fixed in my other PR. Moving it here and improving some code.

nikothedude:
image
please hold me to this

nikothedude:
> There's also probably a much cleaner way to do this but I don't care enough to like, make another complex for loop that has a 4 line long if statement.

i lied.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+		has_alt = TRUE
+		if (!(mutant_variants & STYLE_TAUR_HOOF))
+			mutant_variants |= STYLE_TAUR_HOOF
+	if (icon_state in GLOB.pawed_taur_uniform_sprites)
+		has_alt_sprite[icon_state] = HAS_ALT_SPRITE
+		has_alt_paw_sprite[icon_state] = HAS_ALT_SPRITE
+		has_alt = TRUE
+		if (!(mutant_variants & STYLE_TAUR_PAW))
+			mutant_variants |= STYLE_TAUR_PAW
+	if (!(has_alt)) // If none of the 3 above if statements are true, it has no alt sprite
```
Zonespace27:
```suggestion
	if (!has_alt) // If none of the 3 above if statements are true, it has no alt sprite
```

modular_skyrat/modules/customization/__DEFINES/alt_sprites.dm
```diff
@@ -0,0 +1,2 @@
+#define HAS_TAUR_SPRITE 1
```
Zonespace27: opinion: these shouldn't be in the customization module, they should be in the skyrat defines folder

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+/obj/item/clothing/proc/handle_taur_sprites() //niko-if i do not come back in a week and entirely refactor how this works gitban me
+	if (istype(src, /obj/item/clothing/suit))
+		handle_taur_sprites_for_suits()
+		return
+	if (!(istype(src, /obj/item/clothing/under))) //remove this if we ever make a taur sprite for anything else, this is for performance
+		return
+	if (mutant_variants & STYLE_TAUR_ALL)
+		return
+	if (has_taur_sprite[icon_state] == HAS_TAUR_SPRITE)
+		if ((has_taur_snake_sprite[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_SNAKE)))
```
Zonespace27:
```suggestion
		if (has_taur_snake_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_SNAKE)))
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+ */
+/obj/item/clothing/proc/handle_taur_sprites() //niko-if i do not come back in a week and entirely refactor how this works gitban me
+	if (istype(src, /obj/item/clothing/suit))
+		handle_taur_sprites_for_suits()
+		return
+	if (!(istype(src, /obj/item/clothing/under))) //remove this if we ever make a taur sprite for anything else, this is for performance
+		return
+	if (mutant_variants & STYLE_TAUR_ALL)
+		return
+	if (has_taur_sprite[icon_state] == HAS_TAUR_SPRITE)
```
Zonespace27:
```suggestion
	if (has_taur_sprite[icon_state])
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+		if ((has_taur_snake_sprite[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_SNAKE)))
+			mutant_variants |= STYLE_TAUR_SNAKE
+		if ((has_taur_horse_sprite[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_HOOF)))
```
Zonespace27:
```suggestion
		if (has_taur_horse_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_HOOF)))
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+			mutant_variants |= STYLE_TAUR_HOOF
+		if ((has_taur_paw_sprite[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_PAW)))
```
Zonespace27:
```suggestion
		if (has_taur_paw_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_PAW)))
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+			mutant_variants |= STYLE_TAUR_PAW
+		return // If we already know this icon state has a taur sprite, skip the for loop and take from the cache
+	else if (has_taur_sprite[icon_state] == HAS_NO_TAUR_SPRITE)
```
Zonespace27: leaving these as `(has_taur_sprite[icon_state] == HAS_NO_TAUR_SPRITE)` instead of converting to `(!(has_taur_sprite[icon_state] == HAS_NO_TAUR_SPRITE)` is preferred since the latter will pass if `icon_state` is null.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+ *
+ * Finally, if all of the conditions are false, has_taur_sprite_suit for that icon state will be set to false.
+ *
+ * DO NOT RELY ON THIS PROC TO DO YOUR WORK FOR YOU. If you are adding a new item, or a sprite, you SHOULD be manually setting your mutant_variants flags, otherwise the for
+ * loop will be ran and cause slight performance loss.
+ */
+/obj/item/clothing/proc/handle_taur_sprites_for_suits()
+	if (mutant_variants & STYLE_TAUR_ALL)
+		return
+	if (has_taur_sprite_suit[icon_state] == HAS_TAUR_SPRITE)
```
Zonespace27:
```suggestion
	if (has_taur_sprite_suit[icon_state])
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+		if ((has_taur_snake_sprite_suit[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_SNAKE)))
```
Zonespace27:
```suggestion
		if (has_taur_snake_sprite_suit[icon_state] && (!(mutant_variants & STYLE_TAUR_SNAKE)))
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+			mutant_variants |= STYLE_TAUR_SNAKE
+		if ((has_taur_horse_sprite_suit[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_HOOF)))
```
Zonespace27:
```suggestion
		if (has_taur_horse_sprite_suit[icon_state] && (!(mutant_variants & STYLE_TAUR_HOOF)))
```

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,129 @@
...
+			mutant_variants |= STYLE_TAUR_HOOF
+		if ((has_taur_paw_sprite_suit[icon_state] == HAS_TAUR_SPRITE) && (!(mutant_variants & STYLE_TAUR_PAW)))
```
Zonespace27:
```suggestion
		if (has_taur_paw_sprite_suit[icon_state] && (!(mutant_variants & STYLE_TAUR_PAW)))
```

nikothedude:
aaalright, I think Im done with this PR

GoldenAlpharex:
This isn't stale, I just need to get some time to review it.

code/__DEFINES/~skyrat_defines/inventory.dm
```diff
@@ -4,6 +4,7 @@
 #define STYLE_TAUR_SNAKE (1<<2) //taur-friendly suits
 #define STYLE_TAUR_PAW (1<<3)
 #define STYLE_TAUR_HOOF (1<<4)
-#define STYLE_TAUR_ALL (STYLE_TAUR_SNAKE|STYLE_TAUR_PAW|STYLE_TAUR_HOOF)
+#define STYLE_TAUR_ALL (STYLE_TAUR_SNAKE | STYLE_TAUR_PAW | STYLE_TAUR_HOOF)
+#define STYLE_TAUR (STYLE_TAUR_SNAKE & STYLE_TAUR_PAW & STYLE_TAUR_HOOF)
```
GoldenAlpharex: Move that down from the bitfield, and it might be wise to make the first one more descriptive, or add some AUTODOC documentation to it.

code/__DEFINES/~skyrat_defines/taur_sprite_defines.dm
```diff
@@ -0,0 +1,2 @@
+#define HAS_TAUR_SPRITE 1
+#define HAS_NO_TAUR_SPRITE 0
```
GoldenAlpharex: I'm not quite sure why you need two states for this, just use `!HAS_TAUR_SPRITES` if it doesn't have taur sprites, or something.
```suggestion
#define HAS_TAUR_SPRITE TRUE
#define HAS_NO_TAUR_SPRITE FALSE
```

modular_skyrat/modules/customization/_globalvars/lists.dm
```diff
@@ -70,3 +70,17 @@ GLOBAL_LIST_INIT(color_list_ghoul, list( \
...
+/// Generates a list of pawed taur uniform sprites in text form.
+GLOBAL_LIST_INIT(horse_taur_uniform_sprites, icon_states('modular_skyrat/master_files/icons/mob/clothing/uniform_taur_hoof.dmi'))
+/// Generates a list of hoofed taur uniform sprites in text form.
+GLOBAL_LIST_INIT(pawed_taur_uniform_sprites, icon_states('modular_skyrat/master_files/icons/mob/clothing/uniform_taur_paw.dmi'))
+/// Generates a list of naga taur suit sprites in text form.
+GLOBAL_LIST_INIT(naga_taur_suit_sprites, icon_states('modular_skyrat/master_files/icons/mob/clothing/suit_taur_snake.dmi'))
+/// Generates a list of pawed taur suit sprites in text form.
+GLOBAL_LIST_INIT(horse_taur_suit_sprites, icon_states('modular_skyrat/master_files/icons/mob/clothing/suit_taur_hoof.dmi'))
+/// Generates a list of hoofed taur suit sprites in text form.
+GLOBAL_LIST_INIT(pawed_taur_suit_sprites, icon_states('modular_skyrat/master_files/icons/mob/clothing/suit_taur_paw.dmi'))
```
GoldenAlpharex: You should say icon_states, instead of "sprites in text form".

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,130 @@
...
 
+	var/static/has_taur_sprite = list()
+	var/static/has_taur_snake_sprite = list()
+	var/static/has_taur_paw_sprite = list()
+	var/static/has_taur_horse_sprite = list()
+
+	var/static/has_taur_sprite_suit = list()
+	var/static/has_taur_snake_sprite_suit = list()
+	var/static/has_taur_paw_sprite_suit = list()
+	var/static/has_taur_horse_sprite_suit = list()
```
GoldenAlpharex: It's not casted as a list, and why is a list named like a boolean?

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,130 @@
...
+ * true for that icon state, then it sets the more specific list to true for that state, and finally the specific mutant_variants bitflag for the clothing if the
+ * clothing doesnt already have it.
+ *
+ * Finally, if all of the conditions are false, has_taur_sprite for that icon state will be set to false.
+ *
+ * DO NOT RELY ON THIS PROC TO DO YOUR WORK FOR YOU. If you are adding a new item, or a sprite, you SHOULD be manually setting your mutant_variants flags, otherwise the for
+ * loop may be ran and cause slight performance loss.
+ */
+/obj/item/clothing/proc/handle_taur_sprites() //niko-if i do not come back in a week and entirely refactor how this works gitban me
+	if (istype(src, /obj/item/clothing/suit))
```
GoldenAlpharex: It's been over a week Niko.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,130 @@
...
+		handle_taur_sprites_for_suits()
+		return
+	if (!(istype(src, /obj/item/clothing/under))) //remove this if we ever make a taur sprite for anything else, this is for performance
+		return
+	if (mutant_variants & STYLE_TAUR)
+		return
+	if (has_taur_sprite[icon_state])
```
GoldenAlpharex: Add an empty newline after each return, please. Or before your if statements, I guess would be more accurate.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
GoldenAlpharex: This could probably just be FALSE instead of `HAS_NO_TAUR_SPRITE`, this doesn't really need to be a define, it's clear enough.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,130 @@
...
+ * the icon state EXPLICITELY has no taur sprites. If no, it then checks 3 DMI files for the icon_state, and if it appears in any of them, has_taur_sprite_suit is set to
+ * true for that icon state, then it sets the more specific list to true for that state, and finally the specific mutant_variants bitflag for the clothing if the
+ * clothing doesnt already have it.
+ *
+ * Finally, if all of the conditions are false, has_taur_sprite_suit for that icon state will be set to false.
+ *
+ * DO NOT RELY ON THIS PROC TO DO YOUR WORK FOR YOU. If you are adding a new item, or a sprite, you SHOULD be manually setting your mutant_variants flags, otherwise the for
+ * loop may be ran and cause slight performance loss.
+ */
+/obj/item/clothing/proc/handle_taur_sprites_for_suits()
```
GoldenAlpharex: I'm really curious to see how much of an effect they have on performance, I feel like it won't be great.

GoldenAlpharex (changes requested):

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Great as in, it won't be too taxing or it WILL be taxing?

code/__DEFINES/~skyrat_defines/taur_sprite_defines.dm
nikothedude: Good point.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: uhhhhhhhhhhhhhhhhhhhhhhhh

*cough

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Its technically a list of booleans. The only possible values are 0 and 1.

nikothedude:
I didn't put newlines after EVERY if statement, because when I did that it looked awful.

nikothedude:
may potentially come back later and ACTUALLY compress the lists, functionally. but for now? im full on fuckin' pizza

nikothedude:
btw each "future" change i might do i'd probably do in a future pr if this one is merged, so dont let that stop you

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: An ASSOCIATIVE list of booleans, excuse me.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+ * Finally, if all of the conditions are false, has_taur_sprite for that icon state will be set to false.
+ *
+ * DO NOT RELY ON THIS PROC TO DO YOUR WORK FOR YOU. If you are adding a new item, or a sprite, you SHOULD be manually setting your mutant_variants flags, otherwise the for
+ * loop may be ran and cause slight performance loss.
+ */
+/obj/item/clothing/proc/handle_taur_sprites()
+
+	if (istype(src, /obj/item/clothing/suit))
+		handle_taur_sprites_for_suits()
+		return
```
LemonInTheDark: This should just be an override of `handle_taur_sprites` but for suits

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+		if (has_taur_snake_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_SNAKE)))
+			mutant_variants |= STYLE_TAUR_SNAKE
+		if (has_taur_horse_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_HOOF)))
+			mutant_variants |= STYLE_TAUR_HOOF
+		if (has_taur_paw_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_PAW)))
+			mutant_variants |= STYLE_TAUR_PAW
+		return // If we already know this icon state has a taur sprite, skip the for loop and take from the cache
+
+	else if (has_taur_sprite[icon_state] == FALSE)
+		return
```
LemonInTheDark: you don't need the rest of this check, just use else

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+
+	if (has_taur_sprite[icon_state])
+		if (has_taur_snake_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_SNAKE)))
+			mutant_variants |= STYLE_TAUR_SNAKE
+		if (has_taur_horse_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_HOOF)))
+			mutant_variants |= STYLE_TAUR_HOOF
+		if (has_taur_paw_sprite[icon_state] && (!(mutant_variants & STYLE_TAUR_PAW)))
+			mutant_variants |= STYLE_TAUR_PAW
+		return // If we already know this icon state has a taur sprite, skip the for loop and take from the cache
```
LemonInTheDark: Setting a bitflag is just as expensive as checking to see if it's set, you don't need to check if it's already set before setting it

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+	else if (has_taur_sprite[icon_state] == FALSE)
+		return
+	// Code only goes here if has_taur_sprite[] == nothing, AKA if init has never been ran
+	var/taur_sprite = FALSE
+
+	if (icon_state in GLOB.naga_taur_uniform_sprites)
+		has_taur_sprite[icon_state] = TRUE
+		has_taur_snake_sprite[icon_state] = TRUE
+		taur_sprite = TRUE
+		if (!(mutant_variants & STYLE_TAUR_SNAKE))
```
LemonInTheDark: ditto

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+	var/static/list/has_taur_horse_sprite = list()
+
+	/// has_taur_sprite_suit[icon_state]: has_taur_sprite but for /obj/item/clothng/suit
+	var/static/list/has_taur_sprite_suit = list()
+	/// has_taur_snake_sprite_suit[icon_state]: has_taur_snake_sprite but for /obj/item/clothing/suit
+	var/static/list/has_taur_snake_sprite_suit = list()
+	/// has_taur_paw_sprite_suit[icon_state]: has_taur_paw_sprite but for /obj/item/clothing/suit
+	var/static/list/has_taur_paw_sprite_suit = list()
+	/// has_taur_horse_sprite_suit[icon_state]: has_taur_horse_sprite but for /obj/item/clothing/suit
+	var/static/list/has_taur_horse_sprite_suit = list()
```
LemonInTheDark: Why do these lists need to exist? if you want fast lookup, couldn't you just make the sprites lists assoc = TRUE?

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,154 @@
...
+
+	if (icon_state in GLOB.pawed_taur_uniform_sprites)
+		has_taur_sprite[icon_state] = TRUE
+		has_taur_paw_sprite[icon_state] = TRUE
+		taur_sprite = TRUE
+		if (!(mutant_variants & STYLE_TAUR_PAW))
+			mutant_variants |= STYLE_TAUR_PAW
+
+	if (!taur_sprite) // If none of the 3 above if statements are true, it has no alt sprite
+		has_taur_sprite[icon_state] = FALSE
```
LemonInTheDark: Would save all this headache

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: But if I don't check to see if it already has the bitflag, won't it just be applied twice? Doesn't that have consequences?

LemonInTheDark:
Sorry uh, niko was fretting about the performance of his code in tg coderbus, and I got a bit carried away

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: nope, they're just binary 1s in a number
the | operator says set the output to the 1s from either value. It's binary OR
so 001 |  010 becomes 011. 
the nice thing about | is it doesn't add, it's a bit operation.
So 001 | 001 is just 001, since while they share the same bit in the same position, the | operator doesn't really care

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: by this I mean:
The `has_taur_*_sprite` lists are just the icon states in the global lists, but assoc to true instead.
Why not just generate those global lists to be assoc to true, and access them if you want to check if a sprite is valid or not.

Assoc lookup ought (I hope) to be constant time (s not like these lists are huge), and it would save a lot of cruft and duplicate code that you're doing here.

nikothedude:
lmao, don't worry, i appreciate it and i'm sure the maintainers do too

nikothedude:
/obj/item/clothing/proc/handle_taur_sprites                                                        0.009        0.009        0.010        0.005         3523
/proc/init_taur_sprites                                                                            0.022        0.022        0.022        0.022            6

profiler results, this is at startup

this seems fairly lightweight to me

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Keeping it out of paranoia, there is a non-zero chance of something fucking up somewhere and somehow assigning a value that isnt 0 or null

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,110 @@
...
+ * the icon state EXPLICITELY has no taur sprites. If no, it then checks 3 DMI files for the icon_state, and if it appears in any of them, has_taur_sprite is set to
+ * true for that icon state, then it sets the more specific list to true for that state, and finally the specific mutant_variants bitflag for the clothing.
+ *
+ * Finally, if all of the conditions are false, has_taur_sprite for that icon state will be set to false.
+ *
+ * Has seperate logic for suits, as they use a different set of DMIs as of now.
+ */
+/obj/item/clothing/proc/handle_taur_sprites()
+
+	if (istype(src, /obj/item/clothing/suit)) //override for suits, which use different dmis
```
LemonInTheDark: When I say override, I mean like, move it to `/obj/item/clothing/suit/handle_taur_sprites()`, and don't return there, rather then keep it in this proc, or in a proc with a new name

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,110 @@
...
+		if (has_taur_sprite_suit[icon_state]) // Very slightly speeds up the proc by letting things with no sprites instantly return
+			if (GLOB.naga_taur_suit_sprites[icon_state])
+				mutant_variants |= STYLE_TAUR_SNAKE
+			if (GLOB.horse_taur_suit_sprites[icon_state])
+				mutant_variants |= STYLE_TAUR_HOOF
+			if (GLOB.pawed_taur_suit_sprites[icon_state])
+				mutant_variants |= STYLE_TAUR_PAW
+			return
+
+		else if (has_taur_sprite_suit[icon_state] == FALSE)
```
LemonInTheDark: don't need this, can just be an else

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,110 @@
...
+			mutant_variants |= STYLE_TAUR_HOOF
+
+		if (GLOB.pawed_taur_suit_sprites[icon_state])
+			has_taur_sprite_suit[icon_state] = TRUE
+			taur_sprite_suit = TRUE
+			mutant_variants |= STYLE_TAUR_PAW
+
+		if (!taur_sprite_suit) // If none of the 3 above if statements are true, it has no alt sprite
+			has_taur_sprite_suit[icon_state] = FALSE
+			return
```
LemonInTheDark: You should never need this, global init ought to be run before anything else

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,110 @@
...
+		taur_sprite = TRUE
+		mutant_variants |= STYLE_TAUR_HOOF
+
+	if (GLOB.pawed_taur_uniform_sprites[icon_state])
+		has_taur_sprite[icon_state] = TRUE
+		taur_sprite = TRUE
+		mutant_variants |= STYLE_TAUR_PAW
+
+	if (!taur_sprite) // If none of the 3 above if statements are true, it has no alt sprite
+		has_taur_sprite[icon_state] = FALSE
```
LemonInTheDark: Ditto

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: What do you mean? This IS ran after global init, this is just how I'm checking to see if the icon_state has a taur sprite.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: eh, sure

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: OOOOOOH. OKAY.

nikothedude:
okay it seems im not done

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Oh. Yeah, this is basically just kind of redundant right now, it's a caching system so we can skip the lists if the clothing has no taur sprites. I'm debating removing it.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: this causes a possibly unreachable code error, i'll keep it as it is

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: Should have no reason to exist right?

nikothedude: I honestly think there's no real reason to remove it, honestly, it's probably improving performance, even if VERY slight.

LemonInTheDark: when would it ever run? isn't it just dead code?

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: how? the caching?

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,110 @@
 	var/greyscale_config_worn_taur_paw
 	var/greyscale_config_worn_taur_hoof
 
+	/// has_taur_sprite[icon_state]: Does this icon_state have ANY taur sprite?
+	var/static/list/has_taur_sprite = list()
+
+	/// has_taur_sprite_suit[icon_state]: has_taur_sprite but for /obj/item/clothng/suit
+	var/static/list/has_taur_sprite_suit = list()
```
LemonInTheDark: I don't think you need these at all

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Yep, DM thinks it can never reach that if it's an else.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: > when would it ever run? isn't it just dead code?

I'm not sure what you mean, this runs on the first init of any clothing item, as long as the icon_state hasn't been used in this proc before. After that first init, the cacher is never ran again, but the cache is accessed on every successive init, for the purpose of immediately returning if there is no taur sprite.

It is a bit complicated for only a small performance increase, but again, I already wrote it, so

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Could probably simplify that, yeah.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Specifically, the error appears on var/taur_sprite = FALSE

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Actually, if I want to keep the caching system intact, I _would_ need these. Come to think of it, I'm not sure whatever tiny performance gain the cache would give would be worth having two redundant lists filled with every single icon_state in the game.

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: What are you caching exactly? You already have a list of all valid icon states in those globals, what work needs to be cached?

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,109 @@
...
+	if (mutant_variants & STYLE_TAUR)
+		return
+
+	if (has_taur_sprite[icon_state])
+		if (GLOB.naga_taur_uniform_sprites[icon_state])
+			mutant_variants |= STYLE_TAUR_SNAKE
+		if (GLOB.horse_taur_uniform_sprites[icon_state])
+			mutant_variants |= STYLE_TAUR_HOOF
+		if (GLOB.pawed_taur_uniform_sprites[icon_state])
+			mutant_variants |= STYLE_TAUR_PAW
```
LemonInTheDark: Oh I see you're trying to make this faster.
This isn't expensive enough to be a problem, assoc list lookups are quite fast. way more then you should need here.
Especially since has_taur_sprite will be much larger then any of these other lists ever could be

If you really want to avoid all 4 lookups, you can just do the same thing you did to the other static lists to these, prefill them with appropriate contents once. Profile before you do that though, I doubt it'll make a significant difference

modular_skyrat/modules/customization/modules/clothing/clothing.dm
LemonInTheDark: The other thing to note is even though it's already written, it requires upkeep just by being in the codebase. Most of the work that's done on code is done after it's been merged in

modular_skyrat/modules/customization/modules/clothing/clothing.dm
nikothedude: Without the caching system, the code's significantly simpler and has very similar performance (/obj/item/clothing/proc/handle_taur_sprites                                                        0.004        0.004        0.004        0.003         2988) so I think I'll get rid of it

modular_skyrat/modules/customization/modules/clothing/clothing.dm
```diff
@@ -19,6 +19,49 @@
...
+
+	if (mutant_variants & STYLE_TAUR)
+		return
+
+	if (GLOB.naga_taur_suit_sprites[icon_state])
+		mutant_variants |= STYLE_TAUR_SNAKE
+	if (GLOB.horse_taur_suit_sprites[icon_state])
+		mutant_variants |= STYLE_TAUR_HOOF
+	if (GLOB.pawed_taur_suit_sprites[icon_state])
+		mutant_variants |= STYLE_TAUR_PAW
```
LemonInTheDark, line 65: One final thing.
You have two sets of lists here, each representing the icon states in a dmi.
Is there any reason you couldn't just have one set of lists, where each list represents two dmis?
Then this override would have no reason to exist

nikothedude: Mainly because I've already told golden about 5 times that they are finally free to review it and I don't want to make any more changes that might need a new review.

But yeah theres no real reason. I might fix this later, in another pr.

nikothedude: (or maybe just in this pr who the hell knows)

nikothedude:
ugh

nikothedude:
okay making a new pr

modular_skyrat/modules/customization/modules/clothing/clothing.dm
GoldenAlpharex: I'm almost done reviewing everything, go ahead and make that change, and I think we'd be good for at least a test-merge.

GoldenAlpharex (changes requested):

nikothedude:
It's occured to me, and golden, in dev channels, that this change would break things in a scenario where a icon_state an occurance in one file but not the other, meaning if a jumpsuit has "white" as the icon_state, but theres only a "white" in a suit DMI, it'll cause it to break.

nikothedude:
I have no fucking idea of how I'll fix accessories being misaligned while wearing non-taur suits.

nikothedude:
Yeah, I know the last commits could've been done better, but burnout's starting to hit me hard and I just want to be done with this

modular_skyrat/master_files/code/modules/clothing/non_anthro_clothes.dm
```diff
@@ -274,9 +274,6 @@
 /obj/item/clothing/suit/ianshirt
 	mutant_variants = NONE
 
-/obj/item/clothing/suit/poncho
```
AtoriBirb, line 277: This part is breaking the poncho for anthros with digi legs currently.

AtoriBirb:
The changes to [modular_skyrat/master_files/code/modules/clothing/non_anthro_clothes.dm](https://github.com/Skyrat-SS13/Skyrat-tg/pull/11707/files#diff-a4c051df720646a3bc0dccc22bd2dede153fc727d1b9b44f1fd76e05521d1488) are currently breaking the Poncho for anthros with digi legs.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
```diff
@@ -136,12 +145,34 @@
...
+					U.has_overlays_shifted_due_to_taur = FALSE
+
 		overlays_standing[SUIT_LAYER] = wear_suit.build_worn_icon(default_layer = SUIT_LAYER, default_icon_file = 'icons/mob/clothing/suit.dmi', override_icon = icon_file, override_x_center = x_override, mutant_styles = applied_style, species = dna.species.species_clothing_path)
 		var/mutable_appearance/suit_overlay = overlays_standing[SUIT_LAYER]
 		if(OFFSET_SUIT in dna.species.offset_features)
 			suit_overlay.pixel_x += dna.species.offset_features[OFFSET_SUIT][1]
 			suit_overlay.pixel_y += dna.species.offset_features[OFFSET_SUIT][2]
 		overlays_standing[SUIT_LAYER] = suit_overlay
+
+	// Only fired if the suit is being taken off, shifts accessory overlays back to normal if this suit didnt allow for taur sprites but the jumpsuti did
```
GoldenAlpharex:
```suggestion
	// Only fired if the suit is being taken off, shifts accessory overlays back to normal if this suit didnt allow for taur sprites but the jumpsuit did
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
```diff
@@ -136,12 +145,34 @@
...
+					if (!(U.has_overlays_shifted_due_to_taur) && (U.accessory_overlay))
+						U.accessory_overlay.pixel_x += 16  //If both the suit and uniform support taur sprites but the uniform has not shifted the overlay,
+						U.has_overlays_shifted_due_to_taur = TRUE // shift it.
+
+		else if (istype(w_uniform, /obj/item/clothing/under))
+			var/obj/item/clothing/under/U = w_uniform
+			if (U.applied_style_store & STYLE_TAUR_ALL) // If this suit doesnt support taur sprites but the uniform does (and has the overlay shifted to accomodate),
+				if (U.has_overlays_shifted_due_to_taur) // reset it to normal until we take the suit off
+					U.accessory_overlay.pixel_x -= 16
+					U.has_overlays_shifted_due_to_taur = FALSE
```
GoldenAlpharex:
```suggestion
			var/obj/item/clothing/under/uniform = w_uniform
			if (uniform.applied_style_store & STYLE_TAUR_ALL) // If this suit doesnt support taur sprites but the uniform does (and has the overlay shifted to accomodate),
				if (uniform.has_overlays_shifted_due_to_taur) // reset it to normal until we take the suit off
					uniform.accessory_overlay.pixel_x -= 16
					uniform.has_overlays_shifted_due_to_taur = FALSE
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/human_update_icons.dm
```diff
@@ -136,12 +145,34 @@
...
 			suit_overlay.pixel_x += dna.species.offset_features[OFFSET_SUIT][1]
 			suit_overlay.pixel_y += dna.species.offset_features[OFFSET_SUIT][2]
 		overlays_standing[SUIT_LAYER] = suit_overlay
+
+	// Only fired if the suit is being taken off, shifts accessory overlays back to normal if this suit didnt allow for taur sprites but the jumpsuti did
+	else if (istype(w_uniform, /obj/item/clothing/under))
+		var/obj/item/clothing/under/U = w_uniform
+		if ((U.applied_style_store & STYLE_TAUR_ALL) && (U.accessory_overlay) && !(U.has_overlays_shifted_due_to_taur))
+			U.accessory_overlay.pixel_x += 16
+			U.has_overlays_shifted_due_to_taur = TRUE
```
GoldenAlpharex:
```suggestion
		var/obj/item/clothing/under/uniform = w_uniform
		if ((uniform.applied_style_store & STYLE_TAUR_ALL) && (uniform.accessory_overlay) && !(uniform.has_overlays_shifted_due_to_taur))
			uniform.accessory_overlay.pixel_x += 16
			uniform.has_overlays_shifted_due_to_taur = TRUE
```

GoldenAlpharex (changes requested):

nikothedude:
Okay so basically kapulimbs is changing how all this works and uhhh no thanks. Will continue this after its merged.

PR #11790, opened 2022-02-28
[Non-modular] Allows (Health-)scanners to be used by blind players

desertmote:
## About The Pull Request

Related to our blindness overhaul.

Just squint your eyes really hard.

## Changelog

:cl:
qol: Allows (Health-)scanners to be used by blind players
/:cl:

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -188,9 +188,9 @@ GENE SCANNER
 	if(user.incapacitated())
 		return
 
-	if(user.is_blind())
+/*	if(user.is_blind())	SKYRAT EDIT - Blindness overhaul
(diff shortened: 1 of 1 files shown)


SpaceVampire:
What is the blindness overhaul? I don't see any kind of design document about it.
This is more of a balance change then a QOL change as well. Given that it is clearly an intended feature that comes with being blind. Because you know. Can't look at the screen, it isn't made out of braille.
Don't think you have very good logic for doing this.

ArdentSpark:
I really don't think this is that consequential. I don't think it's that unreasonable to justify it via accessibility audio reading, which is a current technology, let alone one in the age of bluespace

I'm not terribly sure how significant it is of a balance concern either

PR #11697, opened 2022-02-23
The Impact Soundening

Gandalf2k15:
## About The Pull Request

This is part of my love for sound program.

Basically adds a whole variety of hit sounds to the game and is dependent on the material it is made out of.

# SOUND ON

https://user-images.githubusercontent.com/9026500/155427854-8e9c07da-4b57-48c6-9fab-c5f0db877dcb.mp4

(no cats were harmed in the making of this video)

Before soundscaping:

https://user-images.githubusercontent.com/9026500/155603059-201cb3d5-b76a-475d-bb37-eb7856633b76.mp4

After soundscaping:

https://user-images.githubusercontent.com/9026500/155603103-fd42c73b-0870-4baf-adfb-c1d6f2cc20ff.mp4

## Changelog

:cl:
add: Different objects now have different bullet impact sounds.
/:cl:

PR #11793, opened 2022-02-28
[DNM] [TM-Only] Revert "The Impact Soundening"

GoldenAlpharex:
Reverts Skyrat-SS13/Skyrat-tg#11697

At Gandalf's request, he apparently merged it on accident with a logic error that makes every sound the same (whoops!), will get closed whenever he does the fix.


Gandalf2k15:
fiux

PR #11290, opened 2022-02-05
[MIRROR] Tgui input refresh [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/64331
--------------------

## About The Pull Request
This PR addresses a number of issues surrounding tgui inputs (some I caused, some that existed prior- assign gbp however you like)

- Fixes #63777 - the new component autofocuses, auto selects.
- Fixes issues that existed with tgui alerts. Rewritten ground up, the window now responds to user preferences and no longer clips messages.
- Sets all occurrences of number inputs to use a strict sanity check.  
- Fixes #63629 .... which is my mistake entirely

## Why It's Good For The Game
TGUI number input is now much more like the byond numberinput. It's entirely less annoying now that it autofocuses
Adds in some code improvements
Covers more error cases
TGUI inputs are now all consistent with user prefs handling

## Changelog

:cl:
fix: Tgui number input now uses a new component that makes it easier to use. It now autofocuses and autoselects, just like the original. 
fix: Alert modals have been rewritten. Messages no longer clip, they also now respond to TGUI input preferences (UI tab) and you can press ESC to close.
fix: You should now be able to enter a blank text to succumb.
/:cl:

PR #10355, opened 2021-12-30
[MIRROR] TGUI list conversions + bug fixes [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63354
--------------------

## About The Pull Request
- Converts more inputs to TGUI. Possibly all user-facing input lists in the game.
- Did any surrounding text/number inputs as well
- Added null choice support so users can press cancel.
- Added some misc TGUI input fixes
- Fixed custom vendors while I was there

I refactored a lot of code while just poking around.
Primarily, usage of .len in files where I was already working on lists.
Some code was just awful - look at guardian.dm and its non use of early returns
If there are any disputes, I can revert it just fine, those changes are not integral to the PR.

## Why It's Good For The Game
Fixes #63629
Fixes #63307
Fixes custom vendors /again/
Text input is more performant.
Part of a long series of TGUI conversion to make the game more visually appealing

## Changelog

:cl:
refactor: The majority of user facing input lists have been converted to TGUI.
refactor: Tgui text inputs now scale with entered input.
fix: Many inputs now properly accept cancelling out of the menu.
fix: Fixes an edge case where users could not press enter on number inputs.
fix: Custom vendor bluescreen.
fix: You can now press ENTER on text inputs without an entry to cancel.
/:cl:

PR #11808, opened 2022-03-01
Fixes a bug with native inputs.

nevimer:
## About The Pull Request

This looks like one of these intentional bugs people leave lying around to get free GBP later, it's that silly. Anyways, this passes the default value back to input() which no longer presents NULL now. As a result, colour picking no longer gives you #000000, suit sensors no longer default to off, etc. Works as expected.

Fixes https://github.com/Skyrat-SS13/Skyrat-tg/pull/11290
and https://github.com/Skyrat-SS13/Skyrat-tg/pull/10355

from: https://github.com/tgstation/tgstation/pull/63354

Not having bugs is nice. Especially ones that seem intentional. I don't like the bloated library monster known as TGUI, so the less I need to depend on it, the better. 

## Changelog

:cl:

fix: fixed native windows input boxes
/:cl:

### DIFF
--- a/code/modules/tgui/tgui_input_list.dm
+++ b/code/modules/tgui/tgui_input_list.dm
@@ -23,7 +23,7 @@
 			return
 	/// Client does NOT have tgui_input on: Returns regular input
 	if(!user.client.prefs.read_preference(/datum/preference/toggle/tgui_input))
-		return input(user, message, title) as null|anything in items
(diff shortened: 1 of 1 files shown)


nikothedude:
cant this be done upstream

Gandalf2k15:
upstream it

nevimer:
it is upstream but stale

PR #11937, opened 2022-03-07
[NON-MODULAR] Removes Purge and Block Malf AI objectives

Zonespace27:
## About The Pull Request
Removes the following objectives:

- Do not allow any organic lifeforms with sapience to escape on the shuttle alive.
- Ensure no nonhuman humanoid species with sapience are present aboard the escape shuttle.

AI, malf or not, shouldn't be murderboning and these objectives don't fit into what Skyrat is aiming for

## Changelog

:cl:
del: Removed Block and Purge Malf AI objectives
/:cl:

### DIFF
--- a/code/modules/antagonists/malf_ai/datum_malf_ai.dm
+++ b/code/modules/antagonists/malf_ai/datum_malf_ai.dm
@@ -89,23 +89,21 @@
 
 /// Generates a special objective and adds it to the objective list.
 /datum/antagonist/malf_ai/proc/forge_special_objective()
-	// SKYRAT EDIT START
-	var/special_pick = rand(1,2)
+	var/special_pick = rand(3,4) // SKYRAT EDIT - REMOVING PURGE/BLOCK
 	switch(special_pick)
-		/*if(1)
+		if(1)
 			var/datum/objective/block/block_objective = new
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/malf_ai/datum_malf_ai.dm
```diff
@@ -89,21 +89,23 @@
 
 /// Generates a special objective and adds it to the objective list.
 /datum/antagonist/malf_ai/proc/forge_special_objective()
-	var/special_pick = rand(1,4)
+	// SKYRAT EDIT START
+	var/special_pick = rand(1,2)
```
GoldenAlpharex: Just change it into a `rand(3, 4)` instead, that ends up being a one-line change.

GoldenAlpharex (changes requested):
Someone please stop this man he's on fire.

code/modules/antagonists/malf_ai/datum_malf_ai.dm
Zonespace27: You know, this is one of those things that _feels_ like it should've been the obvious solution at the time, but this is what we have code review for

GoldenAlpharex (approved):

PR #12010, opened 2022-03-11
[MODULAR] Fixes wildcat rubbers bleeding.

nikothedude:
## About The Pull Request

Adds 3 vars to the bullets taht shouldve been on there.

police brutality is bad.

## Changelog

:cl:
fix: Wildcat rubbers now don't cause bleeding.
/:cl:


KathrinBailey (approved):
more of a balance than a fix given they're not supposed to be purely stamina and are supposed to still cause hurties, but I'll allow it anyway.

PR #12052, opened 2022-03-13
Combat Overhaul - Slippery component

Gandalf2k15:

As specified in part A of the combat overhaul, this removes the knockdown and item drop of any slip. 

Naturally, because of modularisation, this is done in one file.

## Changelog

:cl:
balance: Combat overhaul part A: Slipping no longer causes paralyze or item drop.
/:cl:

### DIFF
--- a/code/datums/components/slippery.dm
+++ b/code/datums/components/slippery.dm
@@ -29,6 +29,8 @@
 	var/datum/weakref/holder_connect_loc_behalf
 
 /datum/component/slippery/Initialize(knockdown, lube_flags = NONE, datum/callback/callback, paralyze, force_drop = FALSE, slot_whitelist)
+	paralyze = 0 // SKYRAT EDIT ADDITION - COMBAT OVERHAUL
+	force_drop = FALSE // SKYRAT EDIT ADDITION - COMBAT OVERHAUL
 	src.knockdown_time = max(knockdown, 0)
 	src.paralyze_time = max(paralyze, 0)
 	src.force_drop_items = force_drop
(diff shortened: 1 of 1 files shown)


AzyreR:
This really feels like it'll break a bunch of things.. as an example:
A Double Esword user, even after the omniblock chance nerf they got are completely immune to lasers, deal a massive amount of damage on hit, have a chance to block your hit if you try to disarm them with a baton and even then they could just use noslip shoes and/or tape the sword to their hand, essentially with this, fighting a desword user will become practically suicide and leave fighting them with an equal melee weapon or ballistics as the only option, as having seen people with a desword and used it myself batons were actually something you would need to keep in mind of, as just having two sec officers swarm you was already a concern.
While on a different note, it's a core mechanic, this isn't as much related to combat as it is related to just the game itself, you slip on ice and you drop your crusher/PKA and get mauled by dragons, the clown slips you as sec and you have an assistant steal your dropped baton, janitor didn't put down a wet floor sign and you slip while rushing from place to place.
People already run everywhere and sometimes don't even touch the walk hotkey in a whole round while tripping and falling on every single slip hazard they run across, this'll just further enable that giving them the chance to think even less about their actions when i personally believe this game is about the exact opposite. Also there's a couple blobs that rely heavily on freezing floortiles and making everything a slip nightmare, and actual slip clustergrenades.
I don't have a proposal for this, except maybe heavy stamina damage, but I don't think anyone wants to get stamcrit by walking over two tiles of wet floor and/or be able to walk across 10 wet floor tiles knowing full well they have the ability to walk over exactly 13 tiles without caring... and when will you ever even walk across that many wet floor tiles, slipping will just end up turning into a hidden mechanic at that rate.

EDIT: The proposal I made for wallshoving might work here too, have slipping do heavy stamina damage and actually slip you only when you're below a certain stamina threshold, more than likely 50%, as it'll prevent someone from just tripping you with a single soap but will make you slip after doing one or two tiles of the same giving you some tolerance against it making it not be a surefire technique but still being a present threat.

YakumoChen:
what happened to the host request PR? I thought my idea was pretty good.

Gandalf2k15:
> what happened to the host request PR? I thought my idea was pretty good.

It'll be incorporated

GoldenAlpharex (approved):
Looks good to me, very atomic too, I me likies.

ChungusGamer666:
Massive nerf to a disgusting amount of things. 
Consider increasing the knockdown time at the very least, since being knocked down amounts to a temporary speed debuff.

Gandalf2k15:
I'll revisit this another year

PR #12068, opened 2022-03-14
Brings Space Hotel into the modern era, with some QoL.

dekunutkid:
## About The Pull Request
After a recent round playing on the Space Hotel I realised how limited it actually is, and how out of date it is. Having not been updated in so long the Space Hotel is missing things that should now be standard. 
This PR features the addition of an oven in the kitchen, an autolathe to the engineering room, a LustWish vendor and the replacing of the pool with Azarak's liquid system. 
A small QoL change involves the movement of a grille in maint so it isn't liable to shock you at any time.

Keeping up with the times is always a good thing.
The Autolathe is intended as a way of becoming marginally more independent and let the staff make some of their own parts rather than being forced to use what they're given with little recourse besides pandering to the station crew for basic parts when they come.

## Changelog

:cl:
add: Added an oven, LustWish and autolathe to the Space Hotel.
add: The Space Hotel pool now uses Azarak's liquid system, and maintenance grilles in the Hotel no longer shock you.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/spacehotel_skyrat.dmm
+++ b/_maps/RandomRuins/SpaceRuins/spacehotel_skyrat.dmm
@@ -401,6 +401,10 @@
 	},
 /turf/open/floor/wood,
 /area/ruin/space/has_grav/hotel)
+"cP" = (
+/obj/machinery/vending/dorms,
+/turf/open/floor/carpet/black,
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (approved):
Oh yeah, this looks good. Twin Nexus my beloved.

PR #12087, opened 2022-03-15
[SEMI-MODULAR] Adds admin buttons for exploitable overrides, as well as an OPFOR service, letting admins easily give people exploitables. Also cleans up some code I made before.

nikothedude:
## About The Pull Request

Makes a bunch of edits in some core files, does what it says in the title. It also makes a few random changes to the code I made on this.

Won't lie, made this PR while tired off my ass, so don't be surprised if there's some fuckiness in the code.

Exploitables good.

## Changelog

:cl:
add: New OPFOR services for exploitables.
admin: Added new admin buttons in traitor panel for adding or removing exploitable access.
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -723,6 +723,7 @@
 					log_admin("[key_name(usr)] tried and failed to give [current] an uplink.")
 				else
 					log_admin("[key_name(usr)] gave [current] an uplink.")
+			//SKYRAT EDIT ADDITION BEGIN -- EXPLOITABLES
 			if("toggleexploitables")
 				if (has_exploitables_override)
 					has_exploitables_override = FALSE
@@ -739,6 +740,7 @@
 					add_verb(current, /mob/proc/view_exploitables_verb)
 				log_admin("[key_name(usr)] toggled [current]'s exploitable menu override to [has_exploitable_menu_override].")
 				message_admins("[key_name(usr)] toggled [current]'s exploitable menu override to [has_exploitable_menu_override].")
+			//SKYRAT EDIT ADDITION END
 
 
 	else if (href_list["obj_announce"])
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -95,6 +95,10 @@
 	// SKYRAT EDIT ADDITION -- EXPLOITABLE MENU
 	///Tracks if the target has the view_exploitables_verb verb. THIS MUST BE CHANGED IF THE VERB IS ADDED OR REMOVED OR ELSE STUFF BREAKS.
 	var/has_exploitable_menu = FALSE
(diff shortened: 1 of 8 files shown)


modular_skyrat/modules/opposing_force/code/equipment/services.dm
```diff
@@ -57,3 +57,22 @@
 /datum/opposing_force_equipment/service/market_crash/on_issue()
 	var/datum/round_event_control/event = locate(/datum/round_event_control/market_crash) in SSevents.control
 	event.runEvent()
+
+/datum/opposing_force_equipment/service/give_exploitables
```
Zonespace27, line 61: this shouldn't be two different services, consolidate it into one.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -392,7 +392,7 @@
 				return
 
 	//SKYRAT EDIT ADDITION BEGIN - VIEW RECORDS
-	if (is_special_character(usr))
+	if (is_special_character(usr) || usr.mind.has_exploitables_override)
```
Zonespace27, line 395: remove all special character stuff and just give `has_exploitable_override` to all antagonists instead.

code/datums/mind.dm
```diff
@@ -719,6 +723,23 @@
 					log_admin("[key_name(usr)] tried and failed to give [current] an uplink.")
 				else
 					log_admin("[key_name(usr)] gave [current] an uplink.")
+			if("toggleexploitables")
+				if (has_exploitables_override)
```
Zonespace27, line 728: this and 3 lines below can be consolidated into `has_exploitables_override = !has_exploitables_override`

Zonespace27 (changes requested):

SpaceVampire:
Do you think you improve the exploitable menu, so that it doesn't just dump directly into chat?
I feel like it would be a lot better if you could just open/close them inside of the main thing.
Pretty sure that is a lot more close to how it is handled in other codebases.

nikothedude:
> Do you think you improve the exploitable menu, so that it doesn't just dump directly into chat? I feel like it would be a lot better if you could just open/close them inside of the main thing. Pretty sure that is a lot more close to how it is handled in other codebases.

There was actually a reason I had normal record viewings dump to chat, because having a TGUI window suddenly obscuring your view might make fighting a little difficult, although I kinda question who'd be looking at records while fighting, so I might just convert it all to TGUI.

nikothedude:
Drafting this, I'm turning this into a refactor of how we handle records because god damn I was not good at code when I made it.

nikothedude:
Doing this in another PR. Im gonna change how it all works.

PR #8989, opened 2021-10-23
Soft revert of HoP's winter coat resprite.

SpaceVampire:
## About The Pull Request

Removes the changes done in https://github.com/Skyrat-SS13/Skyrat-tg/pull/8978, while keeping the actual sprites in the code for the moment. This also changes the stuff done to allow them to hold the energy guns, along with the armor on the parade jacket.

## How This Contributes To The Skyrat Roleplay Experience

This was done on Maintainer request.

## Changelog

:cl:
fix: The HoP coat has returned to it's old glory.
fix: The parade jacket no longer has random armor, along with gun-sized pockets. Sacrifice for your drip.
/:cl:

PR #12097, opened 2022-03-15
[semimodular] fixes the hop winter coat because I can't trust other people to do it

Seris02:
## About The Pull Request

title.
Codewise, the winter coat doesn't work with reskins, as well as the fact that

# the fuckin sprites were never put back in.

I didn't realise this because I haven't been playing much for about five months
so there are now two hop winter coats, as is required, one normal, classic, the better one, and one of them 'formal'

this is a fix of the PR #8989

better sprites, the glory days are back

## Changelog

:cl:
fix: the HoP winter coat has been separated into two coats, and the old icons that were supposed to be put in in the last fix PR for it are actually now back in the files.
/:cl:


GoldenAlpharex (approved):
Checks out.

PR #12139, opened 2022-03-17
Cleans our new box icons

ghost:
## About The Pull Request
Mostly just removes useless icons from the .dmi and fixes some jank with the icons (offsets, some missing, and in one case one was animated and shit broke)

Though while I was here, I made the plastic box have a different palette+handle (like the old plastic box had) and played around with the shotgun boxes, which looked virtually the same as every other box when we added the sprites; but now
image
They're much more recognizable, but still fit the new style

Sprite cleanup from fixes to just general tidiness

## Changelog

:cl:
fix: fixed the skyrat boxes having a few borked icons
imageadd: shotgun shell boxes and plastic boxes have undergone some respriting
/:cl:


modular_skyrat/modules/aesthetics/storage/storage.dm
```diff
@@ -84,3 +91,22 @@
 /obj/item/storage/box/silver_sulf
 	icon_state = "medbox"
 
+
+/obj/item/storage/box/cyber_implants
+	illustration = NULL //Included in the sprite
```
ghost:
```suggestion
	illustration = null //Included in the sprite
```
capitalization moment

PR #12205, opened 2022-03-21
[MODULAR][QoL] Removes empty Hypovials from standard Hypospray Kits

softcerv:
## About The Pull Request
Removes the empty vials that come standard with the standard hypospray kits, including the one inserted into the hypospray.

Since Hypovials are printable from the chem-master now, having empty hypovials within the kit introduces unnecessary tedium and things to discard. 

## Changelog

:cl:
qol: removes empty Hypovials from Hypospray Kits
/:cl:

### DIFF
--- a/modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
+++ b/modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
@@ -25,26 +25,26 @@
 	var/list/allowed_containers = list(/obj/item/reagent_containers/glass/vial/small)
 	/// Is the hypospray only able to use small vials. Relates to the loaded overlays
 	var/small_only = TRUE
-	//Inject or spray?
+	/// Inject or spray?
 	var/mode = HYPO_INJECT
 	var/obj/item/reagent_containers/glass/vial/vial
-	//If the Hypospray starts with a vial, what vial does it start with?
+	/// If the Hypospray starts with a vial, which vial does it start with?
 	var/start_vial
-	//Does the Hypospray start with a vial?
+	/// Does the Hypospray start with a vial?
 	var/spawnwithvial = FALSE
 
-	//Time taken to inject others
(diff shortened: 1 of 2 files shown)


lessthnthree:
So instead to fill from plumbing factories we'll then carry bottles to the chemmaster, to turn into vials, adding tedium.

softcerv:
> So instead to fill from plumbing factories we'll then carry bottles to the chemmaster, to turn into vials, adding tedium.

Most people end up discarding the empty hypovials on the floor because they use the chem-master to make new ones when making their mixes. From what I have seen, it is very rare that the included empty hypovials actually get used.

Proximitron:
As someone who plays medical often:
I usually take the hypospray out, take the empty vial out of the hypospray and discard the hypospray kit altogehter, putting the hypospray in my belt.
Others i know keep the kit but discard all the empty vials. The reason for this is very simple: The empty vials are not labelt, while the ones from the chemmaster are labled automaticly on production. As the empty vials are insufficient for every medicin you need anyway, there is really no need for them.

modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
```diff
@@ -28,8 +28,8 @@
 	//Inject or spray?
 	var/mode = HYPO_INJECT
 	var/obj/item/reagent_containers/glass/vial/vial
-	var/start_vial = /obj/item/reagent_containers/glass/vial/small
-	var/spawnwithvial = TRUE
+	var/start_vial
+	var/spawnwithvial = FALSE
```
GoldenAlpharex: If you could document those variables while you're there, that'd be great.

GoldenAlpharex (changes requested):
Aside from the required documentation, I'd say this is a good change.

GoldenAlpharex (approved):
Now that's what I'm talking about!

PR #12251, opened 2022-03-24
[MODULAR] Alert level text changes

SingingSpock:
## About The Pull Request

Alters the alert level text to reflect current Skyrat policy rather than tg or outdated policy.

We have a consistent confusion with what is required at different alert levels because the text shown in game is often in direct conflict with what Skyrat policy says. I have moved every alert text into the modular file, and I have edited most of them to reflect the current state of Skyrat policy. The most notable changes are in regard to what weaponry is permitted on what alert level, and clarifying that searches on Blue alert require authorization from the HoS, Warden, or Captain (rather than random searches like our current text says). Additional touch-ups were made as well, such as making leaving affected areas on an Orange alert a recommendation rather than a requirement (sometimes non-engineering crew are perfectly willing and able to help repair damage) and adding internals and distancing recommendations to Violet as it is used almost exclusively for viruses.

## Changelog

:cl:
config: changed alert upto and downto texts
/:cl:


SingingSpock:
I'm not entirely sure whether the config tag was the right one for the changelog, decided to go with it since I was editing a file titled config.dm. Change it if I'm wrong, please.

ghost (commented):
Looks sound to me, the single thing to note is that "Security staff may have lethal weapons." just feels... short, idk? More something to think about than something to worry about though

SingingSpock:
> Looks sound to me, the single thing to note is that "Security staff may have lethal weapons." just feels... short, idk? More something to think about than something to worry about though

The relevant line in Security policy just reads "Security personnel may be equipped with lethal primaries." That's the primary change, gear-wise, from Blue.

PR #12267, opened 2022-03-25
Since Opfor has it's own ban category, it no longer blocks people with an antag ban.

Iamgoofball:
This seems like an oversight from before we added BAN_OPFOR.


Gandalf2k15:
Intended behaviour

PR #12278, opened 2022-03-25
[Modular] Deletes primal podperson while we sort out the weird BSOD issue with 'em

Iamgoofball:
This'll knock any characters with the species back to Human which makes them safe to click on and select.


GoldenAlpharex:
Not really necessary while we have the test-merge revert, I don't think.

GoldenAlpharex:
No need for this.

PR #12306, opened 2022-03-26
[MIRROR] Mystery boxes no longer cut off each other's audio [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/65655
--------------------

## About The Pull Request
I put mystery boxes on their own sound channel so they could cancel out the full music clip when someone took an item from them, but this causes issues because only one box could be playing a sound clip at a time, and they'd cut each other off if there were multiple. Now they reserve their own channels for their duration, then release them, so they can all play at the same time. Accordingly, their sounds now fall off quicker so having multiple of them nearby won't destroy your ear drums.

## Why It's Good For The Game
Makes having multiple mystery boxes less jarring

## Changelog

:cl: Ryll/Shaps
qol: Mystery boxes no longer cut off each other's sound clips when playing their own
/:cl:


code/__DEFINES/sound.dm
```diff
@@ -21,6 +22,10 @@
 #define CHANNEL_HEV 1010
 #define CHANNEL_PCV 1009
 //SKYRAT EDIT CHANGE END
+=======
+#define CHANNEL_TRAITOR 1017
+
+>>>>>>> f8794c441b4 (Mystery boxes no longer cut off each other's audio (#65655))
```
GoldenAlpharex, line 28:
```suggestion
#define CHANNEL_TRAITOR 1017

```

PR #12309, opened 2022-03-26
[MODULAR] Adds No Guns-new quirk that strictly disallows you from using guns. That's it.

nikothedude:
## About The Pull Request
Adds a new quirk, No Guns, with -6 cost, and the effect of giving you NO_GUNS, the same trait sleeping carp gives you. 

The best part of chunky fingers without the dumb parts. I love that quirk, but not the fact that I can't use computers. So, I decided to lazily slap together a PR with the quirk that actually does the good part exclusively.

Also, opt-in mechanical disadvantages good.

## Changelog

:cl:
add: Added a new quirk: No Guns. Makes it so you can't use guns.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/jobs.dm
+++ b/code/__DEFINES/~skyrat_defines/jobs.dm
@@ -4,7 +4,7 @@
 #define JOB_UNAVAILABLE_SPECIES JOB_UNAVAILABLE_QUIRK + 1
 #define JOB_UNAVAILABLE_LANGUAGE JOB_UNAVAILABLE_SPECIES + 1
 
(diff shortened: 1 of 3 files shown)


SpaceVampire:
You need to make it so that you can't take this and clunky fingers at the same time due to the overlap between them. I also think this is a bit too specific of a quirk, but whatever.

YakumoChen:
👎 
Too much overlap with Pacifist and Chunky Fingers. 

Even then it needs a specific trait ban for security and heads.

tf-4:
Heads? What? Sec, naturally, and HoS falls under that. But heads is incorrect.

GoldenAlpharex (changes requested):
I'm fine with this, but that's probably another thing that Security should reasonably not be allowed to have, considering how I don't think you'd hire mercenaries that don't know how to use a gun.

nikothedude:
As much as I'm opposed to quirk restrictions as a whole, sure.

Gandalf2k15 (approved):

PCPort:
https://youtu.be/NzwRTXBRS2g

PR #12317, opened 2022-03-27
GATEWAY OPTIMISATION - MOB SPAWNING

Gandalf2k15:
## About The Pull Request

Mobs no longer instantly spawn on mob_spawn_initialise. They instead spawn when a mob WITH a client enters their trigger range.

This should help with performance a little.

## Changelog

:cl:
code: Some mob spawners have been optimised.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
+++ b/modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
@@ -549,6 +549,8 @@
 	desc = "A 5.56mm surplus bullet casing."
 	projectile_type = /obj/projectile/bullet/a556/weak
 
+#define MOB_PLACER_RANGE 12 // One more tile than the biggest viewrange we have.
+
 /obj/effect/random_mob_placer
 	name = "mob placer"
 	icon = 'icons/effects/mapping_helpers.dmi'
@@ -557,9 +559,22 @@
 
 /obj/effect/random_mob_placer/Initialize(mapload)
 	. = ..()
+	for(var/turf/iterating_turf in range(MOB_PLACER_RANGE, src))
+		RegisterSignal(iterating_turf, COMSIG_ATOM_ENTERED, .proc/trigger)
+
+/obj/effect/random_mob_placer/proc/trigger(datum/source, atom/movable/entered_atom)
+	if(!isliving(entered_atom))
+		return
+	var/mob/living/entered_mob = entered_atom
+
+	if(!entered_mob.client)
+		return
+
 	var/mob/picked_mob = pick(possible_mobs)
 	new picked_mob(loc)
-	return INITIALIZE_HINT_QDEL
+	qdel(src)
+
+#undef MOB_PLACER_RANGE
 
 /obj/effect/random_mob_placer/xen
 	possible_mobs = list(
--- a/modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
+++ b/modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
@@ -549,7 +549,7 @@
 	desc = "A 5.56mm surplus bullet casing."
 	projectile_type = /obj/projectile/bullet/a556/weak
 
-#define MOB_PLACER_RANGE CONFIG_GET(string/default_view) + 1 // One more tile than the biggest viewrange we have.
+#define MOB_PLACER_RANGE 16 // One more tile than the biggest viewrange we have.
 
 /obj/effect/random_mob_placer
 	name = "mob placer"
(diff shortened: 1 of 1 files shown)


modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
```diff
@@ -549,6 +549,8 @@
 	desc = "A 5.56mm surplus bullet casing."
 	projectile_type = /obj/projectile/bullet/a556/weak
 
+#define MOB_PLACER_RANGE 12 // One more tile than the biggest viewrange we have.
```
nevimer: CONFIG_GET(string/default_view) + 1 ?

Gandalf2k15: That relies on the config being set tho

GoldenAlpharex: I don't see many instances where the config wouldn't be set.

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
```diff
@@ -549,6 +549,8 @@
 	desc = "A 5.56mm surplus bullet casing."
 	projectile_type = /obj/projectile/bullet/a556/weak
 
+#define MOB_PLACER_RANGE CONFIG_GET(string/default_view) + 1 // One more tile than the biggest viewrange we have.
```
GoldenAlpharex: Actually I think that config might not give us the thing we actually want, as that seems to be a string that says `15x15`.

I'd probably say something like 8 would be a good value here, and if other servers have for some reason an even bigger distance, well... Sucks to be them I guess.

Unless we can find the define that's already saying the size of our widescreen.

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
```diff
@@ -557,9 +559,22 @@
 
 /obj/effect/random_mob_placer/Initialize(mapload)
 	. = ..()
+	for(var/turf/iterating_turf in range(MOB_PLACER_RANGE, src))
+		RegisterSignal(iterating_turf, COMSIG_ATOM_ENTERED, .proc/trigger)
```
GoldenAlpharex, line 563: What happens if those turfs are deleted?

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
```diff
@@ -557,9 +559,22 @@
...
+
+/obj/effect/random_mob_placer/proc/trigger(datum/source, atom/movable/entered_atom)
+	if(!isliving(entered_atom))
```
GoldenAlpharex, line 567: You should probably add `SIGNAL_HANDLER` on the first line of the proc, as that's typically what's done for procs that are activated by a signal, if I remember correctly?

GoldenAlpharex (changes requested):

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
Gandalf2k15: They just lose their signal register

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
Gandalf2k15: Yee

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
nevimer: The default *should be* 21x15, and the non default should be 15x15, as a fresh server defaults to 21x15.

modular_skyrat/master_files/code/modules/awaymission/mission_code/black_mesa/mesa_mobs.dm
```diff
@@ -549,7 +549,7 @@
 	desc = "A 5.56mm surplus bullet casing."
 	projectile_type = /obj/projectile/bullet/a556/weak
 
-#define MOB_PLACER_RANGE CONFIG_GET(string/default_view) + 1 // One more tile than the biggest viewrange we have.
+#define MOB_PLACER_RANGE 16 // One more tile than the biggest viewrange we have.
```
nevimer, line 552: I dunno, I see issues down the line if TG ever added viewport adjustment or some of the downstreams edit their default config.

PR #12318, opened 2022-03-27
Fixes Emergency Shuttles

Inari-Whitebear:
## About The Pull Request

Currently the system loading the shuttles looks in the wrong folder for most shuttles as Skyrat Emergency Shuttle overrides the path for it. From what I've tested this fix still ends up having the default shuttle, but now the others are buyable again.

**To note: I have no idea if this was maybe intended? i.e. the correct way to enable the other shuttles would be to copy their map files to the Skyrat folder and edit them to fit the Skyrat theme better.**

Allows more shuttles to exist again, as people seem to want them.

## Changelog

:cl:
fix: Fix pathing issue with Skyrat emergency shuttle causing other emergency shuttles to not be found
/:cl:

### DIFF
--- a/modular_skyrat/modules/advanced_shuttles/code/shuttles.dm
+++ b/modular_skyrat/modules/advanced_shuttles/code/shuttles.dm
@@ -62,7 +62,7 @@
 	port_id = "escape_pod"
 	suffix = "default_skyrat"
 
-/datum/map_template/shuttle/emergency
+/datum/map_template/shuttle/emergency/default
 	prefix = "_maps/shuttles/skyrat/"
 	suffix = "skyrat"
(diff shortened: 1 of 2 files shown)


YakumoChen:
finally

GoldenAlpharex:
> From what I've tested this fix still ends up having the default shuttle, but now the others are buyable again.

I just want you to elaborate on that, if you could. By buyable, I assume you mean that they will show up as the appropriate shuttle when bought, rather than the default one every time, correct?

Inari-Whitebear:
Yeah, I mean that they actually show up. Also just pushed a fix as I had accidentally borked the Outpost station shuttle with my original fix.

GoldenAlpharex (approved):

PR #12329, opened 2022-03-28
Head of staff bowman headsets

GuiltyNeko:
## About The Pull Request
Gives heads of staff bowman headsets in their lockers, like the HoP, cap, and HoS. Heads of staff can't be antags, except obsessed, which anyone at all can be including the captain. So you don't need to account for them needing to be chased by security, and and this doesn't necessarily increase the ease of getting a bowman, because checkpoint, and with command keys orderable through cargo, this doesn't make that any easier either. (basically I just want this)

Also I gave the QM a normal spare headset in their locker since they didn't have one. 

Heads of staff with bowmans provides them a reliable defense against basic stuff, and means a blueshield might be able to reliably carry equipment like flashbangs, and most importantly, makes not being allowed to take the captain's bowman headset as acting captain less of an issue.

## Changelog

:cl:
add: Heads of staff will now find bowman headsets in their lockers.
fix: The QM's locker now has a spare normal headset.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
+++ b/modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
@@ -20,6 +20,6 @@
 	icon_state = "com_headset_alt"
 
 /obj/item/radio/headset/heads/qm/alt
-	name = "\proper the quartermaster's headset"
+	name = "\proper the quartermaster's bowman headset"
 	desc = "The headset of the king (or queen) of paperwork. Protects ears from flashbangs."
 	icon_state = "com_headset_alt"
--- a/modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
+++ b/modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
@@ -3,3 +3,23 @@
 	desc = "The headset of the king (or queen) of paperwork."
 	icon_state = "com_headset"
 	keyslot = new /obj/item/encryptionkey/heads/qm
+
+/obj/item/radio/headset/heads/rd/alt
+	name = "\proper the research director's bowman headset"
+	desc = "Headset of the fellow who keeps society marching towards technological singularity. Protects ears from flashbangs."
+	icon_state = "com_headset_alt"
+
+/obj/item/radio/headset/heads/ce/alt
+	name = "\proper the chief engineer's bowman headset"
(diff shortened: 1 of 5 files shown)


Aether217:
In practice the bowman really just exists to protect from flashbangs, which is not something the majority of command members need. There's not much of a point to this addition

GuiltyNeko:
By the way, unrelated, but the head of security's non-bowman headset has flashbang protection.

SpaceVampire:
This is just unnecessary power creep. All of the heads of staff that don't have it, have no reason to have it.

modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
```diff
@@ -3,3 +3,38 @@
...
+	. = ..()
+	AddComponent(/datum/component/wearertargeting/earprotection, list(ITEM_SLOT_EARS))
+/obj/item/radio/headset/heads/qm/alt
+	name = "\proper the quartermaster's bowman headset"
+	desc = "The headset of the king (or queen) of paperwork. Protects ears from flashbangs."
+	icon_state = "com_headset_alt"
+
+/obj/item/radio/headset/heads/qm/alt/ComponentInitialize()
+	. = ..()
+	AddComponent(/datum/component/wearertargeting/earprotection, list(ITEM_SLOT_EARS))
```
GoldenAlpharex, line 40: First of all, those should probably be called `/bowman` rather than `/alt` because that doesn't mean anything.

Secondly, I too think this is some unnecessary powercreep and I don't really see a need for this in the game.

GoldenAlpharex (changes requested):
I'm only reviewing for future reference, but this is both quite non-modular, and in my own personal opinion, really unnecessary.

Heads of staff don't need better protection just because they can't be antags, the goal isn't to encourage security to use flashbangs when heads of staff are involved.

GoldenAlpharex:
So, because I don't think this would really be good for the game, I'll close it for now. We can always re-open it back later if we ever feel like this has become necessary.

modular_skyrat/master_files/code/game/objects/items/devices/radio/headset.dm
GuiltyNeko: I did alt because that's what the other bowmans were called so that was... the thing to do.

PR #12331, opened 2022-03-28
Matter manipulator fix

GuiltyNeko:
## About The Pull Request
We had this funky "matter manipulator" item that was supposed to be a ranged RCD, but it wasn't ranged, because TG code is fucking dumb and the code for ranged RCDs was on the ARCD subtype, which the matter manipulator was not a subtype of, hence no range. I tried making it on the entire thing, all RCDs, but then all RCDs became ranged and that's not what I'm after, so I just made it an ARCD subtype and now it functions the exact same huzzah.

Does anyone use this?

## Changelog

:cl:
fix: The matter manipulator is now ranged as is intended.
/:cl:


GoldenAlpharex (approved):
Good work.

PR #12332, opened 2022-03-28
Delta station atmos door ID requirement undo

GuiltyNeko:
## About The Pull Request
Delta station atmos is either atmos or engineering access, which allows heads of staff and the paramedic access to atmospherics, which is of course not necessary in the slightest. Engineers still have atmos access on skeleton crew and can get the HoP or captain to give them atmos access if they need it.

## Changelog

:cl:
fix: removed generic engineering access from delta station atmos doors
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -36069,7 +36069,7 @@
 /obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/atmos{
 	name = "Atmospherics";
-	req_one_access_txt = "10;24"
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (commented):
Was that a thing upstream too? If so, could be a good idea to fix that there too?

Is this this way on any other map, or is Delta just an outlier in this regard?

GuiltyNeko:
I think this is a silly thing some mapper decided to do for some reason, it's only delta.

GoldenAlpharex:
If we don't want this to be reset every map reset, we might want to make that change upstream too, then.

KathrinBailey:
do this upstream as well if you want it to stay

PR #12343, opened 2022-03-28
[TM-ONLY] Temporary revert of mapping_helpers.dm

GuiltyNeko:
## About The Pull Request
An edit broke airlock cycling. Until it's fixed, upstream, this will help with the issue where doors will just not cycle. I'd like to say testmerge only, as this really should not be permanent. If this lasts long enough to need a permanent fix, it's time to scream at TG. Also I couldn't auto-revert so I did a thing that kinda came close and anyways here we are.

Pressure.

## Changelog
No need since it's a temp-revert

### DIFF
--- a/code/modules/mapping/mapping_helpers.dm
+++ b/code/modules/mapping/mapping_helpers.dm
@@ -101,22 +101,17 @@
 //airlock helpers
 /obj/effect/mapping_helpers/airlock
 	layer = DOOR_HELPER_LAYER
-	late = TRUE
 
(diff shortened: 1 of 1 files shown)


Tastyfish:
Looks like https://github.com/tgstation/tgstation/pull/65779 intends to fix it

GuiltyNeko:
Closing since our maps ALREADY have these mapping helpers and if we revert them, suddenly issues happen, also I hate lazy mappers.

PR #12391, opened 2022-03-30
Opting out of EORG (Ghostcafe Teleport)

Inari-Whitebear:
## About The Pull Request

Ports (https://github.com/Skyrat-SS13/Skyrat13/pull/3459)
* Adds a checkbox to opt out of EORG
* When that is checked, you get teleported to the ghost cafe on round-end

Possible To-Do's:
* Spread out spawning spots better, I concentrated them mostly around the shuttle.
* Find a better solution than a teleport.
* Find a better solution to EORG in general.
* Maybe change the way the spawner is added to the map.
* Adjust descriptions for the checkbox.
* Test some more.

Some people don't like EORG. While a teleport is a little heavy-handed, it might at least be preferable to the alternatives. This was ported from Skyrat13 (https://github.com/Skyrat-SS13/Skyrat13/pull/3459), since someone yesterday in the Discord suggested to do something useful instead of just talking about EORG solutions. And one of the useful things suggested was that someone could port the EORG thing from the old codebase.

## Changelog

:cl:
add: Checkbox to opt out of EORG. Users who have this checked get teleported to the ghost cafe at round-end.
add: Extra text on ghost cafe spawns warning users that spawn in the cafe that on round-end people who opted out of EORG will be teleported here
/:cl:


Gandalf2k15:
This was never signed off, nor was it something we were going to do. Sorry.

PR #12417, opened 2022-03-31
Assault ops armament update and mapping fixes

Cenrus:
## About The Pull Request

Updates and adds a few items to the armament dispenser.

New Items:

-Radio Jammer (4 points)
-Deluxe Agent ID (5 points)
-Ammo Pouch (1 point)
-Combat Gloves Plus (secondary option, gives wearer krav maga, 5 points)
-Modsuit Cloak Module (like the one that science can make, 4 points)
-Modsuit Quick Carry module (2 points)
Added Ion rifle to the ship armory

Rebalanced:

Halved the elite suit price, so it costs 8 instead of 15 points. It wasn't worth it at 15, since it only offers a slight armor advantage.
Reduced marine armor from 6 to 5, it's actually a bit mediocre and I think the basic modsuits you get in the ship are better.
Reduced price of both knives by 1.
Increased energy shield cost from 5 to 8, because it's actually a lot better than I thought it was.

As for fixes, the cruiser has got a few of them. I've fixed the turrets not firing the intended 357 bullets and added vents and scrubbers to deal with atmos issues on board.

If you think anything is overtuned please let me know so I can change some numbers or just outright remove it.

Fixed some stuff that really needed fixing

## Changelog

:cl:
add: Changed some armament vendor item costs
fix: Fixed assault op ship turrets not using ballistics
/:cl:


nevimer:
Are we able to buy magazines for guns yet? AFAIK you can only get the one starting in the gun.

Cenrus:
> Are we able to buy magazines for guns yet? AFAIK you can only get the one starting in the gun.

That was fixed, you get 3 or 2 spare magazines depending on which gun you buy + you can get extra from the vendor.

Zonespace27:
> Are we able to buy magazines for guns yet? AFAIK you can only get the one starting in the gun.

Cobalt has a PR for that

KathrinBailey (approved):
nothing against the map changes, it's all balance, up to @Gandalf2k15

PR #12480, opened 2022-04-01
Removes Teshari

Gandalf2k15:
## About The Pull Request
Due to the fact Teshari have been impacting the server negatively recently, we've come to the decision to remove them.

 The original code that they are based upon is not compatible with our updates nor the upstream, they are also a source of a lot of lag.

We understand this may be a problem for some of you, but I'm afraid you'll just have to cope.

:cl:
del: Teshari have been removed from the codebase.
/:cl:


Ebin-Halcyon:
Zamn. :pensive:

ArdentSpark:
speedmerge

PR #12492, opened 2022-04-01
Removes stormtrooper outfit from PK vendor

tf-4:
# This is not an April fool's PR. Just poorly timed.

## About The Pull Request

Title

The stormtrooper outfit actively detracts from a convincing atmosphere. Way too overt a reference.

## Changelog
:cl:
del: Security no longer gets the stormtrooper outfit in their vendors.
/:cl:


Gandalf2k15:
self thumbs up, instant closure without appeal

tf-4:
denied

hugmeorelse:
Good joke, but yeah, I do think it is too overt of a reference, not really the direction we want to take the server in.

PR #3820, opened 2021-03-03
Removes all dumb stuff from the ghost cafe

Azarak:
Removed sheet stack spawners
Removed the Robco vendor
Removed all protolathes
Removed the medical department there
Removed the plumbing section
Removed ARCD's and RLD's
Consoles and items in genetics and nanite labs have been replaced with useless props

## Why It's Good For The Game
Ghost cafe shouldnt be anything but a hangout place

## Changelog
:cl:
del: Removed all non-hangout gameplay and sandbox features from the ghost cafe
/:cl:

PR #12511, opened 2022-04-02
Adding a fighting pit to Ghost Cafe. For containment following a ruling.

theselfish:
## About The Pull Request

I added a fighting pit to follow 
image
This ruling. You get a fence and knives. OpFor/ahelp/pray for other stuff.
I fixed the alien tools not having a retractor.
Added a Chameleon ID. For whatever reason people want it.
I also spun every single stool around. They were annoying me.

People enjoy fighting in the Ghost Cafe. A ruling has been made that you must make an area for it.
 This gives them an area, with a note for explanation of why it is there.

## Changelog

:cl:
add: Ghost Cafe has a fighting pit. You can also use a Chameleon ID.
qol: Ghost Cafe stools now properly face the correct direction. 
fix: There is now an Alien Retractor to match the rest of the kit.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -5948,6 +5948,10 @@
 /obj/structure/window/spawner/east,
 /turf/open/floor/iron,
 /area/centcom/interlink)
+"bbW" = (
+/obj/effect/decal/cleanable/blood,
+/turf/open/misc/grass/planet,
+/area/centcom/holding/cafepark)
 "bee" = (
 /obj/machinery/cryopod{
 	dir = 1
@@ -6075,10 +6079,6 @@
 	},
 /turf/open/floor/glass/reinforced,
 /area/cruiser_dock)
-"bBe" = (
-/obj/effect/gibspawner/robot,
-/turf/open/floor/plating,
-/area/centcom/holding/cafewar)
 "bDJ" = (
 /obj/machinery/door/airlock{
 	id_tag = "dorm23";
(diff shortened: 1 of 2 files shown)


softcerv:
This is something most likely out of the scope of this PR, but being able to watch this via cameras like you can with the Thunderdome arena (I think you can do this, but I'm not 100% sure) would be pretty neat

theselfish:
> This is something most likely out of the scope of this PR, but being able to watch this via cameras like you can with the Thunderdome arena (I think you can do this, but I'm not 100% sure) would be pretty neat

Most likely. Yes. I have thought of adding an advanced camera console for whatever reason.
But I think that for watching the Arena I'd need to install cameras.
Add in the fact the current thunderdome monitors don't even work and...

tf-4:
We used to have this and it was explicitly and intentionally removed. see https://github.com/Skyrat-SS13/Skyrat-tg/pull/3820 for similar

theselfish:
> We used to have this and it was explicitly and intentionally removed. see #3820 for similar

Yet people still want to fight in the Cafe. To the point a ruling has been made that says *If you want to have a fighting pit with your friends make yourself an area in one of the free parts and keep it there.*

This is cutting out the making part.

theselfish:
I should also mention that in the time since that PR a Medical Wing has been re-added.

hugmeorelse:
I appreciate the enthusiasm, but I think we're fine without it.

hugmeorelse:
There are two reasons for this. This is another element we have to maintain for maps which is something we're already struggling to do, and the value being added here does not supersede what has to be done to maintain this. Lastly, I fail to understand why we want to take up more real estate for something that can be easy built or just done without in one of the free parts. It's not like you need an actual fighting pit to fight in. You can practice mechanics there easily.

PR #2285, opened 2020-12-23
[MIRROR] Syringe gun and rapid fire syringe gun nerfs, uplink reagent gun buffs.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/55642
--------------------

## About The Pull Request

Direct alternative to #55627 - See that PR for a ton of discussion to give context to this PR.

1. Uplink reagent gun gets piercing darts.
2. Botanists get access to the uplink reagent gun.
3. All syringe guns can no longer chamber bluespace syringes.
4. Rapidfire syringe guns will be locked behind illegal tech.

## Why It's Good For The Game

> 1. Uplink reagent gun gets piercing darts.

I originally thought this thing synthesises chems. It does not. It synthesises darts. It's a lot of TC to spend on an item that's worse than using a syringe gun with bluespace syringes in almost every single way and could use a QoL tweak on its own to make it a worthwhile purchase.

Rather than bringing the TC cost down, I've raised its power level by giving it piercing darts. I don't personally think there is an issue with traitors having the ability to spend 12TC to deliver 15u piercing darts filled with chems compared to spending no TC and just using the ordinary syringe guns with 10u piercing/15u ordinary syringes instead.

> 2. Botanists get access to the uplink reagent gun.

Botanists now work with tons of chems. Could potentially enable some antag botanists an additional option for using the chems they make. Makes sense in my head?

> 3. All syringe guns can no longer chamber bluespace syringes.

Some of the problem areas highlighted with syringe guns included a number of comments on how 60u bluespace syringes are what can tip syringe guns from being a useful self defense weapon to the ultimate one-click validhunting tool.

We have been trying to do away with 1-click combat roulette in favour of longer fights with more room for individual player skill beyond who clicks first. I think this change adds enough room for syringe guns to be an effective self defense weapon while restricting their raw 1-click kill potential. See also changes to flashes which limited their offensive potential and focused on their utility as a defensive tool. Chem-related antags can still trade 12TC for an improved gun that generates its own 15u piercing syringes.

Syringe guns in general can continue to chamber 50u lethal injection syringes which are limited in quantity on any shift and only available from secure areas.

> 4. Rapidfire syringe guns will be locked behind illegal tech.

Incredibly powerful chemistry tool that is capable of mass production at any med techfab once researched.

[...]

PR #12534, opened 2022-04-05
[MODULAR] Replacement syringe gun crate

SpookyTheFox:
## About The Pull Request
A replacement syringe gun.

Paramedic took the syringe gun to cryo? Clown stole the other one? Now you can order a replacement!
Due to skyrat's long rounds and high population, map spawn items tend to go missing and stay gone before those that are supposed to have them, get them. Some maps only have one of these spawn. Even when people don't intend to hog it for the round, it tends to disappear into the clutter of the medical storage room. The cost has been set to the same as the cell powered medigun, normally a 20 minute cool down order. This ensures an opportunity cost of not ordering anything else like modsuits or pen crates for a long period of time, or needing the station to cover the cost.

## Changelog

:cl:
add: single replacement syringe gun crate
/:cl:

Verified working on local test.

### DIFF
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -107,6 +107,14 @@
 					/obj/item/weaponcell/medical/toxin)
 	crate_name = "medicell replacement crate"
 
+/datum/supply_pack/medical/syringe_gun
+	name = "Syringe Gun"
+	desc = "A replacement medical syringe gun"
+	cost = CARGO_CRATE_VALUE * 30
+	access = ACCESS_MEDICAL
+	contains = list(/obj/item/gun/syringe)
+	crate_name = "replacement syringe gun crate"
+
 //////////////////////////////////////////////////////////////////////////////
 //////////////////////////// Security ////////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////////
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -113,7 +113,7 @@
 	cost = CARGO_CRATE_VALUE * 30
 	access = ACCESS_MEDICAL
 	contains = list(/obj/item/gun/syringe)
-	crate_name = "replacement syringe gun crate"
+	crate_name = "replacement syringe gun crate."
 
 //////////////////////////////////////////////////////////////////////////////
 //////////////////////////// Security ////////////////////////////////////////
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -109,7 +109,7 @@
 
 /datum/supply_pack/medical/syringe_gun
 	name = "Syringe Gun"
-	desc = "A replacement medical syringe gun"
+	desc = "A replacement medical syringe gun."
 	cost = CARGO_CRATE_VALUE * 30
 	access = ACCESS_MEDICAL
 	contains = list(/obj/item/gun/syringe)
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -113,7 +113,7 @@
 	cost = CARGO_CRATE_VALUE * 30
 	access = ACCESS_MEDICAL
 	contains = list(/obj/item/gun/syringe)
-	crate_name = "replacement syringe gun crate."
+	crate_name = "replacement syringe gun crate"
 
 //////////////////////////////////////////////////////////////////////////////
 //////////////////////////// Security ////////////////////////////////////////


modular_skyrat/modules/cargo/code/packs.dm
```diff
@@ -107,6 +107,14 @@
 					/obj/item/weaponcell/medical/toxin)
 	crate_name = "medicell replacement crate"
 
+/datum/supply_pack/medical/syringe_gun
+	name = "Syringe Gun"
+	desc = "A replacement medical syringe gun"
```
Zonespace27:
```suggestion
	desc = "A replacement medical syringe gun."
```

modular_skyrat/modules/cargo/code/packs.dm
```diff
@@ -107,6 +107,14 @@
...
+	cost = CARGO_CRATE_VALUE * 30
+	access = ACCESS_MEDICAL
+	contains = list(/obj/item/gun/syringe)
+	crate_name = "replacement syringe gun crate."
```
Zonespace27:
```suggestion
	crate_name = "replacement syringe gun crate"
```

softcerv:
I'd advise maybe moving this to goodies or some other category besides medical, mostly because medical now has SmartDart guns which function as syringe guns but without the ability to use harmful chemicals and piercing syringes (they can also be printed at the medical prolathe,) making the standard syringe guns more of an antag/self defense tool.

modular_skyrat/modules/cargo/code/packs.dm
SpookyTheFox: I was trying to keep consistency with the other ones, but they really aren't super consistent.

SpookyTheFox:
> I'd advise maybe moving this to goodies or some other category besides medical, mostly because medical now has SmartDart guns which function as syringe guns but without the ability to use harmful chemicals and piercing syringes (they can also be printed at the medical prolathe,) making the standard syringe guns more of an antag/self defense tool.

I considered that. It's really only valuable to someone with not only access to chemistry equipment, but then also the funds to buy it. The goodie pack route, the cost would be on the purchaser, and the only reasonable values do not jive with what I expect them to earn from bounties in a reasonable amount of time. They can buy shotguns and print reagent dart shells from an autolathe if they really want to go that route, but it just does not make sense in terms of time to earn the money, ways it can go wrong, and final performance. Bounties in general really need attention, but that's way beyond the scope of this PR.

I think putting it in goodie packs also legitimizes it as a weapon, and from an IC standpoint, it's a medical tool not for personal use.

I probably would have passed on making this PR entirely since the rapid syringe gun exists in code and there's supposed to be a route to obtain it even as a non-antag, But you actually can't get it, or anything from illegal tech. Our tech nodes are a complete mess in need of balance, QoL, and bug fixes, with no pending plans for any of those.

softcerv:
>I think putting it in goodie packs also legitimizes it as a weapon, and from an IC standpoint, it's a medical tool not for personal use.

I'm going to be honest, 9/10 times it usually is used as one. Especially now that SmartDarts are an option here. The main reason why I'm against it being a departmental order is that it feels like a very personal oriented purchase, especially when compared to other things like Mediguns and MODsuits, I don't feel like this is something I'd see a CMO buy for their department unless they wanted to arm up.

MassacreDemigirl:
medical has really limited stuff to buy that is actually worthwhile, I say its a good addition

OzzyMcPseudonym:
In all honesty I am all for the goodie pack route. Something akin to the revolver cost would be pretty balanced if you ask me (mostly cause you can just fill an armor-piercing shot with 2 Anacea and 8 Lexorin and practically crit or kill someone off of that alone, that stuff should probably be expensive.) The big thing to consider is "How does the syringe gun compare to other tools that medical has" (Bath salts, meth, ephedrine, smoke and foam, Laser Scalpel, etc). As far as accessibility is concerned it's a rather powerful tool that any doctor can reasonably use if they know a recipe to Chloral Hydrate or Super Melatonin or something else that debilitates or kills targets. 

It applies the darts to one target instead of an AOE, which in my opinion makes it more valuable than smoke and foam because it has less chance of accidentally hitting someone that you didn't mean to hit. Not only more valuable from a "Don't hurt innocents" perspective but also from a policy perspective. It makes much more sense to make a sedative dart IC than to make a whole-ass sedative foam grenade. Without a syringe gun medics are encouraged to instead use Melatonin foam with a Modafinil pill on themselves to prevent themself from falling asleep to their own grenade. Syringe guns make the chem meta more IC reasonable and also provide doctors with a means of targeted self-defense if some DEsword guy is slashing up people in the bay for some reason.

YakumoChen:
> I probably would have passed on making this PR entirely since the rapid syringe gun exists in code and there's supposed to be a route to obtain it even as a non-antag, But you actually can't get it, or anything from illegal tech. Our tech nodes are a complete mess in need of balance, QoL, and bug fixes, with no pending plans for any of those.

https://github.com/Skyrat-SS13/Skyrat-tg/pull/2285

The RSG move to illegal is pretty justified. As was said, syringes can be a one shot kill. Having a mass produced, rapid fire, one shot one kill guns is not something we want to have readily available most rounds.

MassacreDemigirl:
great point @YakumoChen however we have 10mm makarovs on sale as goodies.

YakumoChen:
that's fine, but not a point to justify Rapid Syringe guns being readily obtainable. Out of the scope of the PR, regardless.

MassacreDemigirl:
ah I was not aware we were talking about making RSG buyable, but having a way to replace normal syringer is nice

Gandalf2k15 (approved):

Gandalf2k15:
I see no harm given the way we are planning to go with cargo.

PR #12591, opened 2022-04-07
Adds Export Scanner To Cargo Borg

Reco201:
## About The Pull Request
Adds Export Scanner to the cargo borg a tool all cargo techs have.

image

Cargo Borgs can now check the value of items and crates as well as check manifests for their value.
However they are unable to scan the tip for bounty cubes that will be left up to regular techs.

Also bonus turns out the sm is worth not much.
image

## Changelog

:cl:
add: Cargo Borgs now come equip with a export scanner.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cargoborg/code/robot_modules.dm
+++ b/modular_skyrat/modules/cargoborg/code/robot_modules.dm
@@ -15,6 +15,7 @@
 		/obj/item/dest_tagger,
 		/obj/item/crowbar/cyborg,
 		/obj/item/extinguisher,
+		/obj/item/export_scanner,
 	)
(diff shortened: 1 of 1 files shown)


YakumoChen:
The value is likely actually just the gas contents from you activating the supermarket, fun fact

Reco201:
> The value is likely actually just the gas contents from you activating the supermarket, fun fact

That's really cool

jjpark-kb:
> The value is likely actually just the gas contents from you activating the supermarket, fun fact

actually, I think it might be internal radio maybe?

Gandalf2k15 (approved):

PR #12613, opened 2022-04-08
[NON-MODULAR] Renames the bluespace warning labels to be OOC info

MassacreDemigirl:
## About The Pull Request

Time and time again it has shown that we need to specify things like this clearly to the playerbase. Makes the armor/damage value listings of items into OOC values instead of being "bluespace labels" slapped unto items.
Hopefully removes some cover that people can use to disregard policy.

## Changelog

:cl: ErdinyoBarboza
spellcheck:  Renames the bluespace warning labels to be OOC info.
/:cl:


GoldenAlpharex (changes requested):
I'd say at least just add `OOC: ` before the ever-updating bluespace label or whatever, rather than just making it plain and boring.

PR #12651, opened 2022-04-10
Adds bowman headset to HECU armaments

Zonespace27:
## About The Pull Request
Adds 3 bowman headsets to the HECU armament vendor for 4 points each

There's only one sound-protecting headset among the HECU, they should be able to get more.

## Changelog

:cl:
balance: HECU can now buy bowman headsets
/:cl:


Stalkeros2:
It was intended for HECU to have a single bowman headset exclusive to their squad leader's "loadout"/pile-of-equipment (that's why it's in their tent), to prevent flashbangs from becoming a complete meta, and to give them a loud speaking option.
However, since those are basically paid, I think that it won't be that much of an issue.
~~Also good god I should finally get them HECU-exclusive headsets instead of the faction ones.~~

PR #12687, opened 2022-04-11
Adjusts interlink spawn locations

tf-4:
## About The Pull Request

More spawns on the starboard-side shuttle. No spawns in the cafe area. Shifts a few other ones around for flavour (one by washing machine, another by the arcade, moves a few in the south area to the benches instead of randomly in the middle).

image

Spawning on top of people is bad. Spawning in places that make a little more sense than randomly in the middle of a room is good.

## Changelog
:cl:
qol: Late joiners no longer spawn in chairs in the interlink cafe
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -368,8 +368,14 @@
 /turf/open/misc/grass/planet,
 /area/centcom/holding/cafepark)
 "ahu" = (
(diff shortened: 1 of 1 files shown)


theselfish:
Maybe also move the Bathroom spawners? It's odd looting a pile just to have someone _**poof**_ into existence. 

Not to mention what might happen if someone was doing a bathroom scene.

GoldenAlpharex:
Nice, I always didn't like the one in the game room, it was so stupid. Good work.

PR #12699, opened 2022-04-12
Fixed Veteran-Only species being selectable by non-Veterans

GoldenAlpharex:
## About The Pull Request
Does what it says on the tin. I now added some extra verification so that, now, if the race is veteran-only and you're trying to apply those prefs to yourself as a non-Veteran, it's not going to work, which means nobody is going to be grandfathered in for a bug.

Veteran-Only means Veteran-Only, otherwise it loses all meaning.

## Changelog

:cl: GoldenAlpharex
fix: Fixed Hemophages being selectable by non-Veteran. If you were a non-Veteran and had a Hemophage, it'll convert them to a human when you spawn in, but I'd still recommend making the change to human yourself, just in case something funky happens. The alternative to that is to become a Veteran if you want to keep your Hemophage :)
/:cl:

### DIFF
--- a/code/modules/client/preferences/species.dm
+++ b/code/modules/client/preferences/species.dm
@@ -32,9 +32,14 @@
 	return values
 
 /datum/preference/choiced/species/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/prefs)
+	// SKYRAT EDIT START - Veteran-only races
+	var/datum/species/value_typepath = value
+	if(initial(value_typepath.veteran_only) && !is_veteran_player(target.client))
+		value = create_default_value()
+	// SKYRAT EDIT END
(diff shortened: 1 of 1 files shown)


code/modules/client/preferences/species.dm
```diff
@@ -32,9 +32,14 @@
 	return values
 
 /datum/preference/choiced/species/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/prefs)
+	// SKYRAT EDIT START - Veteran-only races
+	var/datum/species/value_typepath = value
+	if(initial(value_typepath.veteran_only) && !is_veteran_player(target.client))
```
Zonespace27, line 37:
```suggestion
	if(initial(value_typepath.veteran_only) && !is_veteran_player(target?.client))
```

Gandalf2k15: That's actually just checking if target exists

PR #12728, opened 2022-04-13
Fixes Teshari and Vox clothing sprites

Tastyfish:
## About The Pull Request

The birbs are inherently snowflakey and thus naturally did break in the kapulimbs thing.
| Snowflake Bird |
|:--:|
|A visualization of teshari code|

I refactored them to share as much code as possible, with the only non-modular injection being 2 lines in each of the `update_inv_✖` functions, largely the same 2 lines except the arguments (following a regular pattern).

Their special bodytypes are now `BODYTYPE_CUSTOM` rather than vox and teshari specific ones, marking that the species contains custom snowflake code to call. 

I killed the matrix color system and refactored the tesh fallbacks to be GAGs based, so that we don't have 2 competing color systems. That technically means tesh clothing could just use the fallback configs or greyscale sprites freely if they want.

| demos |
|:--:|
| This is the ideal bird form. |

**Outstanding issue:** For some reason, vox have 2 dmi's for heads and glasses with slightly different content in each, (head_vox.dmi vs species/vox/head.dmi) & (eyes_vox.dmi vs species/vox/eyes.dmi). I biased towards the ones in the species/vox folder. A vox spriter will have to figure that mess out.

Fixes some kaplimbs bugs

## Changelog

:cl:
fix: Fixes teshari and vox clothing sprites.
refactor: Redoes how snowflake species do clothing. May cause unforseen edge case issues to look out for.
/:cl:


code/modules/mob/living/carbon/human/human_update_icons.dm
```diff
@@ -423,14 +426,14 @@ There are several things that need to be remembered:
 		var/obj/item/worn_item = head
 		var/mutable_appearance/head_overlay
 		update_hud_head(worn_item)
-		var/handled_by_bodytype = FALSE
+		var/handled_by_bodytype = TRUE
```
Gandalf2k15, line 429: ?

Tastyfish: They're supposed to be TRUE by default. It being FALSE in this proc was a typo as far as I can tell, as that would render the var useless.

Tastyfish:
Note that the linter failed due to Deltastation having duplicated pipes in a tile, not related to this PR. edit: and map merge conflict markers that are also in master*

Gandalf2k15 (approved):

PR #12748, opened 2022-04-14
[MODULAR] Self Defense Medicell

softcerv:
## About The Pull Request
Adds the Self defense Medicell that allows doctors to turn their Mediguns into far weaker energy guns. 

The cell contains both disabler and kill modes. The disabler can only fire 4 shots from a standard Medigun and has reduced stamina damage compared to the standard Skyrat disabler shot. 

The kill mode can only fire 3 shots, it uses the same projectile as the kill mode on e-guns, it is marked as harmful.

It can be ordered through cargo as a goodie and should cost around 2k cr.
 
ingame picture of the new cell texture
image

This is going with the push to make armaments more readily available to the crew, this gives doctors a unique "last resort" type option for defending themselves.

## Changelog

:cl:
add: adds in the Self Defense Medicell, allowing the Medigun to become a weaker energy gun. 
/:cl:


YakumoChen:
now we just need to make this standard on the secmedic instead of giving them a pistol

softcerv:
> now we just need to make this standard on the secmedic instead of giving them a pistol

This is actively worse than a mini-egun or a disabler. This isn't a very good option for them.

YakumoChen:
> > now we just need to make this standard on the secmedic instead of giving them a pistol
> 
> This is actively worse than a mini-egun or a disabler. This isn't a very good option for them.

good, they're a medic, not an officer

ReynTheLord:
So... why do non antag doctors need a reason to make their medicell guns into lethal weapons, instead of, if crew is arming up, using actual weapons? Sorry to be that guy but this has problems

softcerv:
> > > now we just need to make this standard on the secmedic instead of giving them a pistol
> > 
> > 
> > This is actively worse than a mini-egun or a disabler. This isn't a very good option for them.
> 
> good, they're a medic, not an officer

The design philosophy of them has them as more of a officer with a medikit rather than a medic with a gun. They don't have medical access by default anymore because of that.

softcerv:
> So... why do non antag doctors need a reason to make their medicell guns into lethal weapons, instead of, if crew is arming up, using actual weapons? Sorry to be that guy but this has problems

It's weaponizing something they already have access to, instead of getting them a new gun. 

For instance, there was an on-station threat and you wanted to arm a shaft miner, would it be more efficient to get them a new gun that isn't part of their regular job gear, or give them something that can make the weapon they already have into a somewhat viable tool for on-station combat?

That last remark seems very unneeded.

SpookyTheFox:
I recommend taking a look at other weapons available, and their cost to obtain, as well as their opportunity cost to use in combat. You need to be in the range of 160-300 raw damage per magazine to have a credible self defense weapon, and this has... 60. There's a strong danger here of the weapon being totally pointless because anyone who knows the values will know that it's unconditionally safe to face tank the weapon, and thus the operator takes a total loss on opportunity cost versus any other means of fighting. That's also not enough for PvE/maint mobs, which is a key design metric. I recommend reducing your energy cost to one half or one third of the values you've selected, to get more shots into the gun. Switching the beams to heavy beams might be a solution, but it still comes in a little low.

softcerv:
> I recommend taking a look at other weapons available, and their cost to obtain, as well as their opportunity cost to use in combat. You need to be in the range of 160-300 raw damage per magazine to have a credible self defense weapon, and this has... 60. There's a strong danger here of the weapon being totally pointless because anyone who knows the values will know that it's unconditionally safe to face tank the weapon, and thus the operator takes a total loss on opportunity cost versus any other means of fighting. That's also not enough for PvE/maint mobs, which is a key design metric. I recommend reducing your energy cost to one half or one third of the values you've selected, to get more shots into the gun. Switching the beams to heavy beams might be a solution, but it still comes in a little low.

Thinking about it, I'd be fine nerfing the cost amount a bit. I mostly just want to avoid this becoming a replacement for an actual energy gun, I also have to take into account how effective all of this would be inside of the CMO's energy gun (which auto recharges.) That, and I don't want the gun to excel at both healing and killing

YakumoChen:
I think its fine as it is or with only another shot at best. It's more of a deterrent than anything, it shouldn't be optimal for combat at all. Especially since you don't want to give the CMO an easy AEG. 

Doctors have laser scalpels and such to follow up in combat with, if needed.

softcerv:
> I think its fine as it is or with only another shot at best. It's more of a deterrent than anything, it shouldn't be optimal for combat at all. Especially since you don't want to give the CMO an easy AEG.
> 
> Doctors have laser scalpels and such to follow up in combat with, if needed.

That and a large range of anesthetics to use. Medical can fuck people up if they have the proper momentum to do so.

Draganfrukts:
Heavily discourage a "medical" item from being able to lethal people. You could try something akin to the pepper ball bullets

softcerv:
> Heavily discourage a "medical" item from being able to lethal people. You could try something akin to the pepper ball bullets

A lot of medical items can already be shockingly lethal if used in the right hands. Even items like SmartDarts can be used to kill someone. Especially considering that we have guncargo now, I think this is something that's reasonable to have.

softcerv:
I think I have some better ways that I want to handle this, I am going to close this for now while I figure all of that out.

PR #12246, opened 2022-03-23
ACTUALLY FIX RESTRICTED JOBS ROLLING ANTAGONIST

Gandalf2k15:
Yes. This SHOULD work.

:cl:
add: You should no longer roll antagonist as a restricted role.
/:cl:

PR #12778, opened 2022-04-15
[NON-MODULAR] Hopefully prevents Chaplains from rolling Heretic

theselfish:
## About The Pull Request

Title.
I would have looked into the way https://github.com/Skyrat-SS13/Skyrat-tg/pull/12246 done it. But that seems to prevent them from rolling *any* antag. Which isn't what the ruling said.

image

## Changelog

:cl:
del: Chaplain can no longer be heretical to their own religion. However, that worked.
/:cl:


GoldenAlpharex:
That would work, but we can also just change that in the config ourselves.

theselfish:
> That would work, but we can also just change that in the config ourselves.

Go ahead.

I just done this because of the mention in the ruling.

GoldenAlpharex:
I edited the config so this wouldn't be possible in the future, thanks for the PR anyway :)

PR #12780, opened 2022-04-15
Fixes several issues with the Skyrat maps

Cenrus:
## About The Pull Request

Stops the roundstart AT errors caused by skyrat maps, and eliminates a couple of duplicate APCs where I could find them.

The river in the ghost cafe was causing a lot of AT errors because of the air difference in turfs, this should no longer be an issue.

## Changelog

:cl:
fix: fixed multiple active turf errors
/:cl:


_maps/RandomRuins/LavaRuins/skyrat/lavaland_surface_ash_walker1_skyrat.dmm
```diff
@@ -1615,7 +1615,9 @@
 "WN" = (
 /obj/structure/curtain/bounty,
 /obj/structure/table,
-/turf/open/floor/iron,
+/turf/open/floor/iron{
```
GoldenAlpharex: I feel like this should probably not be just a varedit, it should likely be its own tile subtype, so that it can be re-used in the future, if need be.

_maps/map_files/generic/CentCom_skyrat_z2.dmm
```diff
@@ -6549,6 +6549,11 @@
 /obj/effect/turf_decal/vg_decals/numbers/six,
 /turf/open/floor/iron/dark,
 /area/cruiser_dock)
+"dWZ" = (
+/turf/open/misc/beach/coastline_b{
```
GoldenAlpharex, line 6553: Is there no directional variants that already exist in the code, so you don't have to set the dir yourself? If not, it's fine, I suppose, but still could be nice.

GoldenAlpharex (changes requested):

_maps/RandomRuins/LavaRuins/skyrat/lavaland_surface_ash_walker1_skyrat.dmm
Cenrus: I had a look at other ruin maps and they all seem to use var edited gas mixes. Not a very good practice so I've added a lavaland atmos subtype of iron tiles to the modular directory.

_maps/map_files/generic/CentCom_skyrat_z2.dmm
Cenrus: No directional variants, had to do it myself

PR #12781, opened 2022-04-15
[SEMI-MODULAR] Removes redundant genital toggles + splits up "Uses skintones" in mutiple prefs.

Ghommie:
## About The Pull Request
This PR removes the genital toggle prefs and more. The choiced preferences already have None options so the toggles are really redundant. These choiced prefs have also been refactored, to reduce copyasta. To clarify: having them set to "None" makes sure further options such as length, size, colors won't show up.

Moving on, the "uses skintones" pref has been split into "uses skin tone" and "uses skin color" options - GoldenAlpharex suggested this would have been the better course of action - for each one of those visible organs.

Because of these changes, I have had to code a cheap replica of the savefile conversion present in the non-modular preferences_savefile.dm so that the old char prefs cold be safely converted to the newer ones, smoothly and without unwanted extra bits on some players' charaters. And I say cheap because so far it's only used for char prefs and it has no MIN version define and code, but at least it's neither fly-by-night nor one-shot solution like that tgui migration thing and can be used for future savefile updates.

Currently, we have a single skin tone toggle for all reproductive organs that only changes the icon states used to render these organs to ones with a slightly pinkier hue. Players still have to **manually** set the color of each one to match more or less the tone of their character's skin, and that's no good, considering most players would want their human-er characters to be of a consistent tonality, markings aside.
So, code improvements aside, this PR aims to allow these players to check a box instead of having to guess the color of the character's skin through a color selection input box.

PLEASE DO NOT TESTMERGE THIS. IT TOUCHES SAVEFILES, SO ANY ACCIDENTAL OR NOT UNTESTMERGE MAY SCREW THINGS UP. I ALREADY TESTED THE CONVERSION LOCALLY TO MAKE SURE IT WORKS FINE.

## Changelog

:cl:
refactor: Genital prefs have been refactored. As a result, redundant toggle prefs of these organs have been removed.
qol: "Uses Skin Tone" pref has been split: each visible genital now has its own "Skin Tone" and "Skin Color" pref. This means that human-er characters can now have their bits roughly of the same color of their skin without having to guess it through the color input.
/:cl:


modular_skyrat/master_files/code/modules/client/preferences_savefile.dm
```diff
@@ -1,4 +1,22 @@
...
+ */
+#define MODULAR_SAVEFILE_VERSION_MAX 1
+
+/datum/preferences/proc/savefile_needs_update_skyrat(savefile/S)
+	var/savefile_version
+	READ_FILE(S["modular_version"], savefile_version)
+
+	if(savefile_version < MODULAR_SAVEFILE_VERSION_MAX)
+		return savefile_version
+	return -1
```
GoldenAlpharex: Shouldn't there be a define for the `-1` here? Like, I don't exactly know what it stands for, looking at it quickly.

Documenting the new proc you're adding would also be greatly appreciated.
```suggestion
/datum/preferences/proc/savefile_needs_update_skyrat(savefile/save)
	var/savefile_version
	READ_FILE(save["modular_version"], savefile_version)

	if(savefile_version < MODULAR_SAVEFILE_VERSION_MAX)
		return savefile_version
	return -1
```

modular_skyrat/master_files/code/modules/client/preferences_savefile.dm
```diff
@@ -1,4 +1,22 @@
...
+
 /datum/preferences/proc/load_character_skyrat(savefile/S)
```
GoldenAlpharex: Since you're here, yeet that single-letter variable into oblivion and give it a better name.

modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/genitals.dm
```diff
@@ -22,10 +24,12 @@
...
-	if(gen)
-		return  "[gen.sprite_suffix]"
-	else
-		return null
+	return "[gen?.sprite_suffix]"
+
+/datum/sprite_accessory/genital/get_special_render_colour(mob/living/carbon/human/H, render_state)
+	var/obj/item/organ/genital/gen = H.getorganslot(associated_organ_slot)
+	if(gen?.uses_skin_color && H.dna.species.use_skintones)
+		return skintone2hex(H.skin_tone)
```
GoldenAlpharex: Use a better variable name than `H`.

code/__HELPERS/~skyrat_helpers/unsorted.dm
```diff
@@ -46,3 +46,10 @@
 
 	center_turf.luminosity = old_luminosity
 	return .
+
+///This will check if GLOB.sprite_accessories[mutant_part]?[part_name] is associated with sprite accessory with factual TRUE.
+/proc/is_factual_sprite_accessory(mutant_part, part_name)
+	if(!mutant_part || !part_name)
+		return FALSE
+	var/datum/sprite_accessory/accessory = GLOB.sprite_accessories[mutant_part]?[part_name]
+	return accessory?.factual
```
GoldenAlpharex, line 55: What's a "factual" `sprite_accessory`?

GoldenAlpharex (changes requested):
I'll have to test this on my test server just to be sure that everything works properly.

That being said, are the color options disabled when "Uses Skincolor" is checked? (could be a nice UX feature if they were)

code/__HELPERS/~skyrat_helpers/unsorted.dm
Ghommie: https://github.com/Skyrat-SS13/Skyrat-tg/blob/bf4a8e1f4ec68e72b38d308025c5731fec517468/modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories.dm#L21

GoldenAlpharex: :+1:

Ghommie:
> are the color options disabled when "Uses Skincolor" is checked? (could be a nice UX feature if they were)

Not yet, but I can do that.

Gandalf2k15 (approved):

PR #12807, opened 2022-04-16
[MIRROR] [NO GBP] Fixes turbine heat leaking... again [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/66138
--------------------

## About The Pull Request
Turns out that we needed another isnum() check because the if(checked_conductivity) returns false when the checked_conductivity is equal to zero (a valid amount).
Also fix the issue of not resetting the thermal conductivity to the initial value when the turbine is removed

## Why It's Good For The Game
Properly fixes the issue this time

## Changelog

:cl:
fix: turbines don't leak heat. Properly now
/:cl:


GoldenAlpharex (approved):
Nice.

PR #12822, opened 2022-04-16
Adds back the arrival hangars from before the map reset (for meta and delta)

IsaacTheSharkWolf:
## About The Pull Request

Adds back the old arrivals from before the map reset, no fuss.

Works with the centcom shuttle we have i suppose, just generally something that had to be done but hadn't been done yet.

## Changelog

:cl:
fix: Reverts Delta arrivals back to how they used to be for skyrat before the map reset.
fix: Reverts Meta arrivals back to how they used to be for skyrat before the map reset.
/:cl:


GoldenAlpharex:
Can you just update your branch locally? It should hopefully fix the merge conflicts, considering you've got the hooks.

Gandalf2k15:
remake when you get round to it.

PR #12824, opened 2022-04-16
Headshot removal

tf-4:
## About The Pull Request

You can now remove a headshot, unlike before where it wouldn't save.

I totally didn't add a joke headshot image then get stuck with it for days.

## Changelog
:cl:
fix: Headshots can now properly be removed if you so wish.
/:cl:


Zonespace27:
Good catch.

PR #12854, opened 2022-04-17
Donator item edits.

Drifter7371:
This is pretty on the nose.


Zonespace27:
Confirmed with donator that they're okay with this

PR #12862, opened 2022-04-18
[MIRROR] tgui: API improvements + docs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/65943
--------------------
## About The Pull Request

This pull request improves tgui API in many ways.

[Using TGUI for custom HTML popups](https://github.com/stylemistake/tgstation/blob/c390c74cbe478570fe3ea65a9eb433e6364d7a45/tgui/docs/tgui-for-custom-html-popups.md)

This standardizes and simplifies the process of HTML popup creation and `DM <-> JS` communication.

Makes tgui window API a perfect alternative for old-style browser panels. It will be super useful for @ Iamgoofball since he wanted to make a lightweight browser element that plays background music, and this will make his life a lot easier.

It is now possible to create tgui windows with fully inlined JS and CSS, which can be used to make unkillable tgui-based UIs (can't white/blue screen due to network errors). You can split files into JS and CSS, and still serve a single HTML file using this.

Moved `sendMessage` function to the `Byond` API object, where it rightfully belongs, and now supports a shorthand form: `Byond.sendMessage(type, payload)`. This shortens and simplifies a lot of code.

Refactored `window.update` to no longer be public. Now to subscribe to incoming messages, you should use new public APIs: `Byond.subscribe(fn)` and `Byond.subscribeTo(type, fn)`, and TGUI internally uses these functions as well, which reduces boilerplate in `index.js`.

Renamed `window.__windowId__` to `Byond.windowId` (old variable is still available for backwards compatibility).

Byond API now supports `null` id, e.g. `Byond.winget(null, 'url')`, which makes things like this possible:

```js
// Fetch URL of a currently connected server
Byond.winget(null, 'url').then((serverUrl) => {
  // Connect to this server (opens a new dreamseeker instance)
  Byond.call(serverUrl);
  // Close this client because new instance is connecting
  Byond.command('.quit');
});
```

Certain polyfills are now statically compiled (commited into git) and are baked into `tgui.html`. The downside is that HTML went `16 kB -> 50 kB`. The upside is that you can now use a relatively modern level API with full support for IE8 when writing plain old html UIs using `/datum/tgui_window` directly. They are committed into git, because polyfills will never need to be updated (unless of course we randomly decide to get rid of `ie8.js` and `html5shiv.js`).

## Breaking Changes

No breaking changes. This should be tested for regressions. Upgrading is simple if you're on a relatively up-to-date branch - copy paste all affected tgui files and you're good.


stylemistake:
Interesting ping fix. Proper fix is going to be merged soon upstream, so make sure to undo this edit.

PR #12937, opened 2022-04-20
Gives more people access to my own donator item

zydras:
## About The Pull Request

Like the title says, this gives more people access to my own donator item. With permission from Useroth, of course. Three extra names. That's about it.


Useroth:
One day we need to formalize the donator stuff

PR #12945, opened 2022-04-21
jaunters will now work just being on the person

jjpark-kb:
## About The Pull Request
jaunters will now protect you from chasms just being on your person (in your bag, your pocket, etc.) rather than when its only on your belt.

Chasms are the bane of people... jaunters really should be used more to mitigate this frustration.

## Changelog

:cl:
qol: jaunters now work while being on your person (so backpack, pocket, etc.), rather than just in the belt slot (for chasms)
/:cl:

### DIFF
--- a/code/datums/components/chasm.dm
+++ b/code/datums/components/chasm.dm
@@ -83,6 +83,7 @@
 				return FALSE
 		if(ishuman(AM))
 			var/mob/living/carbon/human/victim = AM
+			//SKYRAT EDIT: Making Jaunters Better
 			var/obj/item/wormhole_jaunter/found_jaunter = locate() in victim.get_all_contents()
 			if(!found_jaunter)
 				return TRUE
@@ -99,7 +100,7 @@
 				if(!fall_into_chasm)
 					chasm.visible_message(span_boldwarning("[victim] falls into the [chasm]!")) //To freak out any bystanders
 				return fall_into_chasm
-			*/
+			SKYRAT EDIT: Making Jaunters Better */
 	return TRUE
 
 /datum/component/chasm/proc/drop(atom/movable/AM)
--- a/code/datums/components/chasm.dm
+++ b/code/datums/components/chasm.dm
@@ -83,6 +83,15 @@
 				return FALSE
 		if(ishuman(AM))
 			var/mob/living/carbon/human/victim = AM
+			var/obj/item/wormhole_jaunter/found_jaunter = locate() in victim.get_all_contents()
+			if(!found_jaunter)
+				return TRUE
+			var/turf/chasm = get_turf(victim)
+			var/fall_into_chasm = found_jaunter.chasm_react(victim)
+			if(!fall_into_chasm)
+				chasm.visible_message(span_boldwarning("[victim] falls into the [chasm]!")) //To freak out any bystanders
+			return fall_into_chasm
+			/* SKYRAT EDIT: Original Code
 			if(istype(victim.belt, /obj/item/wormhole_jaunter))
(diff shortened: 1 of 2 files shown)


code/datums/components/chasm.dm
```diff
@@ -83,13 +83,23 @@
 				return FALSE
 		if(ishuman(AM))
 			var/mob/living/carbon/human/victim = AM
+			var/obj/item/wormhole_jaunter/found_jaunter = locate() in victim.get_all_contents()
```
Zonespace27, line 87: skyrat edit comments

Zergspower:
Praise be - makes them actually useful to purchase now too

YakumoChen:
:-1:
They only work on the belt so that using one over using another belt item is a conscious choice instead of an afterthought 

Also take it upstream

jjpark-kb:
> 👎 They only work on the belt so that using one over using another belt item is a conscious choice instead of an afterthought

I realize that there was a conscious decision to do this for balance sake, but they are so under-utilized, it is crazy. It is the solution to chasms, but because of how awful these are people are complaining more and more about chasms. I would rather us make the solutions to chasms better rather than making chasms baby-proofed.

> Also take it upstream

I think this is far from your jurisdiction; you do not have the ability/right to tell anyone to do that.

code/modules/mining/equipment/wormhole_jaunter.dm
```diff
@@ -1,7 +1,8 @@
...
 /obj/item/wormhole_jaunter
 	name = "wormhole jaunter"
-	desc = "A single use device harnessing outdated wormhole technology, Nanotrasen has since turned its eyes to bluespace for more accurate teleportation. The wormholes it creates are unpleasant to travel through, to say the least.\nThanks to modifications provided by the Free Golems, this jaunter can be worn on the belt to provide protection from chasms."
+	//desc = "A single use device harnessing outdated wormhole technology, Nanotrasen has since turned its eyes to bluespace for more accurate teleportation. The wormholes it creates are unpleasant to travel through, to say the least.\nThanks to modifications provided by the Free Golems, this jaunter can be worn on the belt to provide protection from chasms." SKYRAT EDIT: Original
+	desc = "A single use device harnessing outdated wormhole technology, Nanotrasen has since turned its eyes to bluespace for more accurate teleportation. The wormholes it creates are unpleasant to travel through, to say the least.\nThanks to modifications provided by the Free Golems, so long as you have the jaunter on your person it will provide protection from chasms."
```
GoldenAlpharex, line 5:
```suggestion
	desc = "A single use device harnessing outdated wormhole technology, Nanotrasen has since turned its eyes to bluespace for more accurate teleportation. The wormholes it creates are unpleasant to travel through, to say the least.\nThanks to modifications provided by the Free Golems, so long as you have the jaunter on your person, it will provide protection from chasms." // SKYRAT EDIT - ORIGINAL: desc = "A single use device harnessing outdated wormhole technology, Nanotrasen has since turned its eyes to bluespace for more accurate teleportation. The wormholes it creates are unpleasant to travel through, to say the least.\nThanks to modifications provided by the Free Golems, this jaunter can be worn on the belt to provide protection from chasms."
```

GoldenAlpharex (changes requested):
If upstream doesn't want it, honestly, I'm all for it. Chasms are stupid RR anyway, and just in general both lag and noob traps.

Does the jaunter have infinite uses, or is it only usable once? Because if it's the latter, then I'm even more for it, if it's the former, then we could add a restricted amount of uses before you need to get another one (probably just one).

jjpark-kb:
gonna be real, the "take it upstream" annoyed me too much-- I am not taking it upstream nor here I guess.

PR #12954, opened 2022-04-21
High heels sounds

NX115:
## About The Pull Request

     Added, to high heels, the sound of heels striking the ground during the walking cycle. Copied over from dominatrix heels.

     Helps fix the weird sound disparity between the heel sets.

## Changelog

:cl: NX115
expansion: sounds to high heels walking
soundadd: reused sound from dominatrix heels
/:cl:


YakumoChen:
Could you lower the volume on it for the regular heels to be more in line with regular walking sounds? 

I use both sets of heels and I don't think I'd want to wear them on my non-horny character if it comes with very loud, annoying walking sounds associated with an ERP item.

PR #12956, opened 2022-04-21
Ghost role appearance change restriction

tf-4:
## About The Pull Request

Disables selecting your custom character when spawning into a ghost role by default. Enables it on ghost roles that should have it.

## Changelog
:cl:
fix: You can no longer turn into your selected character when spawning as a ghost role that it doesn't make sense for.
/:cl:

### DIFF
--- a/code/modules/mob_spawn/ghost_roles/mining_roles.dm
+++ b/code/modules/mob_spawn/ghost_roles/mining_roles.dm
@@ -15,6 +15,7 @@
 	the fresh air of Earth. These thoughts are dispelled by yet another recollection of how you got here... "
 	spawner_job_path = /datum/job/hermit
 	quirks_enabled = TRUE // SKYRAT EDIT ADDITION - ghost role loadouts
+	random_appearance = FALSE // SKYRAT EDIT ADDITION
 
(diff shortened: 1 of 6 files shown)


SarmentiCampbell:
HECU and scientists too, The randomized character is legit horrid, i do not want to have a gorilla scream with a felinid laugh

Gandalf2k15 (changes requested):
Enable HECU and black mesa roles though

YakumoChen:
HECU and Scientists should be random, they should be pure human with no mutant parts. Same goes for Enclave.

Gandalf2k15:
> HECU and Scientists should be random, they should be pure human with no mutant parts. Same goes for Enclave.

They are already species restricted.

YakumoChen:
> > HECU and Scientists should be random, they should be pure human with no mutant parts. Same goes for Enclave.
> 
> They are already species restricted.

Does that prevent you from using mismatch parts?

Gandalf2k15:
> > > HECU and Scientists should be random, they should be pure human with no mutant parts. Same goes for Enclave.
> > 
> > 
> > They are already species restricted.
> 
> Does that prevent you from using mismatch parts?

no but it sure as hell helps to prevent it

PR #12997, opened 2022-04-22
[MIRROR] Removes canister tiers, now canisters have an option to turn on shielding for power consumption [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/66210
--------------------

## About The Pull Request
Canister tiers were first introduced by me when canister fusion was still a thing. Now they have become only a nuisance and tedium since the main reason they were added is gone. current limits are higher than old t1 but lower than old t2
I've now added an option in the UI to every canister to become shielded and consume the APC power so that its contents won't be able to destroy it, the higher the temperature and pressure inside, the higher the draw (it should peak around 25 kW)
Canisters now have a cell that can hold the shielding in areas without power for a time, can be upgraded, removed, replaced

## Why It's Good For The Game
Increases players (old and especially new) quality of life, removes tedium and follows the oranges will of canisters should draw power to contain high heat and pressure

## Changelog

:cl: Ghilker, shield sprite by @ SnoopCooper
remove: remove canister tier, you still need to build a frame and use iron on it but nothing more.
expansion: canisters now have a toggle for shielding that will start to consume power if the temperature and pressure go over the labeled one (current limits are higher than old t1 but lower than old t2). Peak draw should be around 25 kW with fusion temperatures. Canisters have an internal cell that allows it to hold shielding in areas without power for a time, powercells can be upgraded, removed, replaced.
/:cl:


Zergspower:
ok legit concern - IPC's Synths draw power from an APC and if this peaks at 25kw.. we already knock out lights for a few seconds each time we eat

Zergspower:
> 

the built in battery does give it a bit of time - but that was at room temp nitrogen in my test - whew

PR #13033, opened 2022-04-23
Increases cash briefcase cost

tf-4:
## About The Pull Request

disables cash briefcase pending arconomy

Spending like 2 TC for any gun in gun cargo is a bit much

## Changelog
:cl:
balance: the cash briefcase now costs 5 TC instead of 1
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/badass.dm
+++ b/code/modules/uplink/uplink_items/badass.dm
@@ -34,6 +34,7 @@
 
 // Low progression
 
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (changes requested):
I personally feel like just upping the price might be a better alternative for now, so we don't just strip out content entirely.

Zonespace27:
Make it 5 TC to be on-par with arconomy when it hits

GoldenAlpharex (approved):
Looks good to me.

PR #13085, opened 2022-04-25
Until Skyrat's host faces punishment for the way he treats his staff, his developers, and his players, I will be closing all of my PRs. After what came to light today on the ss13 reddit, I am not comfortable with developing for Skyrat anymore.

Iamgoofball:
# I am not comfortable with my work being used by the host of Skyrat after what he did to players, staff, and developers came to light today on the ss13 reddit. As such, I am closing all of my PRs until change occurs and I feel comfortable working here again. I apologize to any players looking forward to these features.

Crab-17 is supposed to only be a once a round purchase, but after the upstream rework of how uplink items work this feature broke and different uplinks can buy multiple crab-17 phones. 

I've opted to patch this issue temporarily down here via adding a global check for whether the economy has already crashed once until upstream fixes this issue. I've alerted the dev who broke it in the past, but they likely forgot as they've been busy with other projects, so I'll ask them about it again.

This prevents double dumping of the economy if a dump has already occurred.
:cl:
fix: CRAB-17 only works once a round now.
/:cl:

PR #13086, opened 2022-04-25
The rave of the crab (Bye bye CRAB-17)

GoldenAlpharex:
## About The Pull Request
It's run its course, it's not particularly funny, it doesn't add much to the experience, spams the chat of all of those that simply don't want to deal with it, and also creates a bunch of GC failures that I honestly don't feel like investigating, for a lack of caring for the feature in general.

Closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/13085, because it's not necessary as we wait for a fix from upstream.

Honestly, I can see the appeal for /tg/. But on here? Not a lot of roleplay comes out of worming your way in to tap a machine with your card and go back on your merry way. It might have *potential* for roleplay, but that potential basically never becomes reality. It's just... Boring, really.

## Changelog

:cl: GoldenAlpharex
del: Nanotrasen managed to stabilize its assets, so much so that market crashes related to massive asset movements are no longer possible, even through crab-related phones.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/device_tools.dm
+++ b/code/modules/uplink/uplink_items/device_tools.dm
@@ -151,6 +151,7 @@
 	cost = 3
 
 
+/* SKYRAT EDIT REMOVAL - It's laggy and doesn't really add much roleplay value
 /datum/uplink_item/device_tools/suspiciousphone
 	name = "Protocol CRAB-17 Phone"
 	desc = "The Protocol CRAB-17 Phone, a phone borrowed from an unknown third party, it can be used to crash the space market, funneling the losses of the crew to your bank account.\
@@ -159,6 +160,7 @@
 	restricted = TRUE
 	cost = 7
 	limited_stock = 1
+*/ // SKYRAT EDIT REMOVAL END
 
 /datum/uplink_item/device_tools/binary
 	name = "Binary Translator Key"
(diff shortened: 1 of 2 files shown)


softcerv:
This is a good change, especially because It is incredibly difficult to deal with the CRAB-17 on Blueshift.

tf-4:
It seemed to be re-enabled entirely as an oversight from progtot so good change

theselfish:
The purpose of the CRAB-17 is to distract the crew while a traitor does something.

Because *everyone* goes to the C-17.

Zergspower:
> The purpose of the CRAB-17 is to distract the crew while a traitor does something.
> 
> Because _everyone_ goes to the C-17.

sadly it's never used like that, and only serves to just cause 40-50 people screaming on the radio about where is it. Causing all current RP situations to cease as we run there

softcerv:
> The purpose of the CRAB-17 is to distract the crew while a traitor does something.
> 
> Because _everyone_ goes to the C-17.

The crab 17 also just directly filters that money into the account of however set it up, if I remember right. Which is a very powerful ability considering we now have guncargo.

Tastyfish:
Good riddance. Money for most crew members it tight enough as it is without losing half of it before it's physically possible to get to the machine.

Gandalf2k15:
fibe

PR #13173, opened 2022-04-27
[NON MODULAR] Changes the gas experiments to discounts, instead of requirements

RatFromTheJungle:
## About The Pull Request

Changes the gas experiments (all) to discounts, as opposed to hard requirements  with as little edits to the files as possible, namely, because after sitting on it for a good long while, I don't really think there's been a benefit from it at all, besides barring previously available non-issue causing content behind an atmos-knowledge gate that i've struggled to find anyone who knows it, that would even want to play sci.

Idea stolen from (https://github.com/tgstation/tgstation/pull/65933)

what it says on the tin.

## Changelog

:cl:
balance: Made the gas experiments discounts as opposed to hard requirements.
/:cl:

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -1013,7 +1013,7 @@
 	)
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 5000)
 	discount_experiments = list(/datum/experiment/scanning/points/machinery_pinpoint_scan/tier3_microlaser = 4000)
-	required_experiments = list(/datum/experiment/ordnance/gaseous/noblium)
+	// required_experiments = list(/datum/experiment/ordnance/gaseous/noblium) SKYRAT EDIT - Commented out
 
 /////////////////////////Clown tech/////////////////////////
 /datum/techweb_node/clown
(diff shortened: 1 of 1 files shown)


Ghilker:
goddamn this is both lame and he worst way to deal with this

YakumoChen:
just make them discounts

ArdentSpark:
if the experiments are 'too hard' then hyper-nob and nitrium can be swapped between T4 lasers and the implant tech node
because hyper-nob is way easier to make with T4 lasers, and nitrium can be made trivially by just, combining the superheated tritium you harvest from the ordnance chamber (which is a very very basic requirement of ordnance techs anyway) with nitrogen and BZ (nitrogen is in gas storage, BZ is in xenobiology)

GoldenAlpharex (changes requested):
I think I'd prefer you bumping the prices of those techweb nodes by a certain amount and making those experiments provide a discount instead of just flat-out removing them.

YakumoChen:
PR about to be made redundant https://github.com/tgstation/tgstation/pull/66738

Gandalf2k15:
https://github.com/tgstation/tgstation/pull/66738

PR #13176, opened 2022-04-27
Disables time track profiling temporarily

Gandalf2k15:
Just disabled it till we fix the profliling. Might be a source of lag.


Tastyfish:
It would be hilarious if the time tracker caused significant lag.

Gandalf2k15:
probably not needed anymore

PR #13180, opened 2022-04-27
Removes the lathe tax from off-station lathes

Zonespace27:
## About The Pull Request
DS-2 and such can't use their lathes because they don't have bank accounts.

Using them without having to spawn in credits is a good thing

## Changelog

:cl:
fix: Off-station lathes are now free to use
/:cl:


Gandalf2k15:
Upstreamies

PR #13199, opened 2022-04-28
Adds suspenders to autodrobe.

SapphoQueer:
## About The Pull Request
Simple as the title says, the rest of the outfit the mime starts with is all in the autodrobe before this PR, except suspenders, this adds them.

Well, I suppose this brings people the _Full immersive mime experience_ if they decide to dress up as one, or lets a roundstart mime replace suspenders if they get lost.

## Changelog

:cl:
add: Adds suspenders to the autodrobe.
/:cl:

### DIFF
--- a/code/modules/vending/autodrobe.dm
+++ b/code/modules/vending/autodrobe.dm
@@ -75,6 +75,7 @@
 		/obj/item/clothing/mask/gas/sexyclown = 1,
 		/obj/item/clothing/under/rank/civilian/clown/sexy = 1,
 		/obj/item/clothing/mask/gas/sexymime = 1,
+		/obj/item/clothing/suit/toggle/suspenders = 1,
 		/obj/item/clothing/under/rank/civilian/mime/sexy = 1,
 		/obj/item/clothing/under/rank/civilian/mime/skirt = 1,
 		/obj/item/clothing/mask/animal/rat/bat = 1,


code/modules/vending/autodrobe.dm
```diff
@@ -75,6 +75,7 @@
 		/obj/item/clothing/mask/gas/sexyclown = 1,
 		/obj/item/clothing/under/rank/civilian/clown/sexy = 1,
 		/obj/item/clothing/mask/gas/sexymime = 1,
+		/obj/item/clothing/suit/toggle/suspenders = 1,
```
Zonespace27, line 78: Modularize this by putting it in the autodrobe's `skyrat_products` list in the modular_skyrat folder wherever that is defined

Zonespace27 (changes requested):
above

YakumoChen:
I think those come in the standard costume crate, available from cargo. Maybe we should keep incentive to order the crate? For stuff other than just the clown costume, at least.

theselfish:
> I think those come in the standard costume crate, available from cargo. Maybe we should keep incentive to order the crate? For stuff other than just the clown costume, at least.

They do. But honestly?
Not only can you get suspenders in the loadout menu, not many people would be buying them.
If you get a clown crate, you get it for the clown stuff, or the mime mask.

It's also locked to theater access.

PR #13223, opened 2022-04-30
Synthetics mirroring other mutants pay

Zergspower:
*beeps

## About The Pull Request

Updates payrate for synthetics to match the rest of the non-human crew

With the new paycheck system in place, this change is rather needed as the pentalty was far more impactful.

## Changelog

Changes the payrate of synthetics from .50 to .75 to match the other races that are non-human.

:cl:
balance: Makes synthetic species have a pay modifier that matches other mutant non-human races.
/:cl:


GoldenAlpharex (approved):
I'm fine with this, especially considering how we're trying to make economy more relevant, I don't think it's good to hold them back from a sizeable part of the game without much thought put into it.

Zonespace27 (approved):
As much as I disagree with reducing paychecks anyway, this is a step in the right direction

PR #13235, opened 2022-04-30
Makes examine block HR's format better collapse

Tastyfish:
## About The Pull Request

- Makes it so consecutive `<hr>` breaks in an examine block are reduced to a single one, and a block and start or end with one.
- Also eliminated usage of `findtext()` in favor of literal comparison in examinate.
- Also bulleted reagent formatting, and added a break between carbon clothing and all the status stuff.

image
image

### Preview of PR this is for:

image

better formatting

## Changelog

:cl:
refactor: Examine block HR section breaks are now handled such that they don't repeat or show up at the bottom.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -636,7 +636,7 @@
 
 	if(custom_materials)
 		// SKYRAT EDIT ADDITION BEGIN - HR sections
-		if(custom_materials.len > 1)
+		if(length(custom_materials) > 1)
 			. += EXAMINE_SECTION_BREAK //SKYRAT EDIT ADDITION
 		//SKYRAT EDIT ADDITION END
 
@@ -646,7 +646,7 @@
 		. += "<u>It is made out of [english_list(materials_list)]</u>."
 
 		// SKYRAT EDIT ADDITION BEGIN - HR sections
-		if(custom_materials.len > 1)
+		if(length(custom_materials) > 1)
 			. += EXAMINE_SECTION_BREAK //SKYRAT EDIT ADDITION
 		//SKYRAT EDIT ADDITION END
 	if(reagents)
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -628,28 +628,37 @@
  * Produces a signal [COMSIG_PARENT_EXAMINE]
  */
 /atom/proc/examine(mob/user)
-	. = list("[get_examine_string(user, TRUE)].<hr>")
+	. = list("[get_examine_string(user, TRUE)].", EXAMINE_SECTION_BREAK) // SKYRAT EDIT CHANGE - original: list("[get_examine_string(user, TRUE)].")
 
 	. += get_name_chaser(user)
 	if(desc)
 		. += desc
 
 	if(custom_materials)
-		. += "<hr>" //SKYRAT EDIT ADDITION
+		// SKYRAT EDIT ADDITION BEGIN - HR sections
+		if(custom_materials.len > 1)
+			. += EXAMINE_SECTION_BREAK //SKYRAT EDIT ADDITION
+		//SKYRAT EDIT ADDITION END
+
 		var/list/materials_list = list()
 		for(var/datum/material/current_material as anything in custom_materials)
 			materials_list += "[current_material.name]"
 		. += "<u>It is made out of [english_list(materials_list)]</u>."
+
+		// SKYRAT EDIT ADDITION BEGIN - HR sections
+		if(custom_materials.len > 1)
+			. += EXAMINE_SECTION_BREAK //SKYRAT EDIT ADDITION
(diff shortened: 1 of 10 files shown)


theselfish:
Why not do this upstream, if you haven't?

code/game/atoms.dm
```diff
@@ -628,28 +628,37 @@
...
+	. = list("[get_examine_string(user, TRUE)].", EXAMINE_SECTION_BREAK) // SKYRAT EDIT CHANGE - original: list("[get_examine_string(user, TRUE)].")
 
 	. += get_name_chaser(user)
 	if(desc)
 		. += desc
 
 	if(custom_materials)
-		. += "<hr>" //SKYRAT EDIT ADDITION
+		// SKYRAT EDIT ADDITION BEGIN - HR sections
+		if(custom_materials.len > 1)
```
Zonespace27: use `length() over .len`

code/game/atoms.dm
```diff
@@ -628,28 +628,37 @@
...
+			. += EXAMINE_SECTION_BREAK //SKYRAT EDIT ADDITION
+		//SKYRAT EDIT ADDITION END
+
 		var/list/materials_list = list()
 		for(var/datum/material/current_material as anything in custom_materials)
 			materials_list += "[current_material.name]"
 		. += "<u>It is made out of [english_list(materials_list)]</u>."
+
+		// SKYRAT EDIT ADDITION BEGIN - HR sections
+		if(custom_materials.len > 1)
```
Zonespace27: use `length() over .len`

code/modules/mob/mob.dm
```diff
@@ -494,8 +494,13 @@
 	//SKYRAT EDIT ADDITION
 	if(result.len)
 		for(var/i = 1, i <= result.len, i++)
-			if(!findtext(result[i], "<hr>"))
+			if(result[i] != EXAMINE_SECTION_BREAK)
 				result[i] += "\n"
+			else
+				// remove repeated <hr's> and ones on the ends.
+				if(i == 1 || i == result.len || result[i - 1] == EXAMINE_SECTION_BREAK)
```
Zonespace27: use `length() over .len`

Zonespace27 (changes requested):

Tastyfish:
> Why not do this upstream, if you haven't?

Upstream doesn't use HR's in examine text at all.

code/game/atoms.dm
Tastyfish: done

code/game/atoms.dm
Tastyfish: done

code/modules/mob/mob.dm
Tastyfish: done

code/modules/mob/mob.dm
```diff
@@ -493,9 +493,14 @@
 
 	//SKYRAT EDIT ADDITION
 	if(result.len)
-		for(var/i = 1, i <= result.len, i++)
-			if(!findtext(result[i], "<hr>"))
+		for(var/i = 1, i <= length(result), i++)
```
Zonespace27: use `for(var/i in startnum to endnum)` formatting, in this case it would be `for(var/i in 1 to length(result))`

code/modules/mob/mob.dm
```diff
@@ -493,9 +493,14 @@
...
+			if(result[i] != EXAMINE_SECTION_BREAK)
 				result[i] += "\n"
+			else
+				// remove repeated <hr's> and ones on the ends.
+				if(i == 1 || i == length(result) || result[i - 1] == EXAMINE_SECTION_BREAK)
```
Zonespace27:
```suggestion
				if(i == 1 || (i == length(result)) || (result[i - 1] == EXAMINE_SECTION_BREAK))
```
readability

Zonespace27 (changes requested):

code/modules/mob/mob.dm
Tastyfish: I actually wasn't sure if this would work with the `i--` rewinding but amazingly it does. done.

code/modules/mob/mob.dm
Tastyfish: done.

Gandalf2k15:
How does this handle things that you haven't set to use the examine define? Like say something that has lots of additional examine stuff

Tastyfish:
> How does this handle things that you haven't set to use the examine define? Like say something that has lots of additional examine stuff

I believe I've covered every current use of `<hr>`, as well a couple of other cases that made sense to me that currently are just a giant block of text. I could add more if you want specific cases of things being broken up, but otherwise everything else still displays things as normal, as default examine text rendering isn't affected.

PR #13252, opened 2022-05-01
[MODULAR] Gives departmental gurads access to the eyepatch HUDs for their departments.

softcerv:
## About The Pull Request
Gives the departmental guards the ability to use the eyepatch HUDs in the loadout for their departments. The only exception being the mesons, the engineering departmental guard already had access to these. 

This is something that's useful for departmental guards to have so that they can help better assist their department without having to steal a HUD. It also felt inconsistent that every single member of a department could get these items besides the departmental guards.

## Changelog

:cl:
fix: departmental guards can now use the eyepatch HUDs for their departments in the loadout menu.


Iamgoofball:
they're actually issued security huds(not hudglasses) in their lockers, but yeah giving them access to departmental huds is good too

theselfish:
> they're actually issued security huds(not hudglasses) in their lockers, but yeah giving them access to departmental huds is good too

As the patches have no flash resistance, this PR is fine.

tf-4:
Seems more like an oversight tbh

softcerv:
> Seems more like an oversight tbh

sounds good to me, the changelog has been changed.

PR #3223, opened 2021-02-09
Makes Quad-Sec only work inside /area/security

MLGTASTICa:
## About The Pull Request
Makes quad-sec only have its healing effects apply in security areas
## Why It's Good For The Game
Security uses quad-sec to powergame mid-combat because it has no downsides that other cobbychems have to prevent powergaming.

## Changelog
:cl:
balance: Quadruple Security now only heals inside security areas
/:cl:

PR #13275, opened 2022-05-02
Revival of #3223 (Makes Quad-Sec only work inside /area/security)

Cenrus:
## About The Pull Request

So at some point #3223 was reverted at some point, this implements it back. For those who don't know, this PR made quadruple sec only work in security areas to prevent it from being passive healing with no adverse side effects in combat scenarios.

See original PR for their reasoning

## Changelog

:cl: MLGTASTICa
balance: Quadruple Security now only heals inside security areas
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -1732,8 +1732,10 @@ All effects don't start immediately, but rather get worse over time; the rate is
 	//Securidrink in line with the Screwdriver for engineers or Nothing for mimes
 	var/obj/item/organ/liver/liver = drinker.getorganslot(ORGAN_SLOT_LIVER)
 	if(liver && HAS_TRAIT(liver, TRAIT_LAW_ENFORCEMENT_METABOLISM))
-		drinker.heal_bodypart_damage(1 * REM * delta_time, 1 * REM * delta_time)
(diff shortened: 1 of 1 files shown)


desertmote:
original reasoning was wrong and also really old.
you need to give a reason thats not "no downsides" because that is incorrect.

please explain why this fits our current (not years ago like original PR) environment

capsaicinz:
would you like to remove all other general healing chemicals that you can sip away in the middle of combat as well? just the other day, i had to deal with somebody drinking meth every time i started chasing them as sec. people can drink cocktails of healing chems far more effective than quadsec for hardly more work. it ISN’T that much healing. it ISN’T an issue. and it DOES have a downside. quadsec will get you drunk, which isn’t good for combat at all— screen shaking, vomiting if you drink too much. i don’t know why you think this is a good idea, but it isn’t.

Tbowes-dev:
The Quad-sec healing is very very minor in comparison to the risk of being drunk, which isnt something you want in combat with the vomiting.
A guard could get a brute and burn pill from chems and be much more safe and heal way more than what flavorful drinks could offer them

Gandalf2k15:
Things have changed, I'm not for this inconsequential change that brings non-modular codeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #13277, opened 2022-05-02
Makes prefs apply correctly again

ZephyrTFA:
Hey, the is_accessible proc checks a lot of things, most notably it checks that the character select menu is open and on the correct page. This little if statement here makes it so half, or all if the menu isn't open, don't even apply to you. You need to actually test your shit.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -563,8 +563,7 @@ INITIALIZE_IMMEDIATE(/atom/movable/screen/character_preview_view)
 		if (preference.savefile_identifier != PREFERENCE_CHARACTER)
 			continue
 	// SKYRAT EDIT
-		if(preference.is_accessible(src)) // Only apply preferences you can actually access.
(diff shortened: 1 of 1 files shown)


Gandalf2k15 (changes requested):
This will likely cause a lot of unintended behavior because of how is_accessible is incorrectly used. An example: You have taur body unchecked, but within, you have it set up, is_accessible is used to determine if mismatched parts are used.

Gandalf2k15:
So refactor the way its used, add a new function to is accessible. This is not the way to do it

ZephyrTFA:
> So refactor the way its used, add a new function to is accessible. This is not the way to do it

Your bad modular code is outside the scope of my patience

Gandalf2k15:
> > So refactor the way its used, add a new function to is accessible. This is not the way to do it
> 
> Your bad modular code is outside the scope of my patience

It's not my code.

PR #13288, opened 2022-05-03
Fixes teshari and vox clothing (again)

Tastyfish:
## About The Pull Request

Teshari must be protected at all costs.

- Restores BODYTYPE_CUSTOM clothing functionalit  by letting it set mutant_override *if* it provides a custom icon.
  - The logic is actually marginally better than before. Might have fixed a couple of vox edge cases, but I don't remember which exact items they happened with.
- Restores functionalty of ignoring offsets if the icon's already been `mutant_override`n, as the overriden icon should already match the mutant bodypart it was sprited for (teshari and vox sprites broadly rely on this).

bird

## Changelog

:cl:
fix: Fixes teshari again.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_update_icons.dm
+++ b/code/modules/mob/living/carbon/human/human_update_icons.dm
@@ -159,14 +159,18 @@ There are several things that need to be remembered:
 
 		var/icon_file
 		var/woman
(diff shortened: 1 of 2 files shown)


Weblure:
Bless our teshari savoir

PR #13295, opened 2022-05-03
[Semi-Modular] Feline quirk gets a say_mod

desertmote:
## About The Pull Request

Its the same as what felinids get.

Let me know if I have to do it without adding a new tongue subtype.

## Changelog

:cl:
add: Having the feline quirk modifies your speech to say 'meows, "Nya!"'
/:cl:

### DIFF
--- a/code/modules/surgery/organs/tongue.dm
+++ b/code/modules/surgery/organs/tongue.dm
@@ -518,6 +518,13 @@
 		signer.visible_message(span_notice("[signer] lowers [signer.p_their()] eyebrows."))
 
 //SKYRAT MODULAR EDIT BEGIN
+/obj/item/organ/tongue/cat
+	name = "coarse tongue"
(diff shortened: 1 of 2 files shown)


GoldenAlpharex:
Please do it without adding a new tongue.

Instead, just change the say_mod of the tongue so that it does what you want it to do on gain, and reset it to `initial(tongue.say_mode)` on loss.

We have species that depend on their tongue for special behavior, I'd rather avoid handling say_mods from quirks with a new tongue if I can.

GoldenAlpharex (changes requested):
As I said, make it modify the quirk holder's tongue's say_mod on gain, and revert it to its original value on loss.

Thanks!

GoldenAlpharex:
Just ask for this to be re-opened if you want to finish working on it, or just re-open a PR later ;)

PR #13302, opened 2022-05-04
[MIRROR] Removes some unnecessary species mentions from kapuparts [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/66112
--------------------

## About The Pull Request
#65523 made some changes to bodypart names that messed with the grammar of a few messages around the codebase by inserting the species of a bodypart where it's really not needed. For example:

For most general combat/health cases, we don't need to worry about the species of a bodypart. So, this changes combat and wound messages back to simply displaying the relevant bodypart zone being attacked/affected. See below:

[](https://i.imgur.com/nrGXvyK.png)

Let me know if I missed any!

## Why It's Good For The Game
Better grammar

## Changelog

:cl: Ryll/Shaps
spellcheck: Wound and combat messages will no longer mention the species of an attacked bodypart when not appropriate
/:cl:

PR #13310, opened 2022-05-04
[MODULAR] Let us know what limbs we have. 

nevimer:
## About The Pull Request

Undoes the the user-facing effects of https://github.com/Skyrat-SS13/Skyrat-tg/pull/13302

While TGstation may not care what limbs we have, it's a good idea for us given we have TONS of limb types all with different behavior, commonly found. Especially common with synths. 
dreamseeker_xFWWZOtgMh

Flavourful + Helps medical / robotics / players. 

## Changelog

:cl:
code: Overrides the limb-name var from generic to origin-species on initialize. 
/:cl:

### DIFF
--- a/code/modules/surgery/bodyparts/xeno_part_names.dm
+++ b/code/modules/surgery/bodyparts/xeno_part_names.dm
@@ -1,0 +1,3 @@
+/obj/item/bodypart/Initialize()
+	. = ..()
+	plaintext_zone = parse_zone(body_zone)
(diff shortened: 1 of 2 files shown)


YakumoChen:
image

YakumoChen:
the PR you're reverting doesn't actually affect self-examine part names, just combat logs and wounds.

GoldenAlpharex:
I really don't think reverting this is a good thing. At all. It's just more clutter in combat logs, for no actual gain, since that information is already available.

nevimer:
> I really don't think reverting this is a good thing. At all. It's just more clutter in combat logs, for no actual gain, since that information is already available.

Then we should add a way to inspect other peoples limbs, as medical scanners do not show what limbs a person has, nor do visual inspections.

PR #13311, opened 2022-05-04
[NON-MODULAR] Reduces the required rep in objectives for getting a final objective spawn. 

Higgin:
## About The Pull Request

Reduces the required rep for a final objective to appear to a practically achievable level in-round (45 from 120.) 

Final objectives basically didn't exist with my last change while we were still feeling out dynamic (also, they were incorrectly counting rep from time after progression traitors first dropped, not just objectives.) Having them show up is more interesting and contributes towards progtot 'wrapping up' in a way that gets everybody (now much more armed and up to handling them) involved. This still requires a decent amount of objective rep that should be a challenge to traitors to stay active in pursuing. 

## Changelog

Reduces progression traitors' objective rep cost for final objectives to appear from the previous 120 (practically unreachable) to 45 (much more realistic but still quite a bit.) 

:cl:
balance: Progression traitors' final objectives can now appear after they've completed 45 rep worth of objectives (down from an unseen 120.)
/:cl:

### DIFF
--- a/code/modules/antagonists/traitor/objectives/final_objective/final_objective.dm
+++ b/code/modules/antagonists/traitor/objectives/final_objective/final_objective.dm
@@ -11,7 +11,7 @@
 	abstract_type = /datum/traitor_objective/final
 	progression_minimum = 140 MINUTES
 
-	var/progression_points_in_objectives = 140 MINUTES //skyrat edit: original value 20 MINUTES
+	var/progression_points_in_objectives = 45 MINUTES //skyrat edit: original value 20 MINUTES
 
 /// Determines if this final objective can be taken. Should be put into every final objective's generate function.
 /datum/traitor_objective/final/proc/can_take_final_objective()
--- a/code/modules/antagonists/traitor/objectives/final_objective/final_objective.dm
+++ b/code/modules/antagonists/traitor/objectives/final_objective/final_objective.dm
@@ -11,7 +11,7 @@
 	abstract_type = /datum/traitor_objective/final
 	progression_minimum = 140 MINUTES
 
-	var/progression_points_in_objectives = 45 MINUTES //skyrat edit: original value 20 MINUTES
+	var/progression_points_in_objectives = 75 MINUTES //skyrat edit: original value 20 MINUTES
 
 /// Determines if this final objective can be taken. Should be put into every final objective's generate function.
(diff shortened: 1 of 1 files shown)


tf-4:
That's a _very_ significant drop. Would it not be better to go for a middle ground of 90 or so first?

Higgin:
> That's a _very_ significant drop. Would it not be better to go for a middle ground of 90 or so first?

I've never seen it go to 90 either, and the way the final objectives are set up still requires that you do more for them to complete. I have seen rounds with 60 rep in objectives, albeit still very rarely. I don't think I'd be changing anything here if I went any higher than that. 

It'd be nice to have some stat-tracking to calibrate this more. 45 to me seems like a good point from the rounds I've seen of 'traitor is active but not necessarily doing things that guarantee the station is already super on the back foot when they get a final objective.'

Zergspower:
45 minutes is a high-as-hell drop, slash it in half and no lower in my opinion but I agree with TF that starting at 90 is a smarter move. There are two types of progtots, the ones that run through everything and GG and the ones who do a handful here and there.

Higgin:
> 45 minutes is a high-as-hell drop, slash it in half and no lower in my opinion but I agree with TF that starting at 90 is a smarter move. There are two types of progtots, the ones that run through everything and GG and the ones who do a handful here and there.

If I even saw the ones who try to GG gamer mode hit 45 semi-regularly, maybe, but then I don't. Even after getting 45 and getting the final objective unlocked, there's still the matter of actually triggering it - in any case, the events to do so are telegraphed.

code/modules/antagonists/traitor/objectives/final_objective/final_objective.dm
```diff
@@ -11,7 +11,7 @@
 	abstract_type = /datum/traitor_objective/final
 	progression_minimum = 140 MINUTES
 
-	var/progression_points_in_objectives = 140 MINUTES //skyrat edit: original value 20 MINUTES
+	var/progression_points_in_objectives = 45 MINUTES //skyrat edit: original value 20 MINUTES
```
Gandalf2k15: Let's try... 75 minutes

Higgin: ok done

PR #13330, opened 2022-05-04
Puts tacticool skirtleneck next to the tacticool turtleneck in the loadout menu, adds replica centcom turtlenecks as loadout options

Paxilmaniac:
## About The Pull Request
As the title would imply, I've moved the tacticool skirtleneck next to the non-skirt version, as they are basically the same thing with the only difference being a skirt or not, thus it didn't make sense to have them be separate in the menu.
The replica centcom turtleneck and its matching skirt have also been added to the loadout menu, because sometimes you just need a green turtleneck to complete a look.

Makes finding the skirt version of the tacticool turtleneck slightly faster, allows more roundstart outfit options for those turtleneck enjoyers who want something other than black or really dark blue

## Changelog

:cl:
add: Replica centcom turtlenecks and their skirt variants have been added to the loadout menu
qol: The tacticool skirtleneck has been moved next to the non-skirt version, as the two are basically the same thing
/:cl:

### DIFF
--- a/modular_skyrat/modules/loadouts/loadout_items/under/loadout_datum_under.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/under/loadout_datum_under.dm
@@ -419,6 +419,18 @@ GLOBAL_LIST_INIT(loadout_miscunders, generate_loadout_items(/datum/loadout_item/
 	name = "Tactitool Turtleneck"
 	item_path = /obj/item/clothing/under/syndicate/tacticool/sensors
 
+/datum/loadout_item/under/miscellaneous/tactical_skirt
+	name = "Tactitool Skirtleneck"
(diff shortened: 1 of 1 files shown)


Gandalf2k15:
Not a fan of having centcom stuff, even if it's a replica.

Paxilmaniac:
Only reason I think it's alright to put them in is because they can be gotten by the crew anyway by hacking a clothes vendor, much  the same way the tacticool turtlenecks can be

theselfish:
Maybe have it donator locked. Or Vet Locked, if you can.

Gandalf2k15:
Yep

Gandalf2k15:
To ensure we don't saturate the station with fake CC officials, we need that change, reopen once done

PR #13341, opened 2022-05-05
[Missed Mirror][NO GBP] Merge 66649

Tastyfish:
# Original: https://github.com/tgstation/tgstation/pull/66649

## About The Pull Request

- Separated compiling the integration tests and running them as separate steps for organization purposes.
- Added a `TEST_ASSERT_NULL(value, reason)` and `TEST_ASSERT_NOTNULL(value, reason)` because those are conceptually simple tests.
- Makes the PASS and FAIL prefixes in the integration test log be green and red for better readability.
- Failure reasons now display the file and line number.
  - In order to achieve this, direct calls to `Fail()` are now wrapped in a macro, `TEST_FAIL()`, as `Fail()` itself needs preprocessor stuff passed to it.
  - In the midst of updating it, I noticed multiple cases of tests directly calling `Fail()` and returning when they should have used a better macro, so those were updated. There was at least one case where it appeared that the code assumed that the test ended at `Fail()`, but made no attempt to do so, such as with the RCD test.
  - Feel free to double check all of the changed unit tests in case I made a functional behavior change, but they currently pass.
- To take advantage of the previous change, failures are now marked as annotations. Note that outside of github, this creates an ugly-looking line but the primary environment is as a github action.

### Examples with intentionally botched unit test:

image

image

image

## Why It's Good For The Game

Makes inspecting failed unit tests significantly easier.

## Changelog

N/A

### DIFF
--- a/.github/workflows/ci_suite.yml
+++ b/.github/workflows/ci_suite.yml
@@ -98,11 +98,14 @@ jobs:
           sudo apt update || true
           sudo apt install -o APT::Immediate-Configure=false libssl1.1:i386
           bash tools/ci/install_rust_g.sh
-      - name: Compile and run tests
+      - name: Compile Tests
         run: |
           bash tools/ci/install_byond.sh
           source $HOME/BYOND/byond/bin/byondsetup
-          tools/build/build --ci dm -DCIBUILDING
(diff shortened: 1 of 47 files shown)


GoldenAlpharex (approved):
Hell yeah, better CI.

PR #13382, opened 2022-05-06
Adds ashwalkers to icebox

Zonespace27:
## About The Pull Request
Messes with the ashwalker map and adds it to icebox.

Hardmode ashwalker is fun, it changes up the dynamic a lot and even if it doesn't make full sense, I think it's more fun to have ashwalkers on every map.

## Changelog

:cl:
add: Ashwalkers can now appear on icebox
/:cl:


Ebin-Halcyon:
If I ever find a coder, i'll probably replace them with map friendly types.

PR #13465, opened 2022-05-09
Fixes the handheld mirror

Cenrus:
## About The Pull Request

Changes to how icons are updated broke the handheld mirror and it didn't update your hair sprite.

This item could use some improvements but at least it works as intended now.

## Changelog

:cl:
fix: Fixed the handheld mirror
/:cl:


GoldenAlpharex (changes requested):
Hair updating is going to be updated upstream shortly, it was an oversight on Kapu's part and he's fixing it.

PR #13466, opened 2022-05-09
Un-panicfixes the self duplication exploit

YakumoChen:
# About The Pull Request

Makes a chem reaction for blacktar heroin that doesn't endlessly duplicate itself when heated.

Makes the reaction work as intended instead of outright deleting it

## Changelog

:cl:
fix: Blacktar Heroin can be made into rocks again
/:cl:

### DIFF
--- a/modular_skyrat/modules/morenarcotics/code/opium.dm
+++ b/modular_skyrat/modules/morenarcotics/code/opium.dm
@@ -132,7 +132,7 @@
 	..()
 
 /datum/reagent/drug/opium/heroin
-	name = "Heroin"
+	name = "heroin"
 	description = "She's like heroin to me, she's like heroin to me! She cannot... miss a vein!"
 	reagent_state = LIQUID
 	color = "#ffe669"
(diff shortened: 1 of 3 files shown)


modular_skyrat/modules/morenarcotics/code/opium.dm
```diff
@@ -182,7 +195,7 @@
 	export_types = list(/obj/item/reagent_containers/heroinbrick)
 	include_subtypes = FALSE
 
-/datum/export/blacktar
+/datum/export/freebase_blacktar
```
GoldenAlpharex: I don't think this one needs to be removed.

GoldenAlpharex (approved):
This still feels like the scuffiest drug I've ever seen.

YakumoChen:
I should actually test this first so I'm not creating the exact same duplication bug so hold off but

I will when I sleep

YakumoChen:
im not sure why this didn't fail a unit test before...

PR #13536, opened 2022-05-11
[MIRROR] Refactors how legs are displayed so they no longer appear above one-another when looking EAST or WEST [MDB IGNORE]

### LINKED ISSUE #20603: Bodypart sprites render with incorrect layering
All bodyparts are rendered on the same layer, and always in the same order. This means that whenever you're facing east your right leg renders over your left leg, when logically the opposite should be true.

d10_icon

In this screen the ordering has been scrambled a bit due to me swapping out bodyparts but the effect can still be seen. While facing left the right leg is rendered on top. No matter what order the parts are in, one leg will render incorrectly in one dir.

The effect is very very hard to tell on the front/back, but on many races you can still tell if the chest is rendering in front or behind the limbs by if you can see the "seems" of the disembodied chest over the limbs.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/66607
--------------------
## About The Pull Request
So, for ***over 5 years***, left legs have been displaying over right legs. Never noticed it? Don't blame you.
Here's a nice picture provided by https://github.com/tgstation/tgstation/issues/20603, that clearly displays the issue that was happening:
image

It still happened to this day. Don't trust me? See for yourself:
image
image
Notice how the two directions don't look the same? That's because the left leg is always displaying above the right one.

Obviously, that's no good, and I was like "oh, that's a rendering issue, so there's nothing I can do about it, it's an issue with BYOND".

Until it struck me.

*"What if we used a mask that would cut out the parts of the right leg, from the left leg, so that it doesn't actually look as if it's above it?"*

Here I am, after about 25 hours of work, 15 of which of *very* painful debugging due to BYOND's icon documentation sucking ass.

So, how does it work?

Basically, we create a mask of a left leg (that'll be explained later down the line), more specifically, a cutout of JUST the WEST dir of the left leg, with every other dir being just white squares. We then cache that mask in a static list on the right leg, so we don't generate it every single time, as that can be expensive. All that happens in `update_body_parts()`, where I've made it so legs are handled separately, to avoid having to generate limb icons twice in a row, due to it being expensive. In that, when we `generate_limb_icon()` a right leg, we apply the proper left leg mask if necessary.

Now, why masking the right leg, if the issue was the left leg?
Because, see, when you actually amputated someone, and gave them a leg again, it would end up being that new leg that would be displayed below the other leg. So I fixed that, by making it so that bodyparts would be sorted correctly, before the end of `update_body_parts()`. Which means that right legs ended up displaying above left legs, which meant that I had to change everything I had written to work on right legs rather than left legs.

I spent so much time looking up BYOND documentation for MapColors() and filters and all icon and image vars and procs, I decided to make a helper proc called `generate_icon_alpha_mask()`, because honestly it would've saved me at least half a day of pure code debugging if I had it before working on this refactor.

I tried to put as much documentation down as I could, because this shit messes with your brain if you spend too long looking at it. icon and image are two truly awful classes to work with, and I don't look forward to messing with them more in the future.

Anyway. It's nice, because it requires no other effort from anyone, no matter what the shape of the leg is actually like. It's all handled dynamically, and only one per type of leg, meaning that it's not actually too expensive either, which is *very* nice. Especially since it's very downstreams-friendly from being done this way.

Just look, both legs are now looking the same now:

image
image

## Why It's Good For The Game
It fixes https://github.com/tgstation/tgstation/issues/20603, an issue that has been around for over half a decade, as well as probably many more issues that I just didn't bother sifting through.

Plus, it just looks *so* much better. Look at this:
<details><summary>It works with Ethereal limbs</summary>

image
image

</details>

<details><summary>It works with Moth legs</summary>

image
image

</details>

<details><summary>It even works with digitigrade legs!</summary>

image
image

</details>

## Changelog

:cl: GoldenAlpharex
refactor: Refactored how legs are rendered, using dynamically-generated masks to avoid left legs always displaying over right ones, or right legs above left ones.
/:cl:


GoldenAlpharex (approved):
What a very cool and well-documented PR :)

PR #13537, opened 2022-05-11
Freighter Fix 1 (Also cargo outfit fixes)

Deek-Za:
## About The Pull Request

Friends love doing ghost role stuff. Including the freighter but a bunch of changes overtime have caused a number of issues. The first being the outfits are broken. 
Cargo outfits were flagged in a section for not having Digi variants. But we do! This fixes our cargo outfits for all the digi players in the world, including the ones used in said ghost role. 
Also does the same for the bodyarmor as it doesn't need a digi variant to begin with as it's just a top.

Makes the freighter role not have such a terrible start where you're immediately hit with a bunch of broken outfits.
Plus ya know, helps cargo players RP with accurate fashion. Regardless of which outfit they choose.

## Changelog
:cl:
fix: Cargo outfits are once again digi friendly
fix: Bulletproof armor now shows up properly on Digi characters.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/clothing/clothing_variation_overrides/under.dm
+++ b/modular_skyrat/master_files/code/modules/clothing/clothing_variation_overrides/under.dm
@@ -216,5 +216,4 @@
 /obj/item/clothing/under/costume/swagoutfit
 	supports_variations_flags = NONE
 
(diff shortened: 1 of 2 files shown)


twilightwanderer:
I think I'll fix the name of all objects to the standard at the same time

PR #13596, opened 2022-05-14
Transfer votes will no longer be available to be started by everyone

GoldenAlpharex:
## About The Pull Request
As simple as that, I'll mark it for a test-merge first because I need to be sure that it still work in automatic mode (it should, I don't see why it wouldn't, but you never know, and I didn't want to wait forever for it to fire.)

Yeah, I don't think I need to explain why everyone being able to start transfer votes is a bad idea.

## Changelog

:cl: GoldenAlpharex
fix: Fixed everyone being able to start transfer votes. Now, only admins will be able to, if autotransfer is enabled.
config: Added ALLOW_VOTE_TRANSFER config flag, defaults to false.
/:cl:

### DIFF
--- a/config/skyrat/skyrat_config.txt
+++ b/config/skyrat/skyrat_config.txt
@@ -51,8 +51,11 @@ CRYO_MIN_SSD_TIME 15
 ## log subtler emotes in game.txt
 #LOG_SUBTLER
 
-## Enables autotransfer voting.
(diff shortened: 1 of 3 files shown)


softcerv:
wow, this is going to make the game completely unplayable /s

Zonespace27 (approved):
looks fine

PR #13614, opened 2022-05-14
Speeds up the arrivals shuttle

tf-4:
## About The Pull Request

Lowers the travel time of the arrival shuttle from 30 seconds to 15.
Lowers the recharge time from 20 seconds to 15. Still plenty of time to dismount.

The shuttle is cool. It's nice. The amount of time it takes to get there is not nice, and god forbid someone sends it just as you're about to get on and you have to wait the 30s travel time + 20s recharge + 30s travel time + 20s recharge + 30s travel time *just to get on station*.

In that worst-case scenario, the total time you'd spend waiting goes from 130 seconds to just 75. Still a while, but not quite as agonising.

## Changelog
:cl:
qol: The arrivals shuttle got new thrusters, cutting its travel time in half and its recharge time down by a quarter.
/:cl:

### DIFF
--- a/modular_skyrat/modules/advanced_shuttles/code/shuttles.dm
+++ b/modular_skyrat/modules/advanced_shuttles/code/shuttles.dm
@@ -7,9 +7,9 @@
 	dir = WEST
 	port_direction = SOUTH
 
-	callTime = 30 SECONDS
+	callTime = 15 SECONDS
 	ignitionTime = 6 SECONDS
-	rechargeTime = 20 SECONDS
+	rechargeTime = 15 SECONDS
 
 	movement_force = list("KNOCKDOWN" = 3, "THROW" = 0)
 	can_be_called_in_transit = FALSE


tf-4:
couldnt think of a better tag than QOL but this two number change is not worthy of GBP

softcerv:
I am both going to very much miss and not miss the sunk cost fallacy arguments about sending the tram. Good PR.

RatFromTheJungle:
> couldnt think of a better tag than QOL but this two number change is not worthy of GBP

can u make it even slower maybe 45s to fly thanks!

YakumoChen:
Times don't factor in the time it takes the shuttle to launch as well, adds like an extra 6 seconds per flight. Missing the tram takes 56 seconds one way, 56 seconds back, then 36 seconds to finally arrive currently. That's 2 and a half minutes!

This change makes it 36 seconds per cycle. And 21 seconds travel, meaning you only wait a minute and a half to get to station after missing it.

GoldenAlpharex (approved):
Yes.

PR #13619, opened 2022-05-15
Removes veteran lock from Vanguard

Zonespace27:
## About The Pull Request
Removes the requirement to be veteran crew from vanguard slots.

Overall, I don't see the purpose in having Vanguard be veteran locked. It's not a job that requires a high amount of trust (aside from not going valid salad on the antags when they get back on station, but the only away mission we have is Black Mesa, which i have yet to see completed without a shift extension). In addition, when I run an away mission, vanguard being vet locked usually means it takes an hour-ish for all 4 slots to fill, which I do not like.

## Changelog

:cl:
config: Vanguard are no longer veteran-crew locked
/:cl:

### DIFF
--- a/modular_skyrat/modules/exp_corps/code/expeditionary_trooper.dm
+++ b/modular_skyrat/modules/exp_corps/code/expeditionary_trooper.dm
@@ -28,8 +28,6 @@
 
 	family_heirlooms = list(/obj/item/binoculars)
 
-	veteran_only = TRUE
-
 	job_flags = JOB_ANNOUNCE_ARRIVAL | JOB_CREW_MANIFEST | JOB_EQUIP_RANK | JOB_CREW_MEMBER | JOB_NEW_PLAYER_JOINABLE | JOB_REOPEN_ON_ROUNDSTART_LOSS | JOB_ASSIGN_QUIRKS
 
 /datum/job/expeditionary_trooper/after_spawn(mob/living/carbon/human/H, mob/M)


Stalkeros2:
Good.
You guys should probably enable other gateway missions too tbh.

LeonY24:
it is very dangerous though

theselfish:
I have to agree with this.
Either we admit that Expo are just Gateway Miners, or we Vet Lock miners.

softcerv:
> it is very dangerous though

If it is a big issue for the game, we probably shouldn't have the job in the game in the first place.

Zergspower:
What's left for Vets? I think this was the last vet special role outside of NTRep and BS since Fleet was removed which was the goal for a lot of people.

Zonespace27:
> What's left for Vets? I think this was the last vet special role outside of NTRep and BS since Fleet was removed which was the goal for a lot of people.

NTR, BS, hemophage, headshots, and a bit more i'm forgetting. The argument that vet losing out on things for the fun of the much wider server population isn't a strong one

GoldenAlpharex:
Considering the amount of issues we've had in the past with vanguards, even with the fact they were Veteran-only... I'm not so convinced this is a good idea.

***Especially*** since they've been allowed to be less balanced with the *express* reason that they were meant to be Veteran-only, because they'd be more likely to follow the rules.

Zergspower:
> > What's left for Vets? I think this was the last vet special role outside of NTRep and BS since Fleet was removed which was the goal for a lot of people.
> 
> NTR, BS, hemophage, headshots, and a bit more i'm forgetting. The argument that vet losing out on things for the fun of the much wider server population isn't a strong one

it is when we're still at a large net loss, the only added thing versus the many removed is headshots

KathrinBailey (changes requested):
image

I'm unconcerned about "what's left for veteran crew".
If you can get agreement from a bunch of admins and/or above without a headmin explicitly saying no I'll merge.
Headmins with repo privileges feel free to close if you deem this to not be good.

softcerv:
> > > What's left for Vets? I think this was the last vet special role outside of NTRep and BS since Fleet was removed which was the goal for a lot of people.
> > 
> > 
> > NTR, BS, hemophage, headshots, and a bit more i'm forgetting. The argument that vet losing out on things for the fun of the much wider server population isn't a strong one
> 
> it is when we're still at a large net loss, the only added thing versus the many removed is headshots

You still have: Hemophages, Oversized, Clown, NT-Rep, Blueshield, and headshots. Veteran crew doesn't need more things added to it unless there is a good reason for that thing to be vet-crew only.

Zergspower:
> > > > What's left for Vets? I think this was the last vet special role outside of NTRep and BS since Fleet was removed which was the goal for a lot of people.
> > > 
> > > 
> > > NTR, BS, hemophage, headshots, and a bit more i'm forgetting. The argument that vet losing out on things for the fun of the much wider server population isn't a strong one
> > 
> > 
> > it is when we're still at a large net loss, the only added thing versus the many removed is headshots
> 
> You still have: Hemophages, Oversized, Clown, NT-Rep, Blueshield, and headshots. Veteran crew doesn't need more things added to it unless there is a good reason for that thing to be vet-crew only.

i'm... not saying add - we keep removing things from the role - and all of those except 1 isnt really a perk as not everyone wants to be a hemophage or oversized.

Gandalf2k15:
Discussing this with people

Gandalf2k15:
I think this time round we'll be keeping it veteran only, if you need to open it up, you can

PR #13631, opened 2022-05-15
Fixes synths errantly getting inherent armor

Zonespace27:
## About The Pull Request
Turns out limb subtypes of `robot/` got an inherent 4 brute / 5 burn protection. Not intended to my knowledge.

This isn't mentioned anywhere and is absolutely a buff.

## Changelog

:cl:
fix: Synths no longer get inherent armor.
/:cl:


Zergspower:
Huh. so if a limb doesnt have that defined it gets that armor by default?

Zonespace27:
> Huh. so if a limb doesnt have that defined it gets that armor by default?

Yep, robotic limbs have an automatic armor.

PR #13645, opened 2022-05-16
re-adds gas miners to icebox and tram

tf-4:
## About The Pull Request

they got lost at some point I guess
I care not for your opinion on whether they should be added. This is a PR to fix a consistency issue.

## Changelog
:cl:
fix: Tram and Icebox once again have gas miners.
/:cl:


Zonespace27 (approved):
sure

theselfish:
"I know how much you don't like smelly smells, I know how much you think it's DISGUSTING, beetles and bugs, you wouldn't touch them when you were a child. Now you must, because if you cannot be in this room for more than one minute, GASgasGASgas, the gas will start flowing, ONLY after you puke though I gotta make that really clear I've been talking for a while."

PR #13657, opened 2022-05-16
[MIRROR] Adds pedestrian crossing signals to tramstation [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/66966
--------------------

## About The Pull Request

There's nothing inherently wrong with crossing the tramstation tram track, except it's impossible to guess where it is or if it's approaching you since it's silent. So I added little signal lights.

**Green** means it's not approaching you or it's so far away it doesn't matter.
**Amber** means you can safely cross but should do with some vague degree of urgency.
**Red** means NT isn't liable for what happens if you cross.

More specifically:
- If the train's stopped (usually switches to this state when the tram controls unlock):
  - If the signals are at the same station that the tram is at: they are red, as the tram could start moving at any moment.
  - Otherwise, they are green.
- If the train is moving:
  - If the tram is moving *away* from the signal, it is green.
  - If the tram is moving *towards* the signal, it is green, amber, or red as per the description at the beginning.

https://user-images.githubusercontent.com/1185434/168409360-9585de0e-6fbb-4b57-b551-2d9b92167a6d.mp4

## Why It's Good For The Game

So people don't have to guess and get hit by a train on their way to work.

## Changelog

:cl:
add: Tramstation now has crossing signals to help you not get hit by the tram (or jump in front of it at the last minute for comedy value).
/:cl:


Tastyfish:
The merge conflict is just with the upstream tramstation.dmm which just needs to be reset.

PR #13661, opened 2022-05-17
DS-2/Interdyne modsuits no longer have access requirements

Zonespace27:
## About The Pull Request
Pretty sure you can't turn these on since interdyne doesn't have ACCESS_SYNDICATE, this fixes that.

Not having to hack modsuits to use them is probably good

## Changelog

:cl:
fix: DS-2 and Interdyne base modsuits no longer need to be hacked to be used by syndicate personnel
/:cl:


tf-4:
Is there a reason not to give them syndicate access?

YakumoChen:
????
They do have that access, they wouldn't be able to walk around the base without it

Zergspower:
Freighter is the one that had this issue, i didnt think Interdyne did

Tastyfish:
Yeah, Interdyne and DS-2 use SYNDICATE access for all airlocks.

dekunutkid:
However for some reason locking the modsuit, it's unable to be unlocked by any syndicate member, I have experienced this myself and as such I approve of this change.

YakumoChen:
That seems to be an issue for modsuits in general honestly.

Or just don't lock them

Or just find the root cause of the bug

PR #13706, opened 2022-05-18
|

ValuedEmployee:


YakumoChen:
Start with 6, half is a big change and leaves security wanting when large threats actually do occur, like Assualt Ops

hugmeorelse:
This is something we set by config. I have it set to the current number for a reason. Furthermore, not all security officers are necessarily going to be active all the time. 8 to 6 is something I'm happy to try, but 8 to 4 is a no.

ValuedEmployee:
> This is something we set by config. I have it set to the current number for a reason. Furthermore, not all security officers are necessarily going to be active all the time. 8 to 6 is something I'm happy to try, but 8 to 4 is a no.

> Start with 6, half is a big change and leaves security wanting when large threats actually do occur, like Assualt Ops

So I see what you're both saying, and I've made a concession to 5 instead of 4. 5 is what TG is currently running with much higher threat counts than we are. I hope this change is more satisfying.

code/modules/jobs/job_types/security_officer.dm
```diff
@@ -5,8 +5,8 @@
 	auto_deadmin_role_flags = DEADMIN_POSITION_SECURITY
 	department_head = list(JOB_HEAD_OF_SECURITY)
 	faction = FACTION_STATION
-	total_positions = 8 //Handled in /datum/controller/occupations/proc/setup_officer_positions() // SKYRAT EDIT: SET TO 8, WAS 5
-	spawn_positions = 8 //Handled in /datum/controller/occupations/proc/setup_officer_positions() // SKYRAT EDIT: SEE ABOVE
+	total_positions = 5 //Handled in /datum/controller/occupations/proc/setup_officer_positions() // SKYRAT EDIT: SET TO 5, WAS 5
+	spawn_positions = 5 //Handled in /datum/controller/occupations/proc/setup_officer_positions() // SKYRAT EDIT: SEE ABOVE
```
Jolly-66: This is essentially a revert if you're going back to 5, which means you shouldn't need these comments anymore (but I would hang onto them until a maintainer has a say in the matter).

ValuedEmployee: > This is essentially a revert if you're going back to 5, which means you shouldn't need these comments anymore (but I would hang onto them until a maintainer has a say in the matter).

That's what I was thinking, didn't want to change too much.

Zonespace27 (approved):
Remove the skyrat edit comments and we're good

ValuedEmployee:
> Remove the skyrat edit comments and we're good

Alright, those comments are gone!

KathrinBailey:
Config change, talk to host/gandalf to get them to change it, this will have no effect.

PR #13710, opened 2022-05-18
Fixes bad global used by turf emote

Tastyfish:
## About The Pull Request

The turf emote was using an unmanaged global to track which turf was picked, making it entirely luck that a given user spawned the correct object if 2 people were using *turf at once.

Unmanaged globals violate code standards anyway, but this shouldn't be global in any form.

## Changelog

:cl:
fix: Fixed possible edge case issues with multiple uses of turf emote at the same time.
/:cl:

### DIFF
--- a/modular_skyrat/modules/emotes/code/additionalemotes/turf_emote.dm
+++ b/modular_skyrat/modules/emotes/code/additionalemotes/turf_emote.dm
@@ -1,5 +1,3 @@
-var/current_turf
-
 /mob/living
 	var/obj/owned_turf
 	var/list/allowed_turfs = list()
@@ -8,6 +6,7 @@ var/current_turf
 	key = "turf"
 	key_third_person = "turf"
 	cooldown = 4 SECONDS
+	var/current_turf
 
 /datum/emote/living/mark_turf/run_emote(mob/living/user, params, type_override, intentional)
 	. = ..()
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/emotes/code/additionalemotes/turf_emote.dm
```diff
@@ -8,6 +6,7 @@ var/current_turf
 	key = "turf"
 	key_third_person = "turf"
 	cooldown = 4 SECONDS
+	var/current_turf
```
GoldenAlpharex, line 10: Just document this please!

GoldenAlpharex (changes requested):

modular_skyrat/modules/emotes/code/additionalemotes/turf_emote.dm
Tastyfish: yep, done

modular_skyrat/modules/emotes/code/additionalemotes/turf_emote.dm
```diff
@@ -94,8 +94,7 @@ var/current_turf
 	if(do_after(user,10))
```
GoldenAlpharex:
```suggestion
	if(do_after(user, 1 SECONDS))
```

GoldenAlpharex (changes requested):
Once that last little comment is done, we're good to go.

PR #13746, opened 2022-05-20
Remove server room firedoors Delta/Meta

### LINKED ISSUE #13730: RnD server room in a state of alarm all the time
## Testmerges: 
[#12529: '[MODULAR] Status Indicators with Preference' by nevimer at commit e2ed90f05b](https://github.com/Skyrat-SS13/Skyrat-tg/pull/12529)

## Reproduction:
1. Come to the server room on any map (except the ice box, everything is fine there)
2. CRAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAAAANG CRAAAAAAAAAAAAAAAAAANG

The damn airlock has been humming all round
image

Deek-Za:
## About The Pull Request

Fire doors at the server rooms in these two maps cause the alarm to perpetually stay on. This is due to the fact that the server room is supposed to be cold and the fire door is detecting the cooler then average temp and activating.
Much like our other maps(Icebox, Tram and Blueshift), the doors are now removed.

Resolves #13730

Stops everyone in science having to deal with an annoying alarm sound. 
Also, as an engineer, answers the question of "What the fuck is science doing in there?" that I always wondered while hearing an alarm while wandering in maint.

## Changelog

:cl:
fix: Removed fire doors causing perpetual fire alarms on Meta/Delta station server rooms.
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -72354,7 +72354,6 @@
 /turf/open/floor/iron,
 /area/station/security/brig/upper)
 "see" = (
-/obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/command/glass{
 	name = "Server Access";
 	req_access_txt = "30"
--- a/_maps/map_files/MetaStation/MetaStation_skyrat.dmm
+++ b/_maps/map_files/MetaStation/MetaStation_skyrat.dmm
@@ -24866,7 +24866,6 @@
 /area/station/maintenance/port/aft)
 "hrR" = (
 /obj/effect/turf_decal/tile/purple/fourcorners,
-/obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/command/glass{
 	name = "Server Access";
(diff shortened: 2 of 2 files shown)


SingingSpock:
Perhaps add a fire alarm panel so it can be turned off rather than removing the firelocks.

Deek-Za:
> Perhaps add a fire alarm panel so it can be turned off rather than removing the firelocks.

I did go down that route. It lets you disable it, yes. But then the door detects the cold and just re-arms again. So unfortunately, it doesn't actually fix the problem. And you can't permanently disable fire alarms anymore. Though if you could, you'd still have to do that at the beginning of every round which would be annoying. I think putting this in line with the other maps would be the best choice. But that was a good idea, yeah.

KathrinBailey (approved):
image

PR #13751, opened 2022-05-20
[MIRROR] [MDB IGNORE] Reformats Access IDs for accessibility and futureproofing [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/67002
--------------------

## About The Pull Request

This PR changes all access defines from integers to strings and organizes them by department for ease of use. In the process I changed some of the access code for certain items like the airlock electronics or circuits cards to stop converting from text to numbers and just directly reference the access required.

THIS IS MY FIRST REAL REFACTOR so I will need extensive review. I tested pretty much everything and it SEEMS to work but since I know very little about coding I may have unintentionally broken some things or not used the code in an efficient manner.

FOR INSTANCE since we no longer use integers and use strings instead, req_access_txt is no longer necessary. However, that means that we start with varedited lists on doors which I'm told may have an impact (albeit a very minor one) on memory. I intend to go through at a later date and make req_access_txt store the strings and convert to a list properly (as well as manually edit all varedit objects again) if there is, indeed, an impact.

I've also gone through and finished editing the inconsistent standards for varedited windoors/buttons/etc. (biggest one being lawyer accesses on secposts and such)

I also went through and finished fixing many airlocks that still had varedits on them since I could see them all as edited in vsc. Unfortunately the CI didn't detect if it had two req_one_access 's or multiple of a req_access.

## Why It's Good For The Game

This is something that I've wanted to do for a while now but was so work intensive and tedious that I've dropped until now. The purpose of this is to properly standardize how the accesses work and make it very easy for someone looking at the code to understand what does what (without needed to have a separate table to reference the IDs from). It also future-proofs the ability to add IDs onto a department without having to do a significant amount of extra editing in other files besides the access file. It also makes it easier to audit and tweak how these accesses work as a groundwork (for example, now that engineering is properly categorized within itself, it is easier to separate the generic access of ACCESS_CONSTRUCTION into separate, well-partitioned access requirements.)

## Changelog

:cl:
refactor: Compartmentalized defines for access IDs and converts them from integers to strings
fix: Standardizes some access reqs on varedited objects like windoors and shutter buttons
fix: Some missed doors with access reqs instead of mapping helpers were fixed
/:cl:


azrael-x86:
yuh oh

PR #13778, opened 2022-05-21
[MODULAR] Introduces the Sec-Medic Disabler

softcerv:
## About The Pull Request
Meant to serve as either a replacement for the standard Firefly or an additional item, the VeyMed FMG (Field Medic Grade) disabler functions like a standard disabler with the addition of the salve mode from mediguns. 

The salve mode can be fired 10 times at full charge and has the same stats as the salve medigun mode.

_**This PR does not actually make this weapon obtainable, it just puts it in the code for any PR that would to use it.**_

## Ingame screenshot
image
The VeyMed FMG disabler is on the top, the standard disabler is on the bottom. 
This adds in a unique version of the disabler that could serve as a good replacement for the Sec Medic's current weapon, it could also work well as a potential guncargo item too.

## Changelog

:cl:
add: adds the VeyMed FMG disabler, a disabler that can use the same salve mode that mediguns have.
/:cl:


MassacreDemigirl:
I mean add it to the secmed while making the PR
Since you are already: spending GBP to add an item
And its codebloat if its not used anywhere

softcerv:
> I mean add it to the secmed while making the PR Since you are already: spending GBP to add an item And its codebloat if its not used anywhere

The PR that was going to replace it ended up getting closed, so I am not really too sure on what I want to do with it now. I might end up turning it into a guncargo thing (probably with the veymed branding gone)

Zergspower:
> > I mean add it to the secmed while making the PR Since you are already: spending GBP to add an item And its codebloat if its not used anywhere
> 
> The PR that was going to replace it ended up getting closed, so I am not really too sure on what I want to do with it now. I might end up turning it into a guncargo thing (probably with the veymed branding gone)

Given they were returned the (garbage) firefly due to guncargo, i'd probably go that route of having it be an orderable item and give them something to look forward too.

PR #13782, opened 2022-05-21
Non-ashwalkers now only get half the ore from forges

Zonespace27:
## About The Pull Request
If you aren't an ashwalker, you only obtain half the ore from forging. A maxxed forge will now effectively produce 2x the ore instead of 4x.

You can make dumb tiers of ore on-station with forging which really shouldn't be done

## Changelog

:cl:
balance: Non-ashwalkers only obtain half the ore from forging
/:cl:


tf-4:
This does need to be addressed, but I'm not a fan of this approach. It makes no sense for monkeys to get more ore than people.

Zonespace27:
> This does need to be addressed, but I'm not a fan of this approach. It makes no sense for monkeys to get more ore than people.

Okay, made it so just ashwalkers get the 1x production of ore.

PR #13795, opened 2022-05-21
Updates tails.dm, allows tails to show from under more clothing

tf-4:
## About The Pull Request

Reorganisation of `tails.dm` to keep things a little more sane in there.

Replaces `HIDEJUMPSUIT` with a new `HIDETAIL` bitflag. `HIDEJUMPSUIT` was a bit of a problem as it meant a lot of clothing blocked tails unnecessarily, while the flag was also key to a few procs that handle sprite updates.

This new flag should let tails show from beneath stuff that it's fine to have (namely robes, costumes), while I re-added it primarily to suits that make you spaceworthy as you shouldn't really have your tail out in space.
Couldn't decide on giving the clothing the flag in their respective core files vs having one big file to add that flag to a bunch of stuff, went with the former in the end but can reshuffle to the latter.

Also updates the hulk tail swing proc to check for this new flag.

This also fixes `SHOWSPRITEEARS` and `HIDEMUTWINGS` both having `(1<<13)`, which I don't think actually could presently cause any issues but is not ideal nonetheless.

Yes, tail.
image

No burning your tail.
image

Modsuits keep tail sprites
image

But not if you have a tail that has no modsuit tail sprite.
image image

Being able to show your tail is nice. Having tails displayed or not based on a very tenuously related flag is not so good.

## Changelog
:cl:
qol: Tails can now protrude from beneath a wider variety of clothing if you so choose.
/:cl:


nevimer:
does this fix tails under hardsuits, a bug that carrier onto modsuits from like September last year

tf-4:
> does this fix tails under hardsuits, a bug that carrier onto modsuits from like September last year

I'm not sure what you mean, so probably not

nevimer:
> > does this fix tails under hardsuits, a bug that carrier onto modsuits from like September last year
> 
> I'm not sure what you mean

okay so basically non lizard tails dont show up under modsuits

felinids use the unsuited sprite for some reason

and most others while existing simply do not show up

nevimer:
> > > does this fix tails under hardsuits, a bug that carrier onto modsuits from like September last year
> > 
> > I'm not sure what you mean
> 
> okay so basically non lizard tails dont show up under modsuits
> 
> felinids use the unsuited sprite for some reason
> 
> and most others while existing simply do not show up

for example, try having a tail while anthro and picking fox 3

checkout a commit from summer last year, say mid august. do the same. you will see the tails worked, and every suit had a tail

tf-4:
Ah. The answer is no. That's due to a lack of sprites mostly. There may also be tails that don't have associated tailsprites even if they exist, but this PR doesn't touch on that.

nevimer:
> Ah. The answer is no. That's due to a lack of sprites mostly. There may also be tails that don't have associated tailsprites even if they exist, but this PR doesn't touch on that.

The sprites do exist! The rendering got broken in a PR that added luminance to bodyparts. Every single tail has a corresponding suit tail sprite. Just like every suit had a snout sprite.

code/modules/antagonists/cult/cult_items.dm
```diff
@@ -403,7 +403,7 @@ Striking a noncultist, however, will tear their flesh."}
 	armor = list(MELEE = 50, BULLET = 40, LASER = 50, ENERGY = 60, BOMB = 50, BIO = 100, FIRE = 100, ACID = 100)
 	hoodtype = /obj/item/clothing/head/hooded/cult_hoodie/hardened
 	clothing_flags = STOPSPRESSUREDAMAGE | THICKMATERIAL
-	flags_inv = HIDEGLOVES | HIDESHOES | HIDEJUMPSUIT
+	flags_inv = HIDEGLOVES | HIDESHOES | HIDEJUMPSUIT | HIDETAIL // SKYRAT EDIT ADDITION - HIDETAIL
```
Gandalf2k15, line 406: I really want override files for simple shit like this. I think an override module is needed.

tf-4: I can do that in a follow up, I think it's a little outside the scope of this PR particularly

PR #10570, opened 2022-01-10
[MIRROR] Buffs The Ectoscopic Sniffer [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/63797
--------------------

## About The Pull Request
Buffs the robotics ectoscopic sniffer to be much more obvious to spacemen's simple minds. 
I've increased the volume from 25 to 75 - a move aimed to make it so that everyone in robotics can hear it go off.
~~The visuals on blip have changed from green to red, to help draw the eye towards it.~~ Reverted pending different idea.
The examine text now explains how it functions in a more heavy-handed way.

## Why It's Good For The Game
Roboticists seem more interested in spacetiding and making big stompy mechs and modsuits instead of actually supporting the silicon players they were (Assumedly, considering the oldest record of either I can find is from 2012) made to support.
This should help draw their gaze back on objective without being /explicitly/ hostile to them.

## Changelog

:cl:
add: Ectoscopic sniffers are now louder, and explain their purpose on examine.
/:cl:

PR #13886, opened 2022-05-25
Update readme.md

Gandalf2k15:
Updates map reset requirements


jjpark-kb (changes requested):
I disagree highly with these changes. These are items that are skyrat exclusive and we shouldn't be removing them from this list.

Gandalf2k15:
> I disagree highly with these changes. These are items that are skyrat exclusive and we shouldn't be removing them from this list.

The only one that does these is me, and no one has mentioned them being missing for this reason: They are all on the interlink.

jjpark-kb:
> 

You are not the only one who does these-- yes you have recently been doing them; however, it is disingenuous to say you are the only one doing them. Additionally, they are not all on the interlink.

Gandalf2k15:
> 

We don't have stasis beds. Command vendor is being moved to loadout. Lustwish is on the interink. Blueshield+NTrep is on the interlink. Posi consoles were added by TG.

jjpark-kb:
> 

Posi consoles were not added by /tg/, they were added by me to oldbase and ported here by Ranged.

Gandalf2k15:
> > 
> 
> Posi consoles were not added by /tg/, they were added by me to oldbase and ported here by Ranged.

bro they are called Ecto-Sniffer https://github.com/Skyrat-SS13/Skyrat-tg/pull/10570

_maps/readme.md
```diff
@@ -4,29 +4,25 @@
 
 The list of items that are modular requiring adding/modifying to /tg/ maps:
 
-- Positronic Alert Consoles (inside Robotics area)
-- Command Vendor (inside Meeting area)
```
GoldenAlpharex, line 8: I don't really think that moving this to the loadout is the good move to make, I'd rather we just add on to the garment bags in the heads' lockers instead, if we're not going to use the command vendor anymore. Would be relatively easy to do modularly, without having the maintenance cost that the Command vendor has.

jjpark-kb:
> 

that is not a posi console.

_maps/readme.md
```diff
@@ -4,29 +4,25 @@
...
-- LustWish Vendor (inside Recreational/Dormitory area)
```
GoldenAlpharex, line 9: I'm not sold on this one, mostly because I know people are going to complain a lot about having to go back to the interlink every time they want an item from it.

Gandalf2k15:
> > 
> 
> that is not a posi console.

It is a better version and achieves the same thing. Do you not check these things?

_maps/readme.md
```diff
@@ -4,29 +4,25 @@
...
 - Barbershop Room (including the vendors, chairs, mirrors, and more)
-- NanoTrasen Consultant's Office
-- Blueshield Office
 - Cryopods (inside Recreational/Dormitory area)
-- Stasis Beds to Stasis Pods
-- Armory/Gun Change
-- Prison Change
```
GoldenAlpharex, line 16: Seems like you forgot to move over this one, our prisons are often different from /tg/'s, and considering the amount of people that worked on bringing them back every time they were forgotten about, I don't think that it's warranted to remove them from that list.

jjpark-kb:
> > > 
> > 
> > 
> > that is not a posi console.
> 
> It is a better version and achieves the same thing. Do you not check these things?

I highly disagree its a better version-- you can actually ask roboticists and they will say the posi is better as it is more clear. Also, don't treat me like that with the passive aggressive comment, it doesn't help.

Gandalf2k15:
> > > > 
> > > 
> > > 
> > > that is not a posi console.
> > 
> > 
> > It is a better version and achieves the same thing. Do you not check these things?
> 
> I highly disagree its a better version-- you can actually ask roboticists and they will say the posi is better as it is more clear. Also, don't treat me like that with the passive aggressive comment, it doesn't help.

Then stop questioning what things are to be kept lol https://github.com/Skyrat-SS13/Skyrat-tg/blob/master/code/game/machinery/ecto_sniffer.dm

jjpark-kb:
> 

I really am trying to have a conversation with you and you seem to not want to have it. My Posi Alert System is clearer than /tg/s "solution." I do not appreciate you likening your treatment towards me in a *public* manner for me questioning your choices.

hugmeorelse:
Let's take this somewhere more appropriate.

hugmeorelse:
@Gandalf2k15 Please wait until we discuss things.

_maps/readme.md
```diff
@@ -4,29 +4,23 @@
...
+- - Replace Laser Guns With CMGs
+- Security Medics Room:
+- - Security Medics Spawner
+- - Security Medics Locker
+- Arrival Docks:
+- - Arrival Shuttle Docking Port Resize
+- - Arrival Shuttle Console
 - Central Command Ferry Hangar (inside Arrival area)
-- Update roundstart_template variable on mining/public/labour stationary docks: mining/x > mining/skyrat, mining_common/x > mining_common/skyrat, labour/x > labour/skyrat , mining/large > mining/large/skyrat
-- Replace the security outposts with their departmental guards
```
GoldenAlpharex, line 21: I got a feeling this is going to cause issues with spawning the guards, and their roundstart equipment, if it's not properly carried over.

_maps/readme.md
```diff
@@ -4,29 +4,23 @@
...
+- - Replace Energy Guns With MCRS
+- - Replace Laser Guns With CMGs
+- Security Medics Room:
+- - Security Medics Spawner
+- - Security Medics Locker
+- Arrival Docks:
+- - Arrival Shuttle Docking Port Resize
+- - Arrival Shuttle Console
 - Central Command Ferry Hangar (inside Arrival area)
-- Update roundstart_template variable on mining/public/labour stationary docks: mining/x > mining/skyrat, mining_common/x > mining_common/skyrat, labour/x > labour/skyrat , mining/large > mining/large/skyrat
```
GoldenAlpharex, line 20: Won't this mean that those shuttles won't be using the Skyrat ones?

GoldenAlpharex (changes requested):
I know the goal of this is to optimize the map reset workflow, but I think we should also look into other ways to make it easier to manage, but that's more something to discuss amongst ourselves.

_maps/readme.md
Gandalf2k15: It's not been on station for... a long while.

_maps/readme.md
Gandalf2k15: Our current maps don't have any different prisons currently, so I have nothing to take over

_maps/readme.md
Gandalf2k15: They should be using sec guard spawns, if not, that needs changed

_maps/readme.md
Gandalf2k15: No, this was never ever required, it's all code.

_maps/readme.md
Gandalf2k15: It'll be moved to the loadout menu and or into lockers, this single change I always forget about, it's not something we need to keep around, or we can just move it to the interlink

_maps/readme.md
GoldenAlpharex: My main concern would likely be the equipment lockers, in that case, then. Because I agree that they could use the same spawnpoints as the sec guard spawns, if possible.

Gandalf2k15: I don't even know what they get, but the lockers should also be updated in code if they aren't already, but, the stuff I am removing is stuff I haven't even been resetting for like... well, I've never included this stuff. And it's never been brought up as an issue

_maps/readme.md
nevimer: What do you mean, it was on station right up until 2 days ago.

Gandalf2k15: I don't even remember the last time I reset it in this tedious process, so it's not me that has been doing it. But no complaints since it's on the interlink.

_maps/readme.md
nevimer: They had custom prisons up until 2 days ago, specifically on Delta I believe. And Icebox, until second last reset

_maps/readme.md
softcerv: This is something that should still be on station, it's a massive QoL because otherwise people will have to go back and forth between the actual station and interlink

_maps/readme.md
softcerv: without the departmental guards lockers, they lack actual access to their batons or flashes along with their cosmetic gear. With security officer lockers, they also get gear like sunglasses and sec radio keys that they normally otherwise wouldn't have.

Gandalf2k15: Well someone should update the sec lockers to have this, but this is asking too much of priority

_maps/readme.md
ghost: Thats entirely false, it will spawn the TG template per shuttle ID.
Source: I had to troubleshoot and add this line because they, as I said, need to have their ID in the shuttle docks.

_maps/readme.md
Gandalf2k15: Again, it's not a priority item, others can deal with it.

ghost:
The entirety of the removals from this are pure laziness on the part of those who maptain, and wont be followed regardless because all you do is  paste the TG map over in the end anyways, removing our changes. Its happened every single map reset.
You should be following the readme in the first place before changing it to make stuff easier.

Gandalf2k15:
> The entirety of the removals from this are pure laziness on the part of those who maptain, and wont be followed regardless because all you do is paste the TG map over in the end anyways, removing our changes. Its happened every single map reset. You should be following the readme in the first place before changing it to make stuff easier.

It is ultimately down to whoever does the resets. So this is just so I know what to look out for. It means nothing in the end.

_maps/readme.md
Gandalf2k15: This is not something we should be doing map wise. It should be code. So shame on thee.

PR #13968, opened 2022-05-28
she makes 100 gbp every map reset using nothing but mining shuttles, here's how

Paxilmaniac:
## About The Pull Request
Resets our mining shuttles to the fancy ones... again. I will be keeping count of how many times I have to do this

I mean if you really think about it, it doesn't contribute much at all. What it does do however is make a number under my name go up, which is a valid reason

## Changelog

:cl:
fix: both public and private mining shuttles on our tg maps have once again been restored to our fancy variants
/:cl:

### DIFF
--- a/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
+++ b/_maps/map_files/Deltastation/DeltaStation2_skyrat.dmm
@@ -4228,7 +4228,7 @@
 	height = 5;
 	id = "mining_home";
 	name = "mining shuttle bay";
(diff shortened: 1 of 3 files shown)


Gandalf2k15:
Instead of this, can you just override the code

GoldenAlpharex (changes requested):
At this point, bring the change upstream by making those subtypes of `/obj/docking_port/stationary`, so it can be safely overriden in our modules without being affected by every single map reset.

Paxilmaniac:
fiiine I'll do it sometime soon

PR #11709, opened 2022-02-24
Cleans our Floor aesthetics module, re-sorting it and updating all the newer TG floors to our new palette

ghost:
## About The Pull Request
Title

No changes to normal tiles

The downside: textured tiles are redundant, as by default all our floors had that texture

The upside: now all the new tiles actually match ours. No more glaring white cubes in medical.
(Old screenshot, these will appear textured now - the POINT is that its the same color)
image

## How This Contributes To The Skyrat Roleplay Experience
Cleaner and more matching sprites, rejoice

## Changelog

:cl:
imageadd: Updated floor tile overrides to include all of TG's newer tiles, so they now match ours
/:cl:

PR #13997, opened 2022-05-30
Allows players to actually be able to use the tile variety added in #13344

Paxilmaniac:
## About The Pull Request
Hey, you guys remember these?
image

No? Didn't think so. Want to know why? Because when we made the override for our floor aesthetics module however long ago (which removed all the smooth versions of tiles for some reason?), we failed to account for the fact that the list of tiles might change, and by we I mean gandalf when they merged the pr mentioned in the title.

Unintentionally missed options from tg due to some random override generally seem like a bad thing, plus it'll save some mapper somewhere the trouble of placing one of these tiles in a map, only to discover that the sprite is not actually in the modular sprites folder.

## Changelog

:cl:
fix: The new tiles added by tg mirror pr #13344, which were overridden unintentionally by a pr from a while ago, are now available as options for changing the looks of floor tiles into
/:cl:


PCPort:
It is time to make more pretty bars

ghost:
Removing the smooth version was intentional and agreed on by maintainers (I did that override); the reason for this is that every station was mapped with 'smooth'/base tile types, and the 'textured' ones were a subtype - but skyrat aesthetically wants the texture, as "the smooth looks plastic-y"
If the smooth were left as-is? The station would all start smooth at roundstart. So, the smooth tiles have textured sprites, and the ACTUAL textured tiles, as dupes, are inaccessible.

ghost:
See the comments of https://github.com/Skyrat-SS13/Skyrat-tg/pull/11709 for more context

Otherwise though, god bless you for this PR, appreciate someone fixing the stuff we've missed out on

GoldenAlpharex (approved):
Looks good to me!

PR #14017, opened 2022-05-31
[MIRROR] This tail refactor turned into an organ refactor. Funny how that works. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/67017
--------------------

### So yeah, uh, we have /organ/internal now.
External organs are cosmetic, don't process, and are stored on limbs directly. Most cannot be surgically removed.
Internal organs process, and are stored in nullspace until the limb they technically belong to is dismembered. ~~This is legacy behavior, fuck you, I'll change it in a future PR.~~ This is intentional behavior, apparently.

- Rewrites tails to be external_organs instead of whatever the fuck they are now. Changes tail wagging to a simple signal.
- Rewrites how external_organs handle color, now defined by the organ itself. Organs can choose to use no color, their own color, hair color, or the parent limb color.
- Stores a list of all external_organs on /carbon, mostly to prevent nested for loop hell.
- External organ color now behaves like Kapulimbs, meaning they will only change color when forced to.
- External organs are now handled (almost) identically to standard organs, and can be surgically removed if flagged.
- External organs are now responsible for their appearance on Insert()
- External organs no longer sit in nullspace forever if their parent is deleted. I don't know how this issue didn't come up sooner.

## Why It's Good For The Game
Brings more ancient shitcode inline with modern, less shit code.

## Changelog

:cl: GoldenAlpharex for all the modular fixes, Kapulimbs for the original code
refactor: Organs have been refactored, especially external organs. Notify Golden if there's any unexpected behavior.
/:cl:


GoldenAlpharex:
Okay gamers let's get this bread

PR #14040, opened 2022-06-01
Fixes adjustable masks for snouts; also teshari mask and hat improvement

### LINKED ISSUE #13957: Adjustable Balaclava lacking some sprites.
while it may just be when a character with a snout, horns or ears wears it, when you toggle the adjustable balaclava to reveal the mouth or the entire face, the character's face gets replaced with the pink and black squares.

Tastyfish:
## About The Pull Request

- Adds `BODYTYPE_CUSTOM` to the bodytype VV enum list, since I forgot to do that before.

- `CLOTHING_SNOUTED_VARIATION` is unreliable for a lot of our masks and hats (particularly ones with multiple states like balaclavas), so the code now double checks if the icon state actually exists. This wouldn't have been a problem with how the update_inv_ code used to work where the icon_state check turned always cancelled overrides, but whatever. That entire file needs to be rewritten at the upstream level anyway.
  - Fixes #13957

- Makes it so if teshari don't have a mask or hat sprite, it checks if there's a muzzled sprite and uses that instead of the default human one. Neither perfectly matches teshari heads, but muzzled is better than direct human in many circumstances.

### Tesh Example: modsuit helmets completely covering the head
image

### Tesh Example: balaclava basically covering the head
image

the endless cycle of putting little outfits on animals like they're people

## Changelog

:cl:
fix: Balaclavas no longer glitch when adjusted on snouts.
fix: Modsuits on tesharis now cover their snoot.
/:cl:


GoldenAlpharex (approved):
Looks good to me!

PR #14046, opened 2022-06-02
Adjusts the range of the mold foam and smoke effects

Ghommie:
## About The Pull Request
For some reason #13407 changed (clarification, this is a fault of one of the maintainers here, not the guy behind the original PR) the range of the toxic and radioactive foam effects from 40 and 50 respectively (still way too high, I've lower them to 4 and 5) to a whooping 200, which is enough to theorically cover the entire z-level. Also, the same port changed the range of a mold smoke effect from 5 to 4.

That said, I'm changing the values of these fluid effects to 4, 5 and 4 again. Better late than never, this issue has been around for almost a month now, notwithstanding how easily it can be fixed.

Read above, I don't want to repeat myself.

## Changelog

:cl:
fix: Stops the mold foam effects from having a station-covering range of 200. It should be 4/5 now.
/:cl:

### DIFF
--- a/modular_skyrat/modules/biohazard_blob/code/biohazard_blob_structures.dm
+++ b/modular_skyrat/modules/biohazard_blob/code/biohazard_blob_structures.dm
@@ -101,7 +101,7 @@
 				reagents.my_atom = src
 				reagents.add_reagent(/datum/reagent/cordycepsspores, 50)
 				var/datum/effect_system/fluid_spread/smoke/chem/smoke_machine/puff = new
-				puff.set_up(4, location = my_turf, carry = reagents, efficiency = 24)
+				puff.set_up(5, location = my_turf, carry = reagents, efficiency = 24)
 				puff.attach(src)
 				puff.start()
 			if(BIO_BLOB_TYPE_FIRE)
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/biohazard_blob/code/biohazard_blob_structures.dm
```diff
@@ -125,7 +125,7 @@
 				R.my_atom = src
 				R.add_reagent(/datum/reagent/toxin, 30)
 				var/datum/effect_system/fluid_spread/foam/foam = new
-				foam.set_up(200, location = my_turf, carry = R)
+				foam.set_up(4, location = my_turf, carry = R)
```
GoldenAlpharex, line 128: My only question is, why is this one in a radius of 4 when all of the other ones are in a radius of five?

GoldenAlpharex (changes requested):
Thanks a lot, was going to do it myself, glad I didn't have to!

modular_skyrat/modules/biohazard_blob/code/biohazard_blob_structures.dm
Ghommie: Beats me, it was originally 50 and 40 but I've decided to cut them to a tenth.

tf-4 (approved):

PR #14053, opened 2022-06-02
Re-Adds the Savannah-Ivanov

Zonespace27:
## About The Pull Request
Readds the savannah-ivanov by undoing a removal we did

Times have changed, the crew has more equivalent firepower, and the mech's removal felt like a bit of a kneejerk PR.

## Changelog

:cl:
add: Re-added the Savannah-Ivanov
/:cl:

### DIFF
--- a/code/modules/research/techweb/all_nodes.dm
+++ b/code/modules/research/techweb/all_nodes.dm
@@ -1758,8 +1758,6 @@
 	)
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 5000)
 	discount_experiments = list(/datum/experiment/scanning/points/machinery_tiered_scan/tier3_mechbay = 2500)
-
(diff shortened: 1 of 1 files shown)


Nerev4r:
so, like.
i tested it locally awhile ago and it's
outside of the raw stats (which are great) the abilities themselves aren't that wacky.
the skyfall deadass has a 10 second channel with like
5 visible red text emotes about the mech doing it, and i don't think either pilot can do literally anything while it channels
and yes it does ignore walls, but, the landing doesn't gib or even.. really do that substantial damage to even human targets
about the best thing about it is the durability and i guess the missiles?? but they're like
0 heavy 1 light 2 flame explosions
it'll blow some limbs off a naked human, but it's on a 35 cooldown and doesn't come down instantly

Gandalf2k15:
Fine

PR #14081, opened 2022-06-04
[NON-MODULAR] Reduces prisoner slots from 12 to 6*.

Higgin:
## About The Pull Request

Reduces prisoner slots from 12 to 6.

It sucks having up to twelve people join in to potentially require your babysitting/involvement to do anything in the round or even just not die to random events. Especially when your security department is two detectives and an officer. 

One of the problems is that the more people playing prisoner, the more interesting it can become for more people to join in as prisoner. Having it possibly go up to twelve (one less than all the available security slots) with no commensurate requirement for more security (COs in particular) is ridiculous as a cap.

Four is much more manageable even with lowpop in sec but should offer enough slots for COs (who sign up actually expecting to and interested in interacting with prisoners) to have focused RP with their prisoner crews as well (without also being pulled every direction by the management hell of a vortex anomaly spawning in the middle of the prison when it has eight people in it, for instance.) 

Six is a good step in the right direction. Hopefully the state of play won't give me any reason to come back to it - at the very least it should manage the greatest absurdities of prisoner numbers vs. sec.

It should also encourage the people who want to roll prisoner to roll for it roundstart during highpop hours - if the role is so popular, and the mix of people you get in prison is part of what makes it enjoyable, this should satisfy that enjoyment with a more competitive mix of people rolling for it while also making it less prone to being an unasked-for and unmanageable-in-scale pain to manage for people playing regular security. 

## Changelog

:cl:
balance: Reduced prisoner slots from 12 to 6.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/prisoner.dm
+++ b/code/modules/jobs/job_types/prisoner.dm
@@ -3,7 +3,7 @@
 	description = "Keep yourself occupied in permabrig."
 	department_head = list("The Security Team")
 	faction = FACTION_STATION
-	total_positions = 12		// SKYRAT EDIT: Original value (0)
+	total_positions = 4		// SKYRAT EDIT: Original value (0)
 	spawn_positions = 2
 	supervisors = "the security team"
 	selection_color = "#ffe1c3"
--- a/code/modules/jobs/job_types/prisoner.dm
+++ b/code/modules/jobs/job_types/prisoner.dm
@@ -3,7 +3,7 @@
 	description = "Keep yourself occupied in permabrig."
 	department_head = list("The Security Team")
 	faction = FACTION_STATION
-	total_positions = 4		// SKYRAT EDIT: Original value (0)
+	total_positions = 6		// SKYRAT EDIT: Original value (0)
 	spawn_positions = 2
 	supervisors = "the security team"
 	selection_color = "#ffe1c3"


RatFromTheJungle:
Much like with the sec-officer counts, don't try to go below half for the first count, ala, make it 6, because down to 25% of what it was before is a /steep/ drop, especially for a popular role.

Higgin:
Four is manageable from practical experience and seems to me about the right place for it. If it'll fly at six, I'll put it there when I'm up tomorrow and come back to it after it's had some time to play out. Frequently it's when it's 5-7 prisoners and lowpop sec that I see the most issues.

code/modules/jobs/job_types/prisoner.dm
```diff
@@ -3,7 +3,7 @@
 	description = "Keep yourself occupied in permabrig."
 	department_head = list("The Security Team")
 	faction = FACTION_STATION
-	total_positions = 12		// SKYRAT EDIT: Original value (0)
+	total_positions = 4		// SKYRAT EDIT: Original value (0)
```
tf-4:
```suggestion
	total_positions = 6		// SKYRAT EDIT: Original value (0)
```
Like jungle said

tf-4 (changes requested):

code/modules/jobs/job_types/prisoner.dm
Zonespace27: Seconding this

GoldenAlpharex:
That kind of stuff is handled in the config anyway.

tf-4 (approved):
This is done in config but I like it conceptually

Gandalf2k15 (approved):

hugmeorelse:
This is a config change. I also don't think it's too necessary, mostly because we've already made changes to the Prisoner role and have made it very clear they shouldn't be rioting or becoming "unmanageable". I do think having it scale to security population would be ideal however if that's possible rather than a flat decrease.

Iamgoofball:
A better idea would be to apply the sec/assistant scaling system to prisoners/guards. I'll mock up something for this shortly.

Gandalf2k15:
See my comment inhttps://github.com/Skyrat-SS13/Skyrat-tg/pull/14081

PR #14169, opened 2022-06-07
make the current map not appear in map votes

tf-4:
## About The Pull Request

Blocks the current map from appearing in the map vote

Playing a map back to back is generally not great. New players may come in and vote for it while others have to play the same map repeatedly.

## Changelog
:cl:
qol: You can no longer vote for the current map in the end of round map vote.
/:cl:

### DIFF
--- a/code/datums/votes/map_vote.dm
+++ b/code/datums/votes/map_vote.dm
@@ -10,7 +10,7 @@
 	var/list/maps = shuffle(global.config.maplist)
 	for(var/map in maps)
 		var/datum/map_config/possible_config = config.maplist[map]
-		if(!possible_config.votable || (possible_config.map_name in SSpersistence.blocked_maps))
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (changes requested):
This would probably need some extensive testing to ensure that we don't get into more "democracy" votes than we already do, where only one option is presented.

tf-4:
yeah probably

Iamgoofball:
i love democracy etc.

lessthnthree:
> This would probably need some extensive testing to ensure that we don't get into more "democracy" votes than we already do, where only one option is presented.

With Kilo cut off in highpop and Delta, Blueshift in lowpop, that leaves Meta, Tram, and Icebox that would always be in the vote. So minus the recently played map it shoullllllld always have a minimum of 2 choices. Unless a map is taken out of rotation.

Could add a check so in the event there's only one option, add a 'backup' map or something.

tf-4:
this still hasnt been TM'd and needs it

Gandalf2k15:
confirmed working

PR #14173, opened 2022-06-08
[Modular] Yet another guns PR (STG+Mp40 Nerf, armadyne dirt nerfed)

RatFromTheJungle:
## About The Pull Request
Removes burst first from the STG, as it made an already fantastic gun a god weapon, also removes it from the mp40 for the same reason, burst with the partial-auto aim from clicking on a mob made them far-far-far to strong.

Also, lowers the dirt modifier on alot of the armadyne handguns, and the makarov(Izhevsk) so they're less out shined by literally anything that can't get dirty (which is alot, like, functionally anything from oldarms/bolt/even-the-other-guns-in-izvhevsk are better then then), while not entirely removing the mechanic.

## Changelog

:cl:

balance: rebalanced the STG and MP40, by removing the burst-fire modes from them.
balance: Nerfed 'dirt' accumulation rate to armadyne (hand)guns, and the R-C Makarov, 
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/guns.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/guns.dm
@@ -31,7 +31,7 @@
 	lock_back_sound = 'sound/weapons/gun/pistol/slide_lock.ogg'
 	bolt_drop_sound = 'sound/weapons/gun/pistol/slide_drop.ogg'
 	can_flashlight = TRUE
-	dirt_modifier = 1
+	dirt_modifier = 0.5
 	emp_damageable = TRUE
 	fire_delay = 1.90
 	company_flag = COMPANY_CANTALAN
@@ -77,7 +77,7 @@
 	mag_display_ammo = FALSE
 	can_flashlight = TRUE
 	company_flag = COMPANY_CANTALAN
-	dirt_modifier = 1
+	dirt_modifier = 0.7
 
 /obj/item/ammo_box/magazine/multi_sprite/g18
 	name = "\improper GK-18 magazine"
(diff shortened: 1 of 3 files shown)


modular_skyrat/modules/sec_haul/code/guns/guns.dm
```diff
@@ -305,7 +307,7 @@
 	rack_sound = 'sound/weapons/gun/pistol/rack.ogg'
 	lock_back_sound = 'sound/weapons/gun/pistol/slide_lock.ogg'
 	bolt_drop_sound = 'sound/weapons/gun/pistol/slide_drop.ogg'
-	dirt_modifier = 0.75
+	dirt_modifier = 0.3
```
Zonespace27, line 310: this isn't armadyne

RatFromTheJungle: that was to keep it in line with the armadyne guns' new dirt balances, i'll correct the changelog, but it just happened to be in the same file as /everyotherarmadynegun/ so I didn't double-glance it

but in summary, it's a ladon with a (much) smaller mag, so making it get dirty faster is a hard-hit for an already underused gun

Zonespace27: The makarov's not on the same level as armadyne guns, it's part of izhevsk's shitheap guns, you don't buy it because it's good, you buy it because it's cheap

tf-4:
Don't justify the balance changes in the changelog, do it in the PR body. Better to keep the changelog short where possible

Good pr tho

modular_skyrat/modules/sec_haul/code/guns/guns.dm
RatFromTheJungle: I don't mean to trouble you with this information, but it is around equally priced to the armadyne guns, and even, more expensive then two of them (which are better in stats)
image
image

PR #14208, opened 2022-06-09
[NON-MODULAR] Bring back Ninja Cyborg

SarmentiCampbell:
## About The Pull Request
Bring back the selection of ninja cyborgs when a ninja decide to hack a borg. It got tested.

Yeah, It was a old feature and got overwritten by the modsuit pr, I'm unsure if i can do it in a way that is modular but i have no idea for it.

## Changelog

:cl:
add: Ninja Cyborgs are back.
/:cl:

### DIFF
--- a/code/modules/ninja/ninjaDrainAct.dm
+++ b/code/modules/ninja/ninjaDrainAct.dm
@@ -277,6 +277,12 @@
 	return COMPONENT_CANCEL_ATTACK_CHAIN
 
 /mob/living/silicon/robot/proc/ninjadrain_charge(mob/living/carbon/human/ninja, obj/item/mod/module/hacker/hacking_module)
+	//SKYRAT EDIT: ADDITION START
+	var/list/modelselected = list()
+	modelselected["Assault"] = "/obj/item/robot_model/ninja"
+	modelselected["Medical"] = "/obj/item/robot_model/ninja/ninja_medical"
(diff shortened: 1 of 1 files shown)


code/modules/ninja/ninjaDrainAct.dm
```diff
@@ -277,6 +277,12 @@
 	return COMPONENT_CANCEL_ATTACK_CHAIN
 
 /mob/living/silicon/robot/proc/ninjadrain_charge(mob/living/carbon/human/ninja, obj/item/mod/module/hacker/hacking_module)
+	//SKYRAT EDIT: ADDITION START
```
nevimer, line 280: To make this modular, put this in a new file and name the proc 

`/mob/living/silicon/robot/ninjadrain_charge(mob/living/carbon/human/ninja, obj/item/mod/module/hacker/hacking_module)`

and use . = ..() where you want the OG code here to start. You probably don't want that, so simply omitting . = ..() will do to keep it as a modular over-ride.

code/modules/ninja/ninjaDrainAct.dm
```diff
@@ -277,6 +277,12 @@
...
+	var/list/modelselected = list()
+	modelselected["Assault"] = "/obj/item/robot_model/ninja"
```
Zonespace27, line 282: i wish this didn't use strings but that's a problem for another PR

PR #14211, opened 2022-06-09
[MODULAR] Fixes the QM's office on all maps + fixes broken Icebox

Iamgoofball:
## About The Pull Request
The QM was missing head of staff equipment(an ID console, a card swiper, and a supply ID trimmer) on all the maps.
Icebox had a broken access helper on Shipping.

Bugfixes good.

## Changelog
:cl:
fix: Fixes the Quartermaster being unequipped for his job as a Head of Staff.
/:cl:

### DIFF
--- a/_maps/map_files/Blueshift/BlueShift_middle.dmm
+++ b/_maps/map_files/Blueshift/BlueShift_middle.dmm
@@ -4960,6 +4960,9 @@
 /area/station/security/medical)
 "bhr" = (
 /obj/effect/turf_decal/siding/wood/corner,
+/obj/machinery/modular_computer/console/preset/id{
+	dir = 8
+	},
 /turf/open/floor/carpet/royalblack,
 /area/station/cargo/qm)
 "bhA" = (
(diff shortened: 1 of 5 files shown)


Jolly-66 (changes requested):
Either I can't see them or find them, you forgot the card authenticator on:
- BlueShift
- Box

If you did put them in, and they're just not being rendered since MDB2 is still wonky. (I'm too hung up right now to pull locally and verify myself). 

Also you could've slotted the printer better in BlueShifts QM office a bit better. I would reccomend under their computers, but this is up to you. This is just a stylistic critique (also because the table is now disconnected and looks, awfully wonky).

Iamgoofball:
Blueshift has a card swiper already, and Icebox/Tram's renders don't work because multi-z moment.

Jolly-66:
> Icebox/Tram's renders don't work because multi-z moment.

They don't work with MDB**1**, but MDB**2** renders them fine. However, its a bit janky with wall mounted objects, which is why I wanted to comment on that fact.

I'll leave the printer up to you. Thank you for verifying the authenticators are there already.

PR #14231, opened 2022-06-10
[MIRROR] Lemon does weather effects [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/67469
--------------------

## About The Pull Request

Ok so I don't like how weather effects look when they run up against some sort of border, particularly walls and darkness.
They just sorta snap out of existence. This is cause they were recently moved from the AREA_PLANE to the ABOVE_LIGHTING_PLANE. 
This fixed things like snow or ash storms looking dim, but had some annoying side effects.

I'm gonna do my best to mesh the two, at least somewhat.
I've done two things. The first is letting weather choose to apply a second icon state to its area, one that draws above lighting, and moving the current overlay down to AREA_PLANE again.

Oh and the above lighting stuff gets made a bit see through, so it's not too intense.

This way darkness doesn't totally mask effects, it just makes them harder to see.

I've gone through the existing effects for this process. I basically just removed the "background" from any effects that had it. 
This way you see the particles at a reduced intensity, but not the "shadow" that the effect applies.

Oh also can't use glow with lava, cause it draws onto the floor, or under walls.

In the process of this I realized that snow looks really bad right now.
It's really jerky, has a very clear wrap point, and doesn't even complete a full animation. 
So I redid it, added some more flakes and made it "faster"

I also tweaked non smoothed lava (IE: floor is lava and the wizard ship)
It was scrolling a pixel a tick, but in steps of 3 ticks. Looked bad, I think the smoothing is better.

I'm not really a spriter or anything, this stuff just cheesed me off a bit

I wish I could show you video, but because these are all weather effects they compress horribly, even like an 8 second gif won't work. I'm sorry.
For what it's worth here's the current snow, might go some way to explaining my dislike.
Notice how it's fully overlayed over the background, so things look really muddy. Just extrapolate that out to all weather and you'll roughly get me.

dreamseeker_UmGO9CpzCg

Edit: photos
old snow:
image
image
new snow:
image
image
old ash:
image
image
new ash:
image
image
old rad:
image
new rad:
image

## Why It's Good For The Game

Game more pretty

## Changelog

:cl:
add: Weather effects will now be a bit more subtle in darkness, hopefully this looks nicer
imageadd: I've done some resprites to snow and non smoothing lava
/:cl:


code/datums/components/food/decomposition.dm
```diff
@@ -73,7 +73,11 @@
 
 	var/turf/open/open_turf = food.loc
 
+<<<<<<< HEAD
 	if(!istype(open_turf) || istype(open_turf, /turf/open/lava) || istype(open_turf, /turf/open/misc/asteroid/basalt)) //Are we actually in a valid open turf?
+=======
+	if(!istype(open_turf) || islava(open_turf) || istype(open_turf, /turf/open/misc/asteroid)) //Are we actually in a valid open turf?
+>>>>>>> e15e0f3e91b (Lemon does weather effects (#67469))
```
tf-4:
```suggestion
	if(!istype(open_turf) || islava(open_turf) || istype(open_turf, /turf/open/misc/asteroid)) //Are we actually in a valid open turf?
```

PR #14241, opened 2022-06-11
Makes status bar not show when title screen browser up

Tastyfish:
## About The Pull Request

Make the status bar in the bottom left go away when the title screen is up, since it otherwise is just covering up the corner and showing the ref URL's that the menu buttons go to.

Note that these 3 spots in the code are the sum total of places that the title browser is made visible or not.

Looks better

## Changelog

infinitely minor visual tweak


Tastyfish:
@Gandalf2k15 is the actual code owner.

PR #14276, opened 2022-06-12
Removing door slamming in oversized's quirk

Kush1Push1:
## About The Pull Request

Removing (originally commenting) door-slamming mechanic, which relays on the oversized's quirk.
I'm concern I don't know a discussing around the quirk (the flow of it appears to me frustrating enough to create one).
I'm not adding something new, but making the quirk more accessible for the majority of players. As far as it requires veteran status to have one, I don't see major problems with abuse of the quirk (there are no real benefits after adding slowdown, in my opinion).

1) It is nearly unplayable to control a character with the oversized quirk beside jobs without need to move frequently (assistants, for an example) due constant slamming doors. 
2) In current state, there are no round-start options to reach the size which the quirk gives to a character. 

After removing the mechanic, the quirk seems to be more playable. It doesn't give real benefits, as there is a slowdown just from having it.

## Changelog

:cl:
del: Removed door slamming mechanic if a character has the oversized quirk.
/:cl:


Nerev4r:
Oversized is a fully optional quirk. You don't need to run it, and you definitely don't need to run it to play a huge character.
From having months of accumulated experience, especially having been one of the first people to use it, having to walk everywhere really isn't that slow unless you're trying to actively chase someone, which isn't that hard either because I made firelocks not proc the effect. Not to mention, you get incredible punch damage and hulk tier throwing from this quirk.

Kush1Push1:
> Oversized is a fully optional quirk. You don't need to run it, and you definitely don't need to run it to play a huge character. From having months of accumulated experience, especially having been one of the first people to use it, having to walk everywhere really isn't that slow unless you're trying to actively chase someone, which isn't that hard either because I made firelocks not proc the effect. Not to mention, you get incredible punch damage and hulk tier throwing from this quirk.

You cannot setup your character's size above 1.5, which still has graphic artifacts. I don't have problems with the quirk itself but with barrier around 1.5. Maybe there is a better option to rise it to 2, Idk.

You can find objects and items in your backpack with more damage than the quirk gives you, after all. Slowdown and raised hunger makes it barely neutral, to me. But still, I only have the problem with lack of access to size without graphic errors.

Zonespace27 (changes requested):
Remove it or don't. Commenting out is bad.

Kush1Push1:
> Remove it or don't. Commenting out is bad.

Alri-ight... Done. Updating.

Nerev4r:
> > Oversized is a fully optional quirk. You don't need to run it, and you definitely don't need to run it to play a huge character. From having months of accumulated experience, especially having been one of the first people to use it, having to walk everywhere really isn't that slow unless you're trying to actively chase someone, which isn't that hard either because I made firelocks not proc the effect. Not to mention, you get incredible punch damage and hulk tier throwing from this quirk.
> 
> You cannot setup your character's size above 1.5, which still has graphic artifacts. I don't have problems with the quirk itself but with barrier around 1.5. Maybe there is a better option to rise it to 2, Idk.
> 
> You can find objects and items in your backpack with more damage than the quirk gives you, after all. Slowdown and raised hunger makes it barely neutral, to me. But still, I only have the problem with lack of access to size without graphic errors.

The raised hunger really isn't a problem. I've tested this myself and I've had other people like Sharknet and Dan Kelly report back to me on how it impacts them, and it doesn't. The slowdown, again, isn't either: you just have to be patient.
The damage isn't just the thing, your punches also have really high stamina damage. We would definitely need some kind of debuff to pay for that, and the airlock slamming is good enough.

Kush1Push1:
> > > Oversized is a fully optional quirk. You don't need to run it, and you definitely don't need to run it to play a huge character. From having months of accumulated experience, especially having been one of the first people to use it, having to walk everywhere really isn't that slow unless you're trying to actively chase someone, which isn't that hard either because I made firelocks not proc the effect. Not to mention, you get incredible punch damage and hulk tier throwing from this quirk.
> > 
> > 
> > You cannot setup your character's size above 1.5, which still has graphic artifacts. I don't have problems with the quirk itself but with barrier around 1.5. Maybe there is a better option to rise it to 2, Idk.
> > You can find objects and items in your backpack with more damage than the quirk gives you, after all. Slowdown and raised hunger makes it barely neutral, to me. But still, I only have the problem with lack of access to size without graphic errors.
> 
> The raised hunger really isn't a problem. I've tested this myself and I've had other people like Sharknet and Dan Kelly report back to me on how it impacts them, and it doesn't. The slowdown, again, isn't either: you just have to be patient. The damage isn't just the thing, your punches also have really high stamina damage. We would definitely need some kind of debuff to pay for that, and the airlock slamming is good enough.

It looks like overkill.

Again, I would prefer to have a mechanic possibility to don't take the quirk and, still, have x2 of the character's size. I think I'll make a second PR for it in case of denying

Gandalf2k15:
Not the way to do it and I'm blocking this change. This quirk is barely allowed in anyway.

PR #14286, opened 2022-06-13
Raising maximal character's body size

Kush1Push1:
## About The Pull Request

Getting custom character's body size to the same limit as oversized quirk has.

A possibility to _**roleplay**_ as a large character without mechanical necessity to slowdown at each airlock/without debuffs from oversized's quirk, as well as buffs.

## Changelog

qol: Raised maximal character's size from 1.5 to 2.


Gandalf2k15:
Absolutely not, this is not acceptable. This range should be going DOWN not up. The quirk was there to allow people that wanted bigger sprites to have them, while being restricted to vet crew and having many downsides, because the sprite size was obnoxious.

Kush1Push1:
Well, that was fast

PR #14313, opened 2022-06-14
[MODULAR] Fixes xeno-tail wag and MOD visibility

desertmote:
## About The Pull Request
This tail had a unique case where it would get hidden beneath MOD stuff. This was because it lacked '_primary' at the end of its icon_state name, even if it was listed as `color_src = USE_ONE_COLOR`.
It also lost its wagging, which was because it was not a `/mammal/wagging` subtype

Restores content that was lost by a refactor

## Changelog
:cl:
fix: The xeno-tail can wag again, and does not get hidden when wearing a MODsuit
/:cl:


GoldenAlpharex (approved):
I guess that works.

PR #14368, opened 2022-06-16
[MODULAR] Make Vendors Have at Least 5 of Cosmetic Clothing!

RimiNosha:
## About The Pull Request

Makes it so ClothesMate and AutoDrobe vendors have five of every item in a modular way!  
New to DMM, so if there's a preferred way to do this, let me know!

Having to miss out on your favourite outfit cause someone already bought it is kinda meh, right?

## Changelog

:cl:
balance: NanoTrasen has updated their clothing contracts to stock five of every clothing available in AutoDrobe and ClothesMate vendors.
/:cl:


tf-4 (approved):

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -13,7 +15,12 @@
 		premium += skyrat_premium
 	if(skyrat_contraband)
 		contraband += skyrat_contraband
-
+	if(skyrat_amount_override)
+		for (var/item in products)
+			products[item] = 5
```
ghost: This currently doesn't actually use the 'amount override' variable, it just forces 5 no matter what. If you put 'products[item] = skyrat_amount_override' then it'll actually use whatever we supply the vendor

Regardless, this is a bit of a troublesome way of doing it, as it sets ALL items to five - no more, no less.
I can look into helping improve it sometime later but git on phones SUCKS OH GOD

RimiNosha: Yea, I did think about checking if the item is a subtype of clothing and make sure there no armour or special tags for it before setting to 5-8 on every single vendor- though I've no idea how well that'd work.

Also, I can't believe I missed that! Will fix!

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -13,7 +15,12 @@
 		premium += skyrat_premium
 	if(skyrat_contraband)
 		contraband += skyrat_contraband
-
+	if(skyrat_amount_override)
+		for (var/item in products)
+			products[item] = skyrat_amount_override
```
Iamgoofball:
```suggestion
			if(products[item] < skyrat_amount_override)
				products[item] = skyrat_amount_override
```

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -13,7 +15,12 @@
...
+		for (var/item in premium)
+			premium[item] = skyrat_amount_override
```
Iamgoofball:
```suggestion
			if(premium[item] < skyrat_amount_override)
				premium[item] = skyrat_amount_override
```

Iamgoofball: These two changes will prevent already above 5 items from getting reduced down to 5, ie. Families clothing, which is meant to be available in bulk.

RimiNosha: Yup, I do have that in my local repo already, I just haven't pushed it yet!

RimiNosha:
I've now made it so this just applies to every vendor, and excludes items with the any of the following traits: `CHUNKYFINGERS`, `FASTMED`, `QUICK_CARRY`, `QUICKER_CARRY`, along with excluding certain paths of items from being affected, such as `/obj/item/clothing/suit/armor`

Thoughts?

ghost:
> I've now made it so this just applies to every vendor, and excludes items with the any of the following traits: `CHUNKYFINGERS`, `FASTMED`, `QUICK_CARRY`, `QUICKER_CARRY`, along with excluding certain paths of items from being affected, such as `/obj/item/clothing/suit/armor`
> 
> Thoughts?

might be best to make sure it doesnt apply to anything thats not a clothing item period, else vendors such as the medical vendor can vend, say, 5 crew trackers, or the secvend spawn 5 webbings.
Are webbings a clothing item? 🤔

tf-4:
I feel like having it on all vendors is gonna have unintended side effects

RimiNosha:
> > I've now made it so this just applies to every vendor, and excludes items with the any of the following traits: `CHUNKYFINGERS`, `FASTMED`, `QUICK_CARRY`, `QUICKER_CARRY`, along with excluding certain paths of items from being affected, such as `/obj/item/clothing/suit/armor`
> > Thoughts?
> 
> might be best to make sure it doesnt apply to anything thats not a clothing item period, else vendors such as the medical vendor can vend, say, 5 crew trackers, or the secvend spawn 5 webbings. Are webbings a clothing item? 🤔

Your first point is already taken care of, it doesn't touch anything that isn't a subpath of clothing

Secondly, I think belts and bags are not under clothes, but I'll double check and add cases for belts and backpacks if need be!

RimiNosha:
> I feel like having it on all vendors is gonna have unintended side effects

It shouldn't, at least any that'll cause gameplay-oriented items to be increased in amount, unless someone adds a new special item under clothing

I will say though that this code is a little cursed right now, cause I cannot for the life of me figure out to pull a default value of a path without instantiating it- 

If push comes to shove though, I can change it to look for specific items to increase instead, though it'll need updating anytime upstream adds clothes to vendors that we want a higher number of-

RimiNosha:
Okay, I've ran through all the vendors and I think I've gotten everything!

Belts and backpacks are not under clothing, so stuff like security webbing isn't anything to be concerned with!

Ready for merciless review, I think!

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +14,38 @@
...
+	for (var/item in premium)
+		if(premium[item] < 5 && ispath(item, /obj/item/clothing) && skyrat_IsNotGameplay(item))
+			premium[item] = 5
+			
 	QDEL_NULL(skyrat_products)
 	QDEL_NULL(skyrat_premium)
 	QDEL_NULL(skyrat_contraband)
 	return ..()
 
+/obj/machinery/vending/proc/skyrat_IsNotGameplay(var/obj/item/clothing/clothing_path)
```
GoldenAlpharex: That proc name doesn't follow the convention of `snake_case`, which would mean it should be looking like `skyrat_is_not_gameplay()`

Furthermore, that name is very ambiguous, and it shouldn't contain `skyrat` at the start of it, it's entirely unnecessary.

You also don't need to (and shouldn't) add `var` to the typepaths of proc parameters, it's already implied and handled the same way by the compiler.

Lastly, I would appreciate if you documented your proc, aka, added a comment starting with `/// ` on the line before that explain what this proc is supposed to do.

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +14,38 @@
...
+	if(TRAIT_CHUNKYFINGERS in clothing.clothing_traits)
+		return FALSE
+	if(TRAIT_FASTMED in clothing.clothing_traits)
+		return FALSE
+	if(TRAIT_QUICK_CARRY in clothing.clothing_traits)
+		return FALSE
+	if(TRAIT_QUICKER_CARRY in clothing.clothing_traits)
+		return FALSE
+	if(TRAIT_DEAF in clothing.clothing_traits)
+		return FALSE
```
GoldenAlpharex: Here's something a little cleaner that would do the same thing:
```suggestion
	var/banned_traits = list(
		TRAIT_CHUNKYFINGERS,
		TRAIT_FASTMED,
		TRAIT_QUICK_CARRY,
		TRAIT_QUICKER_CARRY,
		TRAIT_DEAF,
	)
	for(var/trait in banned_traits)
		if(trait in clothing.clothing_traits)
			return FALSE
```

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +14,38 @@
 	if(skyrat_contraband)
 		contraband += skyrat_contraband
 
+	/// Time to make clothes amounts consistent!
+	for (var/item in products)
+		if(products[item] < 5 && ispath(item, /obj/item/clothing) && skyrat_IsNotGameplay(item))
```
GoldenAlpharex: That `5` should be put into a define instead, something like `#define MINIMUM_CLOTHING_STOCK 5` at the start of the file, and then replacing all of those `5` with `MINIMUM_CLOTHING_STOCK` and finally adding at the end of the file a `#undef MINIMUM_CLOTHING_STOCK`.

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +14,38 @@
 	if(skyrat_contraband)
 		contraband += skyrat_contraband
 
+	/// Time to make clothes amounts consistent!
+	for (var/item in products)
```
GoldenAlpharex: If you're going to do an `ispath(item, /obj/item/clothing)` anyway, then you should do your for loop like this instead, and remove the `ispath` check from that if() statement, as it's going to do that type check for you and it's only going to take out the clothing items from `products`:
```suggestion
	for (var/obj/item/clothing/item in products)
```

GoldenAlpharex (dismissed):

modular_skyrat/modules/modular_vending/code/vending.dm
RimiNosha: - Oop, I looked at another proc and made assumptions, sorry!
- Yea, I think a better name would be `is_item_increase_banned` or something to that degree- and the `skyrat_` part was me once again assuming from `skyrat_products` and such!
- I already knew this, yet I still did it anyways- oops-
- Will do!
Thanks for taking time to review my coding travesty, pff-

modular_skyrat/modules/modular_vending/code/vending.dm
RimiNosha: Thanks! I'll do this, but I'll have the list be at the object level, cause I think creating a list for potentially 10+ items per vendor is kinda bad for performance.

modular_skyrat/modules/modular_vending/code/vending.dm
RimiNosha: I did really debate this, and this was part of the original code, but I don't see too much of a reason for things to have this number overridden, but I'll add it for customization's sake!

modular_skyrat/modules/modular_vending/code/vending.dm
RimiNosha: Ah, nice, I'm used to java where it throws a fit if you don't perfectly sanitize every little thing- force of habit!

RimiNosha:
Merged master to fix CI!

Funce (approved):
Code nicely commented, Golden's review has been noted and fixed. 

This looks good to go.

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +16,38 @@
 	if(skyrat_contraband)
 		contraband += skyrat_contraband
 
+	/// Time to make clothes amounts consistent!
```
jjpark-kb: use // instead of ///

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +16,38 @@
...
+		if(premium[item] < MINIMUM_CLOTHING_STOCK && is_item_increase_banned(item))
+			premium[item] = MINIMUM_CLOTHING_STOCK
+
 	QDEL_NULL(skyrat_products)
 	QDEL_NULL(skyrat_premium)
 	QDEL_NULL(skyrat_contraband)
 	return ..()
 
+/// This proc checks for forbidden traits cause it'd be pretty bad to have 5 insuls available to assistants roundstart at the vendor!
+/obj/machinery/vending/proc/is_item_increase_banned(obj/item/clothing/clothing_path)
```
jjpark-kb: proc name is really long and somewhat misleading; maybe check_banned() or something

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +16,38 @@
...
+	var/obj/item/clothing/clothing = new clothing_path()
+
+	/// Ignore earmuffs!
```
jjpark-kb: mentioned it in a previous comment: please use // to comment rather than /// which is for documentation.

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +16,38 @@
...
+	if(TRAIT_DEAF in clothing.clothing_traits)
+		return FALSE
```
jjpark-kb, line 38: maybe make this return true (same for others)

modular_skyrat/modules/modular_vending/code/vending.dm
```diff
@@ -14,8 +16,38 @@
...
+	/// Don't touch bodyarmour!
+	if(ispath(clothing, /obj/item/clothing/suit/armor))
+		return FALSE
+	/// Don't touch protective helmets, like riot helmets!
+	if(ispath(clothing, /obj/item/clothing/head/helmet))
+		return FALSE
+	/// Ignore all gloves, because it's almost impossible to check what they do...
+	if(ispath(clothing, /obj/item/clothing/gloves))
+		return FALSE
+	return TRUE
```
jjpark-kb, line 51: if you make the others true, make this false

modular_skyrat/modules/modular_vending/code/vending.dm
jjpark-kb: if you don't change the return trues to falses and vice versa, maybe rename it to allow_increase() or something

PR #14385, opened 2022-06-18
[MODULARIZATION] Surgery Tables now only apply numbing

softcerv:
## About The Pull Request
Moves the code, that was used to make surgery tables apply stasis, to a modular Skyrat file along with reverting the Skyrat edits to the table name and description.

Surgery tables now have two extra variables: stasis_capable and numbing_capable. These determine if a surgery table can apply numbing and/or stasis. 

Surgery tables by default now only apply numbing. 
I am keeping numbing on the surgery tables because without it, stasis beds would become the main go-to option due to our current surgery pain mechanics. 

This is something that makes sense to have considering we have gone back and reverted the stasis pod PR. In their current state, stasis surgery tables are both a cheaper and better option when compared with stasis beds, this invalidates the intended risk/reward balance that surgery tables and stasis beds were supposed to have.  

This PR also makes the code for stasis on surgery tables modular. 

## Changelog

:cl:
balance: surgery tables now only apply numbing
code: code for stasis on surgery tables is now modular. 
spellcheck: following the removal of stasis from surgery tables, they now use their original name and description.
/:cl:

### DIFF
--- a/code/game/objects/structures/tables_racks.dm
+++ b/code/game/objects/structures/tables_racks.dm
@@ -676,8 +676,8 @@
  */
 
 /obj/structure/table/optable//SKYRAT EDIT - ICON OVERRIDEN BY AESTHETICS - SEE MODULE
-	name = "stasis operating table" // SKYRAT EDIT name = "operating table"
-	desc = "Used for advanced medical procedures. Now comes with built in stasis technology, patented by Cinco: A Family Company!" // SKYRAT EDIT desc = "Used for advanced medical procedures."
+	name = "operating table"
+	desc = "Used for advanced medical procedures."
 	icon = 'icons/obj/surgery.dmi'
 	icon_state = "optable"
 	buildstack = /obj/item/stack/sheet/mineral/silver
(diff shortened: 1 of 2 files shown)


nevimer:
How do I turn on stasis now?

softcerv:
> How do I turn on stasis now?

you code in a surgery table that has `stasis_capable` enabled. Standard surgery tables can't do stasis anymore.

lessthnthree:
RIP the days where you got a statis, numbing bed all for the price of 1 silver. It was kind of a easy way out, maybe people will stop beelining to replace all the beds with surgery tables.

Seeing an upgraded surgery table with stasis available through research or something would be neat for another PR.

nevimer:
I know you were talking about removing stasis from surgery tables a week ago. How about adding like, a user attainable way to add stasis. Like to the computer, as a requirement, and science to have researched it with the normal cryo stuff.

softcerv:
> I know you were talking about removing stasis from surgery tables a week ago. How about adding like, a user attainable way to add stasis. Like to the computer, as a requirement, and science to have researched it with the normal cryo stuff.

Right now, I'm not going to do that. But if you or someone else wants to code that in, feel free.

nevimer:
Keep in mind that with this nerf, medical players are extremely unrobust AND this nerfs medical antagonism hard here.

softcerv:
> Keep in mind that with this nerf, medical players are extremely unrobust AND this nerfs medical antagonism hard here.

Stasis beds are still in the game for that reason, I made sure every map had them before I made this PR. 
Maybe this PR will help give medical players here the push to try and improve themselves.

nevimer:
> > Keep in mind that with this nerf, medical players are extremely unrobust AND this nerfs medical antagonism hard here.
> 
> Stasis beds are still in the game for that reason, I made sure every map had them before I made this PR. Maybe this PR will help give medical players here the push to try and improve themselves.

My main concern is that any surgery based objectives and gimmicks will *require* you to convince people to use n2o, and TG has the issue where people don't do surgery on tables - rather stasis beds - for this very reason. With stasis surgery, it blurred the line of "are they evil".

YakumoChen:
Maybe players should actually stop being LRP and being afraid of N2O when it doesn't make sense to want to be awake when doctors are gutting you like a pig

I remember having a fairly easy time convincing people to take morphine or N2O when we didn't have stasis on surgery tables. You can't do antag surgery on someone awake and aware anyway.

YakumoChen:
The only nerf I see here is that this makes revival harder since corpses will still rot on a surgery table.

I'm actually not sure why we're doing numbing instead of stasis on surgery tables? Operating computers can't connect to stasis beds. So the risk reward isn't there. Either you have very slow safe revivals or you have faster revivals with more corpse rot. Both are bad options.

softcerv:
> The only nerf I see here is that this makes revival harder since corpses will still rot on a surgery table.

Epi-pens are abundant and contain formaldehyde. 

>I'm actually not sure why we're doing numbing instead of stasis on surgery tables?

Because surgery tables having stasis disrupts the risk vs reward difference that surgery tables and stasis beds have right now. Not to mention, being able to make a surgery table with stasis for only one iron and one silver is way too good.

YakumoChen:
Stasis was added to surgery tables to make revival less painful for dead people. I don't see a reason to disrupt this. Reviving does not need a nerf, it's already shitty being dead.

The real problem is alive people using surgery beds without needing to be knocked out, I don't see a reason we should continue to encourage this.

> Because surgery tables having stasis disrupts the risk vs reward difference that surgery tables and stasis beds have right now. Not to mention, being able to make a surgery table with stasis for only one iron and one silver is way too good.

There is no risk versus reward, all this does is slow down revivals one way or another.

Zonespace27 (approved):

PR #14408, opened 2022-06-19
Resprites on-mob bust overlays

Ghommie:
## About The Pull Request
I'm respriting most of the icons in the breasts_onmob.dmi file, improving the outline and shading for the most part but also removing some rogue pixels and making so that the sprites for lower sizes actually cover the torso enough (long story short, back on Citadel after it had stopped syncing with tg, the female chest overlays of humans and some other species were resprited to be flatter, so that the sprites for size 0 to 3 or 4 would fit better, but because that can't be done here without making this repo and server's status as a downstream of tgstation more precarious than it currently is, I had to redo them).

Like many other things from clothings to markings to hair and even items, these sprites are supposed to be visual cues of how a character looks like. If can't tell what i'm looking at easily, then they are likely meh or bad sprites. The opposite isn't necessarily but let's not get too serious about it.
TL;DR sprite improvements.

## Changelog

:cl:
imageadd: resprited on-mob chest sprites. Mostly the same, save for a more visible outline and shading around the underside.
/:cl:


theselfish:
TITS

tgstation.dme
```diff
@@ -5341,7 +5341,7 @@
 #include "modular_skyrat\modules\gun_cargo\code\objects\vending.dm"
 #include "modular_skyrat\modules\gun_cargo\code\objects\guns\interdyne.dm"
 #include "modular_skyrat\modules\gun_cargo\code\objects\guns\izhevsk.dm"
-#include "modular_skyrat/modules/gun_cargo/code/objects/guns/oldarms.dm"
+#include "modular_skyrat\modules\gun_cargo\code\objects\guns\oldarms.dm"
```
theselfish:
```suggestion
#include "modular_skyrat/modules/gun_cargo/code/objects/guns/oldarms.dm"
```
I have no idea why this keeps happening.

Ghommie: Because I've used DM, it autocorrected the .dme file to use the right path separators.
About path separators, both \ (backslash) and / (slash) are acceptable on windows, but for linux (and probably other unix OS), only / is accepted, so it's reccomended to use slashes rather than backslashes.
I'll undo the change, only because I don't want conflicts.

Gandalf2k15 (approved):

Ghommie:
Beware of the speedmerge.

PR #14441, opened 2022-06-21
Makes ears/frills/horns match the same hide mutant parts logic of every other accessory.

### LINKED ISSUE #13556: Ears not affected by 'Hide Mutant Bodyparts'
pretty sure it may be an oversight...
## Reproduction:
use any kind of ear and if you look north east or west you can still see them even if you hide mutant parts, its at it's most obvious with larger ears, such as bunny ears (the ones that droop down alongside the head)

Tastyfish:
## About The Pull Request

For some reason, most sprite accessories, eg tails, etc. follow the logic of matching the HIDE* flags, unless Hide Mutant Bodyparts is on, then it's always hidden.

For some reason, mutant ears, horns, and frills followed a different logic of always being visible, unless Hide Mutant Bodyparts is on, then it actually uses the flags. Ironically meaning that unless the clothing is supposed to hide hair, the Hide Mutant Bodyparts command doesn't appear to the user to work.

Fixes #13556

This makes it impossible to make everything sensibly use the item flags which I would think should be the default behavior since ears poking through spacesuits and hoods doesn't really make sense, and following the flags for these slots was the default behavior to my knowledge (at least with teshari) before the external organs refactor stuff.

consistent logic.

## Changelog

:cl:
fix: Fixes visible ear visibility to work the way it did before.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
+++ b/modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
@@ -7,13 +7,13 @@
 
 /datum/sprite_accessory/ears/is_hidden(mob/living/carbon/human/human, obj/item/bodypart/bodypart)
 	if(
-		(human.head && (human.head.flags_inv & HIDEHAIR))
-		|| (human.wear_mask && (human.wear_mask.flags_inv & HIDEHAIR))
-		|| human.try_hide_mutant_parts
-		|| !bodypart
+		(human.head && (human.head.flags_inv & HIDEHAIR)) \
+		|| (human.wear_mask && (human.wear_mask.flags_inv & HIDEHAIR)) \
+		|| human.try_hide_mutant_parts \
+		|| !bodypart \
 	)
 		//This line basically checks if we FORCE accessory-ears to show, for items with earholes like Balaclavas and Luchador masks
-		if(!(H.head && (H.head.flags_inv & SHOWSPRITEEARS) || (H.wear_mask && (H.wear_mask.flags_inv & SHOWSPRITEEARS)) || !HD))
+		if(!(human.head && (human.head.flags_inv & SHOWSPRITEEARS) || (human.wear_mask && (human.wear_mask.flags_inv & SHOWSPRITEEARS)) || !bodypart))
 			return TRUE
 	return FALSE
 
--- a/modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
+++ b/modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
@@ -5,8 +5,13 @@
 	relevent_layers = list(BODY_BEHIND_LAYER, BODY_ADJ_LAYER, BODY_FRONT_LAYER)
 	genetic = TRUE
 
-/datum/sprite_accessory/ears/is_hidden(mob/living/carbon/human/H, obj/item/bodypart/HD)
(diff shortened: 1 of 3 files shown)


modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/frills.dm
```diff
@@ -6,7 +6,7 @@
 	genetic = TRUE
 
 /datum/sprite_accessory/frills/is_hidden(mob/living/carbon/human/H, obj/item/bodypart/HD)
-	if(H.head && (H.try_hide_mutant_parts || (H.head.flags_inv & HIDEEARS) || !HD))
+	if((H.head && (H.head.flags_inv & HIDEEARS)) || H.try_hide_mutant_parts || !HD)
```
Gandalf2k15: now might be a good time to make it human readable :)

desertmote:
this is a revert of [#10643](https://github.com/Skyrat-SS13/Skyrat-tg/pull/10643/files) which means horns are going to be broken off and hidden again. makes 0 sense, even less so than ears outside of suits

Tastyfish:
> this is a revert of https://github.com/Skyrat-SS13/Skyrat-tg/pull/10643/files which means horns are going to be broken off and hidden again. makes 0 sense, even less so than ears outside of suits

In that case, if we want horns to poke through, it would make more sense to make horns use the logic of
```
if(H.try_hide_mutant_parts) || !HD)
```

would it not? I can do that.

modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
```diff
@@ -5,8 +5,13 @@
...
-	if(H.head && ((H.head.flags_inv & HIDEHAIR)  && H.try_hide_mutant_parts) || ((H.wear_mask && (H.wear_mask.flags_inv & HIDEHAIR))  && H.try_hide_mutant_parts) || !HD)
+/datum/sprite_accessory/ears/is_hidden(mob/living/carbon/human/human, obj/item/bodypart/bodypart)
+	if(
+		(human.head && (human.head.flags_inv & HIDEHAIR))
+		|| (human.wear_mask && (human.wear_mask.flags_inv & HIDEHAIR))
+		|| human.try_hide_mutant_parts
+		|| !bodypart
+	)
 		//This line basically checks if we FORCE accessory-ears to show, for items with earholes like Balaclavas and Luchador masks
 		if(!(H.head && (H.head.flags_inv & SHOWSPRITEEARS) || (H.wear_mask && (H.wear_mask.flags_inv & SHOWSPRITEEARS)) || !HD))
```
GoldenAlpharex: You forgot a few variables here, space cowboy

tf-4:
What would be nice is for the user to be able to show / hide bodyparts individually in the same way the show/hide underwear menu works.
Not that I expect you to do that in this PR, it's out of scope, but if anyone wants to.

It would solve a few issues with wanting some parts hidden but not all

modular_skyrat/modules/customization/modules/mob/dead/new_player/sprite_accessories/ears.dm
Tastyfish: 😄

Tastyfish:
> What would be nice is for the user to be able to show / hide bodyparts individually in the same way the show/hide underwear menu works. Not that I expect you to do that in this PR, it's out of scope, but if anyone wants to.
> 
> It would solve a few issues with wanting some parts hidden but not all

I agree that the show / hide bodyparts command is basically useless as it's currently implemented.

PR #14457, opened 2022-06-22
Lowers chem regen for changelings.

IsaacTheSharkWolf:
It's still like a good 0.2 above the standard TG 0.5

## About The Pull Request

Simply changes a single variable to reduce the chemical regen of changelings to something a little more reasonable.
Former: 1.5 chem regen per second.
Now: 0.7 chem regen per second.
Base: 0.5 chem regen per second.

Changelings had tripple the chemical regen over base TG changelings, and there's no major reason for them to have this since even the new guncargo guns have trouble dispatching a changeling.
Changelings should be powerful but tripple the chemical regen is a little much.
Hopefully they're a little less painful to fight as they won't be able to heal as rapidly and constantly, making a changeling player consider their actions more carefully.

## Changelog

:cl:

balance: Rebalances changeling chemical regen to 0.7 from 1.5

/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -41,7 +41,7 @@
 	/// The max chemical storage the changeling currently has.
 	var/total_chem_storage = 100 // SKYRAT EDIT - ORIGINAL: 75
 	/// The chemical recharge rate per life tick.
-	var/chem_recharge_rate = 1.5 // SKYRAT EDIT - ORIGINAL: 0.5
+	var/chem_recharge_rate = 0.7 // SKYRAT EDIT - ORIGINAL: 0.5
 	/// Any additional modifiers triggered by changelings that modify the chem_recharge_rate.
 	var/chem_recharge_slowdown = 0
 	/// The range this ling can sting things.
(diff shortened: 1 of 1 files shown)


YakumoChen:
I can't say I've had a single problem fighting changelings, considering the Security Swarm almost always overpowers them, disagree with this change.

lings are weak to burn damage so lasers are always the best bet to kill them.

tf-4:
Try 1

IsaacTheSharkWolf:
> I can't say I've had a single problem fighting changelings, considering the Security Swarm almost always overpowers them, disagree with this change.
> 
> lings are weak to burn damage so lasers are always the best bet to kill them.

it's mostly to ensure the fleshmend ability isn't quite as spammable, still very potent (changelings are already powerful on base TG, with say, 1 chem regen they have double that of the standard.

YakumoChen:
Fleshmend isn't even really spammable, it has safeguards against that. It barely heals anything if you use it again consecutively. They also regen burn damage at half the rate and can't regen when on fire at all, either, again making lasers the choice to fuck them up.

Gandalf2k15:
We really don't have any issues with changelings, and this was changed for a reason(gunsec) and now that we have gunsec+++ it's even more relevant. Also, there is an upcoming change that will probably render the issues null, so for this reason, I'm closing this.

Xymec:
Simply, guncargo shouldn't be able to destroy antags in one spray, it ruins balance and defeats the point of said antag being a threat to the station.

PR #14465, opened 2022-06-23
Re-adds the one true permabrig to Delta

Jakksergal:
## About The Pull Request
Re-adds @KathrinBailey 's delta permabrig previously obliterated by the access overhaul. Probably butchered it somewhere along the lines but I attempted to fix all the lingering access issues.

The default brig of Delta station, as well as a few others, are not designed to cater towards Skyrat's permabrig prisoners.

## Changelog
Re-adds KathrinBailey's Delta Permabrig from the most recent version I could find.

:cl: Jakkie Sergal
add: Re-adds KathrinBailey Delta Permabrig
/:cl:


Zonespace27:
Dope. I would recommend taking this upstream eventually so it doesn't get wiped next map reset

PR #14469, opened 2022-06-23
Ashwalkers killing megafauna will always drop the crusher trophy

Zonespace27:
## About The Pull Request
If there is an ashwalker in the presence of a mega when it dies, it will drop the crusher trophy, regardless of if a crusher was involved or not.

Jake was planning a ritual thingy for ashwalkers, and I wanted to add some really rare components, so crusher trophies seemed to fit well. Also, if you're killing megas as an ashie, you probably put more effort in than a crusher miner
## Changelog

:cl:
add: Megafauna will always drop their crusher trophy when killed by ashwalkers
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/_megafauna.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/_megafauna.dm
@@ -91,6 +91,13 @@
 	if(crusher_dmg && crusher_loot && crusher_dmg.total_damage >= maxHealth * 0.6)
 		spawn_crusher_loot()
 		crusher_kill = TRUE
+	//SKYRAT EDIT START
+	for(var/mob/living/carbon/human/maybe_ashie in range(7))
+		if(is_species(maybe_ashie, /datum/species/lizard/ashwalker))
+			to_chat(maybe_ashie, span_notice(span_bold("The Necropolis has blessed your kill, dropping its trophy!")))
+			spawn_crusher_loot()
+			break
+	//SKYRAT EDIT END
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/hostile/megafauna/_megafauna.dm
```diff
@@ -91,6 +91,13 @@
 	if(crusher_dmg && crusher_loot && crusher_dmg.total_damage >= maxHealth * 0.6)
 		spawn_crusher_loot()
 		crusher_kill = TRUE
+	//SKYRAT EDIT START
+	for(var/mob/living/carbon/human/maybe_ashie in range(7))
+		if(is_species(maybe_ashie, /datum/species/lizard/ashwalker))
+			to_chat(maybe_ashie, span_notice(span_bold("The Necropolis has blessed your kill, dropping its trophy!")))
+			spawn_crusher_loot()
+			break
```
jjpark-kb: early return instead? if not species then continue?

Paxilmaniac:
Aw sweet kidnapping ashwalkers so you can get the crusher trophy of everything you kill

code/modules/mob/living/simple_animal/hostile/megafauna/_megafauna.dm
Zonespace27: Alternative solution: entirely recode it

jjpark-kb (approved):
all for it.

bluetshirtguy:
ashwalkers try not to power up themselves over and over challenge[Impossible]

PR #14480, opened 2022-06-24
Stops mismatching markings from showing up on the dropdowns without "allow mismatched parts" on

Ghommie:
## About The Pull Request
Ok, it seems the `recommended_species` variable for markings hasn't been put to use on the tgui menu unlike the homonymous var for presets. This should fix that oversight.

Having your lists filled with potentially bad picks isn't a good thing and it deters people from character customization.

## Changelog

:cl:
fix: Stopped mismatching markings from showing up on the dropdowns without "allow mismatched parts" on.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/client/preferences/middleware/limbs_and_markings.dm
+++ b/modular_skyrat/master_files/code/modules/client/preferences/middleware/limbs_and_markings.dm
@@ -193,6 +193,8 @@
 
 /datum/preference_middleware/limbs_and_markings/get_ui_data(mob/user)
 	var/list/data = list()
+	var/datum/species/species_type = preferences.read_preference(/datum/preference/choiced/species)
+	var/allow_mismatched_parts = preferences.read_preference(/datum/preference/toggle/allow_mismatched_parts)
 	if(!robotic_styles)
 		robotic_styles = list()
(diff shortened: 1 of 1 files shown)


Iamgoofball:
Markings shouldn't be species based, they're basically tattoos

Ghommie:
> Markings shouldn't be species based, they're basically tattoos

I hope you come to understand I haven't made this PR to filter out unreccomended markings because they won't look good on the blob of pixels we call a character or stop them from taking them. I'm doing this because most players don't exactly bother about markings that are clearly meant for certain species unless the character's one of said species. And I don't see a concrete reason to include such markings in the list -- unless the "allow mismatched parts" setting is on. Again, they're NOT species-exclusive -- in the face of the awful lot of choices we have that makes this sort of customization even more tedious.

Iamgoofball:
The solution, I think, is we find a better way to preview markings before applying them

Ghommie:
> The solution, I think, is we find a better way to preview markings before applying them

Even something similar to the component for hair and underclothes selection would help, but I wouldn't call it a solution; still an awful lot of options to navigate through, some of which are far from germane to your character.

EDIT: To be clear though, I'm not going to touch the UI tsx files. You can apply your own suggestion in a separate PR if you want to.

Gandalf2k15 (approved):

PR #14531, opened 2022-06-26
Fixes lone infiltrator, makes it a light ruleset

Zonespace27:
## About The Pull Request
This never actually ran, I think? The pref was never given

Its had its weight reduced to a more reasonable 4 (from 5) and moved to a light ruleset instead of a heavy, since this is effectively a traitor+, not worth a heavy roll.

Bugs bad

## Changelog

:cl:
fix: Lone Infiltrator should now run and is a light ruleset, with slightly reduced weight
/:cl:

### DIFF
--- a/modular_skyrat/modules/Midroundtraitor/code/event.dm
+++ b/modular_skyrat/modules/Midroundtraitor/code/event.dm
@@ -1,11 +1,27 @@
 /datum/dynamic_ruleset/midround/from_ghosts/lone_infiltrator
 	name = "Lone Infiltrator"
 	antag_datum = /datum/antagonist/traitor/infiltrator
-	midround_ruleset_style = MIDROUND_RULESET_STYLE_HEAVY
-	antag_flag = ROLE_LONE_INFILTRATOR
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/Midroundtraitor/code/event.dm
```diff
@@ -1,11 +1,27 @@
 /datum/dynamic_ruleset/midround/from_ghosts/lone_infiltrator
 	name = "Lone Infiltrator"
 	antag_datum = /datum/antagonist/traitor/infiltrator
-	midround_ruleset_style = MIDROUND_RULESET_STYLE_HEAVY
-	antag_flag = ROLE_LONE_INFILTRATOR
-	restricted_roles = list(JOB_CYBORG, JOB_AI, JOB_SECURITY_OFFICER, JOB_WARDEN, JOB_DETECTIVE, JOB_HEAD_OF_SECURITY, JOB_CAPTAIN, JOB_CORRECTIONS_OFFICER, JOB_VANGUARD_OPERATIVE, JOB_NT_REP, JOB_BLUESHIELD, JOB_ORDERLY, JOB_BOUNCER, JOB_CUSTOMS_AGENT, JOB_ENGINEERING_GUARD, JOB_SCIENCE_GUARD) //SKYRAT EDIT - Sec_haul
+	midround_ruleset_style = MIDROUND_RULESET_STYLE_LIGHT
+	antag_flag = ROLE_TRAITOR // It's basically this anyway
+	restricted_roles = list(JOB_CYBORG,
```
Iamgoofball, line 6: we should really make this a define

PR #14533, opened 2022-06-26
Attack of Oversized

Microvolnovka19:
## About The Pull Request

Adds "tail sweep"(like xeno queen's) ability to Oversized ones. 

Has 50 seconds cooldown,
range of 2 tiles,
awful sprites of shark and general tails.

Maybe I'll add more sprites if this goes on test merge/merge.

Also I have almost zero experience in coding/spriting. Don't judge me.

Currently the Oversized quirk already has some "unique" mechanics like getting brain damage from airlocks. But I think it's not enough! To compensate the oversized clumsiness and slowness I've added an ability to defend themself. No more situations where you are helpless when someone you fighting with hides under your big sprite.

## Changelog

:cl:
add: Oversized people now can tail sweep(if they have tail)
/:cl:


SarmentiCampbell:
I genuinely do not think they need that to defend themselves
As Oversized are already pretty good at defending themselves using their fists.

Zonespace27:
Oversized already has a fair few buffs for their downsides, they don't need more buffs at this rate. (and this is way too good to give them anyway)

PR #14537, opened 2022-06-26
You must get maintainer+ pre-approval to make changes to oversized

Zonespace27:
## About The Pull Request
This is a policy thing because of how this quirk is fairly controversial and seems to have a proposed rebalance every few weeks. Failure to get pre-approval (ideally screenshotted approval in the PR) may result in your PR being closed at maintainer discretion

A bit more control over a quirk whose existence is hanging by the thinnest of threads
Approved [here](https://discord.com/channels/596783386295795713/681885545210708053/990699192559231016)


Jolly-66:
Not sure if it was or is intentional but the link goes to a private chat, that only [we] can see

Zonespace27:
> Not sure if it was or is intentional but the link goes to a private chat, that only [we] can see

Intended

PR #14563, opened 2022-06-27
[MODULAR] No forced lewd emotes for those who don't want them!

RimiNosha:
## About The Pull Request
Makes it so all forced lewd emotes go through a check before happening, this is a somewhat lazy, but relatively clean solution!  

Also serves as a stopgap for those who suffer from the permanently-horny-itus bug, I suppose-

Moan (x12) during your essay writing lewd adventures gets a little annoying- also allows certain types of folk to do public ERP without being so obvious!

Also makes it so command can stop moaning after some intense ERP cause the alert got set to red.

## Changelog

:cl:
code: Added an ERP preference which allows you to disable automatic lewd emotes for yourself.
/:cl:


modular_skyrat/modules/modular_items/lewd_items/code/arousal_system.dm
```diff
@@ -514,6 +514,12 @@
 				span_purple("You can't have an orgasm!"))
 			return TRUE
 
+// Let's not force lewd emotes from folk who don't want them, mmm~?
+/mob/living/carbon/proc/try_lewd_autoemote(var/emote)
```
Zonespace27:
```suggestion
/mob/living/carbon/proc/try_lewd_autoemote(emote)
```

RimiNosha: Ah, nice catch!

hugmeorelse:
Neato. More control over your character and just having the option on hand is good, given it doesn't add unnecessary bloat.

PR #14015, opened 2022-05-31
[MIRROR] The Humanening?! But cheeked up (enlarged posterior) on a Sunday night?!?! As per our sprite lead's review????? [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/67402
--------------------

## About The Pull Request

Old pr: #67298

It was closed because the last spriter was too afraid to add back asses as per twat's request, and so it died. But I really do want the humanening! Not because I care about ass or assn't but because I like the face and better visuals in general. So I've elected to be the ass restorer and save the pull request, god bless america

## Pictures

My pr, only the side sprite was changed

image

Old pr so you can see the full thing

image

## Why It's Good For The Game

Taken from the old pr

Currently humans are one of the oldest sprites in the game, dating in-part to R4407. This is very obvious if you look at it for more then second. But this, this is brand new! Better looking sprites straight for your eyeballs to enjoy.

This'll prob take some getting used to, which is fair. But I feel this is an important step. So, yeah.

## Changelog

:cl: axietheaxolotl for the human, armhulenn for the butt and nothing else
imageadd: Brand new human sprites!
/:cl:

PR #14593, opened 2022-06-28
[DNM] [TM ONLY] Revert "[MIRROR] The Humanening?! But cheeked up (enlarged posterior) on a Sunday night?!?! As per our sprite lead's review????? [MDB IGNORE]"

jjpark-kb:
Reverts Skyrat-SS13/Skyrat-tg#14015

Why?: The Revert PR is meant to be temporary and is only going to be present until a fix is made.


Nerev4r:
Why.

Acer202:
Oh goodness please I can't look at the humans without their clothes on after the humanening. Too cakey and weird. And the chest sprites just aren't doing it with the new style.

Also, consider reverting #14408 in this as well, since they were sprited for the humanening chest and it looks funky now? Or whichever re-sprited them for the humanening chest in the first place. Am not sure.

Gandalf2k15:
We don't need this anymore.

PR #14602, opened 2022-06-28
Fixes mining and gulag shuttles in a way that will stay through map resets - I lied its modular

Paxilmaniac:
## About The Pull Request
image

I mean fine ok I'll do it

but you've got to map reset again in order for these to work :trollface: 

I mean I guess it's bad if there's more stuff to reset manually every map reset, so that fixes that at the cost of some non-modular edits

## Changelog

:cl:
fix: one day we might finally be free of having to manually put back in every single skyrat specific feature every time we do a map reset
/:cl:


Paxilmaniac:
Actually now that I think about it I could probably make this modular, hold on

PR #14190, opened 2022-06-08
Fixes for changeling transformations. Part 2.

Christmas5:
## About The Pull Request
[A small continuation of #13527.](https://github.com/Skyrat-SS13/Skyrat-tg/pull/13527)

There are things I missed and/or haven't been able to fix in the first PR.
This one adds screams, laughs and age to changeling profiles.

I began working on quirks as well, but I'm not entirely sure how to proceed as copying every single quirk could cause balancing issues. For example, I don't believe a changeling shapeshifting into someone blind should become blind as well.
Someone suggested a toggle button, allowing the changeling to switch between their target's quirks and their own.

Also, the sprite in your examine tab will now shapeshift in sync with your character... and the sprite becomes bald. Not ideal but better than keeping the initial one.

**Issues left:**
- After shapeshifting as a changeling, your sprite in the examine tab will become permanently bald.
- Using lesser form with non-standard body sizes still breaks sprites.
- Clothing overlays do not update properly when using lesser form/human form.
- When using last resort, sometimes the injected egg won't hatch.
- Organs do not update entirely as they should.
- Health scanners can still tell whether you are truly blind and/or deaf, and they also read your true brain traumas.
- The sechud icons for weapon permits do not copy over.

## Changelog

**Changeling profiles save more information about the target.**
- Scream and laugh types.
- Age.
- Quirks.

**A changeling's quirks will now change when shapeshifting.**
Because I'm not yet sure how to balance it, you will not receive all of your targets quirks when shapeshifting. The quirks which you can lose/gain when shapeshifting are defined in _var/list/mimicable_quirks_list_ in _changeling.dm._

`	var/list/mimicable_quirks_list = list(
		"Bad Touch",
		"Sensitive Snout",
		"Ash aspect (Emotes)",
		"Canidae Traits",
		"Excitable!",
		"Feline Traits",
		"Floral aspect (Emotes)",
		"Heterochromatic",
		"Hydra Heads",
		"Oversized",
		"Personal Space",
		"Pseudobulbar Affect",
		"Shifty Eyes",
		"Smooth-Headed",
		"Sparkle aspect (Emotes)",
		"Water aspect (Emotes)",
		"Webbing aspect (Emotes)",
		"Friendly",
	)`

Health analyzers now check if you are a changeling before reading your quirks, and if you are, they will read the quirks of the person you're disguised as, instead of reading your true quirks.

[...]

PR #14615, opened 2022-06-29
Fixes for changelings. Part 3.

Christmas5:
## About The Pull Request
[A further continuation on #14190.](https://github.com/Skyrat-SS13/Skyrat-tg/pull/14190)

Current flesh ID's are able to trick sechuds by displaying an occupation icon. Still, they **cannot** display a gun permit icon. It does not matter if either you or your target actually have a gun permit, a flesh ID just won't display it.

**Security won't mix up an assistant with a changeling, if the assistant has a gun permit: It functions partly like a mindshield.**

Of course, that wasn't intended, right? So boom, let's make flesh ID's copy the target's gun permit.

## Changelog

**Changeling profiles now store the target's gun permit status.**
This does not give your flesh ID any sort of access, it only tricks sechuds into displaying a gun permit.

**Lesser form now resets your body size correctly before monkeyizing you.**
Forgot to add _/update_body_size()._ This prevents the user from becoming a monkey with an irregular body size.

**Added a fair bit of comments to the changes I've made here and in my previous PRs, as per request.**
Also, I acknowledge that there's a lot that I should have put upstream. Initially I wanted to keep everything here because most of the issues I was working with were solely related to shapeshifting as anthros. I'm now moving what I can to /tg/.
https://github.com/tgstation/tgstation/pull/68117

**Added two defines for gun permit states, as per request.**
```
#define HUD_PERMIT_YES "hud_permit"
#define HUD_PERMIT_NO "hudfan_no"
```

:cl:
fix: Shapeshifting now updates your flesh ID's gun permit sechud icon.
fix: Lesser form resets your body size before monkeyizing.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -100,7 +100,7 @@
 
 	var/true_form_death //SKYRAT EDIT ADDITION: The time that the horror form died.
 	
-	/// SKYRAT EDIT START
+	/// SKYRAT EDIT START	( TODO: Move upstream. (Except for some of the quirks in the list.) )
 	///	Keeps track of the currently selected profile.
 	var/datum/changeling_profile/current_profile = null
 	/*	
@@ -495,7 +495,7 @@
 	new_profile.undershirt = target.undershirt
 	new_profile.socks = target.socks
 	
-	// SKYRAT EDIT START
+	// SKYRAT EDIT START	( TODO: Move some of this upstream. )
 	new_profile.underwear_color = target.underwear_color
 	new_profile.undershirt_color = target.undershirt_color
 	new_profile.socks_color = target.socks_color
@@ -745,7 +745,7 @@
 	user.undershirt = chosen_profile.undershirt
 	user.socks = chosen_profile.socks
 	
-	// SKYRAT EDIT START
+	// SKYRAT EDIT START	( TODO: Move some of this upstream. )
 	user.underwear_color = chosen_profile.underwear_color
 	user.undershirt_color = chosen_profile.undershirt_color
 	user.socks_color = chosen_profile.socks_color
@@ -763,7 +763,7 @@
 	user.selected_laugh = new chosen_profile.laugh_type
 	user.age = chosen_profile.age
 	
-	/*
+	/*	
 	 *	Remove old quirks and copy over new ones from the chosen profile.
 	 *
 	 *	Only quirks from the mimicable_quirks_list will be removed and/or copied over.
@@ -899,6 +899,7 @@
 	 *	Usually it's not a major issue: The changeling can just shapeshift again into the same target.
 	 *	The temporary fix is to run transfer_identity and updateappearance twice, effectively updating
 	 *	the changeling's DNA and appearance twice, so that they don't have to shapeshift twice.
+	 *	(Seems to be related to anthro stuff and mutant parts.)
 	 */
 	chosen_dna.transfer_identity(user, TRUE)
 	user.updateappearance(mutcolor_update = TRUE, eyeorgancolor_update = TRUE)
@@ -947,7 +948,7 @@
 	/// ID HUD icon associated with the profile
 	var/id_icon
 	
-	/// SKYRAT EDIT START
+	/// SKYRAT EDIT START	( TODO: Move some of this upstream. )
 	/// The colour of the underwear worn by the profile source.
 	var/underwear_color
 	/// The colour of the undershirt worn by the profile source.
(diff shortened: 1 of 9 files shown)


tf-4:
The radial menu is probably something best for upstream.

Christmas5:
> The radial menu is probably something best for upstream.

That is true. I'll make a PR there when I have the time.

code/modules/antagonists/changeling/changeling.dm
```diff
@@ -504,6 +504,8 @@
 	new_profile.age = target.age
 	for(var/datum/quirk/target_quirk in target.quirks)
 		LAZYADD(new_profile.quirks, new target_quirk.type)
+	if(target.wear_id?.get_gun_permit_iconstate()=="hud_permit")
```
GoldenAlpharex: I feel like `"hud_permit"` should be a define somewhere, because it just sounds to me like an issue waiting to happen.
```suggestion
	if(target.wear_id?.get_gun_permit_iconstate() == "hud_permit")
```

code/modules/antagonists/changeling/changeling.dm
```diff
@@ -934,6 +937,7 @@
 	var/laugh_type
 	var/age
 	var/list/quirks = list()
+	var/gun_permit
```
GoldenAlpharex: So, I know I haven't reviewed the previous ones properly, but I'd like for you to go through those variables and add some documentation that just explains what those variables are.
You can do variable documentation like this:
```
/// A variable that's used to calculate the distance between our mob and the sun.
var/distance_from_sun
```

code/modules/antagonists/changeling/powers/transform.dm
```diff
@@ -113,6 +113,7 @@
 	var/icon/cached_flat_icon
 	/// HUD job icon of the ID
 	var/hud_icon
+	var/gun_permit // SKYRAT EDIT
```
GoldenAlpharex, line 117: Documentation here too please :)

modular_skyrat/modules/gun_cargo/code/objects/access.dm
```diff
@@ -1,6 +1,13 @@
 /obj/item/proc/get_gun_permit_iconstate()
-	var/obj/item/card/id/id_card = GetID()
 
+	if(istype(src, /obj/item/changeling/id))
+		var/obj/item/changeling/id/changeling_card = src
+		if(changeling_card.gun_permit)
+			return "hud_permit"
+		return "hudfan_no"
```
GoldenAlpharex: Those two strings should probably be defines instead.

GoldenAlpharex (changes requested):

code/modules/antagonists/changeling/changeling.dm
Christmas5: > I feel like `"hud_permit"` should be a define somewhere, because it just sounds to me like an issue waiting to happen.

Ahh, can they go in /code/_DEFINES/~skyrat_defines/gun.dm? I assume after that I just change any code with hud_permit/hudfan_no to the defines?

Also, what kind of issues can it cause? Asking because I'm fairly new here.

code/modules/antagonists/changeling/changeling.dm
Christmas5: Went through and commented on everything as best as I could. Let me know if I missed something or if it's not enough!

code/modules/antagonists/changeling/changeling.dm
Christmas5: I wrote two and put them in there, then I renamed the strings in data_huds.dm to the new defines.
hudfan_no is also used for clown/mime fan huds in that same file, but I'll leave that be since it's /tg/.

tf-4:
@GoldenAlpharex

code/modules/antagonists/changeling/changeling.dm
```diff
@@ -916,24 +954,43 @@
...
 	var/scream_type
+	/// The typepath of the selected_laugh used by the profile source.
 	var/laugh_type
+	/// The age of the profile source.
 	var/age
+	/// The quirks of the profile source.
 	var/list/quirks = list()
+	/// Whether the profile source displayed a gun permit icon when looked at with a sechud.
+	var/gun_permit
 	/// SKYRAT EDIT END
```
GoldenAlpharex, line 994: This should likely also all be moved to a modular file instead of being added here non-modularly.

code/modules/antagonists/changeling/powers/transform.dm
```diff
@@ -113,6 +113,8 @@
 	var/icon/cached_flat_icon
 	/// HUD job icon of the ID
 	var/hud_icon
+	/// Gun permit of the ID (TRUE or FALSE, it either has one or it doesn't).
+	var/gun_permit // SKYRAT EDIT
```
GoldenAlpharex, line 117: All things considered, this should also be moved to a modular folder, since it's something that doesn't require it to be non-modular.

GoldenAlpharex (changes requested):

Gandalf2k15:
reopen when you are ready to finish this

PR #14624, opened 2022-06-29
[SEMI-MODULAR] Multilinguist Quirk!

RimiNosha:
## About The Pull Request
Allows you to have five languages total!

Trade off? Ten whole points! That's a hefty price to pay for a positive quirk that only gives you extra fluff with fellow crewmembers, I think!

Disclaimer: I have no idea what the language balance is, but with the ashtongue PR being merged, I think this hopefully should be 100%  crewmember-oriented fluff!

Of course, any tips and suggestions are more than welcomed, along with just telling me my 10-minute PR isn't good for SR!

It's feasible for folk to have a comprehension of several languages, I think, so this would be fun for those who like flexing their language skills!

And with this quirk's cost, it's guaranteed to produce some quirky characters for those who pick it, which'd be fun to see and RP with!

## Changelog

:cl:
add: Multilinguist, which gives you two language points for the cost of 10 balance.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/traits.dm
+++ b/code/__DEFINES/~skyrat_defines/traits.dm
@@ -12,6 +12,7 @@
 #define TRAIT_FREE_GHOST "free_ghost" // Can ghost and return freely with this trait
 #define GLOVES_TRAIT "gloves_trait" //Traits associated with wearing gloves
 #define QUIRK_LINGUIST "Linguist" // Extra language point.
+#define QUIRK_MULTILINGUIST "Multilinguist" // Two extra language points.
 #define GLUED_ITEM_TRAIT "glued-item" // This is for glued items, undroppable. Syndie glue applies this.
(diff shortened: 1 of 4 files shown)


Nerev4r:
This seems like way too much for the way our languages are right now. Half of them are meaningless, and we already have people filling their third slot with nonsense just because they feel the need to fill it. Not to mention, this takes away from the usefulness of curators/pAIs/circuit translators.

RimiNosha:
Yea, figures that'd be an issue...

I'll leave this open for a bit, but now that you brought up that point, I'm likely gonna close this- depends on what the overall consensus is though!

ghost:
We already have a trait for this, trait points for 2 language points .-. This pr is redundant

Zonespace27:
Resoundingly "eh" about this PR, even with the high cost i question why one person needs to know so many languages

RimiNosha:
Yup, closing then- It was an interesting idea in my head, but this does indeed seem pretty useless and unwanted!

PR #14635, opened 2022-06-30
Unit Tests ash rituals

Zonespace27:
## About The Pull Request
This PR unit tests ash rituals to ensure things don't break in future work
Also does some related refactoring, like making rituals use direction defines instead of "north", "east", etc.

Unit tests are good


code/__DEFINES/~skyrat_defines/_HELPERS/directions.dm
```diff
@@ -0,0 +1,11 @@
+/// Simple proc that converts a direction number into human readable text
+/proc/dir_num_to_text(dir, upper_case = FALSE)
```
MrMelbert: a dir to text helper aready exists. it's `dir2text()`.
if you need to capitalized just wrap it in `capitalize()`.

Zonespace27: Figured it existed but couldn't find it, thanks 👍

PR #14642, opened 2022-06-30
Updates /update_body_size() to compensate for a player's rotation.

Christmas5:
## About The Pull Request
Running _/datum/dna/proc/update_body_size()_ on someone who's laying down breaks their sprite **if** their body size changes. This is especially noticable if a changeling lays down and shapeshifts into someone who's smaller or bigger than them.

The reason is that updating someone's size includes moving their sprite a few pixels:
`//We update the translation to make sure our character doesn't go out of the southern bounds of the tile`
`var/translate = ((change_multiplier-1) * 32)/2`
`holder.transform = holder.transform.Scale(change_multiplier)`
`holder.transform = holder.transform.Translate(0, translate)`

The code here does not compensate for when the player is laying down or hanging upside down, in which case it will change the translation in the wrong direction, causing the player's sprite to go semi-permanently off-center until their transform matrix gets corrected manually, presumably by an admin.

I added some code and changed _Translate(0, translate)_ to _Translate(translate_x, translate_y)._

## Changelog

**_/update_body_size()_ now compensates for a player's rotation.**
It now splits _translate_ into _translate_x_ and _translate_y_ with appropriate values based on _transform.b, transform.e_ and _features["body_size"]._

:cl:
fix: Updates /update_body_size() to compensate for a player's rotation.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/datums/dna.dm
+++ b/modular_skyrat/modules/customization/datums/dna.dm
@@ -180,7 +180,20 @@ GLOBAL_LIST_EMPTY(total_uf_len_by_block)
 	//We update the translation to make sure our character doesn't go out of the southern bounds of the tile
 	var/translate = ((change_multiplier-1) * 32)/2
 	holder.transform = holder.transform.Scale(change_multiplier)
-	holder.transform = holder.transform.Translate(0, translate)
+	// Check user's rotation before updating their translation, f.e. whether they're laying down or hanging upside down
+	var/translate_x = 0
+	var/translate_y = 0
+	if(holder.transform.a > 0)
+		translate_y = translate
+	else if(holder.transform.d > 0)
+		translate_x = -translate
+	else if(holder.transform.b > 0)
+		translate_x = translate
+	else if(holder.transform.a < 0)
+		translate_y = -translate
+	else
+		return
+	holder.transform = holder.transform.Translate(translate_x, translate_y)
 	holder.maptext_height = 32 * features["body_size"] // Adjust runechat height
 	current_body_size = features["body_size"]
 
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/customization/datums/dna.dm
```diff
@@ -180,7 +180,20 @@ GLOBAL_LIST_EMPTY(total_uf_len_by_block)
...
+	if(holder.transform.a > 0)
+		translate_y = translate
+	else if(holder.transform.d > 0)
+		translate_x = -translate
+	else if(holder.transform.b > 0)
+		translate_x = translate
+	else if(holder.transform.a < 0)
+		translate_y = -translate
+	else
+		return
```
GoldenAlpharex: What happens if I'm tilted by 45 degrees?

Christmas5: I have not checked how the transform matrix changes when you're tilted anything other than 90, 180 or 270 degrees, so I genuinely don't know.

My guess is that one of the if-statements would still catch it. This **would** give unwanted results and probably break your sprite, but I don't think it'd be any worse than what you get without this PR.

A better way to solve it would be for transform_x and _y to be functions based on transform.a, b, d and e (so that they work for any rotation). 
I'll try to do that.

Christmas5: Made functions based on transform.b and transform.e. Testing it right now and it appears to work fine with any angle, including 45 degrees.

Christmas5:
@GoldenAlpharex PR going stale. It should be ready for merging.

GoldenAlpharex (approved):
Looks good to me!

PR #14667, opened 2022-07-01
Bureaucratic error adjustment

tf-4:
## About The Pull Request

allows bureaucratic error to open 0-3 job slots instead of 1-4

image

## Changelog
:cl:
balance: Bureaucratic error is less likely to open 70 command job slots
/:cl:

### DIFF
--- a/code/modules/events/bureaucratic_error.dm
+++ b/code/modules/events/bureaucratic_error.dm
@@ -8,7 +8,8 @@
 	announceWhen = 1
 
 /datum/round_event/bureaucratic_error/announce(fake)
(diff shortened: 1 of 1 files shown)


Ghommie:
You guys have effectively changed the event to open an average of 2 slots of **every** job, which makes no effing sense with the announcement referencing possible staff shortages. It comes to no surprise everytime it runs roughly 80 (previously 100) slots are added across the board. It's silly.
Take this as a mere suggestion, but perhaps the negative variation should be reintroduced, while leaving at least 1 or 2 open slots (unless the job had fewer to start with) as a precaution, so that jobs won't end up getting locked off by it again.

tf-4:
stopped caring

PR #14679, opened 2022-07-01
Automapper - IMPORTANT, READ ME [MDB IGNORE]

Gandalf2k15:
# IMPORTANT: MAP CONTRIBUTION GUIDELINES HAVE BEEN UPDATED
When you are adding a new item to the map you **MUST** follow this procedure:
Start by deciding how big of a change it is going to be, if it is a small 1 item change, you should use the simple area automapper. If it is an entire room, you should use the template automapper.

See each subsection on how to add these things.

We will no longer have _skyrat map versions.

**DO NOT CHANGE TG MAPS, THEY ARE HELD TO THE SAME STANDARD AS ICONS. USE THE ABOVE TO MAKE MAP EDITS.**

## About The Pull Request

THANK THE LORD WE FINALLY HAVE AUTOMAPPING!!!!!!!

So, this is one of the most valuable tools(to me) that I have created to date. It should enable us to stop worrying about map resets altogether.

## The Automapper
This is basically a two-part autoloading and intelligent placement system.

It has two different loading systems. 

### The simple area automapper 
This system uses datums to load single items such as vendors and pets onto select maps using an "intelligent" algorithm to decide where to place them. This will hopefully result in a rather (in)consistent placement of objects on the map, requiring no manual placements.

Example heatmap:
dreamseeker_tGBTCVqn4P

Instructions on how to add your own auto spawn entry:
1. Make the item
2. Make a new area spawn entry
3. Decide where the object should spawn
4. Profit
Example entry:
image

Wall hug? This is used to determine if your object requires open space on all sides unless it is hugging a wall, which it also needs to do in order to spawn. Great for vending machines.

### The template automapper
This is the more advanced version of a template loader, it uses prebaked templates that can be mapped and then added into a config file. The automapper will then load these based on coordinates and other settings, essentially overriding a map. This should make life so so much easier, and hopefully, remove the need for modular maps.

Instructions on how to add an automapper template:
1. Figure out where it must go, and get the right bottom coordinates of the start position.
2. Figure out how big the edit is, and make a new map to that exact size.
3. Make your map override.
4. Save the map template in the templates folder.
5. Add a new entry into the config TOML, including those coordinates.
Examples:
image
You can see the template is exactly the size it needs to be, don't make it any bigger than it needs to be!
You should make your pipes integrate with the existing pipes that go into your template area to minimise the chance of failures down the line.
image

## Changelog

:cl:
add: A new automapper subsystem, please report any anomalies to: https://github.com/Skyrat-SS13/Skyrat-tg/pull/14679
/:cl:


Tastyfish:
Why not upstream? seems useful

Gandalf2k15:
> Why not upstream? seems useful

Probably not since they manually place items on the map, where as we can't

Tastyfish:
> > Why not upstream? seems useful
> 
> Probably not since they manually place items on the map, where as we can't

Oh, you mean from a map maintenance perspective. That makes sense.

modular_skyrat/modules/area_spawn/code/area_spawn_entries.dm
```diff
@@ -0,0 +1,106 @@
...
+	SHOULD_CALL_PARENT(TRUE)
+
+	if(SSmapping.config.map_name in blacklisted_stations)
+		return
+
+	var/amount_in_world = 0
+	for(desired_atom in world)
+		amount_in_world++
+
+	if(amount_in_world > max_amount)
```
Gandalf2k15:
```suggestion
	if(amount_in_world >= max_amount)
```
Add wall hug

Gandalf2k15:
Make area list stop if first area has populated available turfs

_maps/_basemap.dm
```diff
@@ -15,15 +15,9 @@
...
-		#include "map_files\KiloStation\KiloStation_skyrat.dmm"
-		#include "map_files\MetaStation\MetaStation_skyrat.dmm"
-		#include "map_files\IceBoxStation\IceBoxStation_skyrat.dmm"
-		#include "map_files\tramstation\tramstation_skyrat.dmm"
 		#include "map_files\Blueshift\BlueShift_upper.dmm"
 		#include "map_files\Blueshift\BlueShift_middle.dmm"
 		#include "map_files\Blueshift\BlueShift_lower.dmm"
-		#include "map_files\tramstation\tramstation_skyrat.dmm"
 		// The mining maps
 		#include "map_files\Mining\Icemoon.dmm"
```
GoldenAlpharex, line 22: We really need to remember to remove those at some point.

code/modules/mapping/reader.dm
```diff
@@ -133,7 +133,8 @@
 	parsed_bounds = bounds
 
 /// Load the parsed map into the world. See [/proc/load_map] for arguments.
-/datum/parsed_map/proc/load(x_offset, y_offset, z_offset, cropMap, no_changeturf, x_lower, x_upper, y_lower, y_upper, placeOnTop)
+/datum/parsed_map/proc/load(x_offset, y_offset, z_offset, cropMap, no_changeturf, x_lower, x_upper, y_lower, y_upper, placeOnTop, list/blacklisted_turfs) // SKYRAT EDIT CHANGE
```
GoldenAlpharex: Could be a good idea to add what the original proc was, so we know what was changed.

modular_skyrat/modules/automapper/code/area_spawn_entries.dm
```diff
@@ -0,0 +1,159 @@
...
+	/// The amount we want to spawn
+	var/amount_to_spawn = 1
+	/// The max amount that the world can have
+	var/max_amount = 1
+	/// Do we need to be adjacent to a wall? This also checks the other 3 cardinals for no density. Generally useful for objects with density and anchored.
+	var/wall_hug = FALSE
+	/// Map blacklist, this is used to determine what maps we should not spawn on.
+	var/list/blacklisted_stations = list("Blueshift", "Runtime Station", "MultiZ Debug")
+
+/datum/area_spawn/proc/try_spawn()
```
GoldenAlpharex, line 36: Documentation for this proc would be nice

modular_skyrat/modules/automapper/code/area_spawn_entries.dm
```diff
@@ -0,0 +1,159 @@
...
+	return ..()
+
+/**
+ * Area spawn datums
+ *
+ * Use these to spawn atoms in areas instead of placing them on a map. It will select any available open and entering turf.
+ */
+/datum/area_spawn
+	/// The target area for us to spawn the desired atom
+	var/list/target_areas
```
GoldenAlpharex, line 21: It would be interesting to add to that documentation that they're in order of priority for the selection of the spawning location (if that's still the case, that is).

modular_skyrat/modules/automapper/code/area_spawn_entries.dm
```diff
@@ -0,0 +1,159 @@
...
+	wall_hug = TRUE
+
+/datum/area_spawn/lustwish_dorms
+	target_areas = list(/area/station/commons/locker, /area/station/commons/dorms)
+	desired_atom = /obj/machinery/vending/dorms
+	wall_hug = TRUE
+
+/datum/area_spawn/lustwish_prison
+	target_areas = list(/area/station/security/prison, /area/station/security/prison/shower)
+	desired_atom = /obj/machinery/vending/dorms
```
GoldenAlpharex, line 167: Didn't we have a special version of it that was unique to the prison, which didn't have tools in it?

modular_skyrat/modules/automapper/code/automap_template.dm
```diff
@@ -0,0 +1,27 @@
+/datum/map_template/automap_template
+	name = "Automap Template"
+	/// Our load turf
+	var/turf/load_turf
+	/// The map for which we load on
+	var/override_map_name
+
+/datum/map_template/automap_template/proc/preload(map_file, incoming_override_map_name, incoming_load_turf, template_name)
```
GoldenAlpharex, line 11: Would be nice to have documentation for this.

modular_skyrat/modules/automapper/code/turf_proc.dm
```diff
@@ -0,0 +1,17 @@
+/**
+ * Clear everything
```
GoldenAlpharex: Would be nice to give a bit more details on how each parameters work here.

modular_skyrat/modules/automapper/code/turf_proc.dm
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * A more useful version of empty.
+ */
+/turf/proc/clear_everything(turf_type = /turf/open/space, baseturf_type, list/ignore_typecache, flags, list/ignored_atoms = list(/mob/dead))
```
GoldenAlpharex: Are we actually allowed to use parameters that have default values before parameters that don't have default values? I feel like doing so will force whoever uses this proc to type like `clear_everything(baseturf_type = /what/ever, ignore_typecache = list(/turf/my_turf))` instead of just typing in the values without specifying the arguments.

modular_skyrat/modules/automapper/readme.md
```diff
@@ -0,0 +1,12 @@
+## Credits 
+Code: [@Gandalf2k15]
+
+## About
+This automapper is designed to make mapping far easier for modular maps, it should mean less manual map resets(ideally, never).
+It supports two types of automapping, simple area automapper and template automapper.
+
+## Simple Area Automapper
+What this will do is check an area for a suitable single tile to spawn an object in. It uses "intelligent" heatmapping to choose a suitable place.
```
GoldenAlpharex, line 9: It would be nice for you to define what "intelligent" means here, just so it's easier to understand without looking too deep into the code in the future.

GoldenAlpharex (changes requested):
This is strictly from a code perspective, I have no clue what the maps themselves look like from this.

Also, I think this doesn't cover the NT Consultant, for whom @Jolly-66 recently made some offices too.

_maps/_basemap.dm
Gandalf2k15: Yeee

modular_skyrat/modules/automapper/code/area_spawn_entries.dm
Gandalf2k15: cant find one

modular_skyrat/modules/automapper/code/turf_proc.dm
Gandalf2k15: this isn't actually used

Gandalf2k15:
> This is strictly from a code perspective, I have no clue what the maps themselves look like from this.
> 
> Also, I think this doesn't cover the NT Consultant, for whom @Jolly-66 recently made some offices too.

It just makes the templates actual maps

dawsonkeyes:
what if we add a minuscule non-zero chance for it to try and place an instakill rifle somewhere
i think that’s a great idea

YakumoChen:
I feel like whatever algorithm needs work, it often places stuff like vending machines in very inconvenient locations

good example: 
image

what is the effect of this on server load times?

YakumoChen:
image
The smoking lounge here lacks a ShadyCigs (I suggest you also add the bar smoke scrubber as well). I suggest switching the nearby vendor with the Cola machine

image
OSHA hazard, no guardrail

image
The room above is not map-edited correctly

YakumoChen:
**You NEED to add documentation and guidelines so that people know this exists before they try to make edits to the maps. Update the main README.**

Zonespace27:
> what is the effect of this on server load times?

About 15s on average

YakumoChen:
> > what is the effect of this on server load times?
> 
> About 15s on average

sounds kinda bad

Zonespace27:
> > > what is the effect of this on server load times?
> > 
> > 
> > About 15s on average
> 
> sounds kinda bad

Considering our lobby time is 360 seconds, not really (also it's a price i am very much willing to pay)

Tastyfish:
Perhaps wall placements should favor concave corners? That might help with the whole being in the middle of hallway issue

PR #14692, opened 2022-07-02
[Non-Modular] Accessibility For The Blind: Volume 1 (Allows Them Analyzers/White Canes)

Nerev4r:
## About The Pull Request
This PR allows blind people the use of gas, health, and slime scanners. It also puts the white canes in the loadout.

This is the 'how the fuck' section.
Alright, so.
There's two easy surface-level possible methods in which these devices could easily work.
Method one simply being https://youtu.be/pQO4S1t2uiA, well, screen readers. This is a technology we have in 2022 our current year that can easily read text back to blind people at EXTREMELY quick speeds, this video's one of the slowest screen readers I've seen. With the addition of alt text (the text you get when you mouse over images on accessibility-friendly website) it'd be rather easy to navigate the UIs we're presented.

Method two would simply be refreshable braille displays. 
image
These traditionally use a system of round-tipped pins raised through holes in a flat surface, though functionality is being designed for https://www.youtube.com/watch?v=0fIg4rI4cDw microfluidic displays that can do not only full pages, but can generate new text on the fly. See: https://en.wikipedia.org/wiki/Braille_e-book
I don't think these would be very hard at all to make in the year 2562, especially given we have nanomanipulators and such. We, also, have several technologies in the game which use either hardlight or nanomachines, and these devices can be paired with speech synthesizers as well.

Before you ask about blind surgeons, https://www.9news.com/article/news/blind-student-earns-md-things-are-only-impossible-until-theyre-done/73-344742218 technology could be developed for that and _has been already._

Alright, here's the 'why.'
I don't think any of this stuff is necessary. This was nearly removed on /tg/ as well in PR 58134. Being blind is a horrific downside no matter how you slice it, given that you can only see a very small amount of tiles and the little FOV cues outside of that radius are cues that do not give you visual information about what's going on or who's going on. It's, also, wildly inconsistent that blind people can use _every other device including PDAs,_ but somehow analyzers (a basic part of several jobs) are not devices with accessibility. For some reason. C'mon, even 2022 smartphones have screen readers, and they're a hell of a lot less complex.

## Changelog
:cl:
add: Blind people can use analyzers again.
add: White canes have been added to the loadout.
/:cl:

### DIFF
--- a/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_inhands.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_inhands.dm
@@ -20,6 +20,10 @@ GLOBAL_LIST_INIT(loadout_inhand_items, generate_loadout_items(/datum/loadout_ite
 	name = "Cane"
 	item_path = /obj/item/cane
 
+/datum/loadout_item/inhand/cane/white
+	name = "White Cane"
+	item_path = /obj/item/cane/white
+
 /datum/loadout_item/inhand/briefcase
 	name = "Briefcase"
 	item_path = /obj/item/storage/briefcase
--- a/code/game/objects/items/devices/scanners/gas_analyzer.dm
+++ b/code/game/objects/items/devices/scanners/gas_analyzer.dm
@@ -114,13 +114,13 @@
 	return list("gasmixes" = last_gasmix_data)
 
 /obj/item/analyzer/attack_self(mob/user, modifiers)
-	if(user.stat != CONSCIOUS || !user.can_read(src) || user.is_blind())
+	if(user.stat != CONSCIOUS || !user.can_read(src)) //SKYRAT EDIT: Blind People Can Analyze Again
 		return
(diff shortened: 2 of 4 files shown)


modular_skyrat/modules/loadouts/loadout_items/loadout_datum_inhands.dm
```diff
@@ -20,6 +20,10 @@ GLOBAL_LIST_INIT(loadout_inhand_items, generate_loadout_items(/datum/loadout_ite
 	name = "Cane"
 	item_path = /obj/item/cane
 
+/datum/loadout_item/inhand/cane/white
+	name = "White Cane"
+	item_path = /obj/item/cane/white
```
GoldenAlpharex, line 25: Does the white cane actually do anything?

Nerev4r: Yes, it allows them to examine stuff they aren't holding and examine stuff faster.

GoldenAlpharex (approved):
I'm alright with this.

PR #14728, opened 2022-07-05
[MODULAR] Better Interdyne!

RimiNosha:
## About The Pull Request
So, on rounds when the map is Blueshift or Icebox, the base gets mind-numbingly boring- so this is to help in fixing that!

I also plan on tweaking DS2... cause it's got a lot of glaring flaws, but that's a different PR! (I'm aware of the potential DS3 mapper.)

I'm also super aware of the automapper, but it doesn't affect ruins, last time it was asked, so I'm still making this PR!

### General
- Every door now uses access helpers!
- Added a Library and Art Room!
- Moved the modsuit storage to the vault, <s>gamers</s> miners have a tendency to break into the engineering area to steal it for no good reason, and they shouldn't know it's there!
- Kitchen now has a desk to botany, and has shutters to the main dining area.
- Kitchen now has a dinnerware vendor, but lost an admittedly useless fridge in exchange!
- Kitchen also has a food grinder now!
- Tweaked the locations of some vents, don't remember exactly which ones though.
- Added an AutoDrobe, syndies can be maids too, you know!
- All machine frames start bolted to the floor!
- A functioning, if basic, disposals loop!
- Added an ammo workbench lethal disk, in case the base needs to make it's own ammo, for whatever reason. It's behind a safe, so it'll need to be worked for.
- Shifted a bit of land away from the base wall, cause a 1-block gap is enough for spearing.
- Put syndicate items inside of locked crates instead of open ones, no free Interdyne keys and chameleon IDs for you!
- Deck officer gets a Makarov +1 mag, cause everyone else does!
- Added a morgue!
- Added a public toilet!
- Added a barber shop!
- Changed two cells into a BZ room, along with a BZ canister for xenobio. This can isn't able to be used for toxins in normal gameplay, so I think this is safe!
- Shrank the kill room for slimes, and made it more accessible.
- Added extra turrets- there's still blind spots, but they are nowhere near as egregious as previously!
- Fixed the lockers all having wrong accesses!
- Added lights in places it really needed it.
- The SMES units now output 200k by default.

### IceBox Only
- Icebox specific pods are now in their own room, they shouldn't be in places with machines!
- Icebox has an extra syndiedrobe resupply, cause that's a lot of syndies in one place.
- ATV keys have been moved, one for each miner room, and one in the deck officer's office, to make the keys more obvious, cause the dark floor hides them.

And I think that's all my changes... let me know if I missed anything!

A better place to RP in helps a lot!

## Changelog

:cl:
qol: Added and changed a bunch of things for Interdyne to make life much more pleasant.
/:cl:

### DIFF
--- a/_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
+++ b/_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
@@ -9,24 +9,22 @@
 /turf/closed/mineral/random/snow/underground,
 /area/icemoon/underground/explored)
 "ad" = (
-/obj/machinery/door/airlock/external{
-	req_access = list("syndicate")
-	},
+/obj/machinery/door/airlock/external,
 /obj/effect/mapping_helpers/airlock/cyclelink_helper,
 /obj/structure/fans/tiny,
+/obj/effect/mapping_helpers/airlock/access/all/syndicate/general,
 /turf/open/floor/plating,
 /area/ruin/syndicate_lava_base/cargo)
 "ag" = (
 /turf/open/floor/plating,
 /area/ruin/syndicate_lava_base/cargo)
 "ai" = (
-/obj/machinery/door/airlock/external{
-	req_access = list("syndicate")
-	},
+/obj/machinery/door/airlock/external,
 /obj/effect/mapping_helpers/airlock/cyclelink_helper{
 	dir = 1
 	},
 /obj/structure/fans/tiny,
+/obj/effect/mapping_helpers/airlock/access/all/syndicate/general,
 /turf/open/floor/plating,
 /area/ruin/syndicate_lava_base/cargo)
 "ak" = (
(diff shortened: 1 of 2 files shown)


_maps/RandomRuins/LavaRuins/skyrat/lavaland_surface_syndicate_base1_skyrat.dmm
```diff
@@ -2397,6 +2837,30 @@
...
+"CY" = (
+/obj/effect/turf_decal/siding/thinplating/dark/end{
+	dir = 1
+	},
+/obj/structure/table/reinforced,
+/obj/machinery/door/firedoor,
+/obj/machinery/door/poddoor/shutters{
+	dir = 8;
+	state_open = 1;
+	id = s
```
tf-4, line 2860: 😔

_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
```diff
@@ -425,8 +435,24 @@
...
+"cW" = (
+/obj/effect/turf_decal/siding/thinplating/dark/end{
+	dir = 1
+	},
+/obj/structure/table/reinforced,
+/obj/machinery/door/firedoor,
+/obj/machinery/door/poddoor/shutters{
+	dir = 8;
+	state_open = 1;
+	id = s
```
tf-4, line 452: 🤔

Gandalf2k15:
there is currently a map freeze

RimiNosha:
Ah, I'll reopen when it passes, then-

PR #14751, opened 2022-07-06
Fixes tile sprayer sec colour

tf-4:
## About The Pull Request

Been playing some janitor recently and this has annoyed me. The security colour is not red, and having to input it as a custom colour constantly is annoying.

... Pretty sure I'm the only janitor that actually bothers to spray tiles anyway, but it's nice to be consistent.

## Changelog
:cl:
fix: The tile sprayer now has the correct colour for the security decal preset.
/:cl:


code/game/objects/items/airlock_painter.dm
```diff
@@ -384,7 +384,7 @@
 		list("White", "#D4D4D4"),
 		list("Black", "#0e0f0f"),
 		list("Bar Burgundy", "#791500"),
-		list("Sec Red", "#DE3A3A"),
+		list("Sec Blue", "#486091"), // SKYRAT EDIT CHANGE
```
ghost, line 387: Might be worth leaving a Red in for non-sec red decals, just a generic "Red"

tf-4: You can still set a custom colour of anything you choose, so I think it's probably fine to just replace it straight up

PR #14755, opened 2022-07-06
Prevent Synthetic's from removing their organs to be immune to EMP effects with no drawbacks

nevimer:
## About The Pull Request

Right, so the species have a ton of flags that means they don't really need internal organs to begin with - they only feel the effects when they're dead organs or EMP'd organs - and it's only stunlocking. This means you can gut yourself with  no drawbacks, so why wouldn't you? Plus, this prevents you from giving your OP lungs to organics to circumvent their OWN need to breathe. 

Also makes it so revive for putting a brain into and IPC is 50 or less damage, instead of 49 or less. 
dreamseeker_6LljHiLOwa

Fixes a possible abuse case. 

## Changelog

:cl:
fix: Fixes Synthetics being table to remove their organs for all gain. You can still remove head organs and the brain. 
:/cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/surgery/organs/ipc.dm
+++ b/modular_skyrat/modules/customization/modules/surgery/organs/ipc.dm
@@ -17,7 +17,7 @@
 	if(user.stat == DEAD && ishuman(user))
 		var/mob/living/carbon/human/user_human = user
 		if(user_human?.dna?.species && (REVIVES_BY_HEALING in user_human.dna.species.species_traits))
-			if(user_human.health > 50)
+			if(user_human.health >= 50)
 				user_human.revive(FALSE)
 
 /obj/item/organ/internal/brain/ipc_positron/emp_act(severity)
@@ -38,7 +38,7 @@
 	slot = "stomach"
 	desc = "A specialised cell, for IPC use only. Do not swallow."
 	status = ORGAN_ROBOTIC
-	organ_flags = ORGAN_SYNTHETIC
+	organ_flags = ORGAN_SYNTHETIC | ORGAN_UNREMOVABLE
 
 /obj/item/organ/internal/stomach/robot_ipc/emp_act(severity)
 	. = ..()
@@ -61,7 +61,7 @@
 	slot = ORGAN_SLOT_EARS
 	gender = PLURAL
 	status = ORGAN_ROBOTIC
-	organ_flags = ORGAN_SYNTHETIC
+	organ_flags = ORGAN_SYNTHETIC | ORGAN_UNREMOVABLE
 
 /obj/item/organ/internal/ears/robot_ipc/emp_act(severity)
 	. = ..()
@@ -128,7 +128,7 @@
 	cold_level_2_damage = 0
 	cold_level_3_damage = 0
 	status = ORGAN_ROBOTIC
-	organ_flags = ORGAN_SYNTHETIC
+	organ_flags = ORGAN_SYNTHETIC | ORGAN_UNREMOVABLE
 
 /obj/item/organ/internal/lungs/robot_ipc/emp_act(severity)
 	. = ..()
@@ -144,7 +144,7 @@
 /obj/item/organ/internal/heart/robot_ipc
 	name = "hydraulic pump engine"
 	desc = "An electronic device that handles the hydraulic pumps, powering one's robotic limbs."
-	organ_flags = ORGAN_SYNTHETIC
+	organ_flags = ORGAN_SYNTHETIC | ORGAN_UNREMOVABLE
 	status = ORGAN_ROBOTIC
 	icon = 'modular_skyrat/master_files/icons/obj/surgery.dmi'
 	icon_state = "heart-ipc"
@@ -152,7 +152,7 @@
 /obj/item/organ/internal/liver/robot_ipc
 	name = "reagent processing unit"
 	desc = "An electronic device that processes the beneficial chemicals for the synthetic user."
-	organ_flags = ORGAN_SYNTHETIC
+	organ_flags = ORGAN_SYNTHETIC | ORGAN_UNREMOVABLE
 	status = ORGAN_ROBOTIC
 	icon = 'modular_skyrat/master_files/icons/obj/surgery.dmi'
 	icon_state = "liver-c"


modular_skyrat/modules/customization/modules/surgery/organs/ipc.dm
```diff
@@ -17,7 +17,7 @@
 	if(user.stat == DEAD && ishuman(user))
 		var/mob/living/carbon/human/user_human = user
 		if(user_human?.dna?.species && (REVIVES_BY_HEALING in user_human.dna.species.species_traits))
-			if(user_human.health > 50)
+			if(user_human.health >= 50)
```
Zonespace27, line 20: unmentioned change

nevimer: mentioned now!

tf-4:
Your changelog is broken

nevimer:
> Your changelog is broken

based thanks

Stonetear:
I really, really dislike this since this means you can no longer do any kind of cool surgeries with exotic organs on synths.  Synths already cannot be upgraded with the advanced surgeries, can't get beneficial diseases, and can't use healing chems or healing gasses.
This really removes the only remaining medical content for synths to really do.

If we're going to do this, why not make it so that Plasmamen can't remove _their_ organs (since their organs are a EVEN MORE of a direct upgrade!)  Why not make it so Etherials can't have their organs removed (unfair advantage, of course), too.  This is like the edge case of people picking nerve stapled and having robotics hack their head off to remove it.

YakumoChen:
The actual solution is to tie the species traits to the organs themselves and not give a bunch of strong buffs to the species for existing.

Then make the organs incompatible with non-synth species (make it check for blood type or just species, whichever is saner)

This also future proof existing problems with synths trading their organs in at character selection and realising that it fucks them over.

As far as the other arguments go, it would make sense to tie, say, plasmaman organ functionality to the plasma breathing requirement, or make ethereal organs require a working ethereal heart, but that sounds like an upstream problem.

nevimer:
> The actual solution is to tie the species traits to the organs themselves and not give a bunch of strong buffs to the species for existing.
> 
> Then make the organs incompatible with non-synth species (make it check for blood type or just species, whichever is saner)
> 
> This also future proof existing problems with synths trading their organs in at character selection and realising that it fucks them over.
> 
> As far as the other arguments go, it would make sense to tie, say, plasmaman organ functionality to the plasma breathing requirement, or make ethereal organs require a working ethereal heart, but that sounds like an upstream problem.

The organs are null organs. They don't do anything, at all. It's the species flags that make them immune to everything, so  when the parent organ code runs, it just hits return. The only beneficial organ they have to transplant is the lung, and it doesn't need to exist because despite it being 0'd out, they have nobreathe anyways as a species trait. 

> I really, really dislike this since this means you can no longer do any kind of cool surgeries with exotic organs on synths. Synths already cannot be upgraded with the advanced surgeries, can't get beneficial diseases, and can't use healing chems or healing gasses. This really removes the only remaining medical content for synths to really do.
> 
> If we're going to do this, why not make it so that Plasmamen can't remove _their_ organs (since their organs are a EVEN MORE of a direct upgrade!) Why not make it so Etherials can't have their organs removed (unfair advantage, of course), too. This is like the edge case of people picking nerve stapled and having robotics hack their head off to remove it.

This is part one in a series of pull requests. You have no reason to remove them as it breaks the game. You need to understand right now these organs have no code. They all just hit early returns (species traits), and are otherwise normal-ass basic cybernetics.

nevimer:
Even if you DID put new organs in, the code for them would never run, at all, ever, hence breaking the game and having no reason to do it.

YakumoChen:
> The organs are null organs. They don't do anything, at all. It's the species flags that make them immune to everything, so when the parent organ code runs, it just hits return. The only beneficial organ they have to transplant is the lung, and it doesn't need to exist because despite it being 0'd out, they have nobreathe anyways as a species trait.

Yes, move some species flags to the organs, make the organs require being a synth in some way to function correctly. Make the organs require being present for the synth to function so that they can't just be removed for no reason. Making the organs unremovable opens up a different set of problems that you won't have foreseen. 

At present I can already tell this will affect abductors, who can remove synth hearts and insert their own manually for their experiments.

nevimer:
> > The organs are null organs. They don't do anything, at all. It's the species flags that make them immune to everything, so when the parent organ code runs, it just hits return. The only beneficial organ they have to transplant is the lung, and it doesn't need to exist because despite it being 0'd out, they have nobreathe anyways as a species trait.
> 
> Yes, move some species flags to the organs, make the organs require being a synth in some way to function correctly. Make the organs require being present for the synth to function so that they can't just be removed for no reason. Making the organs unremovable opens up a different set of problems that you won't have foreseen.
> 
> At present I can already tell this will affect abductors, who can remove synth hearts and insert their own manually for their experiments.

The Abductor surgery has failsafes for not finding an applicable organ.

YakumoChen:
You can't insert an abductor heart if there is already a heart in the chest.

It is also not "the abductor surgery". Synth abduction must make use of Manipulate Organs to properly probe a synth.

As before, making organs unremovable will only lead to further problems down the line. It's better to just make synths more dependent on their organs, and to code means to prevent other races from using them. This also makes organ failure more meaningful since it could actually kill the synth.

nevimer:
> You can't insert an abductor heart if there is already a heart in the chest.
> 
> It is also not "the abductor surgery". Synth abduction must make use of Manipulate Organs to properly probe a synth.
> 
> As before, making organs unremovable will only lead to further problems down the line. It's better to just make synths more dependent on their organs, and to code means to prevent other races from using them. This also makes organ failure more meaningful since it could actually kill the synth.

Again, as stated before, this is one PR in a series of PR's. Currently removing any organs breaks the synth completely.

YakumoChen:
What do you plan further in the "series"?

Even knowing that I'm against this change. A temporary fix is often the most permanent solution. This breaks more than it fixes and therefore is a bad solution.

Players who are knowingly exploiting things by having their organs removed open themselves up to bwoinks so I don't believe in this change until better workarounds (such as my own suggestions) are properly implemented.

nevimer:
In their current state, a synthetic has everything to gain by being gutted. This fixes that, and that is the scope of the PR. When the time comes, I will reintroduce the ability to remove their organs. But currently? This is not the time.

nevimer:
> What do you plan further in the "series"?
> 
> Even knowing that I'm against this change. A temporary fix is often the most permanent solution. This breaks more than it fixes and therefore is a bad solution.
> 
> Players who are knowingly exploiting things by having their organs removed open themselves up to bwoinks so I don't believe in this change until better workarounds (such as my own suggestions) are properly implemented.


Synth:

https://user-images.githubusercontent.com/77420409/179124946-b135bd98-44ab-4811-815a-152d2ca8907f.mp4



Human:

https://user-images.githubusercontent.com/77420409/179124898-c191dea0-ce05-48bd-b266-fd506235260e.mp4

YakumoChen:
> In their current state, a synthetic has everything to gain by being gutted. This fixes that, and that is the scope of the PR. When the time comes, I will reintroduce the ability to remove their organs. But currently? This is not the time.

when is this ever relevant in-game? you can't disembowel synths anyway, they're immune to wounding. The only way to remove their organs is to do so manually.

This PR is simply fixing a non-issue by creating a different problem, instead of actually fixing the issue.

PR #14765, opened 2022-07-07
[MODULAR] Better Interdyne!

RimiNosha:
## About The Pull Request
So, on rounds when the map is Blueshift or Icebox, the base gets mind-numbingly boring- so this is to help in fixing that!

I also plan on tweaking DS2... cause it's got a lot of glaring flaws, but that's a different PR! (I'm aware of the potential DS3 mapper.)

Please do provide feedback, **especially** if you play Interdyne a lot!

### Fix
- Every door now uses access helpers!
- Fixed the lockers all having wrong accesses!
- Added lights in places where it really needed it.
- The SMES units now output 200k by default.

### Balance
- Moved the modsuit storage to the vault, <s>gamers</s> miners have a tendency to break into the engineering area to steal it for no good reason, and they shouldn't know it's there!
- All machine frames start bolted to the floor and with wires!
- Added an ammo workbench lethal disk, in case the base needs to make it's own ammo, for whatever reason. It's behind a safe, so it'll need to be worked for.
- Shifted a bit of land away from the base wall, cause a 1-block gap is enough for spearing.
- Put syndicate items inside of locked crates instead of open ones, no free Interdyne keys and chameleon IDs for you!
- Deck officer gets a Makarov +1 mag, cause everyone else does!
- Added extra turrets- there's still blind spots, but they are nowhere near as egregious as previously!
- Adds a single pickaxe to the cargo bay, just in case there's no deck officer or miners to mine, along with no DS2!

### General
- Added a Library and Art Room!
- Kitchen now has a desk to botany!
- Kitchen now has shutters to the main dining area instead of windows!
- Kitchen now has a dinnerware vendor, but lost an admittedly useless fridge in exchange!
- Kitchen also has a food grinder now!
- Tweaked the locations of some vents, don't remember exactly which ones though.
- Added an AutoDrobe, syndies can be maids too, you know!
- A functioning, if basic, disposals loop!
- Added a morgue!
- Added a public toilet!
- Added a barber shop!
- Changed two cells into a BZ room, along with a BZ canister for xenobio. This can isn't able to be used for toxins in normal gameplay, so I think this is safe!
- Shrank the kill room for slimes, and made it more accessible.

### IceBox Only
- Icebox specific pods are now in their own room, they shouldn't be in places with machines!
- Icebox has an extra syndiedrobe resupply, cause that's a lot of syndies in one place.
- ATV keys have been moved, one for each miner room, and one in the deck officer's office, to make the keys more obvious, cause the dark floor hides them.

And I think that's all my changes... let me know if I missed anything!

A better place to RP in helps a lot, especially when DS2 doesn't spawn!

## Changelog

:cl:
qol: Added and changed a bunch of things for Interdyne to make life much more pleasant.
/:cl:


Nerev4r:
Why does this thing need even more turrets when it's hardly even allowed for anyone to mess w/ them?

RimiNosha:
Cause I've had a few rounds where later-game miners raid the engineering area, if the turrets are deemed overkill, I'm more than happy to cut them back-

Jolly-66 (commented):
Just want to preface with "Did you test the access helpers?". I would like to make sure they're working first before this is put onto live, mostly so we don't have to throw up a PR later to fix them.

Secondly, can you put airlock cyclers that lead into the departments? If you need help with them, look at other stations or ping me on Discord `Jolly#0512`.

<details>
<summary> Pictures </summary>

image

image

image

image

This last one isn't too needed, but the jumble of airlocks would just make it nicer/easier to deal with. If you just want to, slap them on the northerner most set and that'll be fine.

image

</details>

Lastly, I know you didn't do it in this PR, but can you please rearrange the 2x3 dorm rooms? The beds are inaccessible.

<details>
<summary> Sad bed </summary>

image

</details>

RimiNosha:
@Jolly-66 Done!

And yes, the access helpers work just fine!

Jolly-66:
> Rimi

For reference, I was referring to the `multi` helpers, not the one-way. Multi work if any door with the same tag in the area is opened, and will close the other one. This'll also make the adjacent doors close as well.
These, for example.

image

YakumoChen:
post a map

RimiNosha:
> post a map

MapDiffBot exists literally for this reason-

RimiNosha:
@Jolly-66 Done!

GuiltyNeko:
> The beds are inaccessible.

They're only inaccessible when nobody has spawned in yet. When someone spawns in those spawners, they disappear, making the bed quite accessible.

Edit: And you don't need to sleep in someone else's bed. You have your own. Shoo.

YakumoChen:
> > post a map
> 
> MapDiffBot exists literally for this reason-

Needs picture in the pr if you're doing a major mapping change

Jolly-66 (approved):
Time escapes from me sometimes. Sorry this got left behind.

PR #14781, opened 2022-07-07
[Modular] Removes a defaced Poster.

theselfish:
## About The Pull Request

Removes the Dogborg poster.

The Poster references content that is no longer in the base (SecBorgs).

## Changelog

:cl: Cursor
del: Removed the Defaced Valeborg Advertisement.
/:cl:


Jolly-66:
I personally don't see why we should have to remove it. The poster can still be a subtle and/or funny nod to something that once existed, and could drum up an interesting conversation.

theselfish:
> I personally don't see why we should have to remove it. The poster can still be a subtle and/or funny nod to something that once existed, and could drum up an interesting conversation.

It also makes some users uncomfortable, but it was thought that 'The Poster references content that is no longer in the base (SecBorgs).' was a more solid argument.

YakumoChen:
its a good reminder why we have them gone

theselfish:
I should also add that @thestubborn asked @Zonespace27 and they couldn't even tell what the poster _was_.

Stonetear:
I know this isn't a democacy but I vote no.  I think Vale is a cool borg, and this is one of the posters I snip down to hang in robotics when I see it.

theselfish:
> I know this isn't a democacy but I vote no. I think Vale is a cool borg, and this is one of the posters I snip down to hang in robotics when I see it.

You're right. This isn't a democracy.

jjpark-kb:
we can have references to things that are no longer here.

PR #14793, opened 2022-07-08
[MIRROR] Refactors cleaning [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68156
--------------------

## About The Pull Request
A cleaner component has been added so that the cleaning code can be put in one place. This cleaner component has been used to replace a lot of the cleaning code of mops/soap/rags/cleanbots. The cleaner component can handle cleaning while accounting for cleaning skill. Any cases where simply washing the target doesn't suffice can also be handled here. Such as washing bloodied windows for example. The cleaner component also has a callback it can call after successfully cleaning something. Soap makes use of this to decrease the amount of uses it has left. Mops make use of this to place reagents on the cleaned floor and reduce the reagents it has left.

Rags are now more in line with soap, because they both use the cleaner component. This means rags can get increased cleaning speed based on your cleaning skill. You will also be getting cleaning experience when you clean with rags from now on.

Mops and cleanbots still only clean turfs. The cleaner component doesn't need to care about this detail though. They just pass the turf to it as a parameter.

Soap used to not be able to clean the clothing you were wearing. According to a code comment this was because it caused overlay bugs. Cleaning worn items seems to be updating the sprites just fine nowadays though, so this restriction has been removed.

There were some code comments stating that mops were supposed to get less experience than soap. This was supposed to be done through rounding the experience gain for some reason. But the experience gain for soap and mops seemed to be the same. These comments are now gone. There's an experience gain modifier on the cleaner datum now though, in case somebody wants to do something like that in the future. And it will be much more readable that way too.

Sorry for the amount of commits, I made the mistake of not starting from master. Relevant commits are from [added basic cleaner datum, applied it to rags](https://github.com/tgstation/tgstation/pull/68156/commits/a8eb5feeee6312f0a50c693019be465b3b404825) and onwards.

## Why It's Good For The Game
Having similar code doing similar things in four different places makes it harder to expand that functionality. It also causes little inconsistencies such as the rags not giving cleaning experience for example. Having the cleaning code in one place makes it easier to expand and fix.

## Changelog

:cl:
fix: rag cleaning speed is now influenced by your cleaning skill
fix: cleaning with a rag now gives you cleaning skill experience
fix: rags can now be used to wash bloodied windows
fix: soap is no longer restricted from washing worn clothing, as this doesn't seem to be causing issues anymore
/:cl:


GoldenAlpharex (approved):
Should be good now.

PR #14831, opened 2022-07-09
Makes it so Miasma should only infect each human once per round

GoldenAlpharex:
## About The Pull Request
Miasma will now only infect each human at most once per shift. That number can easily be changed through a simple define, `MAXIMUM_TIMES_INFECTED_BY_MIASMA`.

It's as shrimple as that.

I tried making it be handled by the lungs after making those commits, only to realize that it would mean you'd only be infected DIRECTLY by miasma once, and that it wouldn't shield you against being infected from miasma diseases coming from other people, hence why I kept it the way it is.

Miasma will still be very annoying for Medbay, but for individual players that can't do shit about it, no more 20-45 minutes of your round wasted in a "get infected > wait in medbay on the brink of death > get cured" loop. You catch one of them, get it treated, and won't catch another from Miasma itself, or any other virus created by Miasma.

After discussions with the other maintainers, this was decided to be the most elegant and less impacting solution to that problem.

## Changelog

:cl: GoldenAlpharex
balance: Miasma can now only infect each human once per shift at most, so as to avoid getting stuck in a virus-loop.
/:cl:

### DIFF
--- a/code/datums/diseases/advance/presets.dm
+++ b/code/datums/diseases/advance/presets.dm
@@ -20,10 +20,6 @@
 /datum/disease/advance/random
 	name = "Experimental Disease"
 	copy_type = /datum/disease/advance
-	// SKYRAT EDIT ADDITION START - Less annoying miasma
-	/// Does this virus come from miasma? Defaults to FALSE, will fail to infect anyone that has already caught a miasma virus before.
-	var/from_miasma = FALSE
-	// SKYRAT EDIT END
 
 /datum/disease/advance/random/New(max_symptoms, max_level = 8, caused_by_miasma = FALSE) // SKYRAT EDIT - Less annoying miasma - ORIGINAL: /datum/disease/advance/random/New(max_symptoms, max_level = 8)
 	if(!max_symptoms)
--- a/code/datums/diseases/advance/presets.dm
+++ b/code/datums/diseases/advance/presets.dm
@@ -20,8 +20,12 @@
 /datum/disease/advance/random
 	name = "Experimental Disease"
 	copy_type = /datum/disease/advance
+	// SKYRAT EDIT ADDITION START - Less annoying miasma
+	/// Does this virus come from miasma? Defaults to FALSE, will fail to infect anyone that has already caught a miasma virus before.
+	var/from_miasma = FALSE
+	// SKYRAT EDIT END
 
-/datum/disease/advance/random/New(max_symptoms, max_level = 8)
+/datum/disease/advance/random/New(max_symptoms, max_level = 8, caused_by_miasma = FALSE) // SKYRAT EDIT - Less annoying miasma - ORIGINAL: /datum/disease/advance/random/New(max_symptoms, max_level = 8)
 	if(!max_symptoms)
 		max_symptoms = rand(1, VIRUS_SYMPTOM_LIMIT)
 	var/list/datum/symptom/possible_symptoms = list()
@@ -37,6 +41,7 @@
 		if(chosen_symptom)
 			var/datum/symptom/S = new chosen_symptom
 			symptoms += S
+	from_miasma = caused_by_miasma
 	Refresh()
 
 	name = "Sample #[rand(1,10000)]"
--- a/code/datums/diseases/advance/presets.dm
+++ b/code/datums/diseases/advance/presets.dm
@@ -37,7 +37,7 @@
 		if(chosen_symptom)
 			var/datum/symptom/S = new chosen_symptom
 			symptoms += S
-	from_miasma = caused_by_miasma
+	from_miasma = caused_by_miasma // SKYRAT EDIT ADDITION - Less annoying miasma
 	Refresh()
 
 	name = "Sample #[rand(1,10000)]"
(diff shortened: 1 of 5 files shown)


RatFromTheJungle:
can we now justify inoculation against miasma, by forcing people to be exposed to it

GoldenAlpharex:
> can we now justify inoculation against miasma, by forcing people to be exposed to it

There's a moral debate to be had here, I don't think it's moral to force people to contract a disease against their will, even if it's to protect them in the long run.

hugmeorelse:
Ah, this should combat the back-to-back-to-back-to-back-to-back-to-back-to-back disease outbreaks we had to deal with earlier today. Nothing like curing one disease and then it turns out there's three other new diseases that's spawned in the meanwhile and spreading.

Stonetear:
This a pretty huge nerf to miasma monkey disease farms.  If we're having problems with Miasma, perhaps we could make it use a really obvious gas overlay?  Or perhaps make some kind of indicator that you are beathing miasma like this?  I really don't like a hard nerf like this.
misma
(I made this in like 30 seconds)

Iamgoofball:
Is there something causing miasma to occur more often than normal on Skyrat? Miasma shouldn't be nearly this common for this to be an issue, it's supposed to be a fairly rare gas.

Stonetear:
> Is there something causing miasma to occur more often than normal on Skyrat? Miasma shouldn't be nearly this common for this to be an issue, it's supposed to be a fairly rare gas.

We had a round on tramstation where there was a extremely large number of diseases.  Like 3, and medical was swamped.

Iamgoofball:
So, good news and bad news.

Good news, miasma isn't being produced excessively without intentional farming and this specific change to Miasma probably isn't needed. I took a look at regal rat miasma production, the source of the miasma problem in that round according to eyewitness reports, and it's really, really low. They make like, four moles of it, tops. That's basically nothing.
chrome_lNEQh7lYZN

Bad news, the actual problem here was either atmos tech incompetence or the waste loop piping on that map being fucked.

My money is on the atmos techs being incompetent, and here's why:

Floor scrubbers only filter Carbon Dioxide by default. You have to manually set them to scrub miasma. However, with two button presses, you can clear 10,000kpa of Miasma from a room in literally a minute.

https://user-images.githubusercontent.com/4081722/178152022-930b6e2e-8c27-4191-8b65-e43ba4069910.mp4

# tl;dr: this is a certified "atmos tech not knowing the basics of the job" moment, tell atmos to read the wiki for their job

Iamgoofball:
also, if you don't want to configure scrubbers manually, just press the Contaminated button on the air alarm operating mode menu:
WEMIsKVUPe

Iamgoofball:
also if you somehow manage to fill up waste with miasma just pump it into a heater it destroys itself into oxygen at like, 375k

Iamgoofball:
@Stonetear three diseases is, fortunately, not very much, especially if you have a competent viro who can push out high infectivity "good" diseases to kill the weaker rng diseases

however not having a competent viro results in much the same issue as not having a competent atmos tech

Stonetear:
> @Stonetear three diseases is, fortunately, not very much, especially if you have a competent viro who can push out high infectivity "good" diseases to kill the weaker rng diseases
> 
> however not having a competent viro results in much the same issue as not having a competent atmos tech

Yes, but this is discouraged by skyrat policy, AFAIK.

@GoldenAlpharex
Perhaps a better option would be to make scrubbers filter miasma at roundstart on Skyrat.

tf-4:
> So, good news and bad news.
> 
> Good news, miasma isn't being produced excessively without intentional farming and this specific change to Miasma probably isn't needed. I took a look at regal rat miasma production, the source of the miasma problem in that round according to eyewitness reports, and it's really, really low. They make like, four moles of it, tops. That's basically nothing.
> chrome_lNEQh7lYZN
> 
> Bad news, the actual problem here was either atmos tech incompetence or the waste loop piping on that map being fucked.
> 
> My money is on the atmos techs being incompetent, and here's why:
> 
> Floor scrubbers only filter Carbon Dioxide by default. You have to manually set them to scrub miasma. However, with two button presses, you can clear 10,000kpa of Miasma from a room in literally a minute.
> 
> https://user-images.githubusercontent.com/4081722/178152022-930b6e2e-8c27-4191-8b65-e43ba4069910.mp4
> 
> # tl;dr: this is a certified "atmos tech not knowing the basics of the job" moment, tell atmos to read the wiki for their job
> 
> 

That wasn't the issue lol there was disease vines

tf-4:
> This a pretty huge nerf to miasma monkey disease farms.  If we're having problems with Miasma, perhaps we could make it use a really obvious gas overlay?  Or perhaps make some kind of indicator that you are beathing miasma like this?  I really don't like a hard nerf like this.
> misma
> (I made this in like 30 seconds)
> 

Corpses haven't generated miasma in over a year. Unless you mean farming miasma diseases, which... There's literally zero reason to do

Iamgoofball:
>That wasn't the issue lol there was disease vines

Everyone said it was rats lmao

tf-4:
I think it was a bit of both, but neither are that bad on their own. It was the two in conjunction - plus I'm pretty sure fake disease ran more than once.

Iamgoofball:
Ah, yeah, that'd do it.

PR #14840, opened 2022-07-10
 d

ghost:


Jolly-66:
If the PR is actually a draft, please click the "Convert to Draft' button under reviewers.

ghost:
> If the PR is actually a draft, please click the "Convert to Draft' button under reviewers.

Woops.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,36 @@
+/mob/proc/store_mob(mob/tostore)
+	if(!istype(tostore))
+		return
+
+	if(!tostore.on_enter_container(src) || !on_container_entered(tostore))
+		return
+
+	ADD_TRAIT(tostore, TRAIT_OXYIMMUNE, REF(src)) // We give the mob inside oxy-damage immunity. Hope this doesn't come back to bite us in the ass.
```
desertmote: :/ power crep

Ghommie: Can we explain why the stored mob is given immunity to such damage in the comment? Also can it be changed to TRAIT_NOBREATH if it's simply an issue with the lack of breathable air?

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,36 @@
...
+	if(!istype(toremove))
+		return
+
+	if(!(toremove in contents) || toremove.loc !== src)
+		return
+
+	if(!toremove.on_exit_container(src) || !on_container_exited(toremove))
+		return
+
+	REMOVE_TRAIT(toremove, TRAIT_OXYIMMUNE, REF(src)) // Remove trait prevents the loss of roundstart traits, and we also have ourselves as a source. So don't worry.
```
Ghommie: ditto on the no breath trait.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,36 @@
+/mob/proc/store_mob(mob/tostore)
```
tf-4:
```suggestion
/mob/proc/store_mob(mob/to_store)
```
does this need to be on the mob level? I'd rather it were on the stomach.

Gandalf2k15 (changes requested):
no no no no no make this a component oh god

Iamgoofball:
This in and of itself isn't controversial, and would actually be useful upstream.

Zonespace27 (changes requested):
Componentize this and it should very likely be on the stomach, not the mob

Iamgoofball:
Probably shouldn't make it on the stomach, arguably I think the better route here might be refactoring storage to be able to contain all types of atoms instead of just /obj/item.

modular_skyrat/modules/mob_container/mob_container.dm
softcerv: extreme role play mechanics don't need balance.

tf-4: yes they do

Iamgoofball: you should instead just redirect the breath out to the tile the mob is on

Zonespace27: > extreme role play mechanics don't need balance.

Nah mate we don't need shit like the leather whip being a worse police baton

softcerv: > > extreme role play mechanics don't need balance.
> 
> Nah mate we don't need shit like the leather whip being a worse police baton

Doesn't that make it far easier to single out bad actors tho?

Ghommie: For once, I agree. That'd be a good idea.

Iamgoofball: no, they can't be bad actors if the system doesn't let them be bad actors in the first place

ghost:
This won't be going onto the stomach: as this PR has no intent to make it related to the stomach whatsoever. This is simply a way for mobs to contain other mobs. However, it is currently being componentized.

Zenitheevee:
> Probably shouldn't make it on the stomach, arguably I think the better route here might be refactoring storage to be able to contain all types of atoms instead of just /obj/item.

"No, Mr Antag, My duffelbag does not contain 20 armed teshari, now i'm just going to leave it here in the corner of the room and go take a wizz. Do not open it please it contains all my confiscated contraband items."

Iamgoofball:
exactly

YakumoChen:
The big question becomes how expensive is this?

Last time we rejected ""mob engulfing"" because it was insanely unoptimised.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,94 @@
...
+	var/atom/movable/screen/fullscreen_overlay = /atom/movable/screen/fullscreen/impaired
+
+	/// Whether or not mobs should be able to breathe inside of us.
+	var/container_breathable = TRUE
+
+/datum/component/mob_container/Initialize(mob/mob_holder)
+	if(!istype(mob_holder))
+		return COMPONENT_INCOMPATIBLE
+
+	src.mob_holder = WEAKREF(mob_holder)
```
Gandalf2k15: two things, don't use src. and secondly, components get added to something and the parent becomes that thing, don't have the thing it's being added to passed in on initialise

code/modules/mob/living/carbon/life.dm
```diff
@@ -74,6 +74,11 @@
 		if(istype(loc, /obj/))
 			var/obj/location_as_object = loc
 			location_as_object.handle_internal_lifeform(src,0)
+		// SKYRAT EDIT BEGIN - MOB CONTAINER
+		if(istype(loc, /mob/))
```
Zonespace27: `ismob()`

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,94 @@
...
+	for(var/mob/contained in mobs_held)
+		contained.forceMove(get_turf(mob_holder))
+		contained.clear_fullscreen("in_mob_container")
+
+	mob_holder = null
+	mobs_held.Cut()
+
+	UnregisterSignal(mob_holder, list(COMSIG_MOB_CONTAINER_ENTERED, COMSIG_MOB_CONTAINER_EXITED, COMSIG_MOB_ENTER_CONTAINER, COMSIG_MOB_EXIT_CONTAINER))
+
+/datum/component/mob_container/proc/mob_breathe_check()
```
Zonespace27: code doc

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,94 @@
...
+	return container_breathable
+
+/datum/component/mob_container/proc/store_mob_explicitly(mob/to_store)
```
Zonespace27: ditto

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,94 @@
...
+		return
+
+	if(!SEND_SIGNAL(mob_holder, COMSIG_MOB_CONTAINER_ENTERED, to_store) || !SEND_SIGNAL(to_store, COMSIG_MOB_ENTER_CONTAINER, mob_holder))
+		return
+
+	to_store.forceMove(mob_holder)
+	to_store.overlay_fullscreen("in_mob_container", fullscreen_overlay, 1)
+	mobs_held |= to_store
+
+/datum/component/mob_container/proc/remove_mob_explicitly(mob/to_remove)
```
Zonespace27, line 124: ditto

Zonespace27 (changes requested):

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,94 @@
+#define COMSIG_MOB_CONTAINER_ENTERED "mob_container_entered"
```
Zonespace27: don't define signals here

Zonespace27 (changes requested):

softcerv:
> The big question becomes how expensive is this?
> 
> Last time we rejected ""mob engulfing"" because it was insanely unoptimised.

We removed it because we rebased and nobody ended up coding it back in. This doesn't look to be a port of the system from oldbase.

tf-4:
> > The big question becomes how expensive is this?
> > Last time we rejected ""mob engulfing"" because it was insanely unoptimised.
> 
> We removed it because we rebased and nobody ended up coding it back in. This doesn't look to be a port of the system from oldbase.

It not being an oldbase port doesn't inherently make it good code. Not saying it isn't, even, but it's a valid question.

ghost:
> > > The big question becomes how expensive is this?
> > > Last time we rejected ""mob engulfing"" because it was insanely unoptimised.
> > 
> > 
> > We removed it because we rebased and nobody ended up coding it back in. This doesn't look to be a port of the system from oldbase.
> 
> It not being an oldbase port doesn't inherently make it good code. Not saying it isn't, even, but it's a valid question.

This will have no noticeable impact on performance. Essentially, all we're doing is forcibly moving this mob into another mob's contents, and applying a fullscreen overlay. Likely the most performance intensive addition will be a subversion inside `breathe()`, which checks if the user is inside of a mob instead of a turf.

I'll upload a profile after I get the final version.

modular_skyrat/modules/mob_container/mob_container.dm
ghost: These will be moved. Everything is being kept in one place for now since it's a draft.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,177 @@
...
+
+#define COMSIG_CONTAINER_STORE_MOB "container_store_mob"
+#define COMSIG_CONTAINER_REMOVE_MOB "container_remove_mob"
+#define COMSIG_CONTAINER_DUMP_ALL "container_dump_all"
+
+#define COMSIG_MOB_LEAVE_CONTAINER "mob_leave_container"
+
+/datum/component/mob_container
+	/// The mob we're attached to.
+	var/mob/mob_holder
```
Ghommie, line 14: This variable is a tad redundant; The value of the mob the component is attached to is already stored in `parent`. You can define a `mob_holder` variable internal to each proc when necessary. Memory-wise, that'd also be a micro optimization.

modular_skyrat/modules/mob_container/mob_containers.md
```diff
@@ -0,0 +1,43 @@
+## A Notice on Mob Containers
+Mob containers are for the express purpose of storing a mob inside of another mob. This is not for antagonistic nor further mechanical purposes, and should not be used for such things. This is not a solution for having atomic storage (see /datum/storage) inside of mobs: this is only for mob inside mob storage.
+# Usage
+## Adding a Mob to a Mob
+Adding a mob to another mob is easy! Simply send the signal as following:
```
Ghommie, line 5: While it's not explicitly stated on tg coding standards, I was repeatedly told in the past that, in general, a signal is better called in one and only one location in the code, mostly because it's better code structure and easier to search this way, so consider defining procs that contain these SEND_SIGNAL calls and tell people to use them instead.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,177 @@
...
+ * mob/to_store | the mob to remove from us
+ */
+/datum/component/mob_container/proc/remove_mob_explicitly(mob/to_remove)
+	if(!istype(to_remove) && !(to_remove in mob_holder || to_remove.loc == mob_holder))
+		return
+
+	if(!SEND_SIGNAL(mob_holder, COMSIG_MOB_CONTAINER_EXITED, to_remove) || !SEND_SIGNAL(to_remove, COMSIG_MOB_EXIT_CONTAINER, mob_holder))
+		return
+
+	to_remove.forceMove(get_turf(mob_holder))
```
Ghommie:
```suggestion
	to_remove.forceMove(mob_holder.drop_location())
```
The go-to proc when you want to remove a movable from another movable safely and sensefully.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,177 @@
...
+	SIGNAL_HANDLER
+
+	return TRUE
+
+/**
+ * Called whenever we exit another mob. If our return value is FALSE, we won't exit the mob.
+ *
+ * mob/mob_were_exiting | the mob we're being taken out of
+ */
+/datum/component/mob_container/proc/on_exit_container(datum/source, mob/mob_were_exiting)
```
Ghommie, line 150: Any good reason this ought to be called by sending a COMSIG_MOB_CONTAINER_EXITED signal rather than doing so right away on line 132? Don't get me wrong, the signals are ok, but calling "native" procs this way is pretty convoluted, inefficient.
Also, since you're adding so many signals already, perhaps you should ditch the idea of having subtypes of this component with their own quirks. Let it be done by other datums, components or elements and keep it simple here.

modular_skyrat/modules/mob_container/mob_container.dm
```diff
@@ -0,0 +1,177 @@
...
+
+	remove_mob_explicitly(to_remove)
+
+/**
+ * Signal handler for dumping every mob inside of us.
+ */
+/datum/component/mob_container/proc/dump_all_mobs()
+	SIGNAL_HANDLER
+
+	for(var/mob/contained in mobs_held)
```
Ghommie:
```suggestion
	for(var/mob/contained as anything in mobs_held)
```
[Avoid unnecessary type checks and obscuring nulls in lists](https://github.com/Skyrat-SS13/Skyrat-tg/blob/master/.github/guides/STANDARDS.md#avoid-unnecessary-type-checks-and-obscuring-nulls-in-lists=)

modular_skyrat/modules/mob_container/mob_containers.md
ghost: I have no idea where you possibly could've heard this. From my experience, signals are not a "one-use" sort of deal. It's a signal to specify an action.

modular_skyrat/modules/mob_container/mob_container.dm
ghost: This exists only to avoid re-pathing the parent every single time it's called. The performance difference is negligible.

modular_skyrat/modules/mob_container/mob_container.dm
ghost: Meh, it isn't a huge deal. Efficiency wise, it isn't too terribly different from calling it natively (unless the load is large), however I think this is something worth changing.

Component subtypes aren't a new thing by any means. I'm going to leave it on the subtype idea, that can be changed in the future if absolutely dire.

ghost: Meh, it isn't a huge deal. Efficiency wise, it isn't too terribly different from calling it natively (unless the load is large), however I think this is something worth changing.

Component subtypes aren't a new thing by any means. I'm going to leave it on the subtype idea, that can be changed in the future if absolutely dire.

modular_skyrat/modules/mob_container/mob_containers.md
Ghommie: >I have no idea where you possibly could've heard this.

From a few maintainers on tgstation, obviously.

But it's a consideration, perhaps too petty to be throughfully enforced as a standard.

Gandalf2k15 (changes requested):
Signal abuse is not allowed, do not abuse signals like this.

PR #14843, opened 2022-07-10
[NON-MODULAR] Air Alarms now start on Contaminated mode and use less power on Contaminated mode, to reduce the negative impact that inattentive atmospherics techs have on a round.

Iamgoofball:
## About The Pull Request

Air Alarms now start on Contaminated mode by default.
WEMIsKVUPe

So, the bottom line is that atmos is terribly documented(the official /tg/ wiki is a disgrace with info from literally 2014) and the consequences of this change(bad gas flooding being 10x harder) doesn't matter at Skyrat because flooding gets you banned anyways.

This resolves any issues with floods never getting fixed by atmos techs who are asleep at the wheel or simply uneducated because they haven't been told the atmos secrets by an ancient atmosian elder.

## Changelog
:cl:
balance: Air Alarms now start on Contaminated mode, to reduce the negative impact that inattentive atmospherics techs have on a round.
balance: Scrubbers set to Contaminated mode no longer drain an extremely large amount of power.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/airalarm.dm
+++ b/code/modules/atmospherics/machinery/airalarm.dm
@@ -78,7 +78,7 @@
 	resistance_flags = FIRE_PROOF
 
 	var/danger_level = 0
-	var/mode = AALARM_MODE_SCRUBBING
+	var/mode = AALARM_MODE_CONTAMINATED // SKYRAT EDIT - original: AALARM_MODE_SCRUBBING
 	///A reference to the area we are in
 	var/area/my_area
 
--- a/code/modules/atmospherics/machinery/components/unary_devices/vent_scrubber.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/vent_scrubber.dm
@@ -396,7 +396,7 @@
 
 	var/new_power_usage = 0
 	if(scrubbing == SCRUBBING)
-		new_power_usage = idle_power_usage + idle_power_usage * length(filter_types)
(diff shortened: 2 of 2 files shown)


tf-4:
Me when I drain all the station's power instantly

Iamgoofball:
set the engine up

tf-4:
you underestimate how much power air alarms set to contaminated use

Iamgoofball:
yeha i just checked the numbers you're right, i'll adjust that down too given this change

GuiltyNeko:
Central primary on contaminated draws either 60 or 100 kilowatts.

tf-4:
I don't think this is even an issue that needs resolving anyway tbh.

Iamgoofball:
@tf-4 it's more reliable than trying to get atmos techs to do their job given the shoddy documentation on said job

the only reason not to do these changes is because it makes plasmaflooding non-viable, but also, like, plasmaflooding is nonviable on skyrat by virtue of "you will be banned" so there's no issue there

YakumoChen:
I feel like this isn't really a needed change. The AI can change these as well.
Plasma flooding also isn't banned, it requires an opfor and localised flooding is still within reason for it.

GoldenAlpharex:
Yeah, no, this isn't the way to fix that kind of issue. If Atmos is too complicated to get into, then there should be more in-game ways to teach people the various intricacies of Atmos.

The solution isn't to just baby-proof the station.

PR #14856, opened 2022-07-11
fixes an issue with indentation in the version of select_job() that Skyrat uses

Ghommie:
## About The Pull Request
Sometimes it happens that all command jobs are full or that someone cannot play them because they're either too new to the server or banned. In these cases, it gets impossible to roll for a random job because of the bad indentation.

This shall fix the random job button.

## Changelog

:cl:
fix: Fixed the 'Random Job' button from the 'Choose Profession' panel.
/:cl:


GoldenAlpharex (approved):
That makes sense, good catch, thanks!

PR #14891, opened 2022-07-13
[Modular] Better area spawn algorithm

Tastyfish:
## About The Pull Request

- Placed the actual subsystem and logic into a seperate file from the area_spawn entries.
- Changed the public lustwish definition to include the laundry room as the secondary area, as Tramstation was placing the lustwish inside of dorm rooms (dorms are still last resort, but reliably dangerous).
- Added `meeting_room/council` to the command drobe areas for the sake of Deltstation, which has that area instead of the base meeting room.

Improves the algorithm for placing Skyrat objects for the new area_spawn system from #14679.
- This first and foremost benefits the `wall_hug` spawns which was a tough cookie. Technically changes it for middle of the room ones as well.
- Removed the `max_amount` var because it didn't do anything and it's not clear to me what it's meant to do.
- There were a few bugs with the old system, such as placing objects on open turfs.
- The SS takes 0-0.1 seconds to run on my computer.

Big fundamental changes to backend:

1. processing an area is now cached, so placing a second object in the same area (and same wall_hug state) is basically an `O(1)` operation. Placing an object automatically disqualifies its neighbors (if they are also the same priority because I'm lazy) to be safe.

2. The turfs now have arbitrarily large numeric priorities to really try to find the best fit. For `wall_hug`, it takes into account the turf being empty, how many walls its next to (it wants to be in a nook as that's out of the way), if the diagonals are clear of objects (this is heavily tweaked to make a best effort for the various odd circumstances), and finally, if it's in a large open space (very naively, but tries to avoid 2 turf wide hallways).
   - Emphasizing that I tweaked `wall_hug` priority logic probably 100 times over 2 days, so if the order of the priorities, or the exact definitions of them seem weird, they are what they are because it seems to work.

I've checked Delta, Meta, Icebox, and Tram and they seemed to reliably produce OK results. Ultimately in many cases, they are placed on one of the main candidates for where I would have manually placed them given I can't move or delete anything already in the room.

### Visualization
There is also now a debug command to dynamically view the priorities called Debug -> Test Area Spawn.
Red numbers are wall_hug priority, blue numbers are middle of the room priority. Note in this example, the algorithm would place a vending machine/closet in the bottom corners, extremely out of the way of everything.
image

### Examples
image
image
image
image
image
image
image
image

Less chance of vending machines in the middle of hallways or in the middle of the air.

## Changelog

:cl:
qol: Automatically placed Skyrat vending machines and closets now have better placement.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -204,6 +204,7 @@ GLOBAL_PROTECT(admin_verbs_debug)
 	/client/proc/cmd_display_overlay_log,
 	/client/proc/reload_configuration,
 	/client/proc/reload_interactions, //SKYRAT EDIT ADDITION
+	/client/proc/test_area_spawner, // SKYRAT EDIT ADDITION - AUTOMAPPER
 	/client/proc/atmos_control,
 	/client/proc/reload_cards,
 	/client/proc/validate_cards,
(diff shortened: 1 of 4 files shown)


Tastyfish:
Actual code owner would be @Gandalf2k15

Gandalf2k15 (approved):
based, code checks

PR #14905, opened 2022-07-15
[MODULAR] Return of Removes Valeborg poster (duplicate) and the other thing

YakumoChen:
## About The Pull Request
Removes a duplicate entry for the valeborg poster.

I don't know how the duplicated entry got there but its gone now

## Changelog
none


YakumoChen:
if you dont merge this btw i will cry because I accidentally commited to master

tf-4:
Committing to master gives you +5 based points

ghost:
removing the icon isnt removing it from the pool, poster/random places subtypes of the poster items. If Upstream removed it, there should have been a mirror blacklisting the lizard poster from being chosen

YakumoChen:
I'm looking into it but I think it's a deal with how "random" posters are selected as far as the gas posters still appearing.

They were removed in a PR though, for sure, from the contraband pool.

YakumoChen:
giving up on the poster thing, just sticking with the duplicate

PR #14917, opened 2022-07-15
[HOST][MODULAR]Adds a pointing cooldown (once per half second)

nevimer:
## About The Pull Request

Adds a cooldown to pointing. Three points per 10 seconds. No more point spamming. 

https://user-images.githubusercontent.com/77420409/179305509-ac40c11e-0d8c-443d-b65a-9767da34aa0b.mp4

Less point spamming, yeah. 

## Changelog

:cl:
code: Overrides pointing code to add a timer on pointing. 
/:cl:


code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,21 @@
...
+	if(pointstacks)
+		pointstacks--
+	addtimer(VARSET_CALLBACK(src, pointstacks, 3), 10 SECONDS, TIMER_OVERRIDE)
+
+/mob/pointed()
+	if(client.pointstacks)
+		. = ..()
+		SEND_SIGNAL(client, COMSIG_MOB_POINTED)
+	else
+		to_chat(usr, "You are pointing too hard!")
```
tf-4:
```suggestion
		balloon_alert(usr, "can't point yet!")
```

ghost: Alternatively something like 'slow down!' would work

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,21 @@
+/client
+	var/pointstacks = 3
```
Zonespace27: code doc comment

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,21 @@
...
+	RegisterSignal(src, COMSIG_MOB_POINTED, /client.proc/point_cooldown)
+
+/client/proc/point_cooldown()
+	SIGNAL_HANDLER
+	if(pointstacks)
+		pointstacks--
+	addtimer(VARSET_CALLBACK(src, pointstacks, 3), 10 SECONDS, TIMER_OVERRIDE)
+
+/mob/pointed()
+	if(client.pointstacks)
```
Zonespace27:
```suggestion
	if(client?.pointstacks)
```

Zonespace27 (changes requested):

code/modules/client/verbs/pointing_cooldown.dm
tf-4:
```suggestion
	var/point_stacks = 3
```

GuiltyNeko:
~~I'm just a guy with an opinion but unironically balloon alerts would be best for this.~~

Already suggested, muh bad.

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,24 @@
+/client
+	/// Your clients allowed pointing stacks.
+	var/point_stacks = 3
```
jjpark-kb: make 3 a define

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,24 @@
...
+	. = ..()
+	RegisterSignal(src, COMSIG_MOB_POINTED, /client.proc/point_cooldown)
+/**
+ * Increments the allowed pointing stacks and resets the pointer cooldown.
+ */
+/client/proc/point_cooldown()
+	SIGNAL_HANDLER
+	if(point_stacks)
+		point_stacks--
+	addtimer(VARSET_CALLBACK(src, point_stacks, 3), 10 SECONDS, TIMER_UNIQUE)
```
jjpark-kb: make 3 a define

ArdentSpark:
three per ten is real harsh, not being to point for an entire ten seconds, seems incredibly high. Bearing in mind you might be in combat.
Something like one per two isn't that much different, or hell one per three would stop you getting locked out for so long

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,24 @@
...
+
+/mob/pointed()
+	if(client?.point_stacks)
+		. = ..()
+		SEND_SIGNAL(client, COMSIG_MOB_POINTED)
+	else
+		balloon_alert(usr, "too fast, calm down!")
```
tf-4: This is kind of long and not necessarily clear, I still prefer

```suggestion
		balloon_alert(usr, "can't point yet!")
```

tf-4 (approved):

code/modules/client/verbs/pointing_cooldown.dm
```diff
@@ -0,0 +1,26 @@
+#define POINT_STACK_LIMIT 3
+/client
+	/// Your clients allowed pointing stacks.
+	var/point_stacks = POINT_STACK_LIMIT
```
Gandalf2k15: making this var on every client is going to add pointless overhead.

nevimer: what do otherwise?

YakumoChen: > making this var on every client is going to add pointless overhead.

> pointless

ba dum tish

YakumoChen:
3 per 10 is way too strange of a cooldown. Just make it one per 3 or something

nevimer:
> 3 per 10 is way too strange of a cooldown. Just make it one per 3 or something

why would I do that?

code/modules/client/verbs/pointing_cooldown.dm
nevimer: > making this var on every client is going to add pointless overhead.

ok but how else are you going to track something on a per client basis

YakumoChen:
having to wait a full 10 seconds after pointing a few times is much more obnoxious. Being able to point once every three seconds feels a lot more sensible and is more in line with other similar systems such as emoting.

tf-4 (changes requested):
one per three does make more sense IMO.

nevimer:
> one per three does make more sense IMO.

That is less and that is not the request that was made.

YakumoChen:
1 per 3.3 seconds then?

I don't believe @hugmeorelse will mind as long as the intent of the request (reducing spam pointing) is honoured, and the cooldown is generally the same.

Potentially having to wait 10 seconds to point again is a social ice age and far more likely to stifle RP, being able to spam point 3 times in a second does not make it any more noticable and is more annoying.

GoldenAlpharex (changes requested):
After discussing this with Xyel, I'd like to try this out with one point every 0.5 seconds, so as not to heavily punish those that use it for legitimate reasons, while preventing spam in a relatively decent way, making it even more obvious when someone's just spamming it to be annoying.

nevimer:
oh woops forgot about this lol

tf-4:
> oh woops forgot about this lol

I can reopen it if you're ready to make the requested changes

nevimer:
> > oh woops forgot about this lol
> 
> I can reopen it if you're ready to make the requested changes

ya sure ill do it

tf-4 (changes requested):
Needs to be in a modular folder

nevimer:
https://user-images.githubusercontent.com/77420409/190879737-b0df80da-5782-4f86-863f-3934182c9264.mp4

please enjoy

tf-4 (approved):

PR #14978, opened 2022-07-17
Stops borers synthesising any reagent

tf-4:
## About The Pull Request

adds a `REAGENT_CAN_BE_SYNTHESIZED` check to the borer 'learn chemical' ability

Synthesising things that are not meant for it to be possible to be synthesised isn't great!
Fixes an oversight and a consistency issue.

## Changelog
:cl:
balance: Borers can no longer synthesise chemicals that are meant to be unsynthesisable.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
@@ -137,6 +137,10 @@
 		to_chat(owner, span_warning("You already know this chemical!"))
 		cortical_owner.chemical_evolution += 5
 		return
+	if(!(reagent_choice.chemical_flags & REAGENT_CAN_BE_SYNTHESIZED))
+		to_chat(owner, span_warning("You struggle to understand the complexities of this chemical!"))
+		cortical_owner.chemical_evolution += 5
+		return
 	cortical_owner.known_chemicals += reagent_choice.type
 	cortical_owner.blood_chems_learned++
 	if(cortical_owner.blood_chems_learned == 5)


jjpark-kb:
this isn't a fix, but rather a balance; please change your CL to the appropriate tag.

It was never an oversight as it was intended for borers being able to learn any chemical so long as their host has it-- yes this is more of a consistency thing but I made the conscious decision for it. This is why I say this is a balance.

tf-4:
There are other things in the game (odysseus) that can copy and reproduce any chemical that don't have the ability to bypass the flag, the whole point of the flag is to prevent this stuff. It's a consistency issue and this fixes it even if it does nerf borers as a side effect

jjpark-kb:
> There are other things in the game (odysseus) that can copy and reproduce any chemical that don't have the ability to bypass the flag, the whole point of the flag is to prevent this stuff. It's a consistency issue and this fixes it even if it does nerf borers as a side effect

Again, I acknowledged that its a consistency thing, but I don't accept it being a "fix."

jjpark-kb:
image
please, change your CL

softcerv:
> > There are other things in the game (odysseus) that can copy and reproduce any chemical that don't have the ability to bypass the flag, the whole point of the flag is to prevent this stuff. It's a consistency issue and this fixes it even if it does nerf borers as a side effect
> 
> Again, I acknowledged that its a consistency thing, but I don't accept it being a "fix."

It's not a fix, but it's a change to make it more consistent, which is commonly typed as a fix. This is a case where most players would consider it a fix too.

tf-4:
dont really care enough to argue about it

PR #15005, opened 2022-07-18
[MIRROR] [MDB IGNORE] Hydroponics tray, shower, and sink improvements [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/67672
--------------------

## About The Pull Request

General to all 3 appliances:
- Generally improved visual behavior of the plumbing connections coming out of offset machines/structures, including sinks, showers, and hydroponics trays so that they visually link up to everything else but still connect to the machine too.
- Level-start plumbing appliances' connections now visually work like atmos machines:
  - If they spawn on a floor tile, the pipe is hidden, if built on a floor tile in-game, the pipe is visible.
  - Un-wrenching an appliance reveals the pipe, placing a fresh floor tile hides the pipe.
  - However in the specific case of showers and sinks, un-wrenching now deconstructs them to be consistent with other structures. Rebuilding them will reveal the pipe.

image
image
image
image

As promised in my other plumbing PR, I'm fixing the plumbing behavior of hydroponics trays:
- The pipe connecting to hydroponics trays now sends water to the mystical water tray correctly.
- Note that if the nutrient level is less than full, but the water level is full, the extra water will be deleted instead of preserved, but this behavior is in parity with using a beaker so I'm not really concerned.
- Note that this will rapidly drain the stationary water tank, but you can also quickly drain that with a bucket in the first place. So if you pipe water, use a chemical synthesizer.
- Fixes #66359
- Fixes #54226

Several shower and sink construction improvements as their code had rotted:
- Showers can now be dismantled with a wrench to match most sinks and most structures.
  - The thermostat adjustment behavior is now down with a screwdriver, and there's an examine tip to indicate that.
- Dismantling showers and sinks now actually drops the water recycler.
- You can now remove the water recycler from sinks and showers with a crowbar, and add one by attacking the appliance with a water recycler.
- As you can now *remove* the water recycler from showers, the function to toggle recycling was removed.
  - Now, the multitool switches between 3 modes, as requested below. The 3 modes are:
    1. The classic and default mode of running until manually turned off, or running out of liquid.
    2. Run until turned off, running out of liquid, or if 15 seconds has passed since turning on.
    3. Run forever, until water has turned off, but turn back on again if there is liquid again.
- In testing this, I found out that showers drop more iron than they cost. That is now fixed.
- I've flipped the facing behavior of placing sinks and showers so you now face the wall instead of away from it. This seems far more intuitive to me, and matches the construction of most other wall-facing/mounted things.
- Sinks are now placed at the same offset as they are in the maps, instead of being inconsistently centered.
  - When figuring out the offsets, I realized the pixel offsets for the sinks were wildly inconsistent between maps and many times within the same map, almost always for no apparent reason and off by 1 pixels.
  - As such, showers and sinks now have directional versions. *This is what all the map changes are.* I've included the update script.
  - I think my big plumbing PR actually is what fixed the backend part of this, but noting this as being fixed regardless:
    - Fixes #67584
  - Sinks and showers are now positioned correctly when rotated on shuttles.
    - Partially Fi*es #39347: Fixes the sinks and showers part, but I haven't checked or changed urinals, since they're out of scope of this PR due to not being plumbing related.

## Why It's Good For The Game

Construction QOL, hydroponics tray plumbing input working as people expect, improved visuals.

## Changelog

:cl:
fix: Shower and sink dismantling is now consistent.
fix: Hydroponics tray plumbing connections now accept water without clogging.
add: Showers now have 3 selectable auto-shutoff modes.
qol: Sinks and showers are now constructed with the player facing the wall.
imageadd: Visuals improved on plumbing connections for offset appliances, like sinks and showers.
/:cl:


ORCACommander:
@GoldenAlpharex @Gandalf2k15 there was an update paths script in this that needed to be run

PR #15013, opened 2022-07-18
[MODULARISATION AUDIT] Layer defines

Gandalf2k15:


code/__DEFINES/~skyrat_defines/layers.dm
```diff
@@ -0,0 +1,5 @@
+///Plane of the "splash" icon used that shows on the lobby screen. Nothing should ever be above this. (Except the logo)
+#define BLACK_FADE 9997
+#define SKYRAT_LOGO 9999
```
GoldenAlpharex, line 3: We should likely fix this, too, while we're thinking about this. Not in this PR, but it's still something that's been broke since you added your HTML menu.

GoldenAlpharex (approved):

code/__DEFINES/~skyrat_defines/layers.dm
```diff
@@ -0,0 +1,5 @@
...
+
+#define CHAT_INDICATOR_LAYER 6 // 1 above BYOND's FLY_LAYER
```
tf-4: Do we still need this since tg added typing indicators?

Gandalf2k15: no

code/__DEFINES/~skyrat_defines/layers.dm
Tastyfish: For that matter, the lobby_cam module should also be removed since it's just burning CPU and net traffic doing something nobody can ever see anymore.

Tastyfish:
owned

code/__DEFINES/~skyrat_defines/layers.dm
tf-4: can we get rid of it then

Gandalf2k15: yes

PR #15016, opened 2022-07-18
[MODULARISATION AUDIT] Strippable defines

Gandalf2k15:


GoldenAlpharex (approved):

code/__DEFINES/~skyrat_defines/strippable.dm
```diff
@@ -0,0 +1,2 @@
+#define THIEVING_GLOVES_STRIP_SLOWDOWN 0.5 //a multiplier for the amount of time it takes to strip someone
```
tf-4:
```suggestion
/// A multiplier for the amount of time it takes to strip someone with thieving gloves
#define THIEVING_GLOVES_STRIP_SLOWDOWN 0.5
```
Might as well while here

PR #15023, opened 2022-07-18
Black mesa ghost roles (sci, guard, HECU) can now speak common again

Zonespace27:
## About The Pull Request
This gives BM's ghost roles their common back, letting them know common and uncommon

It can be challenging to RP when nobody speaks the same language

## Changelog

:cl:
balance: Black mesa ghost roles now speak common as well as uncommon
/:cl:

### DIFF
--- a/modular_skyrat/modules/black_mesa/code/ghost_spawners.dm
+++ b/modular_skyrat/modules/black_mesa/code/ghost_spawners.dm
@@ -12,7 +12,6 @@
 /obj/effect/mob_spawn/ghost_role/human/black_mesa/special(mob/living/carbon/human/spawned_human)
 	. = ..()
 	spawned_human.grant_language(/datum/language/uncommon, TRUE, TRUE, LANGUAGE_MIND)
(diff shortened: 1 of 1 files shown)


Stalkeros2:
I'd advise to give it to only a few select people who you'd expect to be the leading roles of both HECU (captain, obviously) and the Science Team (security guard, I assume), to keep the immersion. Another good way of doing this would be spreading some uncommon translation headsets around the Black Mesa in some really-really obvious places.
Or, honestly, really, just... *don't*. Reason why nobody in Mesa knows common is to both give ghost roles an edge over Vanguard in terms of communication and for them to require a dedicated linguist for additional RP.

theselfish:
I think unless you're going to go fully into the "HECU/Mesa as a whole is in the past so they speak English" thing, giving them their own english language, this PR is a necessity.

Gandalf2k15 (approved):

Gandalf2k15:
having watched encounters, the language barrier generally caused bad conflict

PR #15028, opened 2022-07-18
Disable spiders from rolling in events

Funce:
## About The Pull Request
Good bye spiders! You will no longer appear here.

Spiders are boring, we're dealing with it today.

## Changelog

:cl: Foxtrot / Funce
balance: Spider infestation no longer appears randomly
/:cl:

### DIFF
--- a/modular_skyrat/modules/events/code/event_overrides.dm
+++ b/modular_skyrat/modules/events/code/event_overrides.dm
@@ -43,7 +43,8 @@
  * Upped to ensure lowpop steamroll does not happen
  */
 /datum/round_event_control/spider_infestation
-	min_players = 70
+	// min_players = 70
+	max_occurrences = 0
 
 /**
  * Meteor Waves


PCPort:
Don't mind me, just downvoting the salt PR.

Spiders are:

-Fun to play
-A genuine threat to the entire crew if played well, not just security (rare)
-Not at all impossible to defeat, especially with the advent of gun cargo
-Variety to the antag roster

If you don't like the current state of spiders, then think up some actual CHANGES as opposed to just removing them. Improve, not remove, remember?

Funce:
They're still there, they just don't appear randomly anymore. So only when the admins feel particularly..... vindictive.

PCPort:
They roll rarely enough that it's a non-issue as is. At most you'll see like... One spider infestation a week, if that.

Higgin:
spiders should not only roll randomly but be more frequent, they're a good midround and this is a bad PR

edit: also gating things to admin-spawn-only tends to mean they show up even less than before, which sucks when a lot of those things stand to add more cool variety to the rounds a person might catch

hugmeorelse:
Spider shouldn't be rolling right now. We were planning on bringing it back with a combat rework, but that didn't happen. So, it shouldn't be rolling right now. That may change after discussion. However, it's being turned off until discussions are had.

This should be running in dynamic instead anyhow.

PR #15029, opened 2022-07-18
[NON-MODULAR] de-ratelimits TGUI topic calls

azrael-x86:
## About The Pull Request

TGUI calls are no longer rate-limited, removing the topic call spam warning given recently whilst deadminning/switching mobs

why was this happening? my top guess is that since we have more tabs than upstream does, we're calling `Remove-Tab` more often everytime we deadmin/switch mobs, leading to the rate limiting warning messages

no more topic spam

## Changelog

:cl:
fix: removes topic spam warning whilst deadminning
/:cl:

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -48,6 +48,11 @@ GLOBAL_LIST_INIT(blacklisted_builds, list(
 		if (!asset_cache_job)
 			return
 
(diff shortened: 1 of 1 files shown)


Gandalf2k15:
If this is a problem because of the amount of tabs(which we have no extras) then take it upstream as it will eventually become an issue there.

PR #15036, opened 2022-07-18
[MDB IGNORE] Type Paths Audit: Floor Var Edits

ORCACommander:
## About The Pull Request

was running the type path migration scripts on my downstream. looks like 5 of the automapper files are using the old var edit method of changing flooring appearance instead of the new mapping helpers.

Technical PR. Does not alter gameplay or role play

## Changelog

:cl:
code: NT is now sourcing distressed flooring through a lowest bidder program
/:cl:


Gandalf2k15:
I think you got the wrong place

PR #15047, opened 2022-07-19
[MIRROR] Merge Conflict Markers Proper Variable Addition [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68536
--------------------

## About The Pull Request

Hey there,

I fucked up. I used a comma when I should have used a semicolon. I think I did change it back after a test case but I probably forgot to save, skull emoji. Should be all good now.

## Why It's Good For The Game

image

It wasn't supposed to be this way!!! Noooooooooooooooooo- I think the name showing up like that is because TGM will read it that odd way, this should definitely split up the name|desc pattern.

## Changelog

Nothing particularly player facing.


GoldenAlpharex (approved):
Thanks Funce!

PR #15048, opened 2022-07-19
[MIRROR] Fixes circuit dispensers add_item [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68528
--------------------

## About The Pull Request
- adding items to a dispenser shell would hit the shell with the item first
- little to no feedback from the shell as it just eats your held item
- bulk adding items could theoretically exceed the weight limit for items or add in nested bags

## Why It's Good For The Game
you can add items to the dispenser without permanently damaging it
it's a little more obvious it just ate your item
guards against an edge case exploit that's probably unused

## Changelog

:cl: jlsnow301 MooCow
fix: Adding items to dispenser shells no longer damages the dispenser
/:cl:


GoldenAlpharex (approved):
Thanks Funce!

PR #15053, opened 2022-07-19
Adds better ghost role stuff for centcomm

weesnaw-ops:
## About The Pull Request

This adds the ghost roles from the 'no eorg' test round to the codebase, along with many valuable additions such as ovens, bartending supplies, spawning pods- et cetera. 

This PR opens up a little bit of CC access to the ghost roles on station but nothing too major: It lets cooks get behind the kitchen, lets border patrol guards get behind border patrol, adds a freezer, and lets everyone access the viewing area for the thunderdome if they want to cook or watch admins dick around while they're bored preparing a bunch of food

## How This Contributes To The Skyrat Roleplay Experience

This PR will help decrease the amount of EORG and encourage end of round RP for reasons seen in the round run by Ardent Spark. People actually have a place to go RP instead of sitting around on a bench with an itchy trigger finger. Maybe a border guard could be extra stringent. Maybe the chef could have some weird wacky food. Et cetera.

## Changelog

:cl:
add: spawners for cc chef/waiter/bartender/border patrol
add: ovens, grinders, food processors, booze dispensers, bars for the chef + co
add: stamps for the border patrol
add: cash so the bartender/chef can buy ingredients and such
/:cl:

PR #15059, opened 2022-07-19
Adds better cooking equipment to the CentCom kitchen

weesnaw-ops:
## About The Pull Request

This is mostly the same as #15053, but it does not have the spawners, and I edited a few things that I put in the lobby that I thought just looks stupid.

No spawners are wanted in CC but I don't entirely see the harm in updating the equipment to be able to run more than JUST donk pockets. This makes it so if an admin wants to do another thing where they spawn a spawner like a bardrone, like ardent spark did, they can actually do so without having to do a large amount of maintenance. This does not include any of that thunderdome access as well so there's no roaming around admin testing areas. It just makes it easier for admins to put people in CC lobby.

## Changelog

:cl:
add: Adds an oven, food processor, grinder, fridges, and a minibar to the CC lobby
add: Adds stamps to the CC checkpoint
add: No spawners this time. Just decals to indicate where to spawn if an admin would like to do this.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat.dmm
@@ -7,11 +7,24 @@
 /obj/effect/turf_decal/siding/dark{
 	dir = 5
 	},
+/obj/machinery/vending/boozeomat/all_access,
 /turf/open/floor/iron,
 /area/centcom/central_command_areas/evacuation)
 "ad" = (
 /turf/open/space,
 /area/space)
+"af" = (
+/obj/machinery/door/airlock/centcom{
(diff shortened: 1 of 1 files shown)


theselfish:
I would also change the cc kitchen access to...
Just normal kitchen access? Maybe?

weesnaw-ops:
It is done. Kitchen access

Jolly-66:
I left the design of the kitchen intentionally barebones. This is supposed to just be a small place to relax and recharge before, meteorically, your next shift. 

Me and Orion are already working on a facelift of my current design, as well as moving more things over to the automapper.

PR #15064, opened 2022-07-20
Area Spawn now supports wall mounts. wall_hug replaced with mode.

Tastyfish:
## About The Pull Request

As promised, added wall mount support. Turned out to be vaguely complex to yield OK results.
- Picking the placement mode for a `/datum/area_spawn` is now determined by a `mode` var instead of `wall_hug`.
- Note that wall mount mode *requires* the desired atom to have mapping directional helpers. As a result, the positronic console now has directional helpers.

image
image
image
image

Positronic consoles exist again in upstream maps.

## Changelog

:cl:
add: The McPositronic alert console is back.
/:cl:


Jolly-66:
Thank you for your hard work, Tasty.

Tastyfish:
Legit failure on Kilostation, which I didn't test locally for obvious purposes. Will need to add another fallback area for the robotics console.

Edit: Fixed

PR #15068, opened 2022-07-20
[MODULARISATION AUDIT] Modularise some liquids code

tf-4:


ghost:
While we're here, I beg of you, add some opacity to fluids on the ground and make extinguishing the fire ACTUALLY extinguish the fire 🥺 those two issues ALONE make the fluids horribly unfun 🙏

tf-4:
While that's a good idea, that's not really in the scope of the pr

Gandalf2k15 (approved):

SpaceVampire:
Merged with three downvotes?
Democracy is dead. Real 1984 hours.

dwasint:
> While we're here, I beg of you, add some opacity to fluids on the ground and make extinguishing the fire ACTUALLY extinguish the fire 🥺 those two issues ALONE make the fluids horribly unfun 🙏

if you wish to port that feature from my version of it i have a system similar to that that adds dynamic opacity to fluids.

PR #15101, opened 2022-07-21
[SEMI-MODULAR] Vine Balance Tweaks (Seeding and Pulling)

Higgin:
## About The Pull Request

Seeding vines no longer trigger on mobs getting pulled through seeding vines. ~~Vines can drag people again.~~ This includes from the venus human trap vine pull. To compensate, seeding new venus human traps happens a lot faster when you're moving through them recklessly, and it happens across a range of times (15-30s). Seeding procs from hitting the seeding vines (if you're just fighting your way through them) no longer stack.

This should remove one of the biggest potential abuse-cases for vines and some of* the reason for disallowing them from pulling back downed people/bodies - seeding allowed this to be used to farm up massive numbers of human traps. Now, the greater risk is from running through the vines, and it's much more likely to create more venus traps in the time that you're still in the thick of the fight if you do a lot of it. 

~~Moving people around in the environment is an interesting part of the game - instead of outright prohibiting it (and allowing the human traps' vines to be used to abuse it anyway with seeding) this opens up more interesting play to vines and challenges players to recognize the terrain if they get grabbed, dragged, and more deliberately fight their way into a better situation if caught.~~

Vines can still drag people using the tendrils if they really want to move somebody, but as policy indicates that they generally oughtn't be, this at least helps address some of the abuse and jank of combining their pulling tendrils and seeding vine patches to produce massive numbers of extra venus traps. 

Vine balance good. Stops people vine-booming unless they really, really try to stack it up on themselves. 

## Changelog

:cl:
balance: Getting pulled by projectile vines or other mobs through Seeding vine turfs no longer procs seeding on the person being pulled. 
balance: Seeding vine spawns procced by movement through a turf happen much more quickly and in random span of time (15-30s.) Seeding vine spawns procced by attacking seeding vine turfs no longer stack and happen after a minute.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
+++ b/code/modules/mob/living/simple_animal/hostile/venus_human_trap.dm
@@ -241,7 +241,11 @@
 		if(istype(B.target, /atom/movable))
 			var/atom/movable/AM = B.target
 			if(!AM.anchored)
+				AM.set_pulledby(src) //skyrat edit add - setting pulledby to prevent juggling stamcrit mobs back and forth through seeding vines
 				step(AM, get_dir(AM, src))
+				AM.set_pulledby(null) //skyrat edit start - clearing pulledby - also interupts pulling people away, making it a bit of a tug-of-war
+			if(isliving(AM))
+				reset_pull_offsets(AM) //skyrat edit end - making it not break offsets
 		if(get_dist(src, B.target) == 0)
 			qdel(B)
 
--- a/modular_skyrat/master_files/code/modules/events/spacevine.dm
+++ b/modular_skyrat/master_files/code/modules/events/spacevine.dm
@@ -442,10 +442,10 @@
 /datum/spacevine_mutation/seeding/on_cross(obj/structure/spacevine/vine_object, mob/crosser)
 	if(isliving(crosser))
 		var/mob/living/living_crosser = crosser
-		if(isvineimmune(living_crosser) || living_crosser.stat == DEAD)
+		if(isvineimmune(living_crosser) || living_crosser.stat == DEAD || living_crosser.pulledby)
 			return
 		if(prob(10))
-			addtimer(CALLBACK(living_crosser, /mob/living/proc/plant_kudzu), 1 MINUTES)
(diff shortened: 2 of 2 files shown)


RimiNosha:
Letting venus traps pull people seems like a terrible idea, aren't they meant to only defend the vines?

This would also open the door to them pulling players away into an inaccessible spot, which is the bigger concern here!

Higgin:
> Letting venus traps pull people seems like a terrible idea, aren't they meant to only defend the vines?

Taking down a scythe/chainsaw-wielding maniac frontliner is a key part of playing defense. Contestation of the downed is part of that, and they could already pull with the vine projectile - if you stay within four tiles' range or so and don't break LOS, it'll keep pulling the downed person. 

> This would also open the door to them pulling players away into an inaccessible spot, which is the bigger concern here!

Where it actually got ridiculous was farming runaway seeding procs (which create a new kudzu plant that can start rolling mutations plus a new venus human trap on top.) Contestation is or should be considered fair play - vines can't shuttlegib somebody or disable their sensors.

RatFromTheJungle:
Venus traps aren't supposed to be pulling people period, to the degree it's one of the very-few things banned for them in policy; 
image
so a PR encouraging that goes directly against their intent (and one of the few reasons they haven't just been booted out of vines in general by now)

They weren't banned from dragging corpses just to stop seeding, they were banned from dragging corpses, because it frankly wasn't fun for anyone but themselves for them to down someone, and immediately yank them deep into the vines.

Higgin:
> they were banned from dragging corpses, because it frankly wasn't fun for anyone but themselves for them to down someone, and immediately yank them deep into the vines.

It's less fun to fight them when they're hobbling themselves, but if that's how it goes, that's how it goes. I put the check blocking pulling mobs back in. 

This should at least still make the seeding less abusable (i.e. farming a monkey on a seeding patch,) make it more important to avoid heedlessly running through, and make it less likely to turn into a random minute-later venus trap or six that all reappear in the next maintenance tunnel over or in/on the way to medical (and start another area with spreading vines to worry about when they do.)

PR #15109, opened 2022-07-22
[MODULAR] Bluespace Crystals can now be put in item creation material_storage components

Zenitheevee:
## About The Pull Request

so like
Bluespace crystals couldn't be placed in storage that didn't specify bluespace crystals could be put in them.
Which almost made me thought it was some sort of... Designed behavior... Y'know as one normally should
But like, The autolathe has a proc part that implies bluespace crystals could be placed in it so its clearly not.
but like, what even would i label this? Like its a balance issue but its not like, a balance issue in the way we typically think. it enables balance issues, but also like no one really... does the things? Is a clothing resprite a balance issue if it cuts a few pixels off of it because those are now deadpixels? Would pizza having slightly more food value be a balance issue? Would making space fries require salt be a balance issue? The world may never know, but now i am always seeking resolution.

Made it modular too btw. Yes i could go through TG but like then i'd have to go through them with my no GBP over there and figure out whether THEY consider it a balance issue or a fix or whether tacos are actually sandwiches and that my life was a lie.

idunno. does it have to? Is someone going to RP putting in a bluespace crystal into an autolathe? Do people do that? Sure it enables them to do it without looking like a dunce but *do people even do that?*

## Changelog

:cl:
balance: bluespace crystals can go into mat_storage components that uses MAT_CATEGORY_ITEM_MATERIAL. 
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/datums/materials/basemats.dm
+++ b/modular_skyrat/master_files/code/datums/materials/basemats.dm
@@ -1,0 +1,2 @@
+/datum/material/bluespace
+	categories = list(MAT_CATEGORY_ORE = TRUE, MAT_CATEGORY_ITEM_MATERIAL = TRUE)
--- a/tgstation.dme
+++ b/tgstation.dme
@@ -4720,6 +4720,7 @@
 #include "modular_skyrat\master_files\code\datums\id_trim\solfed.dm"
 #include "modular_skyrat\master_files\code\datums\id_trim\syndicate.dm"
 #include "modular_skyrat\master_files\code\datums\martial\cqcplus.dm"
+#include "modular_skyrat\master_files\code\datums\materials\basemats.dm"
 #include "modular_skyrat\master_files\code\datums\quirks\_quirk.dm"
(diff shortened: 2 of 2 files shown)


ghost:
I know you dont WANT to take this to TG, but -
There's no reason to have it alone here. Just, straight up, it'll be a random un-used piece of information that only applies here.
And if TG decides they dont want it, they'll give a proper reason as to why not, instead of there being little reason as to why.

Paxilmaniac:
this should be an upstream pr i think

Paxilmaniac:
damnit you got it first

Zenitheevee:
> I know you dont WANT to take this to TG, but - There's no reason to have it alone here. Just, straight up, it'll be a random un-used piece of information that only applies here. And if TG decides they dont want it, they'll give a proper reason as to why not, instead of there being little reason as to why.

i really *dont* because at *best* it means i just go take it up there and jump through their logic hoops of whether or not and at *worst* i have to do some small fix PR's before they consider it
But also because it means updating my TG fork

Zenitheevee:
taking it upstream but like i'm gonna re-open it if it gets denied up there

PR #15152, opened 2022-07-23
Adds Jolly and myself to the tools codeowners

GoldenAlpharex:
## About The Pull Request
@Jolly-66 has been added as an UpdatePaths codeowner (so he gets notified when a new one comes around, so we don't miss any on accident). Meanwhile, I've gone ahead and added myself to the whole folder, as I want to know when new tooling is added, as I generally keep an eye to make sure we're not missing any content for those.

This means we should see fewer UpdatePaths scripts being forgotten.

## Changelog

Non player-facing

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,7 @@
 
 # GoldenAlpharex
 /modular_skyrat/ @GoldenAlpharex
+/tools/ @GoldenAlpharex
 
 # Jake Park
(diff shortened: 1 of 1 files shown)


Jolly-66 (approved):
I'll still be pinged regardless since the maps should be retroactively updated per some mirrors, but it should give me a better scope since you'll be pinged regardless, and less should fall in between the cracks.

Issue #14403, opened 2022-06-19
Heterochromia eyes change side based on what way you are facing


TetrakoHateThis:
## Testmerges:
[#12529: '[MODULAR] Status Indicators with Preference' by nevimer at commit 360dfe72a7](https://github.com/Skyrat-SS13/Skyrat-tg/pull/12529)
[#10823: 'Database revision revert' by Gandalf2k15 at commit 30f7853899](https://github.com/Skyrat-SS13/Skyrat-tg/pull/10823)
[#14023: 'Fleshmind [BETA]' by Gandalf2k15 at commit 60a1461cb4](https://github.com/Skyrat-SS13/Skyrat-tg/pull/14023)
[#13400: 'Vox version Beanyvox' by Gandalf2k15 at commit b6f5d423c1](https://github.com/Skyrat-SS13/Skyrat-tg/pull/13400)
[#14290: 'Disables bureaucratic error' by tf-4 at commit b2c27333cc](https://github.com/Skyrat-SS13/Skyrat-tg/pull/14290)

## Reproduction:
Use the heterochromia quirk. Face south. Left eye is colour X, right eye is colour Y. If you turn to face east, your sprite will show your visible right eye as Colour X. If you face west, your visible left eye eye shows Colour Y, meaning your eye colours flip when you face to either side.

https://i.imgur.com/p9WsAAd.png - Image example

ghost:
Comedy. Easy fix, just needs the sides swapped to be correct.
God knows what .dmi its stored in tho

nikita-chekistov:
Fixed by https://github.com/tgstation/tgstation/pull/68730

PR #15166, opened 2022-07-24
Heterochromia Fix

Microvolnovka19:
## About The Pull Request

Fixes the left/right bug.

close:https://github.com/Skyrat-SS13/Skyrat-tg/issues/14403

How it looks now:

image
image
image

Enchances visual experience

## Changelog

:cl: SuperDrish
fix: Fixed Heterochromia not knowing where the right and left eyes are
/:cl:


GoldenAlpharex (changes requested):
This is a good fix! However, I'd like to kindly request that you take this upstream, as considering they're using the same .dmi as ours, they're almost certainly experiencing the same bug as us.

On top of that, this edit to human_face.dmi would likely be lost as soon as we end up having an icon conflict in that file, because we might end up resetting it to how it is upstream. Basically, it's going to save everyone a headache.

nikita-chekistov:
Done

PR #15178, opened 2022-07-25
fauna reagent health

jjpark-kb:
## About The Pull Request
so, reagents typically are just useless on fauna/megafauna... I plan on changing that in some basic form at the moment.

toxins (and medicines) should affect some fauna/megafauna as they are living beings... it's expected reagents would affect them.

## Changelog

:cl:
add: mining mobs and megafauna are now impacted slightly by toxins and medicine
/:cl:

### DIFF
--- a/modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
+++ b/modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
@@ -16,7 +16,7 @@
 /mob/living/simple_animal/Life(delta_time, times_fired)
 	. = ..()
 	if(reagent_health && reagents)
-		for(var/datum/reagent/reagents_within in reagents.reagent_list)
+		for(var/datum/reagent/reagents_within as anything in reagents.reagent_list)
 			if(istype(reagents_within, /datum/reagent/toxin))
 				var/datum/reagent/toxin/toxin_reagent = reagents_within
 				var/toxin_damage = round(toxin_reagent.toxpwr)
--- a/modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
+++ b/modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
@@ -1,0 +1,28 @@
+/mob/living/simple_animal
+	/// whether the simple animal can be healed/damaged through reagents
+	var/reagent_health = FALSE
+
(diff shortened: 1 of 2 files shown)


Whygil:
Isn't it for a balance? Bluespace artillery can't kill megafauna either. Same reaction on chemicals would be weird, considering that they are horrible monsters from hell.

jjpark-kb:
> Isn't it for a balance? Bluespace artillery can't kill megafauna either. Same reaction on chemicals would be weird, considering that they are horrible monsters from hell.

mentioning balance, around me? hah.

Zonespace27 (commented):
What's the overhead on adding reagents to all megas and asteroid mobs

modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
```diff
@@ -0,0 +1,28 @@
...
+/mob/living/simple_animal/hostile/megafauna
+	reagent_health = TRUE
+
+/mob/living/simple_animal/hostile/asteroid
+	reagent_health = TRUE
+
+/mob/living/simple_animal/Life(delta_time, times_fired)
+	. = ..()
+	if(reagent_health && reagents)
+		for(var/datum/reagent/reagents_within in reagents.reagent_list)
```
tf-4:
```suggestion
		for(var/datum/reagent/reagents_within as anything in reagents.reagent_list)
```

jjpark-kb:
> What's the overhead on adding reagents to all megas and asteroid mobs

not sure, would be nice to have a TM to see overhead on the main server.

modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
jjpark-kb: how it is currently written works. I can test later if this is fine as well.

tf-4:
Why does it need a TM to see that? I'm not sure what would differ on live versus local with this PR

jjpark-kb:
you're right, a merge would be better. It literally won't be much at all since all human mobs already do this essentially.

modular_skyrat/modules/fauna_reagent/fauna_reagent.dm
Zonespace27: above method is faster

PR #15186, opened 2022-07-25
[MODULAR] Reduce Interdyne Turrets

RimiNosha:
## About The Pull Request
So turns out I completely missed the fact that turrets autotarget and aggro megafauna... well, time to cut them back until I get around to adjusting their targeting logic!

Roundstart annihilation of the Interdyne isn't fun, and the base getting leveled cause a miner was fighting a colossus nearby is also not fun!

## Changelog

:cl:
del: Reduced the amount of turrets on the Interdyne to appease the local megafauna.
/:cl:


RimiNosha:
Del tag and not balance/rebalance, please!

ghost:
Do you think it might be more effective to make/map a subtype of the current turret, with a lower range or an inability to target megafauna? IIRC projectiles dont even damage megafauna to a point of it being useful against them

RimiNosha:
Closing cause I've got an alternate solution!

PR #15192, opened 2022-07-26
Fixed computers joining together when there's also wall mount computers

Tastyfish:
## About The Pull Request

So that security computer in Metastation Orderly's office was glitched because it's on top of the telescreen, which is also a computer, so the overlay joining code got confused.
image

Fixed it by looking harder for a connecting computer.
image

Aesthetics.

## Changelog

:cl:
fix: Central Command finally gave the orderly the correct computer console case on Metastation.
/:cl:

### DIFF
--- a/modular_skyrat/modules/connecting_computer/code/_computer.dm
+++ b/modular_skyrat/modules/connecting_computer/code/_computer.dm
@@ -11,6 +11,18 @@
 /obj/machinery/computer/proc/callback_proc_issue()
 	update_overlays()
 
+/**
+ * Find and return a computer at this location that can connect to us.
+ * Arguments:
+ * * at - the location to look.
+ */
+/obj/machinery/computer/proc/find_connectable_computer(turf/at)
+	for(var/obj/machinery/computer/computer in at)
+		if(computer.dir == dir && computer.connectable)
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/connecting_computer/code/_computer.dm
```diff
@@ -11,6 +11,18 @@
 /obj/machinery/computer/proc/callback_proc_issue()
 	update_overlays()
 
+/**
+ * Find and return a computer at this location that can connect to us.
+ * Arguments:
+ * * at - the location to look.
+ */
+/obj/machinery/computer/proc/find_connectable_computer(turf/at)
```
tf-4:
```suggestion
/obj/machinery/computer/proc/find_connectable_computer(turf/adjacent_turf)
```

tf-4 (changes requested):

modular_skyrat/modules/connecting_computer/code/_computer.dm
Tastyfish: done

tf-4 (approved):

Funce (approved):

PR #15193, opened 2022-07-26
[NON-MODULAR] Turrets Don't Shoot Megafauna!

RimiNosha:
## About The Pull Request
Turns out turrets shoot megafauna by default, which is causing somewhat regular issues on the Interdyne, so let's make needless megafauna aggroing a thing of the past!

Having the Interdyne get leveled on round start isn't great!

## Changelog

:cl:
code: Megafauna are now no longer targeted by turrets.
/:cl:


GuiltyNeko:
This has certain implications for the literally-never-used aux base and aux base turrets.

RimiNosha:
I see that as a non-issue, the base is getting leveled, turrets or not should a megafauna be nearby and someone isn't careful!

Gandalf2k15:
This is a bit of a bandaid fix, there should just be a larger safe area if it's really a big issue. I don't like how turrets now won't target all mobs they should.

PR #15195, opened 2022-07-26
[Modular] Fixes being able to clear temp flavor text

### LINKED ISSUE #15163: Nonremovable Temp Flavor
You cant remove temporary flavor fully.
image
For example this is my temp flavor. If I try to delete it(set nothing to temp flavor text), nothing changes:
image
Writing a spacebar leaves empty "They look different than usual":
image

Tastyfish:
## About The Pull Request

Fixes #15163

A reminder that `""` isn't truthy.
Also makes it use `tgui_input_text()` as this a player verb. Note that `tgui_input_text` has HTML sanitation built in.

bugfix, code cleanup

## Changelog

:cl:
fix: Set Temporary Flavor Text now lets you clear out the temporary flavor text.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/mob/living/living_verbs.dm
+++ b/modular_skyrat/modules/customization/modules/mob/living/living_verbs.dm
@@ -7,10 +7,9 @@
 		to_chat(usr, span_warning("You can't set your temporary flavor text now..."))
 		return
 
-	var/msg = input(usr, "Set the temporary flavor text in your 'examine' verb. This is for describing what people can tell by looking at your character.", "Temporary Flavor Text", temporary_flavor_text) as message|null
-	if(msg)
-		if(msg == "")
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (changes requested):
I think the better fix here would be to set that text as a placeholder in that text box, instead of being text that's written in there by default. It would make a lot more sense to new players and old players alike.

Tastyfish:
> I think the better fix here would be to set that text as a placeholder in that text box, instead of being text that's written in there by default. It would make a lot more sense to new players and old players alike.

@GoldenAlpharex  Setting their previous temporary flavor text as placeholder text?

Edit: The clarification is somewhat academic, as tgui_input_text doesn't support placeholder text.

Gandalf2k15 (approved):

PR #15198, opened 2022-07-26
[MODULAR] Makes CI sound shorter range.

Higgin:
## About The Pull Request

Adjusts CI sound range from a max of 17 to 10 tiles. Sets CI volume falloff to start at 1 tiles distance from the source and drop off by every tile beyond that taken to the power of four (see graphs.) 

Original function: audible out to 10 tiles pretty easily, as far as 17.
image

New function: roughly only as audible at 8 tiles as 11 previously, hard cutoff at 11. 
image

Because we have about seven tiles up/down and ten east/west, this will still potentially mean being heard off-screen if people are to the north and south of you. Still, it's a lot less and bound to be a lot less audible than what we've had for a while now. 

Know what's obnoxious especially to anyone trying to use a subtle approach in their antagonism (GONE WRONG) or getting into a scuffle? CI serving as a clarion call from a mile and a half away. It's meant to indicate mechanical things are about to start happening to people involved in a scene at an interactive distance, not to everybody on the block and down the road. This helps fix that. Let the combat sounds if any speak for themselves.

## Changelog

:cl:
balance: CI sounds no longer carry nearly so far and drop off much closer to a screen's distance.
/:cl:


Iamgoofball:
:+1: hearing CI from like 3 screens off kinda sucked yeah

PR #15199, opened 2022-07-26
[MODULAR] Better Subtle and Subtler!

RimiNosha:
## About The Pull Request
This PR does a few things:
- Gets rid of that unnecessary "visible or audible" prompt.
- Changes some internal logic to be less jank.
- Allows you to pick the same tile, or a specific user to send your subtler to!

Better QoL for those trying to emote with blind characters, cause you're not gonna accidentally do a visible message!

Also allows for better emote secrecy, cause you can now better control who sees your emote!

Oh, and it does all the `can_emote` checks **before** anything else is done, cause typing out your emote to just be told you can't emote is *dumb*.

<details>
<summary>Image!</summary>
<image src="https://user-images.githubusercontent.com/106692773/181109339-92ef9868-b362-4042-a804-f95038ce5258.png">
</details>

## Changelog

:cl:
qol: Subtle is now less tedious to use, and subtler now has new target options.
fix: Subtle and subtler now have their can_emote checks before you're given any prompts as well.
/:cl:

### DIFF
--- a/modular_skyrat/modules/verbs/code/subtle.dm
+++ b/modular_skyrat/modules/verbs/code/subtle.dm
@@ -52,7 +52,7 @@
 	var/list/viewers = get_hearers_in_view(SUBTLE_DEFAULT_DISTANCE, user)
 
 	for(var/mob/ghost in GLOB.dead_mob_list)
-		if(ghost.stat == DEAD && (ghost.client?.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghost in viewers))
+		if((ghost.client?.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghost in viewers))
 			ghost.show_message(subtle_message)
 
 	for(var/mob/reciever in viewers)
--- a/modular_skyrat/modules/verbs/code/subtle.dm
+++ b/modular_skyrat/modules/verbs/code/subtle.dm
@@ -1,6 +1,9 @@
 #define SUBTLE_DEFAULT_DISTANCE 1
 #define SUBTLE_SAME_TILE_DISTANCE 0
 
+#define SUBTLE_ONE_TILE_TEXT "1-Tile Range"
+#define SUBTLE_SAME_TILE_TEXT "Same Tile"
+
 /datum/emote/living/subtle
 	key = "subtle"
 	key_third_person = "subtle"
@@ -39,7 +42,7 @@
 			emote_type = type_override
 
 	if(!can_run_emote(user))
-		to_chat(user, "You can't emote at this time.", type = MESSAGE_TYPE_WARNING)
+		to_chat(user, span_warning("You can't emote at this time."))
 		return FALSE
 
 	var/prefix_log_message = "(SUBTLE) [subtle_message]"
@@ -81,16 +84,16 @@
 
 /datum/emote/living/subtler/run_emote(mob/user, params, type_override = null)
 	if(!can_run_emote(user))
-		to_chat(user, "You can't emote at this time.", type = MESSAGE_TYPE_WARNING)
+		to_chat(user, span_warning("You can't emote at this time."))
 		return FALSE
 	var/subtler_message
 	var/subtler_emote = params
 	var/mob/target
 	if(is_banned_from(user, "emote"))
-		to_chat(user, "You cannot send subtle emotes (banned).", type = MESSAGE_TYPE_WARNING)
+		to_chat(user, span_warning("You cannot send subtle emotes (banned)."))
 		return FALSE
 	else if(user.client && user.client.prefs.muted & MUTE_IC)
-		to_chat(user, "You cannot send IC messages (muted).", type = MESSAGE_TYPE_WARNING)
+		to_chat(user, span_warning("You cannot send IC messages (muted)."))
 		return FALSE
 	else if(!subtler_emote)
 		subtler_emote = tgui_input_text(user, "Choose an emote to display.", "Subtler" , null, MAX_MESSAGE_LEN, TRUE)
@@ -102,12 +105,12 @@
 			for(var/mob/mob_in_view as anything in in_view)
 				if(!istype(mob_in_view))
 					in_view.Remove(mob_in_view)
-			var/list/targets = list("1-Tile Range", "Same Tile") + in_view
+			var/list/targets = list(SUBTLE_ONE_TILE_TEXT, SUBTLE_SAME_TILE_TEXT) + in_view
 			target = tgui_input_list(user, "Pick a target", "Target Selection", targets)
 			switch(target)
-				if("1-Tile Range")
+				if(SUBTLE_ONE_TILE_TEXT)
 					target = SUBTLE_DEFAULT_DISTANCE
-				if("Same Tile")
+				if(SUBTLE_SAME_TILE_TEXT)
 					target = SUBTLE_SAME_TILE_DISTANCE
 			subtler_message = subtler_emote
 		else
@@ -119,7 +122,7 @@
 			emote_type = type_override
 
 	if(!can_run_emote(user))
-		to_chat(user, "You can't emote at this time.", type = MESSAGE_TYPE_WARNING)
+		to_chat(user, span_warning("You can't emote at this time."))
 		return FALSE
 
 	user.log_message(subtler_message, LOG_SUBTLER)
@@ -133,7 +136,7 @@
 		if(get_dist(user.loc, target.loc) <= SUBTLE_DEFAULT_DISTANCE)
 			target.show_message(subtler_message, alt_msg = subtler_message)
 		else
-			to_chat(user, "Your emote was unable to be sent to your target: Too far away.", type = MESSAGE_TYPE_WARNING)
+			to_chat(user, span_warning("Your emote was unable to be sent to your target: Too far away."))
 	else
 		var/ghostless = get_hearers_in_view(target, user) - GLOB.dead_mob_list
 		for(var/mob/reciever in ghostless)
@@ -173,3 +176,6 @@
 
 #undef SUBTLE_DEFAULT_DISTANCE
 #undef SUBTLE_SAME_TILE_DISTANCE
+
+#undef SUBTLE_ONE_TILE_TEXT
+#undef SUBTLE_SAME_TILE_TEXT
--- a/modular_skyrat/modules/verbs/code/subtle.dm
+++ b/modular_skyrat/modules/verbs/code/subtle.dm
@@ -28,7 +28,7 @@
 		to_chat(user, "You cannot send IC messages (muted).")
 		return FALSE
 	else if(!params)
-		subtle_emote = stripped_multiline_input(user, "Choose an emote to display.", "Subtle", null, MAX_MESSAGE_LEN)
+		subtle_emote = tgui_input_text(user, "Choose an emote to display.", "Subtle", null, MAX_MESSAGE_LEN, TRUE)
 		if(subtle_emote && !check_invalid(user, subtle_emote))
 			subtle_message = subtle_emote
 		else
@@ -93,7 +93,7 @@
 		to_chat(user, "You cannot send IC messages (muted).", type = MESSAGE_TYPE_WARNING)
 		return FALSE
 	else if(!subtler_emote)
-		subtler_emote = stripped_multiline_input(user, "Choose an emote to display.", "Subtler" , null, MAX_MESSAGE_LEN)
+		subtler_emote = tgui_input_text(user, "Choose an emote to display.", "Subtler" , null, MAX_MESSAGE_LEN, TRUE)
 		if(subtler_emote && !check_invalid(user, subtler_emote))
 			var/list/in_view = get_hearers_in_view(1, user)
 			in_view -= GLOB.dead_mob_list
--- a/modular_skyrat/modules/verbs/code/subtle.dm
+++ b/modular_skyrat/modules/verbs/code/subtle.dm
@@ -1,3 +1,6 @@
+#define SUBTLE_DEFAULT_DISTANCE 1
+#define SUBTLE_SAME_TILE_DISTANCE 0
+
 /datum/emote/living/subtle
 	key = "subtle"
 	key_third_person = "subtle"
@@ -14,7 +17,7 @@
 
 /datum/emote/living/subtle/run_emote(mob/user, params, type_override = null)
 	if(!can_run_emote(user))
-		user.show_message(span_warning("You can't emote at this time."))
+		to_chat(user, span_warning("You can't emote at this time."))
 		return FALSE
 	var/subtle_message
 	var/subtle_emote = params
@@ -35,6 +38,10 @@
 		if(type_override)
 			emote_type = type_override
 
+	if(!can_run_emote(user))
+		to_chat(user, type = MESSAGE_TYPE_WARNING, "You can't emote at this time.")
+		return FALSE
+
 	var/prefix_log_message = "(SUBTLE) [subtle_message]"
 	user.log_message(prefix_log_message, LOG_EMOTE)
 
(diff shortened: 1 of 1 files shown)


GuiltyNeko:
> Oh, and it does all the `can_emote` checks **before** anything else is done

And after too, right?

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -13,6 +13,9 @@
 	return FALSE
 
 /datum/emote/living/subtle/run_emote(mob/user, params, type_override = null)
+	if(!can_run_emote(user))
+		user.show_message(span_warning("You can't emote at this time."))
```
tf-4: Why `show_message` over `to_chat`?

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -51,17 +42,18 @@
 
 	subtle_message = span_emote("<b>[user]</b>[space]<i>[user.say_emphasis(subtle_message)]</i>")
 
-	for(var/mob/ghosts in GLOB.dead_mob_list)
-		if(!ghosts.client || isnewplayer(ghosts))
+	var/list/viewers = get_hearers_in_view(1, user)
+
+	for(var/mob/ghost in GLOB.dead_mob_list)
```
tf-4:
```suggestion
	for(var/mob/ghost as anything in GLOB.dead_mob_list)
```

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -51,17 +42,18 @@
...
+		if(!ghost.client || isnewplayer(ghost))
 			continue
```
tf-4:
```suggestion
```
We check for client right afterward and `dead_mob_list` already excludes `/mob/dead/new_player`

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -51,17 +42,18 @@
...
-		var/their_turf = get_turf(src)
-		if(ghosts.stat == DEAD && ghosts.client && (ghosts.client.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghosts in viewers(their_turf, null)))
-			ghosts.show_message(subtle_message)
+		if(ghost.stat == DEAD && ghost.client && (ghost.client.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghost in viewers))
```
tf-4:
```suggestion
		if(ghost.stat == DEAD && (ghost.client?.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghost in viewers))
```
can a ghost have a stat other than dead?

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -24,25 +27,13 @@
...
 			subtle_message = subtle_emote
 		else
 			return FALSE
 	else
 		subtle_message = params
 		if(type_override)
 			emote_type = type_override
-	. = TRUE
-	if(!can_run_emote(user))
-		return FALSE
```
GoldenAlpharex: That check still needs to be done before sending the emote, otherwise you could probably be dead and send a subtler. I know it sounds stupid, but it's still necessary.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
 			return FALSE
 	else
-		subtler_message = params
+		target = 1
+		subtler_message = subtler_emote
 		if(type_override)
 			emote_type = type_override
-	. = TRUE
-	if(!can_run_emote(user))
-		return FALSE
```
GoldenAlpharex: Ditto.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
+			for(var/mob/inviewmob in inview)
+				if(!istype(inviewmob))
+					inview.Remove(inviewmob)
+			var/list/targets = list("1-Tile Range", "Same Tile") + inview
+			target = input("Pick a target.") as null|anything in targets
+			switch(target)
+				if("1-Tile Range")
+					target = 1
+				if("Same Tile")
+					target = 0
```
GoldenAlpharex: I don't like the fact that you're using `target` both as `target_player` or `target_distance`. That's relatively misleading.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
-			var/type = input("Is this a visible or hearable emote?") as null|anything in list("Visible", "Hearable")
-			switch(type)
-				if("Visible")
-					emote_type = EMOTE_VISIBLE
-				if("Hearable")
-					emote_type = EMOTE_AUDIBLE
-				else
-					alert("Unable to use this emote, must be either hearable or visible.")
-					return
+			var/list/inview = get_hearers_in_view(1, user)
```
GoldenAlpharex: The `1` should be a define, like `SUBTLER_MAX_DISTANCE`
```suggestion
			var/list/in_view = get_hearers_in_view(1, user)
```

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
+			inview -= GLOB.dead_mob_list
+			inview.Remove(user)
+
+			for(var/mob/inviewmob in inview)
```
GoldenAlpharex: You'll have to do a few of them manually, but, please keep in mind to use `snake_case` when typing out your variable names. I feel like I've told you about it in a few reviews by now.
```suggestion
			for(var/mob/in_view_mob in in_view)
```

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
+				if(!istype(inviewmob))
+					inview.Remove(inviewmob)
+			var/list/targets = list("1-Tile Range", "Same Tile") + inview
+			target = input("Pick a target.") as null|anything in targets
```
tf-4:
```suggestion
			target = tgui_input_list(user, "Pick a target", "Target Selection", targets)
```
Use TGUI for player-facing menus (outside of very specific instances)

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,47 +75,63 @@
...
+	if(istype(target))
+		user.show_message(subtler_message, alt_msg = subtler_message)
+		if(get_dist(user.loc, target.loc) <= 1)
+			target.show_message(subtler_message, alt_msg = subtler_message)
+		else
+			user.show_message(span_warning("Your emote was unable to be sent to your target: Too far away."))
 	else
-		user.visible_message(message = subtler_message, self_message = subtler_message, vision_distance = 1, ignored_mobs = GLOB.dead_mob_list, separation = space)
+		var/ghostless = get_hearers_in_view(target, user)
+		ghostless -= GLOB.dead_mob_list
```
GoldenAlpharex:
```suggestion
		var/ghostless = get_hearers_in_view(target, user) - GLOB.dead_mob_list
```

modular_skyrat/modules/verbs/code/subtle.dm
tf-4:
```suggestion
			for(var/mob/in_view_mob in inview)
```
```suggestion
			for(var/mob/viewable_mob in inview)
```
Either or, or something else. But this isn't that easily readable.

GoldenAlpharex (changes requested):

modular_skyrat/modules/verbs/code/subtle.dm
tf-4: This isn't immediately clear what it's doing, it looks like it's being set to `TRUE` or `FALSE`. See if you can find a way to make it a bit clearer

tf-4 (changes requested):

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Ack-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Old code that I forgot to remove, woops!

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Not sure, but it was in the original code-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Mmm, why the anything there? That will cause runtimes if the mob is somehow null or isn't *actually* a mob-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Cause there's 10 different ways to send a message to a player, and I can never remember the right one.

modular_skyrat/modules/verbs/code/subtle.dm
tf-4: it's a list of mobs, it can't not be a mob

modular_skyrat/modules/verbs/code/subtle.dm
tf-4: With how some of that old code was looking, that's by no means a guarantee of correctness lol. Worth a look into

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -124,10 +133,9 @@
 		if(get_dist(user.loc, target.loc) <= 1)
 			target.show_message(subtler_message, alt_msg = subtler_message)
 		else
-			user.show_message(span_warning("Your emote was unable to be sent to your target: Too far away."))
+			to_chat(user, type = MESSAGE_TYPE_WARNING, "Your emote was unable to be sent to your target: Too far away.")
```
tf-4:
```suggestion
			to_chat(user, "Your emote was unable to be sent to your target: Too far away.",  type = MESSAGE_TYPE_WARNING)
```
named arguments go after ordered arguments

modular_skyrat/modules/verbs/code/subtle.dm
Iamgoofball: if there are non-mobs in the dead_mob_list we have bigger issues

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Uhhh-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: I used defines in place of numbers, hopefully that'll be a little easier to follow!

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Stat is slated to be moved from `mob` to `mob/living`, so I'm removing the check anyways!

theselfish:
Does this add a bubble at least?

As, subtling w/o popping a bubble up causes various "LOOC: Still typing?"

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -22,26 +28,18 @@
...
 		else
 			return FALSE
 	else
 		subtle_message = params
 		if(type_override)
 			emote_type = type_override
-	. = TRUE
+
 	if(!can_run_emote(user))
+		to_chat(user, "You can't emote at this time.", type = MESSAGE_TYPE_WARNING)
```
GoldenAlpharex: Why did you do both of them differently? The other one used just a span, and this one uses the `type` parameter? Just use one and stick to it, for coherence's sake.

modular_skyrat/modules/verbs/code/subtle.dm
GoldenAlpharex: The reason we recommend using `as anything` is that BYOND won't run an implicit typecheck over each entry in that list, which means that it's going to iterate over them much faster. In this case, I think it'd be warranted for it to runtime if we have non-mobs in the `dead_mob_list`, because that's _absolutely_ not meant to happen.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,35 +80,46 @@
 	return FALSE
 
 /datum/emote/living/subtler/run_emote(mob/user, params, type_override = null)
+	if(!can_run_emote(user))
+		to_chat(user, "You can't emote at this time.", type = MESSAGE_TYPE_WARNING)
```
GoldenAlpharex: I really don't know what's the difference between using the `type` parameter and just using a span like is usually done.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,35 +80,46 @@
...
-				else
-					alert("Unable to use this emote, must be either hearable or visible.")
-					return
+			var/list/in_view = get_hearers_in_view(1, user)
+			in_view -= GLOB.dead_mob_list
+			in_view.Remove(user)
+
+			for(var/mob/inviewmob in in_view)
+				if(!istype(inviewmob))
+					in_view.Remove(inviewmob)
```
GoldenAlpharex: That `!istype()` check is useless if you're not using `as anything`. This is the proper way to do what you wanted to do.
```suggestion
			for(var/mob/mob_in_view as anything in in_view)
				if(!istype(mob_in_view))
					in_view.Remove(mob_in_view)
```

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -83,35 +80,46 @@
...
+			var/list/targets = list("1-Tile Range", "Same Tile") + in_view
+			target = tgui_input_list(user, "Pick a target", "Target Selection", targets)
+			switch(target)
+				if("1-Tile Range")
+					target = SUBTLE_DEFAULT_DISTANCE
+				if("Same Tile")
+					target = SUBTLE_SAME_TILE_DISTANCE
```
GoldenAlpharex: Those string options should be defines, since you're using them in more than one spot, so that it's easier to modify that choice in the future, if it's ever necessary. Minor code quality things.

modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -120,10 +128,18 @@
...
+			target.show_message(subtler_message, alt_msg = subtler_message)
+		else
+			to_chat(user, "Your emote was unable to be sent to your target: Too far away.", type = MESSAGE_TYPE_WARNING)
 	else
-		user.visible_message(message = subtler_message, self_message = subtler_message, vision_distance = 1, ignored_mobs = GLOB.dead_mob_list, separation = space)
+		var/ghostless = get_hearers_in_view(target, user) - GLOB.dead_mob_list
+		for(var/mob/reciever in ghostless)
+			reciever.show_message(subtler_message, alt_msg = subtler_message)
+
+	return TRUE
```
GoldenAlpharex, line 145: What's the purpose of returning `TRUE` here?

GoldenAlpharex (changes requested):

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Ahhh, that makes sense, I'm just used to strictly typed everything, so-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: It already returned true before, and I wasn't about to dive into emote code at the time to figure out why.

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Uhhh, that was one I missed, oops-

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Alright.

modular_skyrat/modules/verbs/code/subtle.dm
RimiNosha: Yea, emote code returns true on a successful emote.

Gandalf2k15 (approved):

PR #15223, opened 2022-07-27
[MIRROR] Fixes some alien actions not working.  [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68625
--------------------

## About The Pull Request

Fixes #68623

#68382 added a cooldown to the start of PreActivate for all cooldown actions by default

Space did this so people implementing new actions wouldn't forget to start their cooldowns

But, this caused problems for actions that sleep (user input), that sanity check availability afterwards

(Basically, actions that ask for user input would call `IsAvailable` to check if the action is still usable after waiting, but because it was put on a really long cooldown, it'd never be available. This wasn't a problem before because it didn't go on the long cooldown, but when the behavior was made generic it became an issue.)

I talked with Space a bit, and we decided to just revert it for the time being, allowing actions to handle cooldowns independently.  

Before: It starts a long generic cooldown, to prevent people from using the button while the action is being ran, and the normal cooldown is triggered in `Activate()`
After: There is no generic long cooldown, if an action wants it they can implement it themselves

I want to improve this to be a bit clearer in the future, but for now I'm just getting this PR up so aliens function again

Should probably be unit tested, idk.

## Why It's Good For The Game

Aliens work

## Changelog

:cl: Melbert
fix: Fixes some alien actions like Evolve not working. 
/:cl:


GoldenAlpharex (approved):
Should work just fine now.

tf-4 (approved):

PR #15224, opened 2022-07-27
[DNM][TM ONLY] pr containing code for an event later, ignore this pr if you don't like spoilers [MDB IGNORE]

Paxilmaniac:
## About The Pull Request
don't say I didn't warn you with the title, it's not too late to not look at the files changed

I realized an event i wanted to run would require some code to do, so instead of just not doing it i decided to do it anyways. This is actually pretty safe to testmerge on rounds the event isn't running, only because the changes will never be seen by players unless the admins force it, EXCEPT for the fact I've removed the latejoin spawners from the interlink, so everyone will latejoin on the interlink shuttle instead. I still don't suggest you testmerge this unless I ask you to though.

## Changelog

I don't think I need one here


Paxilmaniac:
while the map i edited may be failing checks, they are failing for reasons that do not effect the gameplay, thus i do not care

modular_skyrat/modules/mars_event/code/terrain.dm
```diff
@@ -0,0 +1,57 @@
...
+	color = RED_ROCK
+	mineralSpawnChanceList = list(
+		/obj/item/stack/ore/uranium = 5,
+		/obj/item/stack/ore/diamond = 1,
+		/obj/item/stack/ore/gold = 10,
+		/obj/item/stack/ore/silver = 12,
+		/obj/item/stack/ore/plasma = 20,
+		/obj/item/stack/ore/iron = 40,
+		/obj/item/stack/ore/titanium = 11,
+		/obj/item/stack/ore/bluespace_crystal = 1
```
Iamgoofball, line 50: no bananium?

Paxilmaniac: this is default ore spawn with gibtonite removed so yes, no bananium

Paxilmaniac:
ill see if i can get this run sometime earlier in the day for other people, then it'll get closed

Paxilmaniac:
im sorry for all the poor maintainers that this is pinging, i don't know if you can testmerge drafts and im not gonna find out

Paxilmaniac:
That last one was a disaster, at least I get to steal the outfit subsystem code to bring upstream

PR #15240, opened 2022-07-28
[NONMODULAR] Voracious changes

Superlagg:
## About The Pull Request

People with the voracious trait actually eat faster now, and are able to eat even while they are full, albeit much more slowly. This also means voracious people can be force fed faster, and it bypasses the overfed delay.

Voracious doesnt do what it says it does, currently just letting you eat more junk food and get a mood boost when full. This fixes it to have more of an impact on the game.

## Changelog

:cl:
add: Voracious properly speeds up eating. It also allows you to eat and be fed even while you're full, albeit more slowly the fulller you are. 
/:cl:


RimiNosha:
Shouldn't this really be upstream, unless I'm missing something?

PR #15249, opened 2022-07-28
Remove Nanotrasen Rep weapon set.

Gladi-Writes:
## About The Pull Request

Note: SRQ/Gladi on discord. Chiara Nowak ingame.

Proposed change removes the NTR weapon due to:
Over-saturation of weapons when the NTR and Blueshield are around making some antagonism nearly, if not literally, impossible.
"Greenshield" Reps that due to their access to a weapon feel overly obliged to utilize force in a non-fitting way in situations.

The fact that central would not expect what is effectively a paperwork jockey oversight person to be any real target or have any reason to doubt their safety.

The representative can still of course purchase their own weapon, or borrow one from security in an emergency. I think this is more fitting for playing the role how it's meant to be, and fits with the in-character idea of this role.

## Changelog

🆑 
remove: NTR weapon-set.
/🆑

### DIFF
--- a/modular_skyrat/modules/nanotrasen_rep/code/nanotrasen_consultant.dm
+++ b/modular_skyrat/modules/nanotrasen_rep/code/nanotrasen_consultant.dm
@@ -54,7 +54,7 @@
 	head = /obj/item/clothing/head/nanotrasen_consultant
 	backpack_contents = list(
 		/obj/item/melee/baton/telescopic = 1,
-		)
+	)
 
 	skillchips = list(/obj/item/skillchip/disk_verifier)
 
--- a/modular_skyrat/modules/nanotrasen_rep/code/nanotrasen_consultant.dm
+++ b/modular_skyrat/modules/nanotrasen_rep/code/nanotrasen_consultant.dm
@@ -54,7 +54,6 @@
 	head = /obj/item/clothing/head/nanotrasen_consultant
 	backpack_contents = list(
 		/obj/item/melee/baton/telescopic = 1,
-		/obj/item/storage/box/gunset/nanotrasen_consultant = 1,
 		)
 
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/nanotrasen_rep/code/nanotrasen_consultant.dm
```diff
@@ -54,7 +54,6 @@
 	head = /obj/item/clothing/head/nanotrasen_consultant
 	backpack_contents = list(
 		/obj/item/melee/baton/telescopic = 1,
-		/obj/item/storage/box/gunset/nanotrasen_consultant = 1,
 		)
```
GoldenAlpharex: Just because it bothers me.
```suggestion
	)
```

GoldenAlpharex (changes requested):

tf-4 (approved):

HoodMilk:
yeah this makes sense

theselfish:
Wouldn't a one-use hold out weapon fit the Rep more than *no* gun?

Like a Derringer?

TetrakoHateThis:
By the same logic, shouldn't HoP also have their Osprey removed?

tf-4:
> By the same logic, shouldn't HoP also have their Osprey removed?

probably but out of scope

Gandalf2k15:
they're quite important to nanotrasen, a corporate liaison if you will

SingingSpock:
This makes no sense. Firstly, as Gandalf said, ICly the NT rep is a valuable individual worth protecting. Secondly, on an OOC level the NTR has a CC ID with ID management access. That means that taking out the rep and getting their card is an easy way for anyone to get AA if the rep is unable to protect themselves. Every role that has ID edit access has a weapon and for good reason.

Gandalf2k15:
see above

PR #15259, opened 2022-07-29
[MODULAR] Actually Fix Invisible Selected Languages

RimiNosha:
## About The Pull Request
I hate prefs code I hate prefs code I hate prefs code

Actually fixes a bug.

## Changelog

:cl:
fix: Actually fix invisible selected languages.
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -113,10 +113,11 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/loaded_preferences_successfully = load_preferences()
 	if(loaded_preferences_successfully)
 		if(load_character())
-			// Sanitize languages.
+			// SKYRAT EDIT START - Sanitizing languages
 			for(var/datum/language/lang_path as anything in languages)
 				if(initial(lang_path.secret))
 					languages.Remove(lang_path)
+			// SKYRAT EDIT END
 	//we couldn't load character data so just randomize the character appearance + name
 	randomise_appearance_prefs() //let's create a random character then - rather than a fat, bald and naked man.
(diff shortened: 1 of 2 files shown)


code/modules/client/preferences.dm
```diff
@@ -113,7 +113,10 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/loaded_preferences_successfully = load_preferences()
 	if(loaded_preferences_successfully)
 		if(load_character())
-			return
+			// Sanitize languages.
+			for(var/datum/language/lang_path as anything in languages)
+				if(initial(lang_path.secret))
+					languages.Remove(lang_path)
```
GoldenAlpharex:
```suggestion
			// SKYRAT EDIT START - Sanitizing languages
			for(var/datum/language/lang_path as anything in languages)
				if(initial(lang_path.secret))
					languages.Remove(lang_path)
			// SKYRAT EDIT END
```

GoldenAlpharex: Also, we should figure out something where it makes sure that the languages you've got in there are valid languages still, as I'm almost certain the current errors with the language menu bugging out for some people is due to some languages getting renamed at some point or the other.
An `initial(lang_path.name) == default_language_name` check could probably do the trick for that too.

GoldenAlpharex (changes requested):

code/modules/client/preferences.dm
RimiNosha: `.name` is not used anywhere in logic, so this is a moot point!

It's all done via paths which are completely immutable unless I'm missing something.

PR #15282, opened 2022-07-29
[FUCK] [NO GBP] Fix Ghost Pods Speaking Over Common and Showing Ghost Roles on Manifest!

RimiNosha:
## About The Pull Request
WHYDIDITBREAKLIKETHISITWORKEDFINEONMYPCASIDEFROMTHEMANIFESTTHINGAAAAAAA

Anyways, it's fixed now!

Bugfix.

## Changelog

:cl:
fix: Fixed ghost cryopods and cryo consoles, I swear I know how to code.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cryosleep/code/cryopod.dm
+++ b/modular_skyrat/modules/cryosleep/code/cryopod.dm
@@ -9,6 +9,8 @@
  */
 GLOBAL_LIST_EMPTY(cryopod_computers)
 
+GLOBAL_LIST_EMPTY(ghost_records)
+
 //Main cryopod console.
 
 /obj/machinery/computer/cryopod
@@ -31,7 +33,7 @@ GLOBAL_LIST_EMPTY(cryopod_computers)
 
 	/// This is what the announcement system uses to make announcements. Make sure to set a radio that has the channel you want to broadcast on.
 	var/obj/item/radio/headset/radio = /obj/item/radio/headset/silicon/pai
-	/// The channel to be broadcast on, valid values are any of the "RADIO_CHANNEL_" defines.
+	/// The channel to be broadcast on, valid values are the values of any of the "RADIO_CHANNEL_" defines.
 	var/announcement_channel = null // RADIO_CHANNEL_COMMON doesn't work here.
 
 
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -304,8 +306,14 @@ GLOBAL_LIST_EMPTY(cryopod_computers)
...
 		crew_member["job"] = "N/A"
 
-	// Delete them from datacore.
+	// Delete them from datacore and ghost records.
 	var/announce_rank = null
+
+	for(var/datum/data/record/record as anything in GLOB.ghost_records)
+		if(record.fields["name"] == mob_occupant.real_name)
+			announce_rank = record.fields["rank"]
+			qdel(record)
```
GoldenAlpharex: This way you won't have nulls in that list. (I think that's how the remove proc is written).
```suggestion
			announce_rank = record.fields["rank"]
			GLOB.ghost_records.Remove(record)
			qdel(record)
```

modular_skyrat/modules/mapping/code/mob_spawns.dm
```diff
@@ -501,10 +501,10 @@
 //CRYO CONSOLES
 /obj/machinery/computer/cryopod/interdyne
 	radio = /obj/item/radio/headset/interdyne
-	announcement_channel = RADIO_CHANNEL_INTERDYNE
+	announcement_channel = "Interdyne"
```
GoldenAlpharex: Why can't those be defines? If the content of the define wasn't correct, just make a new one with the proper name in it.

GoldenAlpharex (changes requested):

modular_skyrat/modules/mapping/code/mob_spawns.dm
RimiNosha: Announcement channel was set to null on live, and this is the only reason why I can think of, and it works on my machine<sup>:tm:</sup>, so yea...

Zonespace27: on live it is now not null, instead it's not announcing it at all

PR #15315, opened 2022-07-30
walk through curtains

jjpark-kb:
## About The Pull Request
makes curtains walk-through-able, even if they are closed.

They are curtains, not some huge door. They should be able to be walked through as the fabric (cloth or plastic, etc.) they are.

## Changelog

:cl:
qol: you can walk through closed curtains now
/:cl:

### DIFF
--- a/modular_skyrat/modules/walk_curtains/curtains.dm
+++ b/modular_skyrat/modules/walk_curtains/curtains.dm
@@ -1,0 +1,21 @@
+/obj/structure/curtain
+	/// whether curtains will toggle their density when toggled
+	var/toggle_density = FALSE
+
+/obj/structure/curtain/toggle()
+	open = !open
+	if(open)
+		layer = SIGN_LAYER
+		plane = GAME_PLANE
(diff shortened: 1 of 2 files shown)


HoodMilk:
Restricting light but not movement is nice, I think all of service will appreciate being able to do this,

theselfish:
Upstream?

modular_skyrat/modules/walk_curtains/curtains.dm
```diff
@@ -0,0 +1,21 @@
+/obj/structure/curtain
+	/// whether curtains will toggle their density when toggled
+	var/toggle_density = FALSE
+
+/obj/structure/curtain/toggle()
```
Iamgoofball, line 5: If we're gonna override a proc like this can we add a comment over in the original define of this so that people know to find the override here on the offchance we need to adjust this down the line?

jjpark-kb:
someone said they'd rather this upstream... someone can do that.

PR #15316, opened 2022-07-31
Moves sweaters to GAGS

tf-4:
## About The Pull Request

Moves the various sweaters to GAGS. Fun times.
Removes unused variants, black and red are used on Blueshift so they stay.
Renames "cream sweater" to just... "sweater". Obviously. The default is still cream though.

Moves the keyhole sweater away from a sweater subtype. Not sure why it even was one to start with really but here we are. I'll make it GAGS too, soon:tm:.

sweater

## Changelog
:cl:
code: Sweaters are now colourable through GAGS
/:cl:


RatFromTheJungle:
mak pickture more bigly I cannot see!!!

PR #15325, opened 2022-07-31
Nerfs borer HP

tf-4:
## About The Pull Request

Halves borer starting HP
Drops HP per level from 10 to 5

Borers are already extremely difficult to hit given their tiny sprite size, average speed, and ability to pass through mobs and instantly walk through and hide under tables or other furnitures.
Problem is, even if you DO hit them you barely do any damage to them if they've existed for more than fifteen minutes.
They should be able to evade hits OR be tanky, not both, and being evasive makes far more sense.

## Changelog
:cl:
balance: Borers have less HP, and gain less per level
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
@@ -212,7 +212,7 @@
 		to_chat(owner, span_warning("You do not have any upgrade points for stats!"))
 		return
 	cortical_owner.stat_evolution--
-	cortical_owner.maxHealth += 3
+	cortical_owner.maxHealth += 5
 	cortical_owner.health_regen += 0.02
 	cortical_owner.max_chemical_storage += 20
 	cortical_owner.chemical_regen++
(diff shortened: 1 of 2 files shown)


Zonespace27:
Feeling a bit 😕 about this one, it's very common when people hear about borers, for people to get medhuds so they can pull borers from under tables, and it's generally a death sentence to get ejected from a host, the only thing able to save that is the health gained

tf-4:
how are you meant to pull them out from under, they can just move the instant you grab them. The context menu is also unwieldy and the alt-click menu is no better (if it even functions)

You can still get a fine amount of HP, it's just no longer overbearing.
And from what I've seen it's *far* from a death sentence to be ejected. Yes, it's not *great*, but the speed & tiny sprite size & slipping under tables means that it's very difficult to even be able to click on them.

Borers SHOULD be somewhat fucked outside of a host, their entire design is them being powerful with a host. They shouldn't be as strong as they are in a host AND super hard to kill outside of one.

jjpark-kb:
okay, not disagreeing with a nerf, but maybe not down to 3? could we try 6 or something? Also, you can target the hud itself.

modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
```diff
@@ -212,7 +212,7 @@
 		to_chat(owner, span_warning("You do not have any upgrade points for stats!"))
 		return
 	cortical_owner.stat_evolution--
-	cortical_owner.maxHealth += 10
+	cortical_owner.maxHealth += 3
```
tf-4:
```suggestion
	cortical_owner.maxHealth += 5
```

PR #10823, opened 2022-01-19
Database revision revert

Gandalf2k15:
It reverts database code

PR #15330, opened 2022-08-01
Optimizes SQL logging and adds missing related default config entries

GoldenAlpharex:
## About The Pull Request
It took us months to realize why https://github.com/Skyrat-SS13/Skyrat-tg/pull/10823 was needed. But today, I figured it out, thanks to Useroth's flash of genius: 
The culprit of all the lag was some code which wasn't enabled by default for anyone, but that we had turned on for our internal log-viewing tool, that logs nearly all of the logs over the SQL database. The problem here is that it's basically holding back every form of logging because it's not properly async, which obviously is quite bad.

So here's my attempt at fixing that:
Queries done for logging purposes will now be batched up in groups of up to 100 (by default, it's a config that you can change, but it shouldn't be lowered too low if you don't want to experience the same kind of lag as we did for the longest time). This way, it should reduce the amount of queries made by a factor of 100, which surely will be beneficial for performances. Furthermore, those bigger queries are now async, which means that they no longer hold up the procs that call them, which means, in short: No more code being held back by the SQL logging! Great!

~~Do note that I currently haven't set up any code to handle logs being effectively lost when the database subsystem is shut down while it still has items in its mass insertion queue, I'll spend more time doing that once I've tested to see that this works on our test server.~~ It's now been added and was fully tested. It works!

It gets rid of our oldest test-merge without compromising on our private log-viewing tool. I'd call that a win-win?

Closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/10823.

## Changelog

:cl: GoldenAlpharex
refactor: Refactored the SQL-based logging to batch queries together and makes those bigger queries asynchronous, which will massively reduce their impact on performance, and negate any perceptible unusual delays when SQL-based logging is enabled in config.
config: Added missing config entries in their own default config file on the repository, as they were currently missing.
config: Added a new config to control how many log entries will be bundled up together in a singular query to the database.
/:cl:

### DIFF
--- a/code/__HELPERS/logging/_logging.dm
+++ b/code/__HELPERS/logging/_logging.dm
@@ -76,14 +76,17 @@ GLOBAL_LIST_INIT(testing_global_profiler, list("_PROFILE_NAME" = "Global"))
 	//SKYRAT EDIT ADDITION BEGIN
 	#ifndef SPACEMAN_DMM
 	if(CONFIG_GET(flag/sql_game_log) && CONFIG_GET(flag/sql_enabled))
-		var/datum/db_query/query_sql_log_messages = SSdbcore.NewQuery({"
-			INSERT INTO [format_table_name("game_log")] (datetime, round_id, ckey, loc, type, message)
-			VALUES (:time, :round_id, :ckey, :loc, :type, :message)
-		"}, list("time" = SQLtime(), "round_id" = "[GLOB.round_id]", "ckey" = key_name(src), "loc" = loc_name(src), type = message_type, "message" = message))
-		if(!query_sql_log_messages.warn_execute())
-			qdel(query_sql_log_messages)
-			return
-		qdel(query_sql_log_messages)
+		SSdbcore.add_log_to_mass_insert_queue(
+			format_table_name("game_log"),
+			list(
+				"time" = SQLtime(),
+				"round_id" = "[GLOB.round_id]",
+				"ckey" = key_name(src),
+				"loc" = loc_name(src),
+				"type" = message_type,
+				"message" = message,
+			)
+		)
 		if(!CONFIG_GET(flag/file_game_log))
 			return
 	#endif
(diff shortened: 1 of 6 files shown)


modular_skyrat/master_files/code/controllers/subsystem/dbcore.dm
```diff
@@ -38,5 +38,5 @@
 	if(length(queued_log_entries_by_table[table]) < CONFIG_GET(number/sql_game_log_min_bundle_size))
 		return
 
-	MassInsert(table, rows = queued_log_entries_by_table[table], duplicate_key = FALSE, ignore_errors = FALSE, delayed = FALSE, warn = FALSE, async = TRUE, special_columns = null)
+	INVOKE_ASYNC(src, .proc/MassInsert, table, /*rows =*/ queued_log_entries_by_table[table], /*duplicate_key =*/ FALSE, /*ignore_errors =*/ FALSE, /*delayed =*/ FALSE, /*warn =*/ FALSE, /*async =*/ TRUE, /*special_columns =*/ null)
```
tf-4, line 41: why comment it all out?

GoldenAlpharex: Because you can't actually use the variable names there, since it's a define and not a proc. I've left them there so it's easier to figure out which variable they represent, for the sake of the future, if a new variable gets added for whatever reason.

code/__HELPERS/logging/_logging.dm
```diff
@@ -76,14 +76,17 @@ GLOBAL_LIST_INIT(testing_global_profiler, list("_PROFILE_NAME" = "Global"))
...
+			format_table_name("game_log"),
+			list(
+				"time" = SQLtime(),
+				"round_id" = "[GLOB.round_id]",
+				"ckey" = key_name(src),
+				"loc" = loc_name(src),
+				"type" = message_type,
+				"message" = message,
+			)
+		)
```
jjpark-kb, line 89: nonmodular, please add //skyrat edit at the very least.

jjpark-kb: nevermind???

PR #15332, opened 2022-08-01
[MDB IGNORE] [MODULAR] Makes automapper work with Centcom

Tastyfish:
## About The Pull Request

For the sake of @Jolly-66, automapper now actually supports the burned in CentCom level.

This is a complex problem, because all of the CentCom atoms are loaded (but not initialized) before the master controller is. The blacklisting mechanism is irrelevant due to this.

Thus, a hole in the spot has to be manually made. This is tough because you can't qdel() things before they're initialized, however waiting until after `SSatoms` has initialized the atoms creates a huge mess (especially in regards to the shuttle). Thus, the spots for CC templates have to be directly initialized and then torn down via arcane incantations.

Yes, I tested the escape shuttle.

Note for the sake of people doing automapper config things, templates for the central command z level should have their `required_map` set to the magic name "builtin". I imagine people just copy and paste entries so that should be straight forward.

**Notes**:
- I'm not 100% sure if those two template chunks are fully up to date, I'll trust Jolly to figure that out.
- I also hard reset `CentCom.dmm` to upstream.

Maptainability

## Changelog

No player or admin-facing changes.

### DIFF
--- a/_maps/skyrat/automapper/automapper_config.toml
+++ b/_maps/skyrat/automapper/automapper_config.toml
@@ -26,8 +26,6 @@ required_map = "builtin"
 coordinates = [128, 79, 1]
 trait_name = "CentCom"
 
-# CC Stuff is commented out for the time being until Cobalt can fix its spawning behavior. - Jolly
-
 # METASTATION MAP TEMPLATES
 # Metastation Arrivals
 [templates.metastaton_arrivals]
--- a/_maps/skyrat/automapper/automapper_config.toml
+++ b/_maps/skyrat/automapper/automapper_config.toml
@@ -11,20 +11,20 @@
 
 # CENTRAL COMMAND MAP TEMPLATES
 # CC Shuttle Dock - This is where we drop the players off after the round ends
-# [templates.centcom_shuttle_dock]
-# map_files = ["templates.centcom_shuttle_dock.dmm"]
-# directory = "_maps/skyrat/automapper/templates/centcom/"
-# required_map = "CentCom.dmm"
-# coordinates = [186, 71, 1]
-# trait_name = "CentCom"
+[templates.centcom_shuttle_dock]
+map_files = ["centcom_shuttle_dock.dmm"]
+directory = "_maps/skyrat/automapper/templates/centcom/"
+required_map = "builtin"
+coordinates = [186, 71, 1]
+trait_name = "CentCom"
 
 # CC Ferry Dock - Admin Shuttle
-# [templates.centcom_ferry_dock]
-# map_files = ["templates.centcom_ferry_dock.dmm"]
-# directory = "_maps/skyrat/automapper/templates/centcom/"
-# required_map = "CentCom.dmm"
-# coordinates = [128, 79, 1]
-# trait_name = "CentCom"
+[templates.centcom_ferry_dock]
+map_files = ["centcom_ferry_dock.dmm"]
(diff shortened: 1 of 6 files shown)


Tastyfish:
I'll figure out why create_and_destroy is finding exactly 1 broken epi pen or pepper spray bottle when I have time. Probably a single object spawning them in a nonstandard way before it dies.

_maps/skyrat/automapper/automapper_config.toml
```diff
@@ -11,20 +11,20 @@
...
-# coordinates = [128, 79, 1]
-# trait_name = "CentCom"
+[templates.centcom_ferry_dock]
+map_files = ["centcom_ferry_dock.dmm"]
+directory = "_maps/skyrat/automapper/templates/centcom/"
+required_map = "builtin"
+coordinates = [128, 79, 1]
+trait_name = "CentCom"
 
 # CC Stuff is commented out for the time being until Cobalt can fix its spawning behavior. - Jolly
```
Jolly-66: Don't forget to drop this comment.

Tastyfish:
Also fixed an issue with automapper where for multi-file stations (ie Blueshift), `load_templates_from_cache()` was called as many times as there were files but with the full list of all the files each time. Didn't matter before because Blueshift doesn't otherwise use automapper.

modular_skyrat/modules/automapper/code/automapper_subsystem.dm
```diff
@@ -70,12 +72,47 @@ SUBSYSTEM_DEF(automapper)
...
 			log_world("AUTOMAPPER: Successfully loaded map template [iterating_template.name] at [iterating_template.load_turf.x], [iterating_template.load_turf.y], [iterating_template.load_turf.z]!")
 
+/**
+ * CentCom atoms aren't initialized but already exist, so must be properly initialized and then qdel'd.
+ * Arguments:
+ * * parent - parent turf
+ */
+/datum/controller/subsystem/automapper/proc/init_contents(atom/parent)
+	var/static/list/mapload_args = list(TRUE)
+	// Don't even initialize things in this list. Very specific edge cases.
```
Gandalf2k15, line 92: why

Tastyfish: The things in that list either cause irreversible game state changes (in the case of the docking station and bookcase) or randomly spawn things into the void if deleted without the game ticking, for some reason.

Edit: They also happen to not runtime if deleted before being initialized. Could change in the future by random updates.

Gandalf2k15: I don't like it, this should be done in a way where the objects are just never loaded

Tastyfish: That is literally impossible. The CentCom objects are loaded before any code is run.

Gandalf2k15: I guess so, but this will probably cause issues down the line

Gandalf2k15 (approved):

modular_skyrat/modules/automapper/code/automapper_subsystem.dm
Tastyfish: I'll be honest, it might. This isn't fullproof by any means, but we can always temporarily cut automapper tson entries for CentCom if it's real bad.

PR #15334, opened 2022-08-01
Increases blob infection to 50 from 25 required pop

Zonespace27:
## About The Pull Request
Instead of admins calling in red-tier ERTs or outright denying blobs at lowpop, how about we just do this instead

Lowpop blobs aren't liked


SingingSpock:
Does this check players actively playing or just players logged in? I rarely see the actual server pop drop below 50 even in our lowpop, so if it's the latter then it likely won't have much effect.

Gandalf2k15:
This is changed in the json config, not like this

PR #15357, opened 2022-08-02
[MIRROR] Blargedy Blarg Blarg Blarg - Word Filter Error Fix [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68905
--------------------

## About The Pull Request

Hey there,

Apparently, even after five days of thinking about the JSON_DECODE issue we had going from rust-g 1.0.1 to 1.0.2 during #68690, I forgot to check our own code. I think I fudged something where I was using the global proc instead of the actual RUSTG define, so you would always get back an encoded JSON array, which would result in _always_ passing a null for the success variable that rust-g would return (based on what was seen on production), etc. The failure condition we were seeing is that no matter how perfect the word filter was, you would still crash to get this issue.

```txt
[2022-07-31 23:13:49.738] Loading config file word_filter.toml...
[2022-07-31 23:13:49.739] The word filter is not configured correctly! 
```

With a few "null spaces" after that message. So, I'm pretty sure this is the issue since I'm not seeing anything in my config_error.log with this change added. (however, I am not seeing anything on standard codebase conditions? this error was reported based on the production /tg/ server's config, but something is definitely yorked irregardless). Passing through this change, there doesn't appear to be anything in `config_error.log` on my local server, and intentionally fudging the TOML passes this:

image

Whoops.

## Why It's Good For The Game

I fucke dup

## Changelog

:cl:
admin: You should hopefully no longer see an error message that has nothing to say regarding the word filter.
/:cl:


code/controllers/configuration/configuration.dm
```diff
@@ -378,10 +378,16 @@ Example config:
...
+<<<<<<< HEAD
 
 	var/list/word_filter = rustg_read_toml_file("[directory]/word_filter.toml")
 	if (!islist(word_filter))
 		var/message = "The word filter configuration did not output a list, contact someone with configuration access to make sure it's setup properly."
+=======
+	var/list/result = rustg_raw_read_toml_file("[directory]/word_filter.toml")
+	if(!result["success"])
+		var/message = "The word filter is not configured correctly! [result["content"]]"
+>>>>>>> a1e716b60a7 (Blargedy Blarg Blarg Blarg - Word Filter Error Fix (#68905))
```
tf-4:
```suggestion
	var/list/result = rustg_raw_read_toml_file("[directory]/word_filter.toml")
	if(!result["success"])
		var/message = "The word filter is not configured correctly! [result["content"]]"
```

tf-4 (approved):

PR #15365, opened 2022-08-02
[MODULAR] Improves liquid input and output pumps

Tastyfish:
## About The Pull Request

This is something I started working on a while ago, and then put on the backburner to finish until now.

Makes the liquid input and output pumps to be underfloor mounted like air vent pumps and scrubbers, so you can use them as such.
- Combines their common code into a base class.
- The input pumps can drain out their tile and the 8 neighboring ones.
- I also made some mapper prefabs of them already set up and on for the sake of mappers.

Top is input, bottom is output.
Easy to remember because they're the same shape as the air ones, just smaller and a different shade.
ezgif com-gif-maker

### Quick example use
image

https://user-images.githubusercontent.com/1185434/182484982-e61b4eef-ed33-4b5e-b4ff-ebbd2206473d.mp4

More support for the turf liquid system.

## Changelog

:cl:
add: Liquid input and output pumps are now more like vent pumps and scrubbers, and can be more effectively used as such.
/:cl:


Gandalf2k15 (approved):
borat-borat-very-nice

PR #15405, opened 2022-08-04
[FUCK] No more randomized humans

GoldenAlpharex:
## About The Pull Request
Random human bad.

Lets people play their character.

## Changelog

:cl: GoldenAlpharex
fix: Your characters will no longer be randomized when you join.
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -118,6 +118,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 				if(initial(lang_path.secret))
 					languages.Remove(lang_path)
 			// SKYRAT EDIT END
+			return // SKYRAT EDIT - Don't remove this. Just don't. Nothing is worth forced random characters.
 	//we couldn't load character data so just randomize the character appearance + name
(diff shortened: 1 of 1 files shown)


tf-4 (approved):
😄

code/modules/client/preferences.dm
```diff
@@ -118,6 +118,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 				if(initial(lang_path.secret))
 					languages.Remove(lang_path)
 			// SKYRAT EDIT END
+			return // SKYRAT EDIT - Don't remove this. Just don't. Nothing is worth forced random characters.
```
Iamgoofball, line 121: this isn't really a fix, i think we should figure out why character data isn't loading instead because it's just gonna make Generic Bald Spaceman instead of random appearance

tf-4: it is a fix, because if it DID load it without this it would just randomise it anyway. The bug was that it randomised it no matter what, not that character data didnt load
the original code returns here, we added the languages bit but accidentally deleted the return. It's restoring it to /tg/ standard

Iamgoofball: ah okay

PR #15425, opened 2022-08-05
Halves MCR damage versus blobs

Zonespace27:
## About The Pull Request
Halves the damage MCRs deal towards blobs, blobbernauts, and blob spores

As it stands, MCRs _shred_ blobs, capable of dealing extremely high damage without needing to recharge, making the EMP strain nearly mandatory if you wish to fight against security in any meaningful way. The difference between MCRs and regular laser guns is that MCRs allow you to reload, allowing you to output much more fire without needing to stop and recharge. This considered, I feel that halving the damage of MCRs, still allowing them to be useful (but not nearly as dominant) is a good change

## Changelog

:cl:
balance: MCRs do halved damage to blobs, blobbernauts, and blob spores
/:cl:

### DIFF
--- a/modular_skyrat/modules/microfusion/code/projectiles.dm
+++ b/modular_skyrat/modules/microfusion/code/projectiles.dm
@@ -18,6 +18,11 @@
 	icon = 'modular_skyrat/modules/microfusion/icons/projectiles.dmi'
 	damage = 20
 
+/obj/projectile/beam/laser/microfusion/on_hit(atom/target, blocked)
+	if(istype(target, /obj/structure/blob) || istype(target, /mob/living/simple_animal/hostile/blob))
+		damage = damage / 2
+	return ..()
+
 /obj/projectile/beam/microfusion_disabler
 	name = "microfusion disabler laser"
 	icon = 'modular_skyrat/modules/microfusion/icons/projectiles.dmi'


Moostard:
It is true when science does all the tech nodes possible (they did their job, spent a lot of points, usually later in the round).
If you have a basic MCR (science did NOT do the tech at all) they will do close to nothing (10 damage plus very fast overheat and low cell capacity)
Blob got buffed already anyway.
Making a weapon like this tickle a round ending threat sounds funny in a bad way.

SpaceVampire:
Making one type of weapon less effective is a pretty unintuitive fix for a problem that doesn't really exist in the first place. You don't have any reason to suspect that this one type of laser weapon would do less damage to this one type of threat unless you just happened to know about the PR, and I really don't like that type of balance. I also don't see why the weapon with all of the end game tech unlocked needs to be less effective against this single threat. Having used the default MCR without science actually getting the upgrades, it really isn't strong at all. If people managed to get it, then just let it be good. Blobs in general also don't need to be buffed, as fighting them tends to turn into an unfun slog past a certain point.

tf-4:
Basically that,
It's extremely unintuitive and makes no sense ICly why it would do less damage against a blob - and *halving* it is pretty nuts

Zonespace27:
I still don't quite like how late-tier MCRs are capable of outputting continuous fire and melting blobs, might approach this some other way

PR #15437, opened 2022-08-06
Reverts the flip change

Edgar-Allan-ShitPoest:
## About The Pull Request
Simply reverts the change to *flip

Hi! I'm actually unenthused about the change to *flip. A lot of updates recently have been very small QOL changes which remove or lock relatively minor and harmless things. One of these is the *flip emote. Locking it behind the freerunning perk was a small, pointless change that was ultimately unneccessary, and adds fuel to the fire that "coders want to make the game less fun." I don't believe it was a necessary change as flipping annoyed no-one. It's locking however adds to the growing list of minor things that were pointlessly removed, which feeds the opinion that coders only change something when they decide they don't like it.
Suplexing for example was removed, and now flipping, as well as the recent language changes. It's a lot of relatively minor things that start to feel restrictive when you add them together. It's both frustrating and does end up driving people away over time.

please give us our flip back 
## Changelog

:cl:

del: Reverted the change which locks *flip to the freerunning perk

/:cl:

### DIFF
--- a/modular_skyrat/modules/emotes/code/emotes.dm
+++ b/modular_skyrat/modules/emotes/code/emotes.dm
@@ -52,12 +52,6 @@
 		return 'modular_skyrat/modules/emotes/sound/emotes/female/female_sneeze.ogg'
 	return
 
-/datum/emote/flip/can_run_emote(mob/user, status_check, intentional)
-	if(!HAS_TRAIT(user, TRAIT_FREERUNNING))
-		user.balloon_alert(user, "not nimble enough!")
-		return FALSE
-	return ..()
-
 /datum/emote/living/peep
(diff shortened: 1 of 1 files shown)


jjpark-kb:
There are plenty of features that can be added/removed without having to get permission from the host of the server; however, this is a feature that was specifically requested by the host to be removed. It was a compromise for it to be locked behind the trait. This unfortunately will not be merged. It is possible for you to talk to the host about this decision-- but I don't really condone that because it could cause some serious stress/strain on the host if everyone aired their grievances to the host.

Thank you for trying to contribute! I recommend talking to maintainers if you have any more ideas!

Gandalf2k15:
Host request.

PR #15442, opened 2022-08-06
rebalances SEVA somewhat, as it's very ??? at the moment

Paxilmaniac:
## About The Pull Request
I'll go through all of the changes I made:

**Goliath extra stun time gotten rid of -**
The goliath extra stun time was dropped for two reasons, one, why does it even do this, its the only clothing item in the game that does this, and second, the change was entirely non-modular as a direct edit to the goliath tentacle proc.

**Armor increase, well, the addition of any to begin with -**
The armor was bumped up a bit, comparison below:

BEFORE: (MELEE = 0, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 10, BIO = 50, FIRE = 100, ACID = 25, WOUND = 2)

AFTER: (MELEE = 20, BULLET = 10, LASER = 10, ENERGY = 10, BOMB = 30, BIO = 50, FIRE = 100, ACID = 50, WOUND = 10)

SEVA feels like its supposed to be a sidegrade to the default mining suit, and a mining suit sidegrade having NO armor at all, like literally zero, was an interesting choice for sure. The addition of a whopping 20 melee armor (still barely anything), and the fact that you **CANNOT ADD GOLIATH PLATES TO IT STILL, THAT PART HASN'T CHANGED**, means it will always be behind normal explorers suits in protection, even without upgrades, but you're still more protected than going out into the wastes literally naked. Wound armor was raised to 10 only because armor should be multiples of 10, especially if we want the bluespace examine tag to work.

**HIDETAIL and cold protection extension -**
I meant to make the suits hide tails but forgor in the first pr, and the cold protection (despite how little there is) should cover the same areas that the heat protection does.

The SEVA feels like its supposed to be an explorer suit sidegrade, and yet is coded like a complete downgrade. As a consequence I've seen a grand total of two (2) people use the suit since I've been playing. These changes should hopefully make the SEVA a viable alternative to the explorer suit, while not pushing it to strong enough that its a complete replacement for free. In the process, I also get rid of like three lines of non-modular code, which is good I think.

## Changelog

:cl:
balance: The SEVA now has minimal amounts of armor, still nowhere near as good as a base mining suit, but it's more protection than running around in your overalls
balance: The SEVA no longer adds an extra two seconds onto goliath tentacle stun time
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/traits.dm
+++ b/code/__DEFINES/~skyrat_defines/traits.dm
@@ -45,9 +45,6 @@
 //Makes sure that people cant be cult sacrificed twice.
 #define TRAIT_SACRIFICED "sacrificed"
 
-//Adds 2 seconds to the Goliath tentacle stun timer.
-#define TRAIT_GOLIATH_STUN "goliath_stun"
-
 /// The trait that determines if someone has the oversized quirk.
 #define TRAIT_OVERSIZED "trait_oversized"
 
--- a/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
+++ b/code/modules/mob/living/simple_animal/hostile/mining_mobs/goliath.dm
@@ -1,5 +1,3 @@
-#define GOLIATH_STUN_TIME_SEVA 12 SECONDS //SKYRAT EDIT - SEVA Fix
-
 //A slow but strong beast that tries to stun using its tentacles
 /mob/living/simple_animal/hostile/asteroid/goliath
 	name = "goliath"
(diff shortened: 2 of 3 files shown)


Zenitheevee:
So
Not _against_ this, I want this to be known.
But wasn't the Seva made to be shittier in defense and goliath grabs... Solely because its ash-storm proof? Like thats a rather large upgrade ((If you take out the fact it doesn't make visibility in an ash storm less punishing)) for __actual miners__ who dont just rush to megafaunas. (aka: the deranged) 
It is also worth remembering that it was balanced that way as a round-start suit, back when we had mining Hardsuits that would give you good protection and iirc ash-storm proof, but just slower.

But yeah, Good touch of love thats much needed.

Zonespace27:
> Zenitheevee

This consideration's fairly removed due to mining modsuits now existing, getting you ashproof + armor for 2500 (soon to be 3000) points

PR #15489, opened 2022-08-09
Modularizes CrewConsole.js to avoid Mirroring issues and fixes maths

Zergspower:
## About The Pull Request

With TG Mirrors a real probability here, i've added CrewConsoleSkyrat.js to account for this

Better represents the health of players who otherwise would be in peril, this would allow medical or other parties to better gauge if someone is dead or in critical in a more correct manner. Previously alive crew were showing up dead, this clearly was a cause of stress when a paramedic now assumes you are the walking dead.

## Changelog

image

Readdresses the math relating to the status icons that haven't been correct in quite a while, given the healthpool differences we were running off TG's formula. As you 'die' when you hit 200 damage I ensured the critical icon doesn't show until you're actually in critical now rather than the 100-125 zone.

:cl:
qol: Icons now correctly indicate health status
fix: Fixes the maths relating to the health pool, no longer working off 1-125
code: migrated CrewConsole to a Skyrat Modular file to avoid future clashes with mirrors
/:cl:


Zergspower:
i need to close this godamn branch already

PR #15510, opened 2022-08-10
Nerfs drifting contractors a bit

tf-4:
## About The Pull Request

So we've seen a lot of drifting contractors recently, and they're good. Very good. A bit *too* good.
The biggest problem with them, as far as I'm concerned, is that they get a lot of REALLY good gear right when they spawn in. Chatted with some folks and decided on this.
For example, they get a fully upgraded batong, with extra stam damage, muting, and cuffs.
They get SWAT boots (which even have a code comment saying they're OP dsquad gear). These are noslips with very high armour.

I like drifting contractors, quite a bit actually. But they could use a tone down from where they're currently at. So this PR removes:
- the upgraded batong that comes in the holster
- the SWAT boots

To soften the blow a bit, I've given them +3 starting TC which is actually a fairly solid increase over the original 10. Ideally this'll help people to better go for some sort of gimmick.

Drifting contractors were overperforming. This should bring them back in line and make them a bit more manageable for security. It also gives a bit more of a sense of progression as you upgrade the batong over time. A sense of pride and accomplishment, if you will.

## Changelog
:cl:
balance: Replaced the upgraded baton with a regular one for drifting contractors.
balance: Replaced the drifting contractor's SWAT boots with combat boots.
balance: Drifting contractors now get +3 starting TC.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/uplink/uplink_implant.dm
+++ b/modular_skyrat/master_files/code/modules/uplink/uplink_implant.dm
@@ -1,2 +1,2 @@
 /obj/item/implant/uplink/precharged/bonus
-	starting_tc = TELECRYSTALS_PRELOADED_IMPLANT + 5
+	starting_tc = TELECRYSTALS_PRELOADED_IMPLANT + 3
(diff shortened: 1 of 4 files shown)


GoldenAlpharex (approved):
Code checks out for me, and the reasoning is quite good too.

tf-4:
Gonna make some adjustments to this

Zonespace27 (approved):
Good compromise

YakumoChen:
Do contractors actually get an uplink?

Zonespace27:
> Do contractors actually get an uplink?

yes

PR #15531, opened 2022-08-10
[MODULAR] The guy who fucked up areas is back. Cuts down some clutter with multiple area files, condensing them into how its set upstream

Jolly-66:
## About The Pull Request
Hi. I'm responsible for the big area, uhh, "refactor" months ago.
Luckily, I'm not touching area defines today.
Instead, I've gone through our modular system and broke up where all our loose area defines are, since they spanned a good few files. This hopefully cuts down clutter and makes things easier to navigate.
This should also, hopefully, encourage new mappers to put new map area defines in these files instead of making a singular module for something like `/area/station/service/longue` or something.

It doesn't. (coder stuff beware)

## Changelog

Not player facing.


GoldenAlpharex (approved):
Looks good to me!

PR #15550, opened 2022-08-11
[MIRROR] Nukies Update 5: Bulldog Unchained (Wardens hate this) (I gave it a second magazine) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69119
--------------------

## About The Pull Request

Gives the Bulldog Shotgun a secondary magazine. You can press rightclick to fire and swap to the secondary magazine. You can also swap the magazine manually by pressing Alt-Rightclick on the gun, and load in a new magazine by rightclicking the gun with the magazine.

If you fire the gun, and your current magazine runs out, it automatically swaps to the other magazine if it has ammo.

## Why It's Good For The Game

<details>
<summary>Overwritten explanations inside</summary>

The Bulldog...just doesn't have a lot going for it. The Bulldog is meant to be a close quarters weapon, but practically any weapon in the nukie arsenal that burst fires will do just as well in close quarters, with comparable, and frankly, significantly greater damage. It has been an unfortunate victim of shotgun balancing, and while it continues to use the same shells as crew, this pattern will likely continue (hint, we're probably doing something about that, but not right now).

When I'm talking 'close-quarters', what I mean is 'three tiles or less', which is the ideal range for any shotgun. It helps make sure you get the full pellet impact, and ensures you don't lose too much damage midflight. At these same ranges, most burst weapons also don't typically miss their shots unless the entire volley would miss. And if the whole volley would miss, so would a shotgun likely miss their spread, as pellets don't tend to spread out until around this range as well, typically being very tight.

For comparisons sake, a point blank burst of a c20r does 90 damage, and if you have the AP version, also 90 damage but you ignore anything up to captain carapace. (only bulletproof armor can stop it, practically). 

A point blank shot of buckshot currently does 45. Slugs do 50. There is no AP alternative. If they have armor, you simply just have to deal with it while already having a damage disadvantage.

There was also a point in time where the Bulldog had one of the best disabling ammo types for nukies, the taser slug (and meteor slug but that was just a glorified taser slug at the time because it was very buggy). They lost that as well with the stun removal, with no compensation. It still has meteor shot, and the utility certainly helps, but it isn't as good at disabling enemies as you might like, and definitely doesn't hurt enough.

It also cannot be bagged, which was originally one of the nicer benefits of the bulldog over comparable shotguns in its class, and let it be a very sneaky weapon to get in close with and spring on the unexpected. Not a huge deal, but it is a massive slap in the face when compared to the c20r again, which still can be bagged as of right now.

If this wasn't enough of a problem, there is actually another weapon that does quite literally all the these things, only missing rapid fire capabilities. And that's the sniper rifle, which is a hilariously powerful close ranged weapon and a disabling weapon AND a high lethality weapon AND an armor piercing weapon, only held back but needing to wait several seconds between shots. But since you can bag the gun, its actually very practical as a backup gun.

So, the bulldog simply doesn't have a significant identity compared to every other weapon that nukies can get. So what can we do? First, we can give it an upside that the other weapons do not. Dual magazines. 

Being able to change ammo types with the press of a button is the kind of upsides only NT weaponry presently get to benefit from. Since shotgun shells have a lot of variety in projectile type, it stands to reason that the user might only want to use certain ammo types in certain situations. Therefore, giving them the option of preloading two mag types allows them the chance to respond to a dynamically changing fight without having to mess around with bags, hotkeys or needing to free up a hand.

It is a start, but I think it is a strong start to making the bulldog worth a shit.
</details>

TL;DR Two mags makes it easier to do combinations of shells and saves needing to hotswap as much while we work on making the Bulldog less useless from an ammo perspective.

## Changelog

:cl:
balance: The Bulldog Shotgun now has a second magazine. 
balance: You can swap the magazine as you fire by shooting with rightclick.
/:cl:


JohnFulpWillard:
warden can combat this with a compact combat shotgun with roundstart lethals

PR #15552, opened 2022-08-11
[MIRROR] Removes ComponentInitialize() [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69118
--------------------
## About The Pull Request

A remake of https://github.com/tgstation/tgstation/pull/60644

Completely removes ComponentInitialize() as a proc, which was called on every single atom in the game, twice in some instances (like new players). If you want to ComponentInitialize() something, you should do it on Initialize() instead, which we would tell people to do anyways.
This also removes the need to tell new contributors to not use ComponentInitialize() because despite its name, it doesn't actually do anything that Initialize() can't.

## Why It's Good For The Game

stupid initialize die

## Changelog

im johnfulpwillard and i make changelogs for things that players dont see!!!


GoldenAlpharex (approved):
Should be good to go!

PR #15563, opened 2022-08-12
[MODULAR] Fixes sec/guard quirk blacklist oversights.

Higgin:
## About The Pull Request

Security roles can no longer take Nerve Staple as a workaround for pacifism (likely an oversight.)

Guards can no longer be Pacifists, Nerve Stapled, or Foreigners. Dying of a brain tumor or unable to shoot guns is fine. Being functionally incapable of interacting with the world or escorting people from the premises is less so if they're going to do anything guard-ly. 

Seems like an oversight - fix good?

## Changelog

:cl:
fix: Nerve Stapled no longer acts as a workaround to Security not hiring Pacifists.
balance: Department Guards can no longer be Pacifists, Nerve Stapled, or Foreigners.
/:cl:


GoldenAlpharex (approved):
Good change, and I don't think it deserves the GBP sink of a "balance" chance when it's more of a consistency issue.

PR #15564, opened 2022-08-12
[MODULAR] Fixes cryopod timers.

### LINKED ISSUE #15523: SSD players can't be put in cryo until they've been ssd for 300 hours
## Reproduction: 
Find ssd player, drag their sprite into a cryopod, receive funne message about needing another 18,000 minutes to cryo them. I could seethis being intentional so players don't get put into cryo while experiencing connection issues, but I would think a timer closer to 10-15 minutes would be more reasonable. Had to ask an admin to remove a player after they had to unexpectedly leave, they had used an expose verb prior to that and I didn't want to leave the poor bean exposed for the rest of the round.

Higgin:
## About The Pull Request
fixes #15523 
credit to @Jolly-66 for picking out the block that was giving the issues, I think it was that SSD time (18000) wasn't getting re-divided to make sense in MINUTES with the remainder - moving the rounding around fixed it

it's good to know your friends aren't staying standing upright for 300 hours

## Changelog

:cl:
fix: Cryopods now display accurate time before you can cryo somebody who's AFK/SSD.
/:cl:


GoldenAlpharex (approved):
Good change.

PR #15583, opened 2022-08-13
Adds a UI to inject chemicals as a borer

Zonespace27:
## About The Pull Request
image
This PR adds a UI for borers to inject chemicals, instead of going by choosing an element from a list. In addition to this, you are able to upgrade the amount of chemicals you inject at a time, starting out with the default 5, allowing you to spend 2 stat points per upgrade to get 10, then 25, then 50. The cooldown for injecting an amount of chemicals higher than default is equivalent (50u injection cooldown = 10 5u injection cooldowns)

Also cleans up some code from another borer PR

This is much better QOL than choosing from a list and being forced to spam it

## Changelog

:cl:
refactor: Borer chem injection now has a custom UI
/:cl:


nevimer:
yessss

ghost:
You should make reagents into a list, otherwise you could inject any reagent into a player.

Zonespace27:
> You should make reagents into a list, otherwise you could inject any reagent into a player.

That is in fact how borers work, you can inject nearly any chem, if you've learned it from a host's bloodstream

ghost:
> > You should make reagents into a list, otherwise you could inject any reagent into a player.
> 
> That is in fact how borers work, you can inject nearly any chem, if you've learned it from a host's bloodstream

This isn't "nearly" every chem though, you can inject any chem without your host ever injecting it. It uses name2reagent. Clients can send any parameters to a request in TGUI, so they can enter the name of any reagent as a parameter.

Zonespace27:
> > > You should make reagents into a list, otherwise you could inject any reagent into a player.
> > 
> > 
> > That is in fact how borers work, you can inject nearly any chem, if you've learned it from a host's bloodstream
> 
> This isn't "nearly" every chem though, you can inject any chem without your host ever injecting it. It uses name2reagent. Clients can send any parameters to a request in TGUI, so they can enter the name of any reagent as a parameter.

I understood what you meant like 2 days after in the shower, fixed in #15602

PR #15623, opened 2022-08-14
Nunsense

NPC1314:
## About The Pull Request

While I do have fun playing a nun, the costume is a bit lacking. Made an alternative version, that looks better, and can be selected in loadouts for chaplains.

How the new one looks
nunsense

Better looking clothes is always good.

## Changelog

:cl:
imageadd: new nun outfit added
/:cl:


code/modules/clothing/head/costume.dm
```diff
@@ -192,3 +192,10 @@
 	name = "allies helmet"
 	icon_state = "allies_helmet"
 	inhand_icon_state = "allies_helmet"
+
+/obj/item/clothing/head/nun // Alternative version of old nun headdress with a lot more detail. Added 2022-08, paired with Nun robes
```
Zonespace27, line 196: this isn't modular

code/modules/clothing/under/costume.dm
```diff
@@ -347,3 +347,12 @@
 	inhand_icon_state = "tmc_jumpsuit"
 	can_adjust = FALSE
 
+/obj/item/clothing/under/costume/nunrobes // new style of nun robes, only available in loadout. Added 2022-08, paired with Nun headdress
```
Zonespace27, line 350: ditto

modular_skyrat/modules/loadouts/loadout_items/loadout_datum_heads.dm
```diff
@@ -567,6 +567,11 @@ GLOBAL_LIST_INIT(loadout_helmets, generate_loadout_items(/datum/loadout_item/hea
 	item_path = /obj/item/clothing/head/imperial/white
 	restricted_roles = list(JOB_CAPTAIN, JOB_HEAD_OF_PERSONNEL, JOB_BLUESHIELD, JOB_HEAD_OF_SECURITY, JOB_RESEARCH_DIRECTOR, JOB_QUARTERMASTER, JOB_CHIEF_MEDICAL_OFFICER, JOB_CHIEF_ENGINEER)
 
+/datum/loadout_item/head/nun
```
Zonespace27, line 570: the icon files aren't modular either

Zonespace27 (changes requested):

modular_skyrat/modules/loadouts/loadout_items/loadout_datum_heads.dm
NPC1314: So you are saying you are requesting the new art to be placed in a new individual dmi? Located where, if so? Please give more detail.

code/modules/clothing/head/costume.dm
NPC1314: What change exactly are you suggesting? Another item path? Putting it in a new dm? Please elaborate.

Zonespace27: Follow the guide from the [readme](https://github.com/Skyrat-SS13/Skyrat-tg/blob/master/modular_skyrat/readme.md)

modular_skyrat/modules/loadouts/loadout_items/loadout_datum_heads.dm
GoldenAlpharex: You would have to look around for other chaplain items that we have added on Skyrat, and probably add those sprites in that same .dmi file. I can't remember off the top of my head if we do (I'm pretty sure we do though?), and in case we don't, then you'll have to add them to one of the modules.

YakumoChen:
Consider taking it upstream to wholesale replace the existing outfit too

SpaceVampire:
I would also consider adding it chaplain clothes vender, if you don't end up doing that.
It's nice having ways to get items that aren't just starting with them via the loadout system.

PR #15634, opened 2022-08-15
[MIRROR] Makes glass tables actually transparent [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69189
--------------------
They're almost completely opaque right now, to the point where you really can't even see the floor underneath them.

Changed the opacity and redid the legs so you can actually see them

:cl: ShizCalev
imageadd: You can now see through glass tables again!
/:cl:

Old:
image

New:
image


GoldenAlpharex (approved):
Nice!

PR #15638, opened 2022-08-15
[MIRROR] De-hardcodes randomize_human() and fixes some related issues along the way [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/68876
--------------------

## About The Pull Request

randomize_human() now relies on species procs to do the brunt of it's feature randomizing. External organs from non-humans also randomize properly now, meaning mulligan and other features that use the randomize_human() proc will work for them like they should. randomize_human() can now assign a mob plural gender and physique as well and will default to it should the species have no sexes, imitating character preferences.

https://user-images.githubusercontent.com/80640114/181880034-41fbbce5-a614-48f4-a374-7c0964759d5e.mp4

For a slightly more detailed code description: randomize_human() now calls a new species proc called randomize_features() for every species subtype, which takes care of randomizing most species-specific visual aspects (with the exception of mcolor as it's common to all species and is randomized by randomize_human() itself). To that end there's now another species proc, randomize_external_organs(), which randomizes all external organs of a species that are present in the mob. The get_global_feature_list() procs for lizard tails, cat tails and pod hair have been changed for this because they were either mislabelled or missing altogether. Randomize_human() also calls a species' randomize_active_underwear() proc instead of changing the var manually.

## Why It's Good For The Game

The objective of de-hardcoding randomize_human() is to make it easier to add or modify species and species features in the future. Fixing external organ randomizing is a good thing as well for obvious reasons. 
In case it needs justifying, the reason to run through all the species proc instead of just our current species is both to avoid the default configuration assigned by initialize_dna() which has some fixed settings, and to avoid edge-case issues with changing species back and forth. It shouldn't pose a performance problem because the vast majority of species just return and it's a good precaution.

## Changelog

:cl:
code: changed the randomize_human() proc to call species procs instead of being mostly hardcoded and cleaned up some code
fix: randomize_human() can now assign plural gender to a mob and will always do so for species without sexes
fix: the external features of non-humans are now randomized properly by randomize_human() and all mechanics that call the proc, such as mulligans
/:cl:


GoldenAlpharex (approved):
Should hopefully be good to go now.

PR #15647, opened 2022-08-15
[NON-MODULAR] Actually Actually Fix Invisible Languages for Good!

RimiNosha:
## About The Pull Request
Fixes a different unselectable invisible languages bug that is completely unrelated to my languages removal and reflavouring PRs!

Basically checks if a language path exists before trying to check if it's secret, cause `secret` is false by default!

Useroth can finally stop having to reset player files to fix invisible languages!  
Less user frustration when they can't fix the language bug themselves!

## Changelog

:cl:
fix: The invisible languages bug is gone forever, provided something else doesn't break!
/:cl:


GoldenAlpharex (approved):
Looks good!

PR #15674, opened 2022-08-16
Updates the content of the ban reason in the database

GoldenAlpharex:
## About The Pull Request
This should save me countless previous headaches where I had to fetch the actual ban note to display the proper message, when there's nothing that links a ban and its matching note 1-to-1. Now I simply won't have to fetch the ban note, as the reason will actually already be complete.

For players, nothing. For Staff, less instances where bans aren't automatically reported.

## Changelog

Non player-facing.


GoldenAlpharex:
I thought some more about it, and I don't actually need this at all. How convenient!

PR #15679, opened 2022-08-17
barricade open/close time buff

Stalkeros2:
## About The Pull Request
Plasteel barricade's open/close time decreased from 5 to 2 seconds.

The five second delay is too absurd, it's too long to just open and close the barricade without putting yourself into a lot of danger, it's not the first time I've seen that five second timer cause, say, xenomorphs or any other threat to break through, just because the barricade couldn't be closed/opened quick enough. Even the false wall gets better at that point.

## Changelog

:cl: Stalkeros
balance: Plasteel barricade's opening/closing time reduced from 5 seconds to 2 seconds.
/:cl:

### DIFF
--- a/modular_skyrat/modules/barricades/code/barricade.dm
+++ b/modular_skyrat/modules/barricades/code/barricade.dm
@@ -690,7 +690,7 @@
 	if(.)
 		return
 
-	if(do_after(user, 20, src))
+	if(do_after(user, 2 SECONDS, src))
 		toggle_open(null, user)
 
 /obj/structure/deployable_barricade/metal/plasteel/proc/toggle_open(state, mob/living/user)
--- a/modular_skyrat/modules/barricades/code/barricade.dm
+++ b/modular_skyrat/modules/barricades/code/barricade.dm
@@ -657,8 +657,10 @@
 	closed = TRUE
 	can_upgrade = FALSE
 	portable_type = /obj/item/quickdeploy/barricade/plasteel
-	///ehther we react with other cades next to us ie when opening or so
+	///Either we react with other cades next to us ie when opening or so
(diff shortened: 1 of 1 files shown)


tf-4:
You need a better reason than that

Stalkeros2:
(Right ik waiting for them to write it because I personally don't mind the five second delay that much and feeling not good to come up with one.)
Done.

Lukaster42:
> You need a better reason than that

The five second delay is too absurd, it's too long to just open and close the barricade on TG, it's not the first time I've seen that five second timer cause xenomorphs to break through, just because the barricade has that timer, even the false wall gets better at that point, no one uses it because of that.

modular_skyrat/modules/barricades/code/barricade.dm
```diff
@@ -690,7 +690,7 @@
 	if(.)
 		return
 
-	if(do_after(user, 50, src))
+	if(do_after(user, 20, src))
```
GoldenAlpharex: Could be interesting to make this time be a variable of the barricade itself, so it could be modified both by admins and in general by subtypes (for traitor items or special loot, I guess. Not that it's incredibly powerful on its own, but...)
```suggestion
	if(do_after(user, 2 SECONDS, src))
```

GoldenAlpharex (changes requested):

modular_skyrat/modules/barricades/code/barricade.dm
Stalkeros2: I was planning to do this myself, but due to some IRL stuff couldn't. Will do, because I do agree with this idea.

GoldenAlpharex (approved):

PR #15680, opened 2022-08-17
[NO GBP] Fix Languages Being Reset if You Load in Early

RimiNosha:
## About The Pull Request
I tested this on exactly 8 different premade characters so if it breaks it literally is not me anymore.

Also adds in a bit because turns out prefs aren't reloaded when you switch characters. The more you know.

Yes, I did ensure there were invisible languages before I made the fix, and yes, they are now gone.

Yes, the server was rebooted a few times after the fix was made, and no, it does not randomize characters, cause I'd notice real quick if my OCs were scrambled.

Please let this be the end of the language saga so I can continue to do the fun thing I wanted to do now.

Should no one do it before me, please remind me after I've done the couple of fun things I have lined up to make pref unit tests to ensure prefs or parts of prefs do not get reset when they really shouldn't.

Reset languages bad.

## Changelog

:cl:
fix: Languages are actually fixed for good. Also fixes them for all your characters.
/:cl:


code/modules/client/preferences.dm
```diff
@@ -115,8 +115,8 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 		if(load_character())
 			// SKYRAT EDIT START - Sanitizing languages
 			for(var/lang_path as anything in languages)
-				var/datum/language/language = GLOB.language_datum_instances[lang_path]
-				if(!language || language.secret)
+				var/datum/language/language = new lang_path()
+				if(!(language.type in subtypesof(/datum/language)) || language.secret)
```
GoldenAlpharex: I would probably add a comment above this if() to say that this ensures that languages that don't exist in code anymore don't linger in your languages prefs. Just so we don't get back to this situation again in the future.

code/modules/client/preferences.dm
```diff
@@ -233,6 +233,17 @@ GLOBAL_LIST_EMPTY(preferences_datums)
...
+			// SKYRAT EDIT START - Sanitizing languages
+			var/languages_edited = FALSE
+			for(var/lang_path as anything in languages)
+				var/datum/language/language = new lang_path()
+				if(!(language.type in subtypesof(/datum/language)) || language.secret)
+					languages.Remove(lang_path)
+					languages_edited = TRUE
+			if(languages_edited)
+				save_character()
+			// SKYRAT EDIT END
```
GoldenAlpharex:
```suggestion

			if(languages_edited)
				save_character()
			// SKYRAT EDIT END
```

code/modules/client/preferences.dm
```diff
@@ -233,6 +233,17 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 				randomise_appearance_prefs()
 				save_character()
 
+			// SKYRAT EDIT START - Sanitizing languages
+			var/languages_edited = FALSE
+			for(var/lang_path as anything in languages)
+				var/datum/language/language = new lang_path()
+				if(!(language.type in subtypesof(/datum/language)) || language.secret)
+					languages.Remove(lang_path)
+					languages_edited = TRUE
```
GoldenAlpharex: Considering how you have this code duplicated in two places, I'd probably recommend just making it a proc that returns true or false based on whether or not languages were edited. Then it'd be a lot cleaner, and a lot easier to maintain in the future.

GoldenAlpharex (changes requested):

code/modules/client/preferences.dm
RimiNosha: Sounds like a plan!

GoldenAlpharex (approved):
This should be good, but I'll just start with a test-merge first.

GoldenAlpharex (approved):
Looks good, seems to work just fine on the server too!

PR #15702, opened 2022-08-19
[MIRROR] Converts a shitload of istypes to their more concise macros [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69260
--------------------
There were a lot of these so i figured i would change as many as i can find.
This was all done with regex search and replace so it should be easy to review

Did i add way too many commits for no good reason? yeah.
cry about it nerds


GoldenAlpharex (approved):
Should be good to go!

PR #15720, opened 2022-08-20
[MIRROR] Removes Thieves [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69304
--------------------

## About The Pull Request

Removes thieves, and their midround counterpart, opportunists

## Why It's Good For The Game

Plethora of reasons. The short of it is that I don't think /tg/station is the server for a miscreants-level-antagonist and it won't be for a very long time. And because it's not, future headmin candidates are going to struggle to change thief to be something different, and it's going to become antithetical to its original goals.  So away with it!
The long reasons list:
* Thief fails to be open. When I implemented flavor traitors, the flavor supplements the antagonist, but they are still free to do as they please. Sure, after that progression traitors got added, but that changes their access to gear, not necessarily their freedom on actions. MRP traitors always had rules around objective following. Thieves have their flavor tied to objectives, and while they can do whatever and ignore objectives, players do not do that in practice.
  * Says my point well: image
* Thief fails to work in LRP. People are confused about what they are allowed to do, since all other antagonists have no restrictions, but thieves do, it ends up with policy headaches. Some candidates are discussion allowing thieves to kill to steal, which will go against the very notion of a 'miscreant-level' antagonist.
* Thief fails to work in MRP. On MRP, people are overly cautious, and farm objectives without actually committing crime, and repercussions for thieving frequently bring up admin headaches, though not the ones like LRP has.

Sorry for uh... dejecting ss13 to that for awhile. I'll try better to design good antagonists. But at least it's not going to become a headmin candidate talking point, getting desecrated by wide policy changes until it is finally taken out back, that whole process won't be fun for players.

## Changelog

:cl:
del: Thieves are gone, and so are Opportunists, their midround counterpart.
/:cl:


Zonespace27:
Goodbye to this pointless waste of a ruleset

PR #15741, opened 2022-08-22
Optimizes the Examine panel's character preview, while making it a lot more accurate

GoldenAlpharex:
## About The Pull Request
The Examine panel's character preview will now use a copy of the mob's current appearance to display it, rather than the whole mess that was "create a dummy and give it a copy of your clothes".

The way it's handled also means that Cyborgs get rendered in the preview window as well, albeit they don't currently display their lights, so there's still some space for future improvements.

I also removed an edit to the changeling procs that was related to the dummy of the examine panel, so that should also be more performant.

Less lag when opening the panel, better memory management, better performances, simpler code... What's not to love?

## Changelog

:cl: GoldenAlpharex
code: Vastly improved the code of the Examine menu's Character Preview, making it a lot more reflective of what the character currently looks like, while also making it much more performant at the same time!
fix: Cyborgs will now appear in the Character Preview (without their lights, sadly) of the Examine panel too!
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/transform.dm
+++ b/code/modules/antagonists/changeling/powers/transform.dm
@@ -142,17 +142,7 @@
 		return
 	..()
 	changeling.transform(user, chosen_prof)
-	
-	// SKYRAT EDIT START
-	var/mob/dum = user.tgui.dummy_holder
-	dum.name = chosen_prof.name
-	for(var/obj/item/item in dum)
-		if(!dum.dropItemToGround(item))
-			qdel(item)
-			dum.regenerate_icons()
-	changeling.transform(dum, chosen_prof)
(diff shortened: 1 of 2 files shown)


azrael-x86:
woo-yeah-oh-yeah


woo yeah baby goldenalpharex pr yeah baby woohoo

GoldenAlpharex:
This will likely lead to a bigger-scale PR to refactor how get_id_photo() works for security records, but I'll probably do that another day.

GoldenAlpharex:
> This will likely lead to a bigger-scale PR to refactor how get_id_photo() works for security records, but I'll probably do that another day.

(it did, hopefully it trickles down soon enough!)

Zonespace27:
looks fine to me, YOLO

PR #15766, opened 2022-08-23
[MODULAR] Removes Space Ninja from rolling from random events

Jolly-66:
This should only be in dynamics. This was enabled for some reason a while back in random events, hence why we're getting it every single round.

## Changelog

:cl: Jolly
del: Space Ninjas no longer roll in random events, in favor of being controlled entirely by dynamic.
/:cl:

### DIFF
--- a/modular_skyrat/modules/events/code/event_overrides.dm
+++ b/modular_skyrat/modules/events/code/event_overrides.dm
@@ -87,7 +87,7 @@
 /**
  * Ninja
  *
- * Removed:
+ * Removed: It's already apart of dynamic, we don't need it both as an event and as a part of dynamic. Leaving it in dynamic for finer control over how often it rolls.
  * 
  */
 /datum/round_event_control/space_ninja
--- a/modular_skyrat/modules/events/code/event_overrides.dm
+++ b/modular_skyrat/modules/events/code/event_overrides.dm
@@ -83,3 +83,13 @@
  */
 /datum/round_event_control/operative
 	max_occurrences = 0
+
+/**
+ * Ninja
+ *
+ * Removed:
+ * 
+ */
+/datum/round_event_control/space_ninja
+	max_occurrences = 0
+	weight = 0


modular_skyrat/modules/events/code/event_overrides.dm
```diff
@@ -83,3 +83,13 @@
  */
 /datum/round_event_control/operative
 	max_occurrences = 0
+
+/**
+ * Ninja
+ *
+ * Removed:
```
GoldenAlpharex:
```suggestion
 * Removed: It's already apart of dynamic, we don't need it both as an event and as a part of dynamic. Leaving it in dynamic for finer control over how often it rolls.
```

GoldenAlpharex (changes requested):
Just giving a proper removal reason to it, otherwise it should be good to go.

Also, some of the code located in [this proc from event code](https://github.com/Skyrat-SS13/Skyrat-tg/blob/master/code/modules/events/space_ninja.dm#L40) should probably be transferred over to [this proc in the midround dynamic rulesets](https://github.com/Skyrat-SS13/Skyrat-tg/blob/master/code/game/gamemodes/dynamic/dynamic_rulesets_midround.dm#L750) instead, as we won't be using the event anymore. It can likely be handled in a separate PR, however, but if it's not done here, an issue should be filled for it.

twilightwanderer:
I would suggest giving the ninja meaningful goals and some kind of code of conduct to give this antagonist a more adequate appearance.

GoldenAlpharex:
> I would suggest giving the ninja meaningful goals and some kind of code of conduct to give this antagonist a more adequate appearance.

This is a good idea, and something that I'd highly suggest making a Feature Request (in the Issue tab) on, to elaborate on your ideas, unless you would want to make a design document on it. That being said, it goes a little outside the scope of this PR.

GoldenAlpharex (approved):

YakumoChen:
Is this removing ninja from being in rounds entirely?

Stalkeros2:
I will totally **not** miss this entirely NRP incredibly evasive antag with goals that don't promote anything but speedruns.

tf-4:
> Is this removing ninja from being in rounds entirely?

No

PR #15842, opened 2022-08-25
[Tram] New tram, new consequences. Tram collisions actually hurt.

lessthnthree:
## About The Pull Request
This change updates the damaged caused from getting hit by the tram to a random number between 120 and 240, roughly a 10% chance of death.

As an MRP/HRP server, characters are expected to value their bones and lives. Risking the traffic signals or riding the rails should have real consequences. In testing we've seen emergent roleplay of people doing things such as enforcement, barriers, painting if a high number of people are foolish getting hit by the tram.

## Changelog
:cl: LT3
balance: Tramstation's tram now actually packs a punch when it hits you. It will hurt. Think twice before crossing on red!
/:cl:


modular_skyrat/modules/tram/tram_lift_master.dm
```diff
@@ -0,0 +1,2 @@
+/datum/lift_master/tram
```
Zonespace27: don't make a module for this, just put it in a `master_files` override

Zonespace27 (changes requested):

Zonespace27 (approved):
I like it

Paxilmaniac:
I thought this was a nerf until I read it and realized its so, so much better

PCPort:
Time to bucklecuff miscreants to the rails and cackle whilst twirling my moustache

GoldenAlpharex:
I think it's a neat idea. Looking forward to see how people shape up for that change.

Iamgoofball:
This is a terrible idea and is just going to result in you guys putting rules on the tram and further ruining an already working just fine design.

Like, alright sure, it's not a bad idea in theory but I foresee the end result being some admin banning a guy for trying to take the tram somewhere and some idiot being on the tracks.

There are times when it's easier for the health of the game as a whole to not make something realistic, and I think this is a good case of that.

Iamgoofball:
after chatting with folks, hesitant 👍 but definitely keep an eye on it

modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
```diff
@@ -0,0 +1,2 @@
+/datum/lift_master/tram
+	collision_lethality = 4
```
GoldenAlpharex, line 12: Could be nice to just add a comment here saying what range of damage that equates to, just so it's easier to understand why this was done this way in the future.

GoldenAlpharex (changes requested):

modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
lessthnthree: Sounds good.

modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
```diff
@@ -0,0 +1,12 @@
+/**
+ * Damage calculation for getting hit by the tram. 120 to 240 damage to the player. A chance of death, but it's going to hurt.
```
tf-4:
```suggestion
 * Damage calculation for getting hit by the tram. 120 to 240 damage to the player at a lethality of 4. A chance of death, but it's going to hurt.
```

GoldenAlpharex (approved):
I've been happy with the results so far.

PR #15855, opened 2022-08-26
[MIRROR] Chasm Fishing [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69252
--------------------

## About The Pull Request

image

Adds the capability of fishing in chasms.
Chasms contain Chasm Chrabs, which are cute little crustaceans you can eat or put in an aquarium.
More importantly, chasms contain everything which has ever fallen into a chasm.
Falling into a chasm is no longer instant permanent round removal... as long as someone is willing to do some fishing, which means your chances of return haven't necessarily increased that much.

image

Potential rescuers (or people lamenting about how they dropped their necropolis chest into a hole) should also beware, chasm crabs are actually young Lobstrosities and there's a rare but not insubstantial chance that you'll fish one of them up instead, which won't make it very happy.

If you somehow come back to life inside a chasm (generally only likely by being a changeling or some admin interference, but perhaps you have some kind of implant which does it) you'll climb back out again, so holes are no longer a way of disposing of a changeling for good. You think The Thing would be defeated if you threw it down a hole?
Ethereals can't manage this because falling into a chasm causes too much damage and they simply smash on the way down.

My current implementation for this involves chasms instantiating an abstract and invisible object when required to put things inside which as far as I can tell seems to work, but please do tell me this is a terrible idea if there's some side effects to this which I have not foreseen and I can try thinking of something else.

## Why It's Good For The Game

The primary reason for this PR is mostly that I think that it's funnier than how chasms currently work.
I also wanted more fishing content, and adding some weird ancilliary purpose for fishing rods seemed fun.
Instant deletion of a player's brain when they fell into a hole can be pretty deflating sometimes especially if they were a blood brother or something, now there's a way around that if you work for it.

## Changelog

:cl:
add: You can now fish in chasms, and might manage to catch things that careless miners have dropped in there.
add: If your pole is long enough to reach into a deep hole, you might catch crabs.
balance: Changelings can climb out of chasms if you throw them into one.
imageadd: An icon of a nice little crab, who will grow up into something less nice.
add: You can craft fishing equipment from various lavaland materials, so that Ashwalkers can fish up dead miners before you.
add: You can sometimes find worms while digging up tiles, if appropriate for that material.
/:cl:


code/datums/components/chasm.dm
```diff
@@ -92,59 +118,124 @@
...
 
+<<<<<<< HEAD
 		if(iscyborg(AM))
 			var/mob/living/silicon/robot/S = AM
 			qdel(S.mmi)
+=======
+	//Make sure the item is still there after our sleep
+	if(!dropped_thing || QDELETED(dropped_thing))
+		return
+>>>>>>> bcadead566b (Chasm Fishing (#69252))
```
tf-4:
```suggestion
	//Make sure the item is still there after our sleep
	if(!dropped_thing || QDELETED(dropped_thing))
		return
```

PR #15856, opened 2022-08-26
[MIRROR] Fixes strange reagent, and possibly other chems too [MDB IGNORE]

### LINKED ISSUE #69243: Sulfuric acid, fluorosulfuric acid, and nitric acid do not deal damage properly when injected
## Reproduction:

1. Fill a syringe with 5u Sulfuric Acid, 5u Fluorosulfuric Acid, and 5u Nitric Acid
2. Inject human with syringe.
3. Expected result is for brute and burn damage to be dealt, as intended in toxin_reagents.dm
4. All 3 acids in the syringe only do 5 burn damage and 48 toxin damage, with 0 brute damage being dealt.
Moreover, a syringe filled with 15u sulfuric acid does 0 brute damage. However, the brute/burn damage is dealt when the acids are splashed onto the target.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69432
--------------------
## About The Pull Request

``remove_reagent()`` was being called before ``expose_multiple()`` was, which meant people were being exposed to chemcials that had 0 volume in them, something that was needed for something like strange reagent to work.

This was caused when support for multiple tastes were added, so I fixed it by making it remove the chems after they all did what they needed to do.

## Why It's Good For The Game

Strange reagent now works, and possibly other chems that relied on ``reac_volume``

Closes https://github.com/tgstation/tgstation/issues/69243

## Changelog

:cl:
fix: Reagents like Strange Reagent, Sulfuric Acid, blob reagents, and many others, now works again.
/:cl:


GoldenAlpharex (approved):
Should be good now!

PR #15913, opened 2022-08-29
[MIRROR] Monkey trip mechanic removed from AI controller and added to brain. Adds organ based trait system. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69496
--------------------

## About The Pull Request
Alright, big exciting PR.

First big thing! Traits can now be tried to organs which are granted and removed from the mob with the organ.
The traits PRIMATIVE, LITERATE, CAN_STRIP, and ADVANCED_TOOL_USER have been removed from species inherent traits list, because they apply to the species brain in specific, not their whole body.
The baseline brain gives the traits LITERATE, CAN_STRIP, and ADVANCED_TOOL_USER, all species that had these traits have a baseline brain. 
Ashwalkers have been given a "primitive brain" which gives ADVANCEDTOOLUSER, CAN_STRIP, PRIMITIVE, and notably not literacy. These give them the same traits as before, but their privative nature and lack of reading followers their brain no matter which body they inhabit.

Monkeys have been given the "primate brain" which gives CAN_STRIP and PRIMITIVE. The monkey's brain now holds the logic for the monkey tripping mechanic instead of the AI datum which is a huge win. Monkeys with clients can also now toggle if they would like to be tripped by people, in order to better blend in. AI monkeys will always have tripping on. I also have implemented my qol monkey tripping changes from #69478 (the videos on that PR are still accurate)

What this means, is  a human's brain in a monkey's body will still be able to use tools, read, and even operate mechs. but will still have the other non mental monkey traits like being able to crawl though vents while naked. So lovely.

Conversely a monkey's brain (even a sentient one) that is put in a human body will still be illiterate, and will be unable to use complex tools.

Transforming into a monkey via genetics or other means will also change your brain into a monkey brain (this mechanic existed before now, it just wasn't applicable)
 
Some of this code might not be up to standard, please give me your corrections.

## Why It's Good For The Game
This is a pretty good step forward with species code, and the organ based trait system opens up many more possibilities for future PRs.
AI shouldn't introduce behavior the base mob (or in this case the mob's brain) isn't able to do. Armhulenn helped me out with how this could be implemented in a cool way. The trip qol stuff shouldn't change how you interact with monkeys, it just stops them from clambering over eachother and falling over at the slightest breeze. Sentient monkeys being able to choose whether or not they want to trip is based on the idea that players controlling mobs should be able to do anything that the mob could do, even if its mostly detrimental, But also allows them to not be total chumps in any combat scenario if they want.
so dope

## Changelog

:cl: itseasytosee
qol: Monkeys wont trip over each other anymore, and will only trip if you run into them with combat mode on.
qol: Player monkeys can now opt in to tripping like a regular monkey.
refactor: Stores monkey tripping in the brain, rather than the AI controller
refactor: Traits such as literacy and advanced tool use are now tied to the brain rather than the species 
/:cl:


code/modules/mob/living/carbon/human/species_types/snail.dm
```diff
@@ -3,11 +3,12 @@
...
 	species_traits = list(MUTCOLORS, EYECOLOR, HAS_FLESH, HAIR, FACEHAIR) //SKYRAT EDIT - Roundstart Snails - Customization
 	inherent_traits = list(
-		TRAIT_ADVANCEDTOOLUSER,
-		TRAIT_CAN_STRIP,
 		TRAIT_NOSLIPALL,
+<<<<<<< HEAD
 		TRAIT_LITERATE,
 		TRAIT_WATER_BREATHING, //SKYRAT EDIT - Roundstart Snails
+=======
+>>>>>>> 497fa7b23f5 (Monkey trip mechanic removed from AI controller and added to brain. Adds organ based trait system. (#69496))
```
tf-4:
```suggestion
		TRAIT_WATER_BREATHING, //SKYRAT EDIT - Roundstart Snails
```

PR #15950, opened 2022-08-31
Fixes being able to retract an active MODsuit's parts and still be able to use its modules

GoldenAlpharex:
## About The Pull Request
The title says it all. People have been abusing it for too long, and now I decided to finally fix it, because nobody else would.

It doesn't make sense and fixes multiple visual glitches, too.

## Changelog

:cl: GoldenAlpharex
fix: You can no longer use MODsuit modules when your MODsuit isn't fully deployed, unless the modules are intended to be usable when the MODsuit is inactive.
/:cl:

### DIFF
--- a/code/modules/mod/modules/_module.dm
+++ b/code/modules/mod/modules/_module.dm
@@ -99,6 +99,14 @@
 	if(((!mod.active || mod.activating) && !allowed_inactive) || !mod.get_charge()) //SKYRAT ADDITION: INACTIVE USE
 		balloon_alert(mod.wearer, "unpowered!")
 		return FALSE
+	// SKYRAT EDIT START - No using modules when not all parts are deployed.
+	if(!allowed_inactive)
+		for(var/obj/item/part as anything in mod.mod_parts)
+			if(part.loc == mod)
+				balloon_alert(mod.wearer, "deploy all parts first!")
+				return FALSE
+	// SKYRAT EDIT END
+
 	if(!allowed_in_phaseout && istype(mod.wearer.loc, /obj/effect/dummy/phased_mob))
 		//specifically a to_chat because the user is phased out.
 		to_chat(mod.wearer, span_warning("You cannot activate this right now."))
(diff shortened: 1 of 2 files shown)


code/modules/mod/modules/_module.dm
```diff
@@ -99,6 +99,14 @@
 	if(((!mod.active || mod.activating) && !allowed_inactive) || !mod.get_charge()) //SKYRAT ADDITION: INACTIVE USE
 		balloon_alert(mod.wearer, "unpowered!")
 		return FALSE
+	// SKYRAT EDIT START - No using modules when not all parts are deployed.
+	if(!allowed_inactive)
+		for(var/obj/item/part as anything in mod.mod_parts)
+			if(part.loc == mod)
```
Zonespace27, line 105: early continue

Zonespace27 (changes requested):

Acer202:
Aww, could this perhaps exclude the helmet? Having the big ugly helmet up all the time is no fun.

Jezzared:
Does this affect everything, or are things like HUDs attached to helmets and translator gloves to the gloves?

If not, maybe that's something that could be looked into, not needing to full deploy for things like the gloves was nice.

Nerev4r:
Yeah, I think the main issue was people using the modules with the chestpiece retracted. I don't think you should need the helmet or gloves or boots.

GoldenAlpharex:
> Aww, could this perhaps exclude the helmet? Having the big ugly helmet up all the time is no fun.

It then looks even stupider anytime you have something that modifies the look of the visor, or adds literally anything to it. There's currently no support for having certain modules require only certain parts up, but I wouldn't be against someone adding that, I just don't want to spend the time on it right now.

code/modules/mod/modules/_module.dm
GoldenAlpharex: This is already an early return, making it an early-continue would just make the proc bigger and make it less obvious what it aims to do.

PR #15967, opened 2022-09-01
[MIRROR] Save 2.2s minimum (with zero ruins, likely a good bit more in production) of atom init time [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69564
--------------------

## About The Pull Request

1. Pre-sort smoothing_groups and canSmoothWith

Without any ruins, these sorts were taking more than 0.6s, and the bulk of the runtime cost of sortTim during init time.

This only happens on init and they are never changed apart from that, so pre-sorts everything and adds a unit test (in the form of `#ifdef UNIT_TESTS`, because you can't `initial` a list) to ensure that they are proper.

2. Keep visibilityChanged() to mapload only for turf/Initialize

Saves about 0.4s worst case scenario (e.g. with no ruins). Very expensive code (175k loop iterations) for 0 side effects.

3. Space areas now have the fullbright overlay, not the space turfs

Saves about 0.8s worst case scenario. Seems to work fine with starlight.

4. Remove `is_station_level` check for window spawners assigning RCD memory.

Saves about 0.3s worst case scenario. The logic for this isn't consistent since neither walls nor floors check this (for performance), plus some minor micro-opts to spawners.

5. Optimize `is_station_level`

Doubles in speed, used heavily in /turf/open/floor and in other initialization procs. Bit hard to tell exactly how much is saved, though.

:cl:
qol: Windows from outside the station now get the RCD reconstruction benefits, just like walls and floors.
/:cl:


GoldenAlpharex:
Freaking finally!

PR #15972, opened 2022-09-01
Fixes security levels playing incorrect sounds

strandsofivy:
## About The Pull Request

like #68620 on tg, this PR essentially just takes out the duplicate sound parameter
 
essentially just bgug fix ported from commented code on tg

## Changelog

:cl:
fix: Fixed security levels not playing their sounds when switching to them.
/:cl:

### DIFF
--- a/modular_skyrat/modules/alerts/code/security_level_datums.dm
+++ b/modular_skyrat/modules/alerts/code/security_level_datums.dm
@@ -68,5 +68,3 @@
 	elevating_to_configuration_key = /datum/config_entry/string/alert_gamma
 	shuttle_call_time_mod = 0.25
 	sound = 'modular_skyrat/modules/alerts/sound/security_levels/gamma_alert.ogg'
-	looping_sound = 'modular_skyrat/modules/alerts/sound/misc/alarm_delta.ogg'
-	looping_sound_interval = 8 SECONDS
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/alerts/code/security_level_datums.dm
```diff
@@ -68,3 +68,5 @@
 	elevating_to_configuration_key = /datum/config_entry/string/alert_gamma
 	shuttle_call_time_mod = 0.25
 	sound = 'modular_skyrat/modules/alerts/sound/security_levels/gamma_alert.ogg'
+	looping_sound = 'modular_skyrat/modules/alerts/sound/misc/alarm_delta.ogg'
+	looping_sound_interval = 8 SECONDS
```
GoldenAlpharex: Why would we play the delta alarm when we already have the gamma alert sound? If you want the gamma alert to be looping, just make it looping, don't use the delta one as a "placeholder" that'd probably overlap with the regular `gamma_alert.ogg` one (which is by far my favorite alert sound)

GoldenAlpharex (changes requested):

modular_skyrat/modules/alerts/code/security_level_datums.dm
strandsofivy: these short eerie chimes that play once every 8 seconds on delta sounded fitting for me on gamma, and they helped keep up the tension long after the gamma alert sound expires (which is what happened to me during armageddon CMEs)

for reference:
https://user-images.githubusercontent.com/47710522/188508721-8901b78f-d40f-4f95-a080-1e5a51f27541.mp4

I'll remove it if the video still doesn't convince you though :p

GoldenAlpharex: I still think we should keep it to the gamma one instead, and make sure that it loops, because I still think it's one of the most immersive alert sounds we have (I still get the chills every time it plays).

strandsofivy: i'll do that once I am home tmrw
btw, do we have a gamma looping sound?

PR #15974, opened 2022-09-01
[MIRROR] Dimensional Anomaly [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69512
--------------------

## About The Pull Request

Everyone has been asking: "When will there be an anomaly like the bioscrambler, but for the space station? Please, we need more things which replace objects with different objects from the same typepath."
Well I made it and it looked like ass because non-tiling floor and walls look terrible, so then I made this instead.

https://user-images.githubusercontent.com/7483112/187051900-6aca0c32-58ed-469c-92d9-129293129793.mp4

The "dimensional anomaly" shifts matter into a parallel dimension where objects are made out of something else.
Like the Bioscrambler anomaly, it does not expire on its own and only leaves when someone signals it or uses an anomaly remover.
When it spawns it picks a "theme" and converts terrain around it until it covers a 7x7 square, then it teleports somewhere else and picks a new theme.

A lot of these themes are relatively benign like "meat", "fancy carpet", or "gold". Some of them are kind of annoying like "icebox" because it creates floor which slows you down, or "clown" because bananium is intentionally annoying. Some of them are actively dangerous, mostly "uranium" and "plasma".
The main problem this will _usually_ cause for crewmembers is decreasing area security. When it replaces doors it replaces them with ones which don't have any access control, and it will also replace RWalls with normal and much more vulnerable walls which will make breaking and entering significantly easier until someone has taken the time to fix the damage. But also sometimes it will irradiate them, you never know.

The fact that sometimes the changes are benign (or provide uncommon materials) and might be happening in places you don't care about access to might encourage people to push their luck and leave it alone until it starts turning the captain's office into a bamboo room or repainting medbay a fetching shade of flammable purple, which I would consider a success.

https://user-images.githubusercontent.com/7483112/187051988-999af657-abeb-4e27-b36c-d3809c2b4733.mp4

If you successfully harvest the anomaly core you can place it into the reactive armour to get Reactive Barricade Armour, which shifts your dimension when you take damage and attempts to place some randomised (not terribly durable) objects between you and hopefully your attacker (it really just picks up to four random unoccupied tiles next to you). If you're EMPed then the changes it make to the environment will often be as unpleasant for you as they are for a pursuer, and significantly more likely to harm both of you rather than just provide obstacles.

Other changes:
I split anomalies out into their own dmi file, seems to be all the rage lately.
I moved the anomaly placing code into a datum instead of the event because I wanted to reuse it but if you have a better idea about where I could have put it let me know.
This also fixes a bug where the material spreader component wasn't working when I applied plasma materials to something, the extra whitespace was parsing as another argument for some reason and meant it would runtime.
Supermatter delamination was still pointing to Delimber anomalies instead of Bioscrambler.

## Why It's Good For The Game

It's enrichment for engineers which doesn't (usually) cause a hull breach (unless it replaces glass windows with paper ones).
It creates novel openings for antagonists by messing with area security.
It adds another source of a few of the more odd kinds of materials in the game and might show players content they weren't aware existed (like making furniture out of pizza).
It's an event which can be a little bit of a pain in the ass but also probably won't kill a bunch of people.
It's funny (this is my real motivation).

## Changelog

:cl:
add: A new kind of anomaly which transforms the area into some other kind of material.
fix: Supermatter delaminations can spawn bioscramblers again.
/:cl:


code/game/objects/effects/anomaly_dimensional_themes.dm
```diff
@@ -0,0 +1,328 @@
...
+	icon = 'icons/obj/food/meat.dmi'
+	icon_state = "meat"
+	material = /datum/material/meat
+	sound = 'sound/items/eatfood.ogg'
+
+/datum/dimension_theme/pizza
+	icon = 'icons/obj/food/pizza.dmi'
+	icon_state = "pizzamargherita"
+	material = /datum/material/pizza
+	sound = 'sound/items/eatfood.ogg'
```
GoldenAlpharex, line 231: I got a sneaky suspicion those four are going to cause us troubles in the future.

GoldenAlpharex (approved):
Should be good to go, it was failing because of another PR that I believe since then has been merged.

code/game/objects/effects/anomaly_dimensional_themes.dm
tf-4: _Couldn't possibly imagine what you mean_.

PR #16002, opened 2022-09-01
[MIRROR] Wings and tails give anyone better tackling values. Moths and lizards that lose theirs will have worse values instead. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69078
--------------------

## About The Pull Request
Late and atomized version of #67791 with some extra changes.

Changes are:

1 - Anyone with wings will have a +2 tackle offense (same as riot armor or gigantism mutation)
2 - Moths without or burnt wings will have -2 tackle offense (same as dwarfism or clumsy)
3 - Anyone with a lizard tail that is wagging it, will have 1+ tackle defense (same as being fat)
4 - Lizards without a tail will have -1 tackle defense (same as being slightly drunk)
5 - Fly people have a higher chance of breaking their spine after splatting themselves into a window/wall (+6 same as clumsy)

Made 3 new traits and gave them to Lizards, Moths and Flypeople, swapped the old `islizard` to check for one of these traits instead.

This way it is easier for anyone that reads these species' code to know that they have an unique interaction with our tackling system without having to dig into component code.
And thanks for the suggestion Kapu.

## Why It's Good For The Game
This makes Moths have an interesting round start niche as security and push them to behave differently than humans if they decide to opt into tackling.

It is fun to steal flashbangs and exploit the weakness that Moth officers have right now, if moths are better at tackling and more likely to use it, it opens another thing you can exploit when engaging against security.

The breakpoints for tackling are on windows of 2 so while it might look like a lot (Riot suit value), if I gave only 1 for moths they would have a 50% chance of being better than other species while tackling.
Inconsistencies are not interesting on species. It would be quite annoying if plasmamen had a coin toss on being cold resistant or self igniting every time they engage with the system.

The flypeople change is there just because the idea of a Fly officer smacking themselves into a window and breaking their spine is funny.

And as was requested by maintainers, it now checks for anyone with wings and lizard tails for the tackle buff, go have your fun Frankstein builders and Officers that can get Xeno to give you a wing potion.

## Changelog

:cl: Guillaume Prata
balance: Anyone with wings or a wagging lizard tail will have a small bonus to their tacklling offense/defense.
balance: Moths and lizards that lose their wings/tail will have worse tackling values instead.
balance: Fly people don't take splatting themselves on windows/walls easily and have a higher chance to get a bad result from that.
/:cl:


GoldenAlpharex (approved):
Should work now.

PR #16011, opened 2022-09-02
[PR PR][MDB IGNORE] merge #15865

Tastyfish:
Fixes merge conflicts for #15865

Also fixes that we used to use `world.realtime` where `REALTIMEOFDAY` is the appropriate and accurate real time value for this context.

Update: ran the update path script as well


Tastyfish:
mapdiffbot is just uphappy because the original PR state had the wrong paths in the SR maps

tf-4:
I should really pay more attention, mb

PR #16022, opened 2022-09-03
Admin action to send play in cryo

AnywayFarus:
## About The Pull Request

This PR adds an admin action to send the player into a cryopod.

Can help / simplify the work for the administration, so as not to drag lazy players into cryo by infiltrating their character.

## Changelog

:cl: Vishenka0704
admin: add send to cryo function in vv dropdown list
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -856,6 +856,7 @@
 	VV_DROPDOWN_OPTION(VV_HK_MOD_QUIRKS, "Add/Remove Quirks")
 	VV_DROPDOWN_OPTION(VV_HK_SET_SPECIES, "Set Species")
 	VV_DROPDOWN_OPTION(VV_HK_PURRBATION, "Toggle Purrbation")
+	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send in Cryopod") //SKYRAT EDIT ADDITION - CRYO SEND
 
 /mob/living/carbon/human/vv_do_topic(list/href_list)
 	. = ..()
@@ -936,6 +937,17 @@
 			var/msg = span_notice("[key_name_admin(usr)] has removed [key_name(src)] from purrbation.")
 			message_admins(msg)
 			admin_ticket_log(src, msg)
+	//SKYRAT EDIT ADDITION BEGIN - CRYO SEND
+	if(href_list[VV_HK_SEND_CRYO])
+		if(!check_rights(R_SPAWN))
+			return
+		src.send_to_cryo()
+		to_chat(usr, "Put [src] in cryopod.")
+		log_admin("[key_name(usr)] has put [key_name(src)] to cryopod.")
+		var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] in cryopod.")
+		message_admins(msg)
+		admin_ticket_log(src, msg)
+	//SKYRAT EDIT ADDITION END
 
 /mob/living/carbon/human/limb_attack_self()
 	var/obj/item/bodypart/arm = hand_bodyparts[active_hand_index]
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -856,7 +856,7 @@
 	VV_DROPDOWN_OPTION(VV_HK_MOD_QUIRKS, "Add/Remove Quirks")
 	VV_DROPDOWN_OPTION(VV_HK_SET_SPECIES, "Set Species")
 	VV_DROPDOWN_OPTION(VV_HK_PURRBATION, "Toggle Purrbation")
-	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send in Cryopod") //SKYRAT EDIT ADDITION - CRYO SEND
+	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send to Cryogenic Storage") //SKYRAT EDIT ADDITION - CRYO SEND
 
 /mob/living/carbon/human/vv_do_topic(list/href_list)
 	. = ..()
@@ -941,9 +941,9 @@
 	if(href_list[VV_HK_SEND_CRYO])
 		if(!check_rights(R_SPAWN))
 			return
-		src.send_to_cryo()
+		send_to_cryo()
 		to_chat(usr, "Put [src] in cryopod.")
-		log_admin("[key_name(usr)] has put [key_name(src)] to cryopod.")
+		log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
 		var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] in cryopod.")
 		message_admins(msg)
 		admin_ticket_log(src, msg)
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -941,12 +941,20 @@
 	if(href_list[VV_HK_SEND_CRYO])
 		if(!check_rights(R_SPAWN))
 			return
-		send_to_cryo()
-		to_chat(usr, "Put [src] in cryopod.")
-		log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
-		var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] in cryopod.")
-		message_admins(msg)
-		admin_ticket_log(src, msg)
+		var/choisen = tgui_alert(usr, "Add a paper notice about sending [name] into a cryopod?", "Leave a paper?", list("Yes", "No", "Cancel"))
+		if(choisen == "Cancel" || !choisen)
+			return
+		else
+			//log/message
+			to_chat(usr, "Put [src] in cryopod.")
+			log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
+			var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] into a cryopod from [ADMIN_VERBOSEJMP(src)].")
+			message_admins(msg)
+			admin_ticket_log(src, msg)
+			
+			choisen = choisen == "Yes"
+		send_to_cryo(choisen)
+
 	//SKYRAT EDIT ADDITION END
 
 /mob/living/carbon/human/limb_attack_self()
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -856,7 +856,6 @@
 	VV_DROPDOWN_OPTION(VV_HK_MOD_QUIRKS, "Add/Remove Quirks")
 	VV_DROPDOWN_OPTION(VV_HK_SET_SPECIES, "Set Species")
 	VV_DROPDOWN_OPTION(VV_HK_PURRBATION, "Toggle Purrbation")
-	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send to Cryogenic Storage") //SKYRAT EDIT ADDITION - CRYO SEND
 
 /mob/living/carbon/human/vv_do_topic(list/href_list)
 	. = ..()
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -936,6 +937,17 @@
 			var/msg = span_notice("[key_name_admin(usr)] has removed [key_name(src)] from purrbation.")
 			message_admins(msg)
 			admin_ticket_log(src, msg)
+	//SKYRAT EDIT ADDITION BEGIN - CRYO SEND
+	if(href_list[VV_HK_SEND_CRYO])
+		if(!check_rights(R_SPAWN))
+			return
+		src.send_to_cryo()
+		to_chat(usr, "Put [src] in cryopod.")
+		log_admin("[key_name(usr)] has put [key_name(src)] to cryopod.")
```
GoldenAlpharex: I'd like it if you could add the location to this log as well.

Also, the content of the log itself should be stored into a variable instead, so that you can add the `span_notice()` around it only for the `message_admins()` portion.
```suggestion
		log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,38 @@
...
+	var/datum/effect_system/spark_spread/quantum/sparks = new //I don't like having to duplicate this
+	sparks.set_up(10, 1, src.loc)
+	sparks.attach(src.loc)
+	sparks.start()
+	return ..()
+
+// give protection from move to player, and give info to portal
+/mob/living/carbon/human/proc/send_to_cryo() 
+	src.move_resist += 1000
+	src.Stun(6 SECONDS, TRUE)
```
GoldenAlpharex:
```suggestion
	move_resist += 1000
	Stun(6 SECONDS, TRUE)
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,38 @@
...
+				break
+
+	playsound(src.loc, 'sound/magic/Repulse.ogg', 100, 1)
+	var/datum/effect_system/spark_spread/quantum/sparks = new //I don't like having to duplicate this
+	sparks.set_up(10, 1, src.loc)
+	sparks.attach(src.loc)
+	sparks.start()
+	return ..()
+
+// give protection from move to player, and give info to portal
```
GoldenAlpharex:
```suggestion
/// Give protection from move to player, and give info to portal
```

modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -160,6 +162,8 @@ GLOBAL_LIST_EMPTY(ghost_records)
 
 /obj/machinery/cryopod/Initialize(mapload)
 	..()
+	if(!quiet && src.z != 2) // maybe I should change the list to not_quiet_cryopods and dont block interlink cryo?
```
GoldenAlpharex: I think your comment is valid, and you could just make the list be `GLOB.valid_cryopods` instead.
```suggestion
	if(!quiet)
```

modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -168,6 +172,8 @@ GLOBAL_LIST_EMPTY(ghost_records)
 
 // This is not a good situation
 /obj/machinery/cryopod/Destroy()
+	if(!quiet && src.z != 2)
+		GLOB.station_cryopods -= src
```
GoldenAlpharex: We don't need to have a check for this, BYOND is actually smart enough to not cause an error when trying to remove something that wasn't present in the list to begin with.
```suggestion
	GLOB.station_cryopods -= src
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,38 @@
...
+/mob/living/carbon/human/proc/send_to_cryo() 
+	src.move_resist += 1000
+	src.Stun(6 SECONDS, TRUE)
+	var/obj/effect/portal/cryo/portal = new /obj/effect/portal/cryo(get_turf(src))
+	portal.target = src
+			
+
```
GoldenAlpharex: My question for all of this:

Why do we have to go through having some sort of portal appearing like this? Couldn't we just have the exact same effect we already have with the admin deletion sparks, and immediately send them to cryo, rather than waiting five seconds for it? I think it'd look a little less weird, and would behave a lot more intuitively than a portal that you'd expect comes from a hand teleporter taking them to cryo.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -936,6 +937,17 @@
 			var/msg = span_notice("[key_name_admin(usr)] has removed [key_name(src)] from purrbation.")
 			message_admins(msg)
 			admin_ticket_log(src, msg)
+	//SKYRAT EDIT ADDITION BEGIN - CRYO SEND
+	if(href_list[VV_HK_SEND_CRYO])
+		if(!check_rights(R_SPAWN))
+			return
+		src.send_to_cryo()
```
GoldenAlpharex: DM already implicitly does the `src.` here, you don't need to explicitly add it.
```suggestion
		send_to_cryo()
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -856,6 +856,7 @@
 	VV_DROPDOWN_OPTION(VV_HK_MOD_QUIRKS, "Add/Remove Quirks")
 	VV_DROPDOWN_OPTION(VV_HK_SET_SPECIES, "Set Species")
 	VV_DROPDOWN_OPTION(VV_HK_PURRBATION, "Toggle Purrbation")
+	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send in Cryopod") //SKYRAT EDIT ADDITION - CRYO SEND
```
GoldenAlpharex: Should be a bit more intuitive that way.
```suggestion
	VV_DROPDOWN_OPTION(VV_HK_SEND_CRYO, "Send to Cryogenic Storage") //SKYRAT EDIT ADDITION - CRYO SEND
```

GoldenAlpharex (changes requested):
Even if I left a lot of comments, I reaaaaally like this PR, it's going to be a big administrative QoL for sure.

modular_skyrat/modules/cryosleep/code/admin.dm
AnywayFarus: deleted

modular_skyrat/modules/cryosleep/code/admin.dm
AnywayFarus: deleted)

modular_skyrat/modules/cryosleep/code/admin.dm
AnywayFarus: changed

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -936,6 +937,25 @@
 			var/msg = span_notice("[key_name_admin(usr)] has removed [key_name(src)] from purrbation.")
 			message_admins(msg)
 			admin_ticket_log(src, msg)
+	//SKYRAT EDIT ADDITION BEGIN - CRYO SEND
+	if(href_list[VV_HK_SEND_CRYO])
+		if(!check_rights(R_SPAWN))
+			return
+		var/choisen = tgui_alert(usr, "Add a paper notice about sending [name] into a cryopod?", "Leave a paper?", list("Yes", "No", "Cancel"))
+		if(choisen == "Cancel" || !choisen)
```
GoldenAlpharex:
```suggestion
		var/send_notice = tgui_alert(usr, "Add a paper notice about sending [name] into a cryopod?", "Leave a paper?", list("Yes", "No", "Cancel"))
		if(send_notice != "Yes" && send_notice != "No")
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -936,6 +937,25 @@
...
+		else
+			//log/message
+			to_chat(usr, "Put [src] in cryopod.")
+			log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
+			var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] into a cryopod from [ADMIN_VERBOSEJMP(src)].")
+			message_admins(msg)
+			admin_ticket_log(src, msg)
+			
+			choisen = choisen == "Yes"
+		send_to_cryo(choisen)
```
GoldenAlpharex: Since you did an early return, you don't need the `else` statement, the goal of doing those is to help reduce indentation.
```suggestion

		//log/message
		to_chat(usr, "Put [src] in cryopod.")
		log_admin("[key_name(usr)] has put [key_name(src)] into a cryopod.")
		var/msg = span_notice("[key_name_admin(usr)] has put [key_name(src)] into a cryopod from [ADMIN_VERBOSEJMP(src)].")
		message_admins(msg)
		admin_ticket_log(src, msg)

		send_to_cryo(send_notice == "Yes")
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,22 @@
+// Send player in not-quiet cryopod. If with_paper = TRUE, place a paper with notification under player.
```
GoldenAlpharex: It needs to be `///` and not `//` for it to count as documentation for the autodoc parser :)
```suggestion
/// Send player in not-quiet cryopod. If with_paper = TRUE, place a paper with notification under player.
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,22 @@
...
+/mob/living/carbon/human/proc/send_to_cryo(with_paper = FALSE)
```
GoldenAlpharex: Does it really have to be just humans? Couldn't we just make it be at `/mob` level (if any mob can be sent to cryo)?

Otherwise, that will still mean we won't be able to send silicons to cryo using this, nor sentient simple mobs (albeit I don't know if they could cryo to begin with)

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,22 @@
+// Send player in not-quiet cryopod. If with_paper = TRUE, place a paper with notification under player.
+/mob/living/carbon/human/proc/send_to_cryo(with_paper = FALSE) 
+	//effect
+	playsound(src.loc, 'sound/magic/Repulse.ogg', 100, 1)
+	var/datum/effect_system/spark_spread/quantum/sparks = new 
+	sparks.set_up(10, 1, src.loc)
+	sparks.attach(src.loc)
```
GoldenAlpharex:
```suggestion
	playsound(loc, 'sound/magic/Repulse.ogg', 100, 1)
	var/datum/effect_system/spark_spread/quantum/sparks = new 
	sparks.set_up(10, 1, loc)
	sparks.attach(loc)
```

modular_skyrat/modules/cryosleep/code/admin.dm
```diff
@@ -0,0 +1,22 @@
...
+	sparks.start()
+
+	//make a paper if need
+	if(with_paper)
+		var/obj/item/paper/cryo_paper = new /obj/item/paper(loc)
+		cryo_paper.name = "Notification - [name]"
+		cryo_paper.add_raw_text("Our sincerest apologies, [name], [job ? "[job]," : ""] had to be sent back in Cryogenic Storage for reasons that cannot be elaborated on at the moment.<br><br>Sincerely,<br><b>NanoTrasen Anti-Sudden Sleep Disorder Agency</b>")
```
GoldenAlpharex:
```suggestion
		cryo_paper.add_raw_text("Our sincerest apologies, [name][job ? ", [job]," : ""] had to be sent back in Cryogenic Storage for reasons that cannot be elaborated on at the moment.<br><br>Sincerely,<br><i>NanoTrasen Anti-Sudden Sleep Disorder Agency</i>")
```

modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -11,6 +11,8 @@ GLOBAL_LIST_EMPTY(cryopod_computers)
 
 GLOBAL_LIST_EMPTY(ghost_records)
 
+GLOBAL_LIST_EMPTY(valid_cryopods)
```
GoldenAlpharex, line 15:
```suggestion
/// A list of all cryopods that aren't quiet, to be used by the "Send to Cryogenic Storage" VV action.
GLOBAL_LIST_EMPTY(valid_cryopods)
```

GoldenAlpharex (changes requested):
Yeah, I've been curious if we couldn't make it so this could be used on all mobs, and not just humans (as it'd probably be even more useful). If sentient mobs can't go in cryo, then at the very least, making it usable on silicons as well.

GoldenAlpharex (approved):
Looks good to me, just going to ask you one thing in DMs first.

GoldenAlpharex (approved):
We're good. Let's send it!

PR #16028, opened 2022-09-03
[MIRROR] Limit icon_states checks in spritesheet creation to UNIT_TESTS (saves 1.6s of init time) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69635
--------------------

## About The Pull Request
Limits a branch that always runtimes to UNIT_TESTS only. This runtime has only ever occurred in one round this entire year, where it seems everything was exploding anyway. Any time this would runtime should be caught by unit tests, as this proc is only ran once. Saves 0.9s.

Does the same to research_designs, and promotes its warning to a stack trace so we can see it better, since it actually has failed recently (but nobody actually reads the logs for it, apparently). Saves 0.7s.


GoldenAlpharex (approved):
Should pass CI now that it's updated!

PR #16033, opened 2022-09-03
Buffs agent ID, removes deluxe agent ID

tf-4:
## About The Pull Request

Removes the deluxe agent ID card from the code, restoring `/obj/item/card/id/advanced/chameleon/black` to what it is on stock /tg/.

Buffs the regular agent ID from `3, 1, 1` to `4, 2, 1`. That's four regular accesses, two command accesses, and one captain access.

The deluxe agent ID card, while somewhat more expensive than the basic agent ID (5TC vs 2TC) was *very* much overtuned. The ability to get all access just by standing next to the Captain for a few seconds while you tick all the little access boxes, not even having to steal their ID, is *extremely* powerful.
The entire point of the ID rework was to move away from people having 'hidden AA', AKA all access and zero way to know if they have it. If you want access into somewhere, you should have to go a bit louder than that either by social manipulation or more obvious tools like jaws, doorjacks, or explosives.
Being able to silently slip in and out of anywhere you like after such little effort needed to be put in to acquire it is not something resembling balance.

That said. The stock agent ID is very much lacklustre in my opinion so I've given it a tentative additional access slot on the common and command levels. This could go up more in the future, but I'd rather not do too much in one go.

## Changelog
:cl:
del: Removed deluxe agent IDs.
balance: Regular agent IDs can now hold one addition common-level access and one additional command-level access.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/id_cards.dm
+++ b/code/__DEFINES/~skyrat_defines/id_cards.dm
@@ -1,0 +1,6 @@
+/// Buffed wildcard slot define for Chameleon/Agent ID grey cards. Can hold 4 common, 2 command and 1 captain access.
+#define WILDCARD_LIMIT_CHAMELEON_PLUS list( \
+	WILDCARD_NAME_COMMON = list(limit = 4, usage = list()), \
+	WILDCARD_NAME_COMMAND = list(limit = 2, usage = list()), \
+	WILDCARD_NAME_CAPTAIN = list(limit = 1, usage = list()) \
+)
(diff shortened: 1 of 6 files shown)


GoldenAlpharex (commented):
Looks good to me, I also thought it was always overtuned.

GoldenAlpharex (approved):
Looks like I'm unable to press the "Approve" button huh

weesnaw-ops:
You can't just stand next to the captain and touch him and it immediately took his access I thought. I thought his ID had to manually be on the ground and you had to swipe over it? At which point it would literally be the same as taking the ID since you're within grab range?

PR #16066, opened 2022-09-05
[NON-MODULAR] fix brig door merged logic

Tastyfish:
## About The Pull Request

Fixes tf4's manual merging of my fixes into #15865, which sort of kept upstream timing and previous SR timing, neither of which being what SR actually intended and currently the doors are broken with the combination of the two (it will give a countdown of multiple years).

As stated my the fix PR I originally made, to explain using `REALTIMEOFDAY`, `world.realtime` should only be used if you absolutely need something resembling a full real world time and date, but it's only precise to the nearest 7 seconds or so (optimistically) and shouldn't be used for in-game timing. (That's why the `REALTIMEOFDAY` macro was created.)

Bugfix

## Changelog

Unnecessary if this is merged before the server updates to have the previous PR.

### DIFF
--- a/code/game/machinery/doors/brigdoors.dm
+++ b/code/game/machinery/doors/brigdoors.dm
@@ -70,7 +70,7 @@
 	if(!timing)
 		return PROCESS_KILL
 
-	if(world.time - activation_time >= timer_duration)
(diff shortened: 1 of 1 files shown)


tf-4 (approved):
👍 Thank you, sry I missed the original pr

Tastyfish:
> 👍 Thank you, sry I missed the original pr

You're good.

PR #16086, opened 2022-09-06
[MIRROR] Stomach get_availability() now checks for NO_HUNGER instead of broken trait that no mobs have. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69674
--------------------

## About The Pull Request
Stomachs will check for the TRAIT_NO_HUNGER to see if they should be in a species by default. It used to check for NOSTOMACH in innate traits, but stupidly enough NOSTOMACH was defined as a species trait not an innate one. Its been like this for _years_. It hasn't caused any bugs though, because a total of zero (0) specise have this trait. It is completely unused and broken at it has existed for five (5) years.
Using the new organ trait system, The nohunger trait on plasmamen is now attached to their lungs rather than their innate traits so they can keep their boney bundle  I touched up the ethereal's no hunger stomach while I was at it.
Also makes the appendix check to see if the species has blood, because skeletons having appendixes are dumb.

## Why It's Good For The Game
Creatures who don't eat don't have stomachs.

## Changelog

:cl:
fix: Would you believe me if I told you androids had functioning human stomachs in them for the past 5 years? No? Well anyways its fixed.
/:cl:


Zonespace27:
manual mirroring tgstation/tgstation#68399

GoldenAlpharex:
Let's goooooo!

GoldenAlpharex (approved):

PR #16115, opened 2022-09-07
Reduces the downsides of borers

Kubisopplay:
## About The Pull Request
Reduces the absolutely unreasonable amounts of brain damage caused by borer abilities, allowing for potential coexistence once again. Gives them ability to learn mannitol once again, allowing for friendly borers to stay somewhat hidden.

Stops borers from being murderworms. Adds any opportunity for RP other then finding yourself dead with 10 traumas.

## Changelog

:cl:
balance: Turns borer lethality down a bit
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
@@ -78,8 +78,6 @@
 			cortical_owner.reagent_holder.reagents.add_reagent(reagent, cortical_owner.injection_rate_current, added_purity = 1)
 			cortical_owner.reagent_holder.reagents.trans_to(cortical_owner.human_host, cortical_owner.injection_rate_current, methods = INGEST)
 			var/obj/item/organ/internal/brain/victim_brain = cortical_owner.human_host.getorganslot(ORGAN_SLOT_BRAIN)
-			if(victim_brain)
-				cortical_owner.human_host.adjustOrganLoss(ORGAN_SLOT_BRAIN, 5)
 			to_chat(cortical_owner.human_host, span_warning("You feel something cool inside of you and a dull ache in your head!"))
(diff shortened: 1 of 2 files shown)


Higgin:
I don't think you should do this, and I don't think you should do it quite like this if you are -

At best, if you want to make it a bit easier to manage, I'd say unblacklist mannitol, don't add it to the basic "chems you get for free" list (so there's something to guide you in getting your host to cooperate by finding some - I could go either way on unblacklisting neurine since it allows negating a lot of the poor management practices of basic traumas, but it does only negate basic.) 

I'd leave the damage #s alone for the time being seeing as you're trying to add a way to negate the damage. If you do anything else, maybe add a longer CD to evolve/egg? As a way to prevent stacking up big damage all at once with saved points.

Kubisopplay:
This is not a final balance PR, it is just so we don't need to delete borers on lowpop fearing a random murderbone. I want to have it on the safer side, since it's much better to have them less dangerous, then more. If you have a better idea you can make a PR too, it's just few lines anyway

modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
```diff
@@ -78,8 +78,6 @@
 			cortical_owner.reagent_holder.reagents.add_reagent(reagent, cortical_owner.injection_rate_current, added_purity = 1)
 			cortical_owner.reagent_holder.reagents.trans_to(cortical_owner.human_host, cortical_owner.injection_rate_current, methods = INGEST)
 			var/obj/item/organ/internal/brain/victim_brain = cortical_owner.human_host.getorganslot(ORGAN_SLOT_BRAIN)
-			if(victim_brain)
-				cortical_owner.human_host.adjustOrganLoss(ORGAN_SLOT_BRAIN, 5)
```
jjpark-kb, line 82: with this removal, you don't need the above variable.

Kubisopplay: Fixed

PR #16116, opened 2022-09-07
[MIRROR] [MDB IGNORE] Shuttle engine code improvement and fixes [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69516
--------------------
This is a remake of https://github.com/tgstation/tgstation/pull/69362 because github desktop has fucked over my branch 

## About The Pull Request

I hated how shuttles worked while I was working with them on a downstream, so instead I gave up and started refactoring shit here.

My original goal was solely just to make engines handle being added to lists themselves, but it turned into this instead.

* Makes use of ``as anything`` in many places
* Renames many vars, including shuttle 'id' var to 'shuttle_id' and engine 'state' to 'engine_state', so it is easier to search for them and manage them at once.
* Engines now weakref their attached ship, and disconnect when unwrenched from it, allowing engines to do more things to their shuttle without constantly checking for being inside of one.
* Removed shuttles constantly checking for how many engines they are, the engine will add themselves to the list instead.
* Adds comments to port_id and shuttle_id to help understand what is needed with these two vars, because currently admins can only replace Emergency shuttles and Ferries, the rest are broken entirely, because people named them differently.
* Adds a ``mapload`` arg to ``connect_to_shuttle()`` for when a shuttle is first created and is syncing machines to it, since in some occasions, stuff may be connected to shuttles afterwards.

In this PR, I fixed Cargo, Arrivals, and Pirate ships. Emergency and Whiteship ones are already functional, the rest I am leaving because:
1-lazy
2-they arent really that important
3-some shouldn't even be replacing eachother? fugitive hunters one seems a little odd, and I don't really want to get too deep into that, I wanted to focus more on the ship parts itself.

## Why It's Good For The Game

I hate the current state of shuttles, especially when they are straight-up broken. Hopefully I can make it slightly better and make it easier to work with in the future, while fixing a few things on the way. Like my vault console PR, this is somewhat selfish since it will also help downstream, for a voidcrew TG rebase.

## Changelog

:cl:
refactor: Shuttle engines now handle engine stuff themselves, rather than delegating it to the shuttle, which would periodically scan all their tiles for engines.
fix: Cargo/Arrival/Pirate shuttles are now admin-replacable, and won't stack on top of eachother.
/:cl:


Paxilmaniac:
holy shit it touches every fucking shuttle file

Paxilmaniac:
@GoldenAlpharex I'm bringing you here because this touches a bit of scary code (shuttles)

PR #16127, opened 2022-09-08
improves escape shuttle accessibility

press-mark:
## About The Pull Request
An emergency exit in a structure is a special exit for emergencies such as a [fire](https://en.wikipedia.org/wiki/Structure_fire): the combined use of regular and special exits allows for faster [evacuation](https://en.wikipedia.org/wiki/Emergency_evacuation), while it also provides an alternative if the route to the regular exit is blocked.

The qualifications for an emergency exit are as follows: it must be in a location that is easily accessible, the exit must have an area or location that it can bring people to in the event of any emergency, it must be controlled by the inside of the building, it must be well managed and regularly up kept, and it must be in a permanent location.

It is usually in a strategically located (e.g. in a [stairwell](https://en.wikipedia.org/wiki/Stairwell), [hallway](https://en.wikipedia.org/wiki/Hallway), or other likely places) outward opening door with a [crash bar](https://en.wikipedia.org/wiki/Crash_bar) on it and with [exit signs](https://en.wikipedia.org/wiki/Exit_sign) leading to it. The name is a reference to when they were frequently used, however, a fire exit can also be a main doorway must be able to be unlocked from the inside of the room. A [fire escape](https://en.wikipedia.org/wiki/Fire_escape) is a special kind of emergency exit, mounted to the outside of a building.

allows 50 people to use 3 doors instead of 1 door in 3 minutes

image

**New shuttle on the left. Current shuttle on the right.**

## Changelog

:cl:
qol: You can now use 2 more doors to escape on the shuttle.
/:cl:


ghost:
image
Love the PR, hate the sprite issues :( Damn these single-acceptable-shape shuttles

ghost:
Also don't remove the emergency lockers please!! They (should) have 4dirs so you can rotate them to fit in A or B (B needing a hole filled with a wall. Again, damn this shuttle.)
image
image

press-mark:
okay here's an update:
image

i think stretching the shuttle to the north by one tile is probably going to be the best solution
the """"other"""" option is to make an automap template and remap departures on Every Single TG Map. but that would be dreadful

note that the tiles i highlighted will need a spriter's touch-ups

Paxilmaniac:
This is a prime example of why I don't like having sprites for each individual tile on shuttles, it adds just far too much sprite and map overhead for when someone wants to make the smallest shuttle change.

press-mark:
closing in favour of https://github.com/Skyrat-SS13/Skyrat-tg/pull/16171

PR #16144, opened 2022-09-09
[MIRROR] Makes AnimatedNumber overperform (at sixty frames per second) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69731
--------------------

## About The Pull Request

Following discussion from #69665, this replaces the animation algorithm of `AnimatedNumber` with the version I put together for techfabs. It also cleans up every remaining usage of `AnimatedNumber`.

Here's a chem dispenser:

https://user-images.githubusercontent.com/12202230/188538741-fec6613c-5f25-4dfc-9982-1e48420ccfe0.mp4

**However**, _this PR also removes the ability to specify custom children inside of an `AnimatedNumber`._ This comes as a result of the performance hit of re-rendering (even very small) inferno trees at 60 Hz. I've taken the liberty of updating any code depending on this behavior--in virtually every case, the `AnimatedNumber` was unnecessary or could be moved into the component hierarchy (instead of acting as a container).

## Why It's Good For The Game

Labels now update at sixty frames per second. There's also a solid performance gain here, since labels stop animating when it's reasonable, and we aren't hitting inferno's vdom diffing at all.

## Changelog

:cl:
refactor: animated numbers now animate at 60 fps
/:cl:


tf-4 (approved):
Nice

PR #16153, opened 2022-09-09
Return of APC arcing, modular edition

tf-4:
## About The Pull Request

This was lost in an APC refactor at some point, now it's un-lost

Broken features ... bad!

## Changelog
:cl:
fix: APCs will now arc again when grid excess exceeds 2.5MW. Use a bronze sheet on an APC with the wires exposed to install shielding!
/:cl:


code/modules/power/apc/apc_attack.dm
```diff
@@ -167,6 +167,20 @@
 		wires.interact(user)
 		return
 
+	// SKYRAT EDIT ADD START - APC ARCING
```
RimiNosha: This can be easily modular, by the way, the `else` on the next line is unnecessary, as well.

ghost:
Pleaseeee add some examines of some sort, at least when its open, to say 
- if its arc shielded
- how to shield it if its not
- how to remove a shield if it is

Iamgoofball:
i don't like arcing

tf-4:
> Pleaseeee add some examines of some sort, at least when its open, to say
> 
>     * if its arc shielded
> 
>     * how to shield it if its not
> 
>     * how to remove a shield if it is

good idea

modular_skyrat/modules/apc_arcing/apc.dm
```diff
@@ -0,0 +1,55 @@
...
+		shock_chance = 15
+	else if(excess >= APC_ARC_MEDIUMLIMIT)
+		shock_chance = 10
+	if(prob(shock_chance))
+		var/list/shock_mobs = list()
+		for(var/creature in view(get_turf(src), 5)) // We only want to shock a single random mob in range, not all.
+			if(isliving(creature))
+				shock_mobs += creature
+		if(shock_mobs.len)
+			var/mob/living/living_target = pick(shock_mobs)
```
Kapu1178:
```suggestion
	if(prob(shock_chance))
		var/list/shock_mobs = list()
		for(var/mob/living/creature in viewers(get_turf(src), 5)) // We only want to shock a single random mob in range, not all.
			shock_mobs += creature
		if(length(shock_mobs))
			var/mob/living/living_target = pick(shock_mobs)
```

tf-4: Won't `viewers()` mean the blind are arc-immune? That doesn't make sense.

Kapu1178: > Won't `viewers()` mean the blind are arc-immune? That doesn't make sense.

byond's concept of blindness isnt what ss13 uses for blind users. blindness is for BYOND darkness, aka byond-FOV. You can also use hearers() to ignore this, which is basically just mobs-in-range.

PR #16169, opened 2022-09-11
Remove duplicated buildmode buttons

AnywayFarus:
## About The Pull Request

Removed unnecessary icons, leaving only the necessary one, and made a way to select the desired path to the icon

Oh, maybe a little bit faster load on server to start RP.

## Changelog

:cl: Vishenka0704
add: Added ways to determine the modularity of an icon in builmode
del: Removed duplicated buildmode icons
/:cl:


code/modules/buildmode/bm_mode.dm
```diff
@@ -5,7 +5,7 @@
 
 /datum/buildmode_mode
 	var/key = "oops"
-
+	var/button_icon = 'icons/misc/buildmode.dmi' // SKYRAT ADDITION - remove buildmode duplicate icons, divine tg and skyrat
```
GoldenAlpharex: This can be defined in the modular file for buildmode mode datums, with the offercontrol one. It should also be using /// to explain what the variable is there for.

GoldenAlpharex (changes requested):

GoldenAlpharex (approved):
Looks good to me!

PR #16197, opened 2022-09-13
Removes obsolete Skyrat_Wiki variable (Fixes CorpReg link) and sets wikiurl default config to the Skyrat wiki (helps test servers/downstreams)

ghost:
## About The Pull Request
So. 
Uh.
We fixed this already at some point, but, uh.
There was an extra space in our link. Our link we don't actually need to use.

Because we already changed the `wikiurl` config to the Skyrat wiki, the entire Skyrat_Wiki variable was obsolete. I've axed it, and to help private servers or downstreams, wikiurl also has a default config of the Skyrat Wiki

_**MAINTAINERS READ THIS NOTE:**_
I've no clue how to modularly do default config OVERWRITES, so i've just done the semi-modular comments. If this NEEDS to be changed, you may have to do it yourself, I'll be out of town until the 20th and even then will be busy.

I sure love reading my favorite Corporate Regulations - oh it's blank.

## Changelog

:cl:
fix: After several failed attempts, Nanotrasen has simply begun outsourcing the printing of their Corporate Regulations (alongside plenty of NDAs of course). Expect the booklets to actually have words from now on!
/:cl:


code/game/objects/items/manuals.dm
```diff
@@ -248,13 +247,13 @@
...
 	//var/wikiurl = CONFIG_GET(string/wikiurl) //ORIGINAL
-	//SKYRAT EDIT CHANGE BEGIN
+	//SKYRAT EDIT ADDITION START - Wiki overrides
 	var/wikiurl
 	if(skyrat_wiki)
 		wikiurl = CONFIG_GET(string/wikiurlskyrat)
 	else
 		wikiurl = CONFIG_GET(string/wikiurl)
-	//SKYRAT EDIT CHANGE END
+	//SKYRAT EDIT ADDITION END - Wiki overrides
```
GoldenAlpharex: So, I did some more digging on the server, and turns out that this is entirely unnecessary, because we have nearly all of the pages already on our wiki, with edits to account for Skyrat-exclusive content as well.

The proper fix here would just be to remove the whole `wikiurlskyrat` config, as it seems it was never used on the server either way.

I verified, and turning the `skyrat_wiki` variable to `0` on the Corporate Regulations book fixes it, and makes it display what it's meant to display.

To make it work on test-servers too, you could instead change the default value of the `wikiurl` entry, with a modular comment (or completely modularly, up to you).

GoldenAlpharex (changes requested):

code/game/objects/items/manuals.dm
ghost: God this has been an adventure

If I override it to default to the skyrat wiki, I have to add 'wiki.' to the beginning lmao
This suckssss

~~I'll have it done sooner or later~~
Ended up redoing the whole PR, easiest way. Should be ready.

code/controllers/configuration/entries/general.dm
```diff
@@ -291,7 +291,7 @@
 /datum/config_entry/string/banappeals
 
 /datum/config_entry/string/wikiurl
-	default = "http://www.tgstation13.org/wiki"
+	default = "https://wiki.skyrat13.tk/w/index.php" //SKYRAT EDIT - Original: "http://www.tgstation13.org/wiki"
```
GoldenAlpharex, line 294: Something like that might actually benefit from being changed in `config.txt`, no? Unless you're telling me you tested this in-game and it worked, in which case it's fine.

GoldenAlpharex (changes requested):
Good PR!

code/controllers/configuration/entries/general.dm
ghost: Oof I thought config.txt was local, didn't think the change would go forward to the server correctly if I did it from github 😅 
~~I can get a commit up in a second~~
Commit is up, hopefully done right. Please re-review if I did one wrong!!

config/config.txt
```diff
@@ -275,7 +275,7 @@ CHECK_RANDOMIZER
 # FORUMURL http://tgstation13.org/phpBB/index.php
 
 ## Wiki address
-# WIKIURL http://www.tgstation13.org/wiki
+# WIKIURL https://wiki.skyrat13.tk/w/index.php
```
Zonespace27: don't change this, 'tis config

GoldenAlpharex: Indeed, since it's commented out, it doesn't actually need to be changed, here.

GoldenAlpharex (changes requested):

GoldenAlpharex (approved):
Awesome, thank you!

PR #16237, opened 2022-09-15
Moving iRCD 3 pack out of goodies.

Plum-but-gayer:
## About The Pull Request

Moves the replacement for the industrial RCD, the Improved RCD out of the 'goodies' section and into the 'engineering' section of cargo ordering, allowing it to be purchased with departmental funding or the departmental order console. When the industrial RCD was removed for balance it was replaced with the improved RCD, however due to the improved RCD both being gated behind a steep goody crate price, we haven't seen it used often in-round. As the intial steep pricing was set with the intention of changing after we figured out balance, I think giving engineering access to the iRCD through departmental funding/orders will allow this fun new item to actually be used by players. As the other reason of iRCDs being in goodies was about the concern of departmental order power creep, they have been set as a 'long cooldown' item while their price is set that they do not make cargo any more money than the existing maximum value departmental order (the gas miner) - this will encourage their purchase using departmental funds, rather than the order console, but still leaves the console as an option. Additionally, only the 3-pack of iRCDs have been placed into engineering section, the single iRCD pack is still available at it's original price in the goodie section for those who are more bounty-inclined.

4

More toys for engineering to play with! With the removal of the original iRCD, engineering has been using departmental orders less. Ideally with this patch interdepartmental RP can increase as the departments have more things to spend their money on. Additionally, less items that require extensive personal bounty-grinding means more time for people to RP with eachotherr! 

## Changelog

:cl:
balance: Nanotrasen is happy to now offer the Improved RCD to the engineering department. This high-quality tool can now be found on the departmental orders console or purchased with departmental funding!

/:cl:


RatFromTheJungle:
the AFAD probably shouldn't be in the medical options (just to limit their availability), you can drop the price or whathaveyou, but they shouldn't be in the one-press console, because two of them still explode like normal medibeams, 

(unless you wanna do what I couldn't work out, and find a way to disable the exploding part)

modular_skyrat/modules/cargo/code/packs.dm
```diff
@@ -107,6 +107,14 @@
 					/obj/item/weaponcell/medical/toxin)
 	crate_name = "medicell replacement crate"
 
+/datum/supply_pack/medical/afad
+	name = "Automated First Aid Device"
+	desc = "Someone mildly hurt and it's too much of a bother to manually handle their burns or cuts? Look no further than the AFAD, a state-of-the-art pain-relief device!"
+	cost = CARGO_CRATE_VALUE * 20
+	access = ACCESS_MEDICAL
+	contains = list(/obj/item/gun/medbeam/afad)
+	crate_name = "AFAD crate"
```
GoldenAlpharex: I'd prefer if it stayed as goodie-only, because it's meant for people to use for space exploration, and just generally personal use, I don't want it to become the meta because it's basically free, even if very slow, healing. It would also bring back the beam-crossing explosions in Medbay as a regular occurrence, which I'd really like to avoid as much as I can.

GoldenAlpharex (changes requested):

modular_skyrat/modules/cargo/code/packs.dm
Plum-but-gayer: Sounds good! I'm going to throw it back in goodies. The main purpose of this PR is the iRCD anyway so I just won't touch it.

Plum-but-gayer (commented):
i think this is either going to update my code or just leave a comment idk but im pushing the button now :)

Plum-but-gayer:
okay i pulled before pushing which is apparently bad but I moved the AFAD back in goodies and everything should be good now. probably. :)

Zonespace27:
doesn't look good to me
image

GoldenAlpharex (approved):
Should be fine now!

Plum-but-gayer:
thank you golden!!

tf-4 (approved):
Sure

PR #16257, opened 2022-09-15
[MIRROR] Removes light from airlocks with an unrestricted side, makes the overlay glow in the dark [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69758
--------------------

## About The Pull Request
- Airlocks that have unres_side set will no longer have a light source
- The overlay associated with unrestricted access is now on the ABOVE_LIGHTING_PLANE plane, making it glow in the dark.
~~- Replaced an `if` chain with math.~~

## Why It's Good For The Game
When I originally made the unrestricted access feature, the overlay was meant to be something like a floor light, and I added a light source to the airlock so that the floor light wouldn't be hidden in the shadows if the airlock was otherwise in the dark.

That was many years ago and I know how to make glow-in-the-dark overlays now. As an aside, this helps nightmares not be blocked from shadow-step by a not-obvious light source.

~~The old method also had some repeated code due to wall of `if` statements. I have replaced them with math.~~

## Changelog

:cl:
balance: Airlocks with an unrestricted side no longer function as a light source. The indicator will now be visible even in the dark.
/:cl:

image


code/game/machinery/doors/airlock.dm
```diff
@@ -477,12 +477,15 @@
...
 	. = ..()
+<<<<<<< HEAD
 	/* SKYRAT EDIT REMOVAL
 	if(hasPower() && unres_sides)
 		set_light(2, 1)
 	else
 		set_light(0)
 	*/
+=======
+>>>>>>> 4de3fc7d210 (Removes light from airlocks with an unrestricted side, makes the overlay glow in the dark (#69758))
```
tf-4:
```suggestion
```

PR #16272, opened 2022-09-16
Makes Interdyne's base less aggressive

Zergspower:
## About The Pull Request

My original intent and probably goal is still to make a flag for "TURRET_FLAG_IGNORE_HUMAN" so that we could avoid having situations that frankly are becoming too frequent especially on Icebox. This change is a compromise to outright removing the turrets from the entire perimeter, instead I've left the two door turrets as-is and they will continue to fire on Miners/borgs etc - however the rest of the turets will now ignore humans to avoid the continued problems and issues that crop up with them hitting miners walking by since Interdyne is too damn thicc.  The screenshot below was taken after I deleted said turrets after they killed a 4th person.

image

Interdyne is supposed to be non-descript, with the turret layout that was added it's inviting far too much attention and far too many ahelps/issues of retaliation that simply aren't warranted.  This change is to reduce the incidents between the two to avoid removal of the turrets, a future goal is to add a new bitflag to avoid targeting carbon/human entities.

## Changelog

:cl:
balance: changes just the outer turrets on Interdyne Ice and Lavabase to target anything not in the Neutral faction (station crew and non-hostile fauna)
Fix: Couches on both will now face correctly (Couch PR goofed their directions)
/:cl:


Zergspower:
Btw if anyone can help me understand how to get a new BitFlag to register - Everything's added and it's able to be toggled but it's simply not reading correctly.

`if(ishuman(C) && !in_faction(C) && !(turret_flags & TURRET_FLAG_IGNORE_HUMAN))`

HoodMilk:
until that new flag is up and working you could slap the wildlife defense turrets from the aux base on it

Zergspower:
> until that new flag is up and working you could slap the wildlife defense turrets from the aux base on it

I did poke at that at first, but it required more VV'ng to work with the syndie system than just adding the faction - the end-goal ofc is getting that bitflag there so not only this particular issue is solved but opens up some PVE focused stuff

Elli-Skala:
The issue would still remain with just the turrents near the entrance just alot rare of people walking south to suddenly run into 2 guns hitting them at once
Because it was happening before when we did only have 2 turrents.

Zergspower:
> The issue would still remain with just the turrents near the entrance just alot rare of people walking south to suddenly run into 2 guns hitting them at once Because it was happening before when we did only have 2 turrents.

You're far more likely to see the front two after they've shot and killed someone, this PR's a compromise of us not ripping them out across the entire perimeter

PR #11509, opened 2022-02-14
Heretic Buffs, the Rebuffening

SpaceVampire:
## About The Pull Request

This PR gets Heretics back up to the same kind of level that all of our other main antags (Tators and Clings) have been pushed to in order to actually stand a chance in the current balance environment (Health changes, and security changes). Many of the changes in this PR had been made to them already, but were removed with the major rework of the code from upstream.

Buffs their heretic blade to no longer be way worse then basic bone tools, or the standard melee weapon of the other two antags, by giving it some extra damage and AP.
This also lets them make a couple extra copies if the weapon if they end up getting one-clicked disarmed into dropping it, because at that point you are already falling behind and not being able to create a new weapon makes it impossible to come back. I might lower the limit back down if it really becomes an issue, but the current one is very punishing for any mistakes.
Sacrifices now give more points, because of how much harder it is to pull off when we expect them to do more then just teleporting into the room and suddenly ganking people. Less punishing if you try to RP it out.
I have really only seen people pull off two of them per round, max. Most people are still doing less then that.
They also have one extra starting point, partly because I suspect a lot of people are going to want to buy the reroll ritual for the heart, because now you can't just make a second one if your target is busy in dorms.

Expect this thing to change as I get feedback on it, and it gets test-merged hopefully.

## How This Contributes To The Skyrat Roleplay Experience

Having the antag actually be playable, and not a total joke of a threat is good.

## Changelog

:cl:
balance: Heretic sickle-sword now does 8 more damage, and has slight armor pen.
balance: Heretics are now able to have more of their weapons summoned at once.
balance: Heretics start with 1(one) extra free bonus point.
balance: The gift of Mansus becomes ever more temping, as those granted with the power of a Heretic now receive greater strength from sacrificing targets.
/:cl:

PR #16275, opened 2022-09-16
Reverts the heretic sacrifice point buff

RatFromTheJungle:
## About The Pull Request
They were buffed in: #11509 to contend with, at the time (including having been midround only), being usually the sole antag in the round, but, as with the changes to dynamic (alot of antag rolls per round), having anymore then 1 person, whom can ascend before even finishing their original 3 sacrifices, is proving far-to-much, and hasn't exactly lead to more RP, but instead, more sudden-heretic-ascensions (**especially on lowpop**), especially now that they get 3 hours to try, instead of just 2:20, which is more then enough time if they're doing even marginally well.

(not even accounting they get essences ontop of this, making it extremely easy to get a vast excess in points)

by bringing heretics back into their slow snowball, instead of onesacrifice = snowball done, while still leaving them their bonus starting point, to ease over the start.

## Changelog

:cl:
balance: heretics, once again, get only two points per sacrifice, instead of 4.
/:cl:


SpaceVampire:
For the record.
"whom can ascend before even finishing their original 3 sacrifices"
Isn't actually true.
Given that you need to finish all of your objectives in order to ascend.
Which include stuff like "sacrifice three people" one of which has to be command staff.
image
So I don't think it should be that much of an issue, outside of low-pop.
Which in that case it might be better just to raise the pop requirements.

RatFromTheJungle:
> For the record. "whom can ascend before even finishing their original 3 sacrifices" Isn't actually true. Given that you need to finish all of your objectives in order to ascend. Which include stuff like "sacrifice three people" one of which has to be command staff. image So I don't think it should be that much of an issue, outside of low-pop. Which in that case it might be better just to raise the pop requirements.

that would be a fair point, if the objectives weren't immediately circled around by the point-buff: 
image
while yes, you have to sacrifice 3 people, the points buff also puts you in the greenzone for objective one, if-not within one or two influences from getting it, just from doing the 3 sacrifice objectives which means, functionally, you do infact finish your ascension by sacrificing your three.

SpaceVampire:
I just don't think that Heretics are very strong outside of low-pop, where you don't have any security around at all. Which is why I suggested that it might be better to just change the pop requirement config for them. I feel like they might have got in increase in cost lately, which is another half of that equation. Having access to more points makes them a lot more interesting to play, because you can pick some of the less meta options. If your issue is with ascending you could always just increase the difficulty of the objectives, making it sacrifice three instead of two for example.

I do also agree that you can currently finish your ascension just by sacrificing your three, as long as one of them is a head of staff which can be harder/easier depending on the head in question. Normally is pretty loud either way. You just said "before even finishing their original 3 sacrifices" the first time which wasn't correct. I feel like I rarely see people get more than one, but I can only view so many rounds so there is always the issue of sample bias on my behalf. Especially as I don't play during low-pop hours, which is where I mostly seem to hear about this issue.

RatFromTheJungle:
> I just don't think that Heretics are very strong outside of low-pop, where you don't have any security around at all. Which is why I suggested that it might be better to just change the pop requirement config for them. I feel like they might have got in increase in cost lately, which is another half of that equation. Having access to more points makes them a lot more interesting to play, because you can pick some of the less meta options. If your issue is with ascending you could always just increase the difficulty of the objectives, making it sacrifice three instead of two for example.
> 
> I do also agree that you can currently finish your ascension just by sacrificing your three, as long as one of them is a head of staff which can be harder/easier depending on the head in question. Normally is pretty loud either way. You just said "before even finishing their original 3 sacrifices" the first time which wasn't correct. I feel like I rarely see people get more than one, but I can only view so many rounds so there is always the issue of sample bias on my behalf. Especially as I don't play during low-pop hours, which is where I mostly seem to hear about this issue.

Having seen them ascend decently-regularly on lowpop, and get close if not ascend fairly often on highpop, i'd go so far as to say the fact they're rolling alot more, has lead to people getting past the initial knowledge gap, and our drastically extended rounds compared to TG, lead to them simply having to much time, and getting their points in general far-to-fast, and getting to ascension far-far to fast.

SpaceVampire:
> Having seen them ascend decently-regularly on lowpop, and get close if not ascend fairly often on highpop, i'd go so far as to say the fact they're rolling alot more, has lead to people getting past the initial knowledge gap, and our drastically extended rounds compared to TG, lead to them simply having to much time, and getting their points in general far-to-fast, and getting to ascension far-far to fast.

I still think that it would be better to just increase the actual ascension requirement, and possibly increase the amount of population needed for them to roll in general to knock them out of the lowpop range. Having less points is more of an encouragement to only pick the best possible options which is kind of boring. Raising the requirement would extend the amount of time where they actually have some powers but it isn't round ending. Which is probably the best game state.

RatFromTheJungle:
> > Having seen them ascend decently-regularly on lowpop, and get close if not ascend fairly often on highpop, i'd go so far as to say the fact they're rolling alot more, has lead to people getting past the initial knowledge gap, and our drastically extended rounds compared to TG, lead to them simply having to much time, and getting their points in general far-to-fast, and getting to ascension far-far to fast.
> 
> I still think that it would be better to just increase the actual ascension requirement, and possibly increase the amount of population needed for them to roll in general to knock them out of the lowpop range. Having less points is more of an encouragement to only pick the best possible options which is kind of boring. Raising the requirement would extend the amount of time where they actually have some powers but it isn't round ending. Which is probably the best game state.

I disagree, because them having easy access to ascension, the majority of their spells (which are supposed to be hard to get in multitude), and also using the rest of the stuff available on station is simply, to much, again, even at highpop, and especially at lowpop, especially-espiecally because it's been spawning between 2 and 5 of them, which is also way-to-many for this level of power.

SpaceVampire:
> I disagree, because them having easy access to ascension, the majority of their spells (which are supposed to be hard to get in multitude), and also using the rest of the stuff available on station is simply, to much, again, even at highpop, and especially at lowpop, especially-espiecally because it's been spawning between 2 and 5 of them, which is also way-to-many for this level of power.

I think that you could start with a by lowering it to three points, instead of two points. I also think that sounds like an issue which should be fixed by the dynamic configs just spawning less of them. My general high-pop experience is pulling off a single sacrifice and then getting hunted by the entire security force for the rest of the shift, but you know, there is the bias of only being able to comment on the types of shifts that I have personally seen with them.

RatFromTheJungle:
> > I disagree, because them having easy access to ascension, the majority of their spells (which are supposed to be hard to get in multitude), and also using the rest of the stuff available on station is simply, to much, again, even at highpop, and especially at lowpop, especially-espiecally because it's been spawning between 2 and 5 of them, which is also way-to-many for this level of power.
> 
> I think that you could start with a by lowering it to three points, instead of two points. I also think that sounds like an issue which should be fixed by the dynamic configs just spawning less of them. My general high-pop experience is pulling off a single sacrifice and then getting hunted by the entire security force for the rest of the shift, but you know, there is the bias of only being able to comment on the types of shifts that I have personally seen with them.

most of them are doing pretty-fine if not great, partially due to abusing jobs like paramed/engineer for it, partially due to just using job equipment as them, and partially due to the point buff, and the first change to be make, is to lower what's making them a direct issue, and not try to beat around the bush with it, it's just simpler and to the point; the issue is that they, collectively, with influence and even one sacrifice, scale out of control faster then what even tg could have intended, while having 2 more hours of time to do it, i'm sure they can make do with two points, in exchange for having triple the time. 

Even if they spend 30m talking to each sacrifice, they'd still have an hour and a half left in the round, if they get to unbelievably weak in post, it wouldn't be that hard to knock it up to 3 later.

PR #16282, opened 2022-09-16
Fixes molds creating foam that could cover the entire Z-level

GoldenAlpharex:
## About The Pull Request
That's about it, really. It bugged me that they could turn all of space green, or orange. So now they won't do that anymore.

Green space isn't fun for time-dilation.

## Changelog

:cl: GoldenAlpharex
fix: Molds can no longer spread their foam across the entirety of space, and have a much more manageable radius instead.
/:cl:


Zergspower:
Can it be true?

PR #16300, opened 2022-09-18
Added Prison Gates, and ID lockable lockers to Perma

Mundang:
## About The Pull Request
Replaced some of the airlocks in Perma to Prison gates, and changed the lockers to ID lockable lockers

Allows for the function to the ID timer, and removes the need to have a separate timer.

## Changelog

:cl: Mundang
add: replaced airlock in Perma with prison gates
add: replaced lockers with genpop lockers
/:cl:

### DIFF
--- a/_maps/map_files/Blueshift/BlueShift_middle.dmm
+++ b/_maps/map_files/Blueshift/BlueShift_middle.dmm
@@ -2605,9 +2605,12 @@
 	id_tag = "permainner";
 	name = "Permabrig Transfer"
 	},
-/obj/effect/mapping_helpers/airlock/access/all/security/brig,
 /obj/effect/mapping_helpers/airlock/cyclelink_helper_multi{
 	cycle_id = "perma-transfer"
+	},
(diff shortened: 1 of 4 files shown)


Jolly-66 (changes requested):
This PR is incompatible with how we do map edits (with the exception of BlueShift).

Delta already has a modular prison segment, and should have these items already (if they don't then I'm wrong)
As for the other maps, I won't permit a single tile auto mapper segment just to add these, as this could possibly get pushed upstream.

I would close this PR with just this comment, but I'm on hiatus right now, so Pax will be in charge here.

Iamgoofball:
they can't take it upstream though the turnstile gates are a down here thing

you could however use the permabrig doors from tramstation instead

Mundang:
That is what I have done. 

> they can't take it upstream though the turnstile gates are a down here thing
> 
> you could however use the permabrig doors from tramstation instead
That is what I have done.

Zonespace27:
Feel free to re-PR this with the changes made.

PR #16304, opened 2022-09-19
[MODULAR] Adds Atmos Access to Station Engineers

RimiNosha:
## About The Pull Request
This is a thing I've had engis complain to me about, and whenever I've played engi, this gets annoying having to ask for this nearly every round there's any breach bigger than a pinhole.

Also adds a new modular folder **just for job access overrides**, named `job_access_overrides`. How creative.

Engis can now access air alarms, and atmos in general, but mostly air alarms. This allows for engineers to do what they inevitably do every shift with less friction, which is fix basic atmos into at least a breathable condition when there are no atmos techs immediately available.

## Changelog

:cl:
balance: Station Engineers now have atmos access.
/:cl:

### DIFF
--- a/modular_skyrat/modules/job_access_overrides/code/station_engineer.dm
+++ b/modular_skyrat/modules/job_access_overrides/code/station_engineer.dm
@@ -1,0 +1,17 @@
+/datum/id_trim/job/station_engineer
+	minimal_access = list(
+		ACCESS_AUX_BASE,
+		ACCESS_CONSTRUCTION,
+		ACCESS_ENGINEERING,
+		ACCESS_ENGINE_EQUIP,
+		ACCESS_EXTERNAL_AIRLOCKS,
+		ACCESS_MAINT_TUNNELS,
+		ACCESS_MECH_ENGINE,
+		ACCESS_MINERAL_STOREROOM,
+		ACCESS_MINISAT,
+		ACCESS_TCOMMS,
+		ACCESS_TECH_STORAGE,
+		ACCESS_ATMOSPHERICS, // Move atmospherics to standard access.
+		)
+	extra_access = list(
+		)
--- a/modular_skyrat/modules/job_access_overrides/readme.md
+++ b/modular_skyrat/modules/job_access_overrides/readme.md
@@ -1,0 +1,13 @@
+## Title: Job Access Overrides
+
+MODULE ID: JOB_ACCESS_OVERRIDES
+
+### Description:
+
+Job access overrides.
+
+Create one file per job, and document any access changes via comments to prevent having to code dive to find the original accesses.
+
+### TG Edits:
+
+- Replaces `minimal_access` and `extra_access` on various job datums.
--- a/tgstation.dme
+++ b/tgstation.dme
@@ -5646,6 +5646,7 @@
 #include "modular_skyrat\modules\interaction_menu\code\click.dm"
 #include "modular_skyrat\modules\interaction_menu\code\interaction_component.dm"
 #include "modular_skyrat\modules\interaction_menu\code\interaction_datum.dm"
+#include "modular_skyrat\modules\job_access_overrides\code\station_engineer.dm"
 #include "modular_skyrat\modules\jukebox\code\dance_machine.dm"
 #include "modular_skyrat\modules\jukebox\code\jukebox_subsystem.dm"
 #include "modular_skyrat\modules\jungle\code\flora.dm"


nevimer:
People have been asking for this for over a year, the separation really makes no sense when broken down.

tf-4:
This just makes engineers better atmos techs. Why would you play atmos tech with this change?

Acer202:
> This just makes engineers better atmos techs. Why would you play atmos tech with this change?

Why do atmos techs have engi access? Why are they even separate jobs?

tf-4:
Currently, atmos techs have something engis don't have (atmos access) and engis have something atmos techs don't have (the skillchip). With this PR, there will be _zero_ reason to play atmos technician, and you'll even be putting yourself at a _disadvantage_ for doing so.

nevimer:
> 



> Currently, atmos techs have something engis don't have (atmos access) and engis have something atmos techs don't have (the skillchip). With this PR, there will be _zero_ reason to play atmos technician, and you'll even be putting yourself at a _disadvantage_ for doing so.

Dunno, it's unique to TG that atmos and engineering are seperate, and despite everyone else having the same two jobs (atmos and engi) they both share the same accesses with very minor almost insignificant access differences, and it works, it's a designation of what you want to do not what you are allowed to do.

tf-4:
> > This just makes engineers better atmos techs. Why would you play atmos tech with this change?
> 
> Why do atmos techs have engi access? Why are they even separate jobs?

Atmos techs do not have full engineering access, and part of the access they even _do_ have they only have because we gave it to them modularly with the reasoning "they had it on oldbase".

Deek-Za:
I love atmos. But with this then I'll just play engineer for the skillchip.   
Cause currently an engineer can access air alarms by hacking it.  Much like a tech would have to hack to get in to tech storage or mess with an apc.
There should be inherent perks for playing each role otherwise why have two different roles to begin with?

RimiNosha:
> > > This just makes engineers better atmos techs. Why would you play atmos tech with this change?
> > 
> > 
> > Why do atmos techs have engi access? Why are they even separate jobs?
> 
> Atmos techs do not have full engineering access, and part of the access they even _do_ have they only have because we gave it to them modularly with the reasoning "they had it on oldbase".

False. We do not touch job accesses, not since the refactor.

RimiNosha:
> I love atmos. But with this then I'll just play engineer for the skillchip. Cause currently an engineer can access air alarms by hacking it. Much like a tech would have to hack to get in to tech storage or mess with an apc. There should be inherent perks for playing each role otherwise why have two different roles to begin with?

Atmos tech gets priority to their equipment, that's a pretty big upside. Not all jobs have to be equal in access if the one with less access gets a sizeable bonus. If push comes to shove, we could always add some kind of skill chip, but they already get the biggest pre-made project room on the station.

tf-4:
> Atmos tech gets priority to their equipment, that's a pretty big upside.

Yes, they get priority _because they have access_. That is their advantage.

> Not all jobs have to be equal in access if the one with less access gets a sizeable bonus.

Yes, that would be the engineering skillchip that engineers with less access get as their sizeable bonus.

modular_skyrat/modules/job_access_overrides/code/station_engineer.dm
```diff
@@ -0,0 +1,17 @@
+/datum/id_trim/job/station_engineer
```
Zonespace27, line 1: this doesn't need to be its own module, `master_files` exists for a reason

Zonespace27 (changes requested):
Not sold on the change itself.

Iamgoofball:
this change is less needed here because skyrat atmos techs actually do their job

modular_skyrat/modules/job_access_overrides/code/station_engineer.dm
Iamgoofball: i dunno, i think it's better from a modularization standpoint to keep big sweeping changes to core code in master_files if we can modularize it, and put individual changes in the skyrat modules for easier searching given the volume of changes you guys have

HoodMilk:
you could change the access on the actual air alarms themselves to something that both atmos techs and engis share. So atmos gets to keep their atmos room full of complex machinery and mixing setups while still allowing engineers to fix air alarms when there are no atmos techs.

tf-4:
> you could change the access on the actual air alarms themselves to something that both atmos techs and engis share. So atmos gets to keep their atmos room full of complex machinery and mixing setups while still allowing engineers to fix air alarms when there are no atmos techs.

While still unnecessary imo, this is a far better solution to the supposed problem.

Zergspower:
Thought about this more, i'd personally not bother with the PR. If there's a staffing issue then they can get access changes, you're signing up to be an Engineer or to handle Atmospherics, blending the two together kind of makes it pointless to have two seperate roles/uniforms etc and would make sense to just make it an alt-title at that point. As others above noted, i'd change the air alarms themselves to take engineering access (which is what makes total sense) and vice-versa if Atmos doesnt have APC access.

YakumoChen:
IC issue, station engineers have an ingame solution called "Ask the Chief Engineer for it"

modular_skyrat/modules/job_access_overrides/code/station_engineer.dm
RimiNosha: Not going to put this in `master_files` unless it's really wanted, it will make the change far harder to find. Plus, in the past, we had a ton of job access overrides, so I feel it'd just be better to get a module set up for if that were to ever happen again.

RimiNosha:
I'm pretty neutral on this overall, but I made this PR cause it's brought up often enough, and it'd be nice to settle this properly in an easily trackable format.

I could override air alarms accesses if that's something enough people want instead.

softcerv:
> I'm pretty neutral on this overall, but I made this PR cause it's brought up often enough, and it'd be nice to settle this properly in an easily trackable format.
> 
> I could override air alarms accesses if that's something enough people want instead.

That seems like a pretty good compromise, if you have a good way that you could code it in.

RimiNosha:
Closing for a different solution, seeing as this one is wildly unpopular.

PR #16356, opened 2022-09-21
[MIRROR] Removes persistence of items through changing species when not allowed to [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70008
--------------------
## About The Pull Request

Makes use of item's ``mob_can_equip`` instead of mob's ``can_equip``, making it take the item's restrictions into account.
Fixes the args for ``mob_can_equip``, removing 'equipper' because it was unused, and adding 'ignore_equipped' so we can use it for mob's ``can_equip``. This also has the side effect of being consistent with can_equip, so it's pretty cool.
Also, fixes the inventory's color, so it's properly red when you can't equip such an item, by making it also use mob_can_equip.

Finally, expands the species clothing unit test to take that into account, to prevent it breaking in the future.

## Why It's Good For The Game

At the moment this solely affects the collar, which only monkeys can wear, but who knows what can happen in the future.
This removes people's exploit of monkifying themselves, putting a collar on, then using mutadone to have a collar as a non-monkey.

## Changelog

:cl:
fix: You no longer keep items you shouldn't be wearing (collars) when changing into a species that can't wear them anymore.
/:cl:


Tastyfish (approved):
Updated the items we have that override mob_can_equip()

GoldenAlpharex (approved):
Looks good!

PR #16375, opened 2022-09-22
[MIRROR] Macro optimizes SSmapping saving 50%  [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69632
--------------------

## About The Pull Request

I went through every major cost in SSmapping, and managed to lower each by either 50% or more.
Before I get too into it, there's 2 sources of major lag in SSmapping.
#1: Loading and placing map files.
#2: Randomly modifying said map files, things like cave generation or lava rivers.

There's a lot here, and I have introduced a TAD bit of different behavior, so lemme go through step by step

Lets start with some misc, and then map loading.
I """optimized""" space link generation and parsed map New().

Most of this was small minor changes, or just code cleanup. It doesn't save a whole ton of time but it makes space linking much easier to read, and... idk about parsed map New. Check that yourself.

parsed_map/_load_impl takes an ingested map file and actually well, loads it into the world.
There's 3 bits of cost here. 
#1: Iterating the format we store the map file in, and turning that into coords to spawn.
This was done in a really naive way. We did our max and min checks for x and y coords EVERY iteration, and also did out of bounds checks there too. Silly.

So rather then doing that, we pull deciding the max and min coords of x and y out into the grid iteration, and instead do from x to y loops inside the actual loading. Better.
While we're here, lets pull out the 3 sets of max/min pairs into once per line block, rather then once per tile. 
Ah and... ok so you're not gonna like this but
TGM and DMM maps are ingested and stored in different ways, giving the two different performance characteristics.
When we load a map file, we store "blocks" of lines. Lines are the literal newlines in the file, and blocks are like `(72,1,1) = {"` sort of thing.
TGM will generate a BLOCK for every column, and a line for every y coord in that column. That means iterating the line itself for keys is pointless in tgm. This lets us save time for TGM loading, but is a bit messy.
In contrast, DMM generates one BLOCK per z level, which makes its grid loading much less expensive, but requires cutting lines up.

I'm honestly just making this pr to clap back at mothblocks for their work so far, but I want to unify how these two formats are handled before this is merged. Having the X iterating point in the loop vary BY FUCKING FORMAT is stupid and unneeded I think. We will see.

Anyway moving on:
#2: Taking map keys (strings that link to a string that contains list of things to place there, including var edits) and transforming their target string into an indexable list for vving and spawning things.

Outside of using regex for trim unoptimally (I made an optimal trim() proc for this usecase), most of our time here was spent in `find_next_delimiter_position`, a generic proc for finding the next index in a string that counts as a new entry. 
We needed this because vv'd lists or strings can include ","s, our delimiter for paths. The same applies for var edits, which sped up processing that significantly too.

Fortunately, there's some extra resolution there. VV'd strings CANNOT include "\n,". This means rather then a proc call into like 4 text searches, I can JUST do a splittext call, and then iterate the output of that. That saves maybe 0.5 seconds.

Lemme see uh, lots of little things, stuff that's suboptimal or could be done cheaper. Some "hey you and I both know a \" is 2 chars long sort of stuff
Few other things, I think that's the main of it. Gives me the fuzzy feelings
Oh and I fixed a parsing bug where inserting \ into a dmm with no valid target would just less then loop the string. Dumb.

Alright that's map parsing, onto the third bit, actually placing things onto the map.
This is where I change behavior a bit, by dropping support for an old feature. Bear with me.

There's a feature in byond where you can use stack multiple turfs onto one tile. This effectively underlays all the preceeding turfs under the last one. Nice and all, but cost a lot of time, and the mappers tell me they lint against it, so it's GONE.

I made turf_blacklist an assoc list rather then just a normal one, so lookups are O(log n) instead of O(n). Also it's faster for the base case of loading mostly space.

Instead of toggling the map loader right before and right after New() calls, we toggle at the start of mapload, and disable then reenable if we check tick. This saves like 0.3 seconds

Rather then tracking an area cache ourselves, and needing to pass it around, we use a locally static list to reference the global list of area -> type. This is much faster, if slightly fragile.

Rather then checking for a null turf at every line, we do it at the start of the proc and not after. Faster this way, tho it can in theory drop area vvs.

Avoids calling world.preloader_setup unless we actually have a unique set of attributes. We use another static list to make this comparison cheap. This saves another 0.3

Rather then checking for area paths in the turf logic, or vis versa, we assume we are creating the type implied by the index we're reading off. So only the last type entry will be loaded like a turf, etc.
This is slightly unsafe but saves a good bit of time, and will properly error on fucked maps.

Also, rather then using a datum to hold preloader vars, we use 2 global variables. This is faster.

Alright, that marks the end of the maploader optimizations themselves. Now onto generation optimizations.
First, cave generation. Uses rustg perlin noise maps to generate a map, and probs to place mobs and megas.
I've slightly changed behavior here, changed how spawning logic works. I'm doing my best to match the intent conveyed by comments, since we wasted a lot of time doing not that, and floyd's not around to ask.

Cave generation was insanely more expensive then it had any right to be.
Maybe 0.5 seconds was saved off not doing a range(12) for EVERY SPAWNED MOB. 0.14 was saved off using expanded weighted lists (A new idea of mine)
This is useful because I can take a weighted list, and condense it into weight * path count. This is more memory heavy, and costs more to create, but is so much faster then the proc.

I also added a naive implementation of gcd to make this a bit less bad.
It's not great, but it'll do for this usecase.

Oh and I changed some ChangeTurfs into New()s. I'm still not entirely sure what the core difference between the two is, but it seems to work fine. (See below. MSO tells me that without changeturf you can just randomly copy vars over from the old turf, I need to test this. I believe this is safe because ruin loading uses New()s too, and they drop tiles in the same scenario. Eh)
I believe it's safe because the turf below us hasn't init'd yet, there's nothing to take from them. It's like 3 seconds faster too so I'll be sad when it turns out I'm being dumb

This reduces cave generation from 3.6 seconds to 0,5.

Now onto river generation.

This uses the same sort of concepts as the last change, mostly New being preferable to ChangeTurf at this level of code.
This bit isn't nearly as detailed as the last few, I honestly got a bit tired. It's still like 0.4 seconds saved tho

Oh and I touched ruin loading a tad.
Turns out it saves time if you don't check area type for every tile on a ruin. Not a whole ton faster, like 0.03, but faster.

Saves even more time (0.1) to not iterate all your ruin's turfs 3 times to clear away lavaland mobs, when you're IN SPACE who wrote this.

Oh it also saves time to only pull your turf list once, rather then 3 times

## Why It's Good For The Game
SPEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEED

## Changelog

:cl:
refactor: Maploading has been optimized significantly, as has ruin generation and other lavaland stuff. Let me know if anything goes fucky
/:cl:


Tastyfish:
> I made turf_blacklist an assoc list rather then just a normal one, so lookups are O(log n) instead of O(n). Also it's faster for the base case of loading mostly space.

This will likely require a change to automapper.

Tastyfish (approved):
Indeed faster. I updated automapper to use the new turf blacklist set format.

GoldenAlpharex (approved):
Nice!

PR #16381, opened 2022-09-23
[MIRROR] Removes AI and Charge parts from tablets, adds support for more later. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70012
--------------------
## About The Pull Request

Mostly a copy paste from commit comments:

### Charge parts

Removes charge parts from tablets entirely
Machine computers now directly use power from the machine they're in, while the rest uses power cells.
Silicon tablets don't use power at all, since they are needed to see and state laws, we don't want it running out of power.

### AI stuff

Removes the intellicard slot, moves it to the AI restorer app
AI restorer app now requires RD access instead of Bridge access. No access is needed to open it, so RD's console is still an option. This is because the app will make it possible to restore AI's from anywhere now that it doesn't require unique parts, so tying it to at least RD access means you need to use a console that has the app pre-installed, or get the RD/Captain to install it, keeping those limitations sorta.
Adds support for apps holding stuff (currently only used in AI restorer holding intellicards)
Calls kill_program on apps that are being deleted, so we can clear references or drop held things on an app's deletion

### Examine stuff

Moves examine to the modular tablet
Machine computers now examine the processor inside of it
Fixes examine text for tablet apps (only Nuke op app, now additionally AI restorer)

### Misc tablet stuff

Removes 'advanced' tablet subtypes that we used before PDAs were added, in some jobs.
Replaces Roboticist's advanced tablet mail with a laptop
Moves the notepad's note var from the tablet, to the note app
Moves modular computer's defines into their own file
Removes ``is_modular_computer()`` and ``get_modular_computer_parts_examine()``
Replaces PDA's ``insert_type`` with just checking if they should have a pen roundstart or not.

## Why It's Good For The Game

Currently my focus is to remove all PC parts, when that's done we can remove all the code that it requires, making tablets run much better and be way less confusing than it currently is.
I should write a hackmd to explain my end goal here, but in a single sentence: 'simplify PDAs so it's not as confusing to work with'. I'll write it when I finish removing all PC parts, which I did mention some should be removed in my old hackmd https://hackmd.io/-523YJ5qRqigIS_zqns2pA but I think it's better if we gut it all entirely. Once that's done things will be more clear to write what I would like to do with it.

## Changelog

:cl:
balance: AI restorer app now requires RD access to download.
fix: Nukie tablets now properly show when a nuke is going off in their examine.
balance: Tablet's charge and AI holder parts have been removed entirely. Tablets now only have a cell, and computers only use their area's APC.
code: Modular computer's power handling, the AI restorer app's functionality, and modular computer's examining has been refactored, please report any bugs you might find.
/:cl:


Tastyfish:
Note, I had to replace some tablets sitting around Cargodise Lost and the pirate fort to either be the PDA equivalent, or for advanced tablets, I replaced them with clear PDA's because those an underused variant.

I had to remove the Modular Tablet loadout item as baseline tablets no longer exist. I replaced it with a Clear PDA item. I figure if we were OK with tablet and laptop loadout items, that'd be fine, especially since it's a variant crew don't get issued.

Tastyfish (approved):
After discussion, removed the loadout item. PDA's work fine.

PR #16387, opened 2022-09-23
EM closet adjustment change thing(?)

tf-4:
should hopefully fix the fckn hard del issue
why was this whole proc modularised just to change like two things, it just makes it harder to maintain IDK


Tastyfish:
I don't think this will fix the issue, as it also happens upstream https://github.com/tgstation/tgstation/actions/runs/3115312476
It's likely because of the upstream 1% chance to INITIALIZE_HINT_QDEL but still populating contents regardless.

tf-4:
Ah I didn't realise it was also an upstream issue. Well I still want this regardless

PR #16389, opened 2022-09-23
[AutoMapper] updates delta rec area + cryo

press-mark:
## About The Pull Request
updates the south end of the rec area since tg changed it idk

BEFORE:
image

AFTER:
image

<details>
<summary>Outdated</summary>

AFTER:
dreamseeker_rivZROc8ck

</details>

## Changelog

:cl:
fix: fixed some out of place chairs and windows in the rec room on delta
/:cl:


Paxilmaniac (approved):
Looks good to me, we could probably cut down on the size of these cryo templates if we just made them use the area of the room they’re connected to, but I won’t make you do that in this pr.

Jolly-66:
> Looks good to me, we could probably cut down on the size of these cryo templates if we just made them use the area of the room they’re connected to, but I won’t make you do that in this pr.

Deltas was left like this due to the power net not having a nearby wire.

ghost:
I wonder, if we use the noop_area and noop_turf on all the tiles we DONT want to change, will it space the area/break, or properly paste the template over?
If that works, i wonder if cables would properly place if they were the only thing added

Jolly-66:
> I wonder, if we use the noop_area and noop_turf on all the tiles we DONT want to change, will it space the area/break, or properly paste the template over? If that works, i wonder if cables would properly place if they were the only thing added

I tested this but my idea was still right.
AutoMapper is just a hard copy/paste, its not like the modular map loading we already have in code, which makes turf passthrough and all that incompatible. 
AM just overrides whatever in the X by Y zone.

Gandalf2k15 (approved):

Gandalf2k15 (changes requested):
Do not have it be a separate area, just leave it as the same area, there is no cable to charge the APC

press-mark:
sure i can do that, ill just need to see if the navigation destinations can be renamed

press-mark:
image

alright the template is now just the actual cryo room itself and it uses the recreation room area
also moved the vents / scrubbers out from underneath the pods cause that sucks

GoldenAlpharex (approved):
Looks good to me. Thank you!

PR #16392, opened 2022-09-23
[MIRROR] Illiterate human-like mobs can send shuttles to a random destination + refactors snowflake checks on the shuttle consoles [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69836
--------------------

## About The Pull Request

Currently there are a bunch of snowflake checks on the mining shuttle console's attack_hand proc, including a check for TRAIT_ILLITERATE which is hardcoded to specific destinations and is causing issues with the public lavaland shuttle in #69641. When I started looking at that issue I realized that this should all probably be done in ui_interact() instead, and also that checking specifically for TRAIT_ILLITERATE when there's a proc for this (can_read) makes little sense. So I moved it all to ui_interact and cleaned up the TRAIT_ILLITERATE check. I also moved some code related to getting the list of destinations of shuttle consoles into a proc which is shared between the ui_data proc and the can_read check.

Now any illiterate mob which isn't otherwise blocked from interacting (such as monkies, ash lizards, and humans with a quirk) which interact with the shuttle console will spend 10 seconds "randomly mashing buttons" before sending it to a random valid destination. This is (essentially) the current behavior for humans with illiteracy and was (per @ timothymtorres ) the intended behavior for ash lizards when he added the illiteracy quirk to begin with in #66648. I'm just making it less snowflakey and I guess technically adding it to monkies too since they could also use the shuttle before that PR and it doesn't make sense to exclude them arbitrarily.

## Why It's Good For The Game

Fixes #69641

Generally makes the code more standardized, attack_hand checks are legacy from before ui interact was unified into can_interact and ui_interact. Making the code apply to all shuttle consoles and randomly pick a valid destination makes the more maintainable and less prone to random issues than a hardcoded list.

Also makes the other existing checks more consistent, for example the labor shuttle will now also warn rev heads, block free golems, and let illiterates move them just like the mining and public lavaland shuttles do.

## Changelog

:cl: VexingRaven
fix: The illiteracy quirk will no longer break the mining shuttle
refactor: Refactored some checks on the shuttle console to apply to all consoles
/:cl:


GoldenAlpharex (approved):
Should be good to go, CI failure is unrelated and already fixed on master :)

PR #16410, opened 2022-09-24
[MIRROR] Adds macros to help with common `set_`- and `adjust_timed_status_effect` uses [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69951
--------------------
## About The Pull Request

Adds some defines to replace some common cases of `adjust_timed_status_effect` and `set_timed_status_effect`.

I had someone mention that the arguments weren't always very clear and it was annoying to have to include the entire typepath of the status effect when using them. 

So, for the most common status effects, I added some macros to help
`adjust_x`,
`adjust_x_up_to`, (will not adjust values past the second argument)
`set_x`, 
`set_x_if_lower` (will only set values if the current duration is lower than the passed duration)

If you have any suggestions to improve these names to be more clear feel free to drop them in it's only a find/replace

Yes I hate the preprocessor, blah blah compile times
If another maint thinks this'll negatively impact compile times too much (probably like a tenth of a second at most) they're free to close it

## Why It's Good For The Game

These should help people developing with status effects a bit. 
instead of having to do `adjust_timed_status_effect(duration, /datum/status_effect/jitter)`, you can just do `adjust_jitter(duration)`.

They're also entirely optional, since they're just helper macros, so if people want to apply it the old way they can. 

If they do go through however, eventually the CC status effects can be converted to use these instead of their own snowflake procs. It'd be kinda neat to send all status effects through the same framework. 

## Changelog

:cl: Melbert
code: Replaces some uses of adjust_timed_status_effect and set_timed_status_effect to use macros, makes it easier to develop for.
/:cl:


Tastyfish (approved):
There were a lot of nested merge conflicts, it all seems to be in parity with upstream now.

Updated modular code to use new macros.

PR #15911, opened 2022-08-29
Remaps Atmos, Engi and surrounding maints (NSS Void Raptor)

QuacksQ:
## About The Pull Request
Re-maps Atmospherics, Engineering and the maintenance surrounding both. 

An effort has been made to increase the project space in Atmospherics for atmos techs to play with. Engineering has also been overhauled to be more spacious. The maints south of Engi and Atmos now connect and loop with no dead ends to get stuck in.

**Old Atmos and Engi:**
OLD

**New Atmos and Engi:**
hej3cz7u

- [x] Check for errors (Active turfs, runtimes, etc)
- [x] Fix found errors

## How This Contributes To The Skyrat Roleplay Experience

Dead ends in maintenace have been removed and all the southern maints around Atmos and Engi now connects. As said earlier atmos now also has more project space for people to play around with and the layout of engineering has been made less claustrophobic.

## Changelog

:cl: 
add: Remaps Atmos and Engi on the NSS Void Raptor. 
/:cl:

PR #16416, opened 2022-09-25
Remaps Atmos, Engi and surrounding maints (NSS Void Raptor)

QuacksQ:
## About The Pull Request
**This is the same PR as #15911 but I broke that one so here we are.**

Re-maps Atmospherics, Engineering and the maintenance surrounding both. 

An effort has been made to increase the project space in Atmospherics for atmos techs to play with. Engineering has also been overhauled to be more spacious. The maints south of Engi and Atmos now connect and loop with no dead ends to get stuck in.

**Old Atmos and Engi:**
OLD

**New Atmos and Engi:**

- [x] Check for errors (Active turfs, runtimes, etc)
- [x] Fix found errors

Dead ends in maintenace have been removed and all the southern maints around Atmos and Engi now connects. As said earlier atmos now also has more project space for people to play around with and the layout of engineering has been made less claustrophobic.

## Changelog

:cl: 
add: Remaps Atmos and Engi on the NSS Void Raptor. 
/:cl:


Paxilmaniac (changes requested):
Hey! Preliminary review without opening it in sdmm:

I feel like you could re-arrange the wires in this room so that the SMES output goes under that north door, rather than under an rwall. You can also cut down on the number of wires total in that room since wires under terminals will still connect to the other terminal wires, reducing the need for that second line of cables.
image

Don't have that scrubber under the CE's locker, there's plenty of floorspace to move it to be out in the open.
image

The shape of secure storage relative to where the blast doors are feels a bit strange, if anything  could you try and re-arrange stuff in there so that you don't have to pull other items out of the way to reach the breach shields or field generators?
image

Another pipe accessibility thing, rather than have the pipes run under a window here could you either make them the visible type and run them over a wall, or just route them under the nearby door?
image

This room here by the HFR area can probably get a scrubber and vent since it has the ability to be sealed off by the firelocks. Also, make that line of engine heaters on the south wall just rwalls.
image

This applies to the rest of raptor too, but tg is gonna be getting rid of the engine router or whatever those 4 way engine pipes are called, so we'll need to start phasing those out.
image

The pipes under windows where they don't need to be thing again, move the window to the side and have the pipe be a visible (overfloor) subtype that runs over the wall, since atmos techs can actually reach that without breaking the window.
image

Not sure if I like having decals on fake stairs tiles like this, doesn't look right.
image

Please just use a valve, digital or manual, that defaults to on for the waste into the space vent line, otherwise we're gonna get delams from people not realizing that this is the problem until its too late.
image

And lastly, I don't think the windows with grass and flowers under them really fit the theme of engineering. I won't make you change it but I really think they should just be the standard of rwindow with a grille under them.
image

Keep in mind I likely didn't catch everything here. Look around and make sure you don't have any vents or scrubbers that are under tables, lockers, big structures, so on. Secondly, pipes around the atmos area should be as accessible as possible. Visible subtype pipes running over walls can be touched by players, unlike pipes running under windows which require the window being removed to mess with.

Zergspower:
image
I mentioned it a few times, but if we're having walkways that close to engineering we should consider moving APC's and Air Alarms out of EMP range

QuacksQ:
@Paxilmaniac 

Cleaned up the wires and ran it under the door on the north.


Scrubber is no longer under the locker and moved the vent a bit.


Removed 2 shield generators and 1 tesla coil while moving them around a bit for accessibility as requested.


Window turned into an R-wall with visible subtype pipes being used now.


Placed reinforced walls on the south side of the room and piped in some atmos.


Windows have been slid to the side and pipes on the walls use the visible subtype.


Decals on the stairs have been removed here. Other spots in atmos and engi that used this have been made to look like this too.


Waste release is now a digi valve thats turned on by default.



Aft maint APC has been moved west a bit to be nearby the abandoned garden rather than south of the SM.



In various spots where pipes run through walls I've swapped them to the visible subtype now as requested.

Paxilmaniac (approved):
Looks pretty good now, get that merge conflict fixed

QuacksQ:
@Paxilmaniac Merge conflict has been resolved 👍👍

Tastyfish:
If it's not too late, `/obj/structure/sign/delamination_counter` now exist in master if you want to throw 1 or 2 in the engine room.

QuacksQ:
> If it's not too late, `/obj/structure/sign/delamination_counter` now exist in master if you want to throw 1 or 2 in the engine room.

Added two of these in the engine room

GoldenAlpharex (approved):
Paxil said it was good so we can blame them if anything happens :)

Paxilmaniac:
Golden merged it so we can blame them if anything happens :)

PR #16422, opened 2022-09-25
[Tram] The crossing lights actually indicate the danger of crossing

lessthnthree:
## About The Pull Request
We had a previous PR that increased damage, but it turns out that the traffic lights weren't changing fast enough for the new TG tram with the various character types we have.

They don't get hit by the tram walking across on a green/amber light.

## Changelog

:cl: LT3
fix: The Tramstation crossing lights will now properly reflect if you're going to get smacked by it if you try to cross.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
+++ b/modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
@@ -1,0 +1,6 @@
+	/// Skyrat edit for people being slow to cross the tracks of the tram, due to character size or whatever else. Make the lights change faster with the tram.
+
+	/// Proximity threshold for amber warning (slow people may be in danger)
+	var/amber_distance_threshold = 90
+	/// Proximity threshold for red warning (running people will likely not be able to cross)
+	var/red_distance_threshold = 80
--- a/modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
+++ b/modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
@@ -1,6 +1,4 @@
-	/// Skyrat edit for people being slow to cross the tracks of the tram, due to character size or whatever else. Make the lights change faster with the tram.
-
-	/// Proximity threshold for amber warning (slow people may be in danger)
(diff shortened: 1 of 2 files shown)


modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
```diff
@@ -0,0 +1,6 @@
+	/// Skyrat edit for people being slow to cross the tracks of the tram, due to character size or whatever else. Make the lights change faster with the tram.
```
Zonespace27: include the DME you goof

modular_skyrat/master_files/code/modules/industrial_lift/crossing_signal.dm
```diff
@@ -0,0 +1,6 @@
...
+
+	/// Proximity threshold for amber warning (slow people may be in danger)
+	var/amber_distance_threshold = 90
```
Zonespace27: you're not setting these vars on anything

Zonespace27: for changing vars you don't need `var/`, only on creating one

Zonespace27 (changes requested):

Elli-Skala:
> The crossing lights actually indicate the danger of crossing
So you're making it permanently red?

Tastyfish:
I have no problem with it being adjusted, as I originally designed it for pre-doom-tram, as you said.

A couple of little things:
- the Amber distance should be at least 20 higher than the red or else it's not enough time to mentally seperate them, or depending on the mood of the fast processor, might not even show up.
- ideally in my opinion, amber should mark the time it's unsafe to cross walking, and then red and save to cross running. Otherwise runners will just ignore the traffic light and go while it's in the middle of red. 

See if 90/70 or 80/60 work.

lessthnthree:
Yeah after sleeping on it, I am going to keep playing with it, Ideally I'd like it to be amber in central when the tram is stopped on east or west. I'll work on it more and take it upstream.

lessthnthree:
60/35 ended up as the sweet spot, so doing that upstream.

PR #16428, opened 2022-09-25
[MDB IGNORE] Touches up floor tiles a tiny bit

### LINKED ISSUE #16008: The texture in black tiles and black + color tiles is not consistent (and possibly others?)
The color for the tiles "textured" bits are not the same.

image

ghost:
## About The Pull Request
Fixes #16008

Title
They weren't starting to show age or anything, I just got totally sidetracked while trying to fix the linked issue.
Added a bit of the depth TG gave them, without that weird shine that makes them feel like plastic. Also touched up colors on a few, updated the ones TG made significant progress on (god bless the mech bays now) and even redid the cursed chapel flooring

These todos are more discussion points than actual checkboxes
- [ ] Medical floors are sliiiightly brighter. Not pure white, but not the off-gray they were before. Yay or nay? Could test-merge to see how it's received.
- [ ] Kitchen and Terracotta still use TG style because it makes them pop out as if they're tile rather than panels ~Terracotta never got changed to our shading actually, so this PR doesn't technically change it~. Does it stand out too much? Could also use test-merge. (Are they even used in any maps? IIRC maps don't use Kitchen but just Checker and Cafeteria)
- [ ] What the hell should we do with reinforced floors; our old one had always been crusty, and TGs current is not only shiny butjust screams Trypophobia (See below)
<details><summary> (By the way, Xyel? God damn you. I copy-pasted Trypophobia into google to make sure it was spelled right for this note and it came up with images. I feel sick, SICK.)</summary>

dreamseeker_2enpW4qnXA
</details>

<details><summary>Screenshots!</summary>

<details><summary>Chapel Flooring</summary>

dreamseeker_QqEGy8XFPO

</details>

<details><summary>Kitchen and Main Hall (kitchen slightly outdated, now has full black outlines)</summary>

dreamseeker_DSjFV8uuFh

</details>

<details><summary>Medbay Before</summary>

dreamseeker_1POy5D7pt1

</details>

<details><summary>Medbay After</summary>

dreamseeker_Xf518g2ah4

</details>

</details>
Most of the changes are pretty small, but make it feel a bit cleaner
(The checker floors are outdated, ignore the second screenshot. The issue was that the 'depth' TG added just happened to be incompatible with those two colors and our alternate shading method - it's been fixed now, with them just using the dark outline in all places like they used to. Makes them pop a bit too much but that's only if they're mixed with other tiles which doesn't usually happen anyways, and its better than... well, the above.

## Changelog

:cl:
imageadd: NT used some of their 'employee vacation funds' to clean the gunk from between the floor tiles, so they should appear cleaner and secure. Consider this an official reminder that Mining staff are expected to wipe their feet before boarding their shuttle.
/:cl:


Paxilmaniac:
IDK even these still look weird to me, I'd really prefer if we just used default tg tiles without any change rather than the poopendarken "The white medbay tiles are darker than tg's normal tiles" we have right now. Maybe I've just gone mad from mapping but having stared at floor tiles in sdmm for longer than most people should, and having to both map around and code fixes for the fact we just completely removed 'smooth' tiles, it seems like a whole lot of work for a worse result overall.

GoldenAlpharex (approved):
I'm down to give this a test-merge, to see how it's like.

ghost:
> IDK even these still look weird to me, I'd really prefer if we just used default tg tiles without any change rather than the poopendarken "The white medbay tiles are darker than tg's normal tiles" we have right now. Maybe I've just gone mad from mapping but having stared at floor tiles in sdmm for longer than most people should, and having to both map around and code fixes for the fact we just completely removed 'smooth' tiles, it seems like a whole lot of work for a worse result overall.

I'm just gonna break this into bullets to respond to lmao
-  Regarding TG's tiles, they were agreed to be extremely plastic-like; it makes the station feel like a toy rather than a gritty facility. The main solution to this is just shifting the shading to be on the top/side of each tile instead of a small line on the bottom, and a deeper gray. At a MINIMUM these would need to stay; texture is covered in the last bullet.
- White medbay tiles are addressed in this PR, sorry I forgot the screenshot. I made them whiter, it makes the place feel more sterile. Personally I like it, it's still not PURE white but it's definitely brighter than TG's gray now.
- Mapping always sucks. I've tried it before, I know your pain but can do nothing for it.
- Smooth/Textured tiles in our tileset are exactly the same - maintainer decision on the last floor cleanup (#11709), the smooth tiles were just a no-go. I DID try them though. Instead our smooth and textured tiles are *exactly the same icons* - don't feel too worried about which one you use in your maps so long as they're not mismatched and right by eachother (You should be hotkey-selecting the floor tiles to place the same ones anyways in this case I hope). Think like, per room, you can have one or the other. Ingame it won't be noticed, and even picking up Textured tiles will only give access to our tileset.

ghost:
Skyratbot hates me :(

Gandalf2k15:
ok

PR #16019, opened 2022-09-03
[MODULARIZING] Partial Refactoring of Organic Interface Code

RimiNosha:
## About The Pull Request
Uhhhh, does many things related to UI that is both user and non-user facing.

### Quick Overview:
- Moves organic interface item slots to the Interact menu (CTRL+SHIFT+CLICK)
  <details>
  <summary>Images (Mild NSFW)</summary>
  Clothes on<br>
  
  <br>Clothes off<br>
  
  </details>
- Removes **a lot** of non-modular UI edits.
- Fixes some odd grammar, and makes some messages less crude.

### Technical Details:
- Frees up four HUD slots, which doesn't sound like a lot until you realise cap is 24, and we hit said cap thanks to lewd items.
  - Having those limited slots used up by something half the playerbase has disabled all the time is bad.
- Above all else, removes a ton of invasive edits, and cuts down the amount of code used by the organic interface system by a *huge* amount. 
- Makes organic interface code much more easy to expand, should someone want to add more organic interface item slots, for whatever reason.
- Yeets the three megafiles into the sun in favor of having many smaller and actually organized files.
- Lots of code reformatting for better readability, as well as making some hardcoded things not hardcoded anymore.
- Adds value limits in both the negative and positives for pain, pleasure and arousal. This should help with random breakages of the system rendering people permanently (un)horny.
- Makes breast size a dropdown, and use letters instead of numbers.

### TODO:
- [x] Fix overlays.
  - [ ] Figure out why the heck nipple overlays aren't working...
- [x] Add more descriptive messages to everything.
- [x] Actually add checks for what's being inserted into an organic interface slot.
- [x] Hide and disable slots properly for no organic interface pref users.
- [x] Remove some shitcode on my part.
- [x] Remove shitcode on other people's part without rewriting organic interface.
  - [ ] Check this if I end up rewriting organic interface anyways.

## How This Contributes To The Skyrat Roleplay Experience
Less HUD icons that ultimately should be in their own UI, and makes maintainers complain slightly less about organic interface code.

LOTS less random breakages, and lots less complaints by the maintainers that no one's maintaining this code.

## Changelog

:cl:
code: Removed a ton of non-modular code related to organic interface.
del: Removed organic interface UI slots.
add: Organic interface slots now exist inside the Interactions menu. (CTRL+SHIFT+CLICK)
qol: Organic interface related messages regarding stripping and wearing are now more descriptive and informative.
fix: Permahorny(tm) and permaunhorny(tm) has been fixed.
[...]

PR #16433, opened 2022-09-25
[MODULAR] Fixes bug where opening Game Preferences and then spawning causing you to be random human. Also SR bodypart prefs refactor

Tastyfish:
## About The Pull Request

For some reason there doesn't seem to be an issue report for this long-standing issue, whatever.

The reason why opening Game Preferences and then spawning made you random was because we added a check when applying the prefs on spawn if the pref `is_accessible()`. Problem is that `is_accessible()` also filters if the pref is applicable to the current window. Since none of the character prefs are used in Game Options, it filtered out all of them. And also tried to apply game prefs to the player.

All the mutant body parts and cursed body parts prefs relied on this clumsy logic to support being toggled, piggybacking off is_available().

Refactors mutant body part options so that `apply_to_human()` itself handles visibility logic, as it should:
- Almost all the toggles had identical functionality. There is now a basetype `/datum/preference/toggle/mutant_toggle`.
  - The only two that were different are the two donator-only slots. Retained functionality.
- The body part choices are now under `/datum/preference/choiced/mutant_choice`.
  - The only one with added functionality was the snout slot. Updated it to support the current loigic.

Updated `/datum/preference/choiced/genital` likewise.

Also updated the emissive eyes setting.

Note that I tested a bunch of settings, but the volume of preferences and combinations of actions to do probably necessitates this as a test merge.

Big, long-standing bug.

## Changelog

:cl:
fix: Opening Game Options and then spawning no longer makes you a random human.
/:cl:


GoldenAlpharex (approved):
Finally, we can squash this bug.

I wish our usage of is_accessible wasn't so sinful, but... Oh well.

Gandalf2k15:
This is awful btw and is a bandaid to a much larger issue, don't hide the big main bug

Tastyfish:
> This is awful btw and is a bandaid to a much larger issue, don't hide the big main bug

If anything, the main bug is us adding is_accessible() here. I don't know why we did, but I imagine there was a point to it.

Tastyfish:
Oh and bodypart options getting hidden but stuck in their old state should be less of an issue now.

modular_skyrat/master_files/code/modules/client/preferences/_preference.dm
```diff
@@ -75,3 +75,88 @@
...
+/datum/preference/choiced/mutant_choice/proc/is_part_enabled(datum/preferences/preferences)
+	return preferences.read_preference(type_to_check)
+
+/**
+ * Actually rendered. Slimmed down version of the logic in is_available() that actually works when spawning or drawing the character.
+ *
+ * Returns if feature is visible.
+ *
+ * Arguments:
+ * * The character this is being applied to.
```
GoldenAlpharex:
```suggestion
 * * target - The character this is being applied to.
```

modular_skyrat/master_files/code/modules/client/preferences/_preference.dm
```diff
@@ -75,3 +75,88 @@
...
+	var/part_enabled = is_part_enabled(preferences)
+	return (species_available || overriding) && part_enabled
+
+/datum/preference/choiced/mutant_choice/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/preferences)
+	if(!preferences || !is_visible(target, preferences))
+		value = create_default_value()
+		. = FALSE
+
+	if(!target.dna.mutant_bodyparts[relevant_mutant_bodypart])
+		target.dna.mutant_bodyparts[relevant_mutant_bodypart] = list(MUTANT_INDEX_NAME = value, MUTANT_INDEX_COLOR_LIST = list("#FFFFFF", "#FFFFFF", "#FFFFFF"), MUTANT_INDEX_EMISSIVE_LIST = list(FALSE, FALSE, FALSE))
```
GoldenAlpharex, line 157: Couldn't we use the default `Mutant Colors` here rather than just going to white? Just a thought, would make a bit more sense to me personally, but it's really whatever.

GoldenAlpharex (changes requested):
A very great fix, and I'd be over the moon happy if this actually fixes the problem, because it's _much_ saner code, and will probably be useful in many other places in the code, probably more than we'd expect.

modular_skyrat/master_files/code/modules/client/preferences/_preference.dm
Tastyfish: The code this was replacing is white - I'm guessing it doesn't matter.

Tastyfish:
On hold for (test) merge until #16019 is merged due to conflicting body part changes.

GoldenAlpharex (approved):
Should hopefully be ready for a test-merge now that there's no longer any conflicts :)

glitchbow:
For some reason on this round (ID: 6699) my synthetic humanoid spawned in as a human. I actually wasn't sure why I couldn't use *beep, *buzz, *dwoop, and other robotic emotes until I looked and saw this as a bugfix, and thought they might be related. Not a random one, it still has her design and FT, but she got to be a real human for just this round. It is probably related to the fact I hit Game Preferences to change ghost types before joining the round.

yellowarrowpam

Tastyfish:
> For some reason on this round (ID: 6699) my synthetic humanoid spawned in as a human. I actually wasn't sure why I couldn't use *beep, *buzz, *dwoop, and other robotic emotes until I looked and saw this as a bugfix, and thought they might be related. Not a random one, it still has her design and FT, but she got to be a real human for just this round. It is probably related to the fact I hit Game Preferences to change ghost types before joining the round.
> 
> yellowarrowpam

Bizarre, since everything else appears to be applied. Applying the species is still vanilla to upstream.

GoldenAlpharex:
We've seen enough from this for me to be satisfied with the point it's at, let's merge it so we can not have to worry about fixing this test-merge again.

PR #16379, opened 2022-09-23
[MIRROR] Replaces obj flag `being_shocked` and flag_1 `shocked_1` with `TRAIT_BEING_SHOCKED` [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69978
--------------------
## About The Pull Request

- Replaces the object flag `being_shocked` with a trait
- Replaces the flag_1 flag `shocked_1` with a trait
- Mild cleanup around these places
   - Reactive tesla armor uses clothing traits

## Why It's Good For The Game

Why did we have two flags on two different levels dedicated to tracking when a movable was shocked, which only happens a second at a time?

Guess it's a tradeoff between proc overhead of having to set and reset the flags vs the access speed of bitflags. Seems kinda pointless

Frees 1 `flags_1`, pushes us away from `flags_2` slightly

## Changelog

:cl: Melbert
code: Changed some shocked flags into one shocked trait
/:cl:

PR #16449, opened 2022-09-26
[MIRROR] Removes overlay queuing, saves 6/7 seconds of initialize. Lightly modifies stat tracking macros [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69696
--------------------

## About The Pull Request

So we have this overlay queuing system right? It's built with the assumption that the "add to overlay list" operation is real expensive, and is thus useful to queue removals or additions.

It turns out that it just isn't, at least during init. In my testing the operation of queuing took LONGER then the actual overlay add/remove did.

That's ignoring the cost of the subsystem's work.

I've also modified part of the stat tracking macro, since it took a good bit of cpu time, and didn't seem to well, do anything. So far as I can tell it always evaluates to 1

## Why It's Good For The Game

Speeds up initialize, and potentially the rest of the game too. Also now we can always trust the overlays list, which is useful


Tastyfish:
I think #16379 needs to be merged first

Tastyfish (approved):
The only Skyrat content effected was code that was duplicated from upstream that was a straightforward removal of COMPILE_OVERLAYS().

PR #16451, opened 2022-09-26
[MIRROR] Removes network cards and printers from tablets [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70110
--------------------
### Network

There was a lot of unused code around Network cards (because networks as a whole is underused) so a lot of code was gutted here, I tried updating as much as I can to use better code, or to just be carried over at all, but sadly networks are networks.

I also renamed NETWORK_CARDS to NETWORK_TABLETS

It never F'ing worked
image

(This is supposed to say the hardware name instead of SYSTEM, but they never had a hardware name SET, EVER)

Post-my PR
image

### Printers

- Removes printer parts and affiliated shitcode
- Simplifies emptying paper bins into tablets, adds a new proc to remove papers from paper bins while taking into account already generated paper
- Moves Messenger's photo uploading to the new try_insert proc in hopes to simplify the attackby because I added two new interactions to it for adding papers to tablets, which made it look worse than it currently is.
- Moves Welder and Wrench interactions to welder_act/wrench_act for the same reason

## Why It's Good For The Game

I'm gonna try to keep the ball rolling to remove all PC parts, looking at tablet code makes me cry, and makes me wonder why the hackmd to transition PDA to tablet considered this to be 'good'.

## Changelog

:cl:
balance: You can no longer upgrade your tablets to have signals to the NTnet despite not being anywhere near telecomms. Instead, the Mining base has a lesser connection, and all computers have Ethernet connections.
balance: All tablets now have a printer in them, rather than it being an external part.
removal: NTNet tablet cards has been removed.
/:cl:


Tastyfish:
#16381 should be merged first

Tastyfish (approved):
The only change to our code was the contractor tablet. Behavior retained.

PR #16468, opened 2022-09-27
Cyborg Model Advertisement Update, or How I Stopped Having To Guess If You're A Janitor Or A Miner

GoldenAlpharex:
## About The Pull Request
Have you ever looked at a borg and shift-clicked to figure out what kind of borg they were, because they're either black, or purple, and subsequently went like "fuck, how do I know which model they are???" as you noticed that nothing was indicating their model in there?

Well, boy do I have the solution for you.

### Introducing: Cyborg Models In Examine
image

It's simple, it's slick, it's incredibly convenient and useful!

It even properly switches between `a` and `an` accordingly! (BYOND text macros my beloved)
image

We have a lot of cyborg models, it's annoying to try and play Guess Who: Cyborg Model Edition every time there's a model that's a bit ambiguous color-wise.

## Changelog

:cl: GoldenAlpharex
qol: After several incidents related to miscommunications regarding robot models, Nanotrasen's Robotic Appearance Normalization Department has decided to make it extra-obvious which model each cyborg is - Just give them a closer look, it's that easy!
/:cl:


RatFromTheJungle:
Does this affect the ninja borgs ability to disguise itself?

GoldenAlpharex:
> Does this affect the ninja borgs ability to disguise itself?

I honestly haven't tested it.

code/modules/mob/living/silicon/robot/examine.dm
```diff
@@ -3,6 +3,11 @@
 	if(desc)
 		. += "[desc]"
 
+	// SKYRAT EDIT START - Borg models in their examine text
+	var/model_name = model ? "\improper [model.name]" : "\improper Default"
+	. += "\nIt is currently \a \"[model_name]\"-type cyborg.\n"
```
ghost, line 8: Might be easier to just put 
```suggestion
	. += "\nIt is currently \a \[model_name]\ model.\n"
```
Appearing as "It is currently a Department model." More concise and still gets the point across

Alternatively we could clean some models out, we've got quite a few of varying standards and they should at the least follow departmental colors

GoldenAlpharex: I decided to go with this, because I couldn't find a nice way to word it that contained "model" without it looking too weird. That's why I decided to go with the way I've done it.

I'm not saying that I'm against some re-coloring of a few of them to follow departmental colors better, but I think this is going to be useful regardless.

GoldenAlpharex:
> Does this affect the ninja borgs ability to disguise itself?

(I have tested it and updated the code to make it so that was also taken into account)

tf-4:
Upstream?

GoldenAlpharex:
> Upstream?

I haven't taken it upstream because I don't know how relevant it would be there, considering how the upstream cyborg models are all visually different.

PR #16482, opened 2022-09-27
Heavily nerfs the APS' fire-rate, and also, the bulldogs

RatFromTheJungle:
## About The Pull Request

Currently: the APS can decap an HoS in fullmodsuit in less then two seconds, or disembowel in one-and-a-third, and the bulldog is just a 16 shot combat shotty with no delay on the shooting, and even at 6 firedelay (it before this PR, has 0, no delay between shots) it was still putting out 5 shots in like two seconds.

This brings them into the real of sanity when it comes to fire-rates, while still keeping them better then most crew available guns.

Makes the APS a direct upgrade to the makarov, and not a direct upgrade to literally everything else, brings the bulldog down to the level of most other shotguns, while keeping it's gimmick of alot more ammo, and portability, for a slight delay between shots (it's really not that bad at all)

## Changelog

:cl:
balance: The Stetchkin-APS has had it's fire-rate cut to a third, and the bulldog shotgun actually has a firedelay now.

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic/pistol.dm
+++ b/code/modules/projectiles/guns/ballistic/pistol.dm
@@ -93,7 +93,7 @@
 	mag_type = /obj/item/ammo_box/magazine/m9mm_aps
 	can_suppress = TRUE
 	burst_size = 3
-	fire_delay = 1
+	fire_delay = 3 //SKYRAT EDIT - Original: 1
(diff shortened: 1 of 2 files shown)


Zonespace27:
Split the two balance changes into separate PRs

PR #16483, opened 2022-09-27
Nerfs the Ebows recharge time

RatFromTheJungle:
## About The Pull Request
doubles the recharge time per ebow shot, to prevent the current situation where they can almost-literally stunlock, while also doing stamina and toxin damage atop.

Remember when we nerfed batons because they over performed (three times), and then forgot there's a ranged baton that does damage, slurring, and blurry-vison aswell? Oops.

## Changelog

:cl:
balance: The mini-energy bow now takes 4 seconds to recharge it's shot, instead of two.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/recharge.dm
+++ b/code/modules/projectiles/guns/energy/recharge.dm
@@ -102,7 +102,7 @@
 	custom_materials = list(/datum/material/iron=2000)
 	suppressed = TRUE
 	ammo_type = list(/obj/item/ammo_casing/energy/bolt)
-	recharge_time = 2 SECONDS
+	recharge_time = 4 SECONDS //SKYRAT EDIT - Original: 2 SECONDS
 	holds_charge = TRUE
 	unique_frequency = TRUE
 	can_bayonet = TRUE


YakumoChen:
I feel this makes it almost completely useless as anything remotely like a weapon.

 It doesn't stun, it does stamina damage and takes several shots to stamcrit. With this you'd almost certainly be better with a disabler or just a baton and stunning that way.

You're going to have to sharply reduce the TC cost because right now it's 10TC and this is already a pretty terrible weapon when we have loads of guns.

Zonespace27:
I don't feel this is necessary, another maint is free to override me if they feel stronger about it

Zonespace27:
Also get your GBP up jesus christ

RatFromTheJungle:
> I feel this makes it almost completely useless as anything remotely like a weapon.
> 
> It doesn't stun, it does stamina damage and takes several shots to stamcrit. With this you'd almost certainly be better with a disabler or just a baton and stunning that way.
> 
> You're going to have to sharply reduce the TC cost because right now it's 10TC and this is already a pretty terrible weapon when we have loads of guns.

if by terrible, you mean the best performing because it knocks down (for a second and a half), makes you drop items, makes you slur, makes your vision blurry, and does 20 toxin, along with 45 stamina, then it might be terrible

and, don't forget, atop all this, it fits in your pocket

YakumoChen:
> > I feel this makes it almost completely useless as anything remotely like a weapon.
> > It doesn't stun, it does stamina damage and takes several shots to stamcrit. With this you'd almost certainly be better with a disabler or just a baton and stunning that way.
> > You're going to have to sharply reduce the TC cost because right now it's 10TC and this is already a pretty terrible weapon when we have loads of guns.
> 
> if by terrible, you mean the best performing because it knocks down (for a second and a half), makes you drop items, makes you slur, makes your vision blurry, and does 20 toxin, along with 45 stamina, then it might be terrible
> 
> and, don't forget, atop all this, it fits in your pocket

3 hits to stamcrit, 10 toxin (20 is the crossbowl absolutely worthless against any number greater than one, costs half your tc, does not stop people from crying over radio, and completely outclassed by a 2 tc telebaton.

Not to even begin that 4 seconds is so long that stamina regen will make it take 4 or even 5 shots _that hit_ to even stun. That's upwards of 20-30 seconds of a fight. For a stun.

PR #16484, opened 2022-09-27
[MODULAR] Fixes very bad flag conflict for the borg boop module

Tastyfish:
## About The Pull Request

For an extremely long time, the boop module (`/obj/item/dogborg_nose`) had an item_flags flag set in flag_1, which didn't cause a panic but due to a recent change, happens to line up to INITIALIZE_1 which will give a warning stacktrace every time one is initialized and cause SSatom to think the item is broken.

buggggggggggg

## Changelog

Hopefully will get merged before it matters


GoldenAlpharex (approved):
Good fix.

PR #16487, opened 2022-09-27
[MIRROR] Properly checks flags with & instead of == [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70130
--------------------
## About The Pull Request

Byond ref: https://www.byond.com/docs/ref/#/operator/& - Basically, flags should use & instead of ==
This is because we can have more than 1 slot on any item, so it's preferred that we check if they have the flag, rather than check if it's using solely that one flag.
Even if it doesn't immediately fix any problems, it's something that should be the standard anyways, to prevent it from ever being a problem in the future.

In general this makes it easier to de-hardcode stuff that uses these things, like defibrilators, because someone at one point wanted to make it fit on the CMO's suit storage slot.

#### Regex

``(flag|flags|slot) (==|!=) (?!NONE)([a-zA-Z]+)``
I manually read through them, no mass replace used, because I didn't want to touch damage_flags & armor_flags because I fear in-game effects, and because there's some cases (like mechs, savefiles and fishing) which uses flags for non-flags.
Fair warning, there's some cases where it should be ==, if you want flags to 100% be the same.

## Why It's Good For The Game

^

## Changelog

:cl:
fix: Mobs with more than one pass flag (cockroaches, alien larva, parrots, bots, butterflies, any many other animals) should no longer knock bots down, if they're somehow on harm intent.
fix: Some female uniforms (white skirt, lawyer skirt) are now properly uniformed.
/:cl:


Tastyfish:
While fixing merge conflicts with `changeling.dm`, I noticed that `/datum/antagonist/changeling/proc/create_profile()` is missing, without skyrat edit notes,

```byond
	new_profile.age = target.age
	new_profile.physique = target.physique

	// Grab the target's quirks.
	for(var/datum/quirk/target_quirk as anything in target.quirks)
		LAZYADD(new_profile.quirks, new target_quirk.type)
```

Is this intentional?

Edit: nevermind, we re-arranged the code and moved it further down, for some reason.

Tastyfish (approved):
Updated modular items with slot checks.

PR #16504, opened 2022-09-28
[MIRROR] Reverts #68155 (Fix simple mob deaths causing deadchat notifications), Fixes living level mobs being able to die while dead [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70103
--------------------
## About The Pull Request

The issue #27188 involved living mobs producing a death message when `death()` was called on them while they were dead

As it turns out, this was caused as `/living/proc/death()` did _not_ check that the mob was dead or not: meaning `mob.death()` on a dead mob was valid

This is obviously, not supposed to be valid - `/carbon/death()` and `/human/death()` both implement early do-nothing-returns if the mob's stat is dead

The living level does not, and it probably should. I can't imagine it breaks anything by preventing a dead mob from dying again. But let me know.

Anyways. #68155 closed that issue, but closed it in a way I disagree with - it basically made all playermobs not give death messages even ones which are pretty useful to know

Such as: Wabbajacked humans

I reverted that PR and implemented and early return in living death.

Bonus code improvement, I dehardcoded an area check for CTF while I was there

## Why It's Good For The Game

Mobs with minds should pretty much always have a death message. It's important to know.
Wabbajacked players, slaughter demons, a lot of other mobs were completely missing from them which was sorely missed

## Changelog

:cl: Melbert
fix: All player controlled simple animals now show a death chat broadcast when they die again. 
fix: Living mobs cannot die while they are dead. Also fixes a bug where dead player controlled mobs will spam death chat broadcasts. 
code: Dehardcoded an exact type check for areas which shouldn't show death chat messages (CTF)
/:cl:


Tastyfish (approved):
Poppy approved.

PR #16516, opened 2022-09-28
[MIRROR] "It has been 0 days since the last incident" sign [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70099
--------------------

## About The Pull Request

image

Adds a fun sign to various places in engineering which tracks how many days (read: rounds) it has been since last time the engine delaminated.

image
Example image here shows an unrealistically large number for testing purposes.

image
It ticks up by one at the end of every round and resets to zero on rounds where the engine blows up (or technically to -1, then ticks up to 0 because the round ended).
Not much more to say about it.

## Why It's Good For The Game

I'll be honest I don't have any deep justification for this one except that I think it's cute.
I can make up something about how I think it might encourage people to try not to blow up the engine but I don't believe that nor would I want them to stop.

## Changelog

:cl:
add: Nanotrasen has started tracking exactly how often you're blowing up the supermatter engine. Surely they can't be sending as many new crystals over there as their financial reports imply.
/:cl:


Jolly-66:
Don't forget to add this to Void Raptor (and maybe BlueShift if you're feeling kind)

Tastyfish (approved):
Void Raptor has an engineering remodeling PR currently and Blueshift is being rebuilt from the ground up.

PR #16557, opened 2022-09-30
[MIRROR] Fixes multi-Z ruins (Ice Moon Mining Site) not spawning [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70097
--------------------

## About The Pull Request

Multi-z ruins currently do not spawn, this is because they call `SSmapping.get_turf_below()` (or above, if somebody is truly deranged) which checks for `SSmapping.initialized` which is of course false because ruin spawning is done during init. After looking at the code I don't see any dependency on SSmapping being initialized, only for `multiz_levels` existing so I updated the check to reflect that. I also moved `generate_z_level_linkages()` (which populates linkage information in `multiz_levels` much further up the initialize process just to be sure that info is populated. It isn't needed further down as new z-levels populate their linkage info already in `add_new_z_level()`, it's mostly just needed for default z-levels which don't call `add_new_z_level()` on their own. I talked with Potato in discord to confirm this makes sense.

## Why It's Good For The Game
Fixes #68579

## Changelog

:cl: VexingRaven
fix: Multi-Z Ruins work again (Ice Moon Mining Site spawns again with the Demonic Frost Miner)
/:cl:


Gandalf2k15:
Waiting for multi z rework.

PR #16561, opened 2022-09-30
Removes cell component from the radio jammer

tf-4:
A questionably balanced feature ~~that is causing issues~~ upstream issue seemingly?
Either way I don't really like this being in here.
Either the cell it's given is too high capacity and it just gets a buff to a 20 tile radius for basically free (like now) or it's too low capacity and it's useless

## Changelog
:cl:
del: Radio jammers no longer use the cell component
/:cl:


jjpark-kb:
and if we want this again, lets make sure we are modular...

PR #16565, opened 2022-09-30
balance the dragonslayer

jjpark-kb:
## About The Pull Request
By multiple people and through some testing, it seems the dragonslayer sword is outperforming. This should hopefully keep its power, but make it not so strong!

Less overpowered weapons means combat can be longer and the story can be more enjoyable.

## Changelog

:cl:
balance: dragonslayer sword does less damage and berserk lasts for less time.
/:cl:


dawsonkeyes:
armor penetration and block chance changes are good. why change the bonus force? that doesn't apply at all the actual player versus player combat, which seems to be what your contribution thing suggests this is trying to touch on

dawsonkeyes:
you can remove the "charged" tracking var as well, as the bubble alert only plays when it hits 100%, and that var was used when charge got over 100%

PR #16568, opened 2022-09-30
Makes the food and some of the snack items from the imported vendors more affordable

Paxilmaniac:
## About The Pull Request
I severely underestimated the effects of economic inflation on higher population rounds, turning a meal that was supposed to be 125 credits into one that costs 210.

One of the biggest complaints I got was that these would end up being too expensive for the amount of actual nutritional value they supplied, so I thought I'd lower them to be a bit more affordable.

## Changelog

:cl:
qol: The snacks and meals from several of the imported vendors around the station have been reduced in price.
/:cl:


jjpark-kb:
this is balance, not qol

Paxilmaniac:
change the tags then

tf-4 (approved):

PR #16570, opened 2022-09-30
Lets the guard rail barricade type actually be built

Paxilmaniac:
## About The Pull Request
This was added when we got the deployable barricades but then was never actually added to any building lists, lol, lmao even.

Also it changes the pixel shift on init to be a single pixel less so it stays on the tile.

I love having completely inaccessible features (Lie).

## Changelog

:cl:
add: You can now build the guardrail 'barricades' with a stack of metal rods
/:cl:


tf-4 (approved):

ghost:
Maybe we should have these take the balance of normal railings into account, dont want a clearly better one or people will only use the better one
(Also keep https://github.com/Skyrat-SS13/Skyrat-tg/pull/16580 in mind in that case, it touches railing balance)

PR #16602, opened 2022-10-02
[Tram] Tram better syncs to new crossing signal timing

lessthnthree:
## About The Pull Request
So it turns out, the horizontal_speed var is actually a delay, so the lower the number, the faster the tram goes. Higher number, slower tram! So presently we have a max speed tram that can crit or kill you, this is an oversight.

We don't want a blindingly fast tram that can crit or kill you.

## Changelog

:cl: LT3
balance: Tramstation's tram speed is now better synced with the crossing lights.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
+++ b/modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
@@ -13,5 +13,5 @@ Also because of tram code, lower horizontal_speed is actually faster, which... y
  */
 /datum/lift_master/tram
 	collision_lethality = 4
-	base_horizontal_speed = 0.8
-	horizontal_speed = 0.8
+	base_horizontal_speed = 0.6
+	horizontal_speed = 0.6
--- a/modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
+++ b/modular_skyrat/master_files/code/modules/industrial_lift/tram_lift_master.dm
@@ -7,6 +7,11 @@ var/damage = rand(5, 10) * collision_lethality
 				collided.apply_damage(0.5 * damage, BRUTE, BODY_ZONE_R_LEG)
 				collided.apply_damage(0.5 * damage, BRUTE, BODY_ZONE_L_ARM)
 				collided.apply_damage(0.5 * damage, BRUTE, BODY_ZONE_R_ARM)
+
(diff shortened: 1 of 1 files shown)


Zonespace27:
Disagree with this. If it's an issue with lethality, lower the change you made instead of kneecapping the tram itself.

lessthnthree:
The tram _should_ hurt if it hits you, but needs to be a bit slower so that players can reasonably cross. We could do 0.7 to keep that risk of danger.

Paxilmaniac:
Beware of the pipeline:
Increase lethality of the tram to basically instakill you --> We suddenly have to slow the tram because its instakilling people

What zonespace said, revert the change we made on the tram's damage if its a problem, don't slow it down.

lessthnthree:
After a round of testing, it's only a slight speed reduction to 0.6.

lessthnthree:
> Beware of the pipeline: Increase lethality of the tram to basically instakill you --> We suddenly have to slow the tram because its instakilling people
> 
> What zonespace said, revert the change we made on the tram's damage if its a problem, don't slow it down.

The concept is that the tram should be damaging, but the lights should give you a proper indication if it's okay to cross. The proc is only every 2 seconds, so slowing the tram a bit, not a lot, solves it.

Zonespace27:
> > Beware of the pipeline: Increase lethality of the tram to basically instakill you --> We suddenly have to slow the tram because its instakilling people
> > What zonespace said, revert the change we made on the tram's damage if its a problem, don't slow it down.
> 
> The concept is that the tram should be damaging, but the lights should give you a proper indication if it's okay to cross. The proc is only every 2 seconds, so slowing the tram a bit, not a lot, solves it.

fix that issue instead of taking a half-measure then

lessthnthree:
Okay.

lessthnthree:
Increasing the subsystem processing cost on CPU by checking the crossing signals more often instead of just slowing the tram a bit that you probably wouldn't notice unless we told you.

Iamgoofball:
you should just fix the crossing signal being inaccurate if the tram speed is adjusted, yeah

Gandalf2k15:
Fix the crossing stuff instead, also wtf is tha file and why is it there

lessthnthree:
I don't want to put additional load on the server by decreasing the time between when the crossing signals update, and a slight slowdown of the tram seems like a better course. 0.5 to 0.6 you're not going to notice unless you knew it happened.

lessthnthree:
A slight change in tram speed is not worth this.

Iamgoofball:
It's not gonna be that expensive and it's a bugfix but alright

PR #16608, opened 2022-10-03
fix: null languages are also removed in sanitize

larentoun:
## About The Pull Request

Removes runtimes when `sanitize_languages` proc tries to create a language with a null path

Less runtimes, easier to find other runtimes

## Changelog

:cl:
fix: null languages are also removed in sanitize
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/client/preferences.dm
+++ b/modular_skyrat/modules/customization/modules/client/preferences.dm
@@ -6,6 +6,7 @@
 			languages.Remove(lang_path)
 			languages_edited = TRUE
 			continue
+
 		var/datum/language/language = new lang_path()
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/customization/modules/client/preferences.dm
```diff
@@ -2,6 +2,10 @@
 /datum/preferences/proc/sanitize_languages()
 	var/languages_edited = FALSE
 	for(var/lang_path as anything in languages)
+		if(isnull(lang_path))
+			languages.Remove(lang_path)
+			languages_edited = TRUE
+			continue
```
GoldenAlpharex, line 8:
```suggestion
		if(isnull(lang_path))
			languages.Remove(lang_path)
			languages_edited = TRUE
			continue

```

GoldenAlpharex (changes requested):

PR #16614, opened 2022-10-03
Removes some duplicate items from nukeop uplinks

Zonespace27:
## About The Pull Request
Some tot items that we added were cheaper than the nukeop version that was **exactly the same**, this corrects that

Bugs bad

## Changelog

:cl:
fix: Traitor-only items no longer show up in nukeop uplinks
/:cl:


modular_skyrat/master_files/code/modules/uplink/uplink_items.dm
```diff
@@ -52,6 +45,7 @@
 	item = /obj/item/gun/ballistic/automatic/c20r/unrestricted
 	cost = 12
 	progression_minimum = 35 MINUTES
+	purchasable_from = ~(UPLINK_NUKE_OPS | UPLINK_CLOWN_OPS)
```
GoldenAlpharex, line 49: This should be a define at the top of the file or something, to explain what it does, and to avoid repetitions if we want to change that behavior.

GoldenAlpharex (changes requested):

modular_skyrat/master_files/code/modules/uplink/uplink_items.dm
Zonespace27: done

tf-4 (approved):

PR #16616, opened 2022-10-03
[MIRROR] TGUI for Techfabs II: The Great Recategorizing (AND ICONS) (AND MECHFABS) (AND AUTOLATHES) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69990
--------------------

### HEY KID YOU WANT SOME ICONS? HOW ABOUT SOME CATEGORIES? IVE GOT ICONS AND NEW CATEGORIES

https://user-images.githubusercontent.com/12202230/190893973-07bcf1e6-4a58-424f-b4f9-e76e03be68ea.mp4

### I ALSO HAVE ICONS IN MECHFABS AND NEW CATEGORIES

https://user-images.githubusercontent.com/12202230/190894084-41217e85-447d-493c-a6db-03deec16acf1.mp4

### DO YOU OR A LOVED ONE SUFFER FROM AUTOLATHE CATEGORIES NOT EVEN LISTING EVERY ITEM YOU CAN PRINT IN AN AUTOLATHE? WELL DON'T WORRY I DID THAT TOO

https://user-images.githubusercontent.com/12202230/190894204-4663ee5d-3a7a-4a84-8652-4646a92e95c6.mp4

### SERIOUS TALK

- I recategorized **EVERY `/datum/design/` IN THE GAME** to be more UX friendly and **I HATE MYSELF FOR IT**
- I refactored techfab UI to **WORK ANYWHERE** for **ANY MACHINE THAT USES `/datum/design` as a SET OF MODULAR COMPONENTS**
- I moved a lot of **DESIGNS EXCLUSIVE TO THE AUTOLATHE** to also work **IN PROTOLATHES**
- I made **MATERIAL ICONS** animate between **ICON STATES** for **STACKS**
- **I PUT ICONS IN ALL OF YOUR FABRICATORS**
- I SOMEHOW DID ALL OF THIS **WITHOUT LOSING ANY PERFORMANCE**
- ALSO SUPPORTS COMPONENT PRINTERS AND MODULE DUPLICATORS

Other garbage:

- Fixed numerous spelling and consistency issues in designs
- Removed `Machine Design (<x>)` and `Computer Design (<x>)` from all relevant designs
- All designs are now in title case
- Numerous designs that were formerly autolathe exclusives can now also be printed at a protolathe (but not all); this is mostly just service equipment like drinking glasses and plates and silverware
- Circuits components can no longer be printed at a circuit imprinter (fixes #67758)
[...]

PR #16648, opened 2022-10-04
[MIRROR] Adds bot launch pad [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70267
--------------------

## About The Pull Request
- A launch pad for bots, inspired by the orbital mech launch pad.
- Requires linking to a bot pad controller (using a multitool) to function (linking process is similar to linking a machine to the ore silo).
- Available after researching the Advanced bots tech node (this node is unlocked after the robotics research and cost 2500 research points).
- Pad doesn't require any extra components to build. Just the circuitboard.
- Pad Controller can be printed from the science protolathe after researching. Costs 500 iron and 500 glass.
- When the controller is used (left-click) and only mob on the pad is a bot, it's launched to the users location.
- Launch is performed by creating an invisible drop pod on the pad and landing it to another location with the payload.
- This drop pod does not have an effect on the drop location (explosion, damage, stun etc.)
- After launching a bot you can right click to the controller in order to make the bot move back to the pad that launched it.
- Has notice and warning messages for diffrent situations.

#### Downsides (good for balance)
- Bots have to be on the pad in order to launh so their patrol mode should be disabled.
- Pads should not be located in a busy place in order to assure that the bot stays on the pad. Or you can turn off the bot before placing it.
- You need to wait for bot to move back to the pad before launching it again. So you can't spam it.

Might require some balance changes (launch delay, building/printing/researching costs, damage on drop).

### Icons
 Controller |  Pad  | Pad open
:-------------------------:|:-------------------------:|:-------------------------:
xaaCapture | xxbbbbbbCapture | xxxccccccCapture

### Low quality gif
Descend animation is not visible here.
ezgif com-gif-maker (1)

## Why It's Good For The Game
When you call a bot using a pda there are usually things blocking it and even there is not you don't know when it will reach you and you might have to wait for a while. This makes calling secbots or medbots pretty much useless because you usually need them to be there in a short time span.
By using a bot pad you can make a bot come to you after a short delay.

## Changelog

:cl:
add: Added bot launchpad and bot launch pad controller.
add: New research node for bot launchpad and bot launch pad controller.
imageadd: added icons for bot launchpad and bot launch pad controller.
/:cl:


Tastyfish:
Depends on #16616

PR #16655, opened 2022-10-04
Removes most of skyrat fishing in favor of tg fishing, keeps the ability to upgrade and fish out of water basins (they just use tg fishing now)

Paxilmaniac:
## About The Pull Request
As cool as it was to have a modular fishing system, now that tg has added one (I say now, it was a whole ago) ours is just redundant. Not just abundant, but significantly easier to spam than tg fishing is, to the point where the entire mining department can be replaced by someone with two fishing rods, or that someone can make an immense pile of fish/maint loot with very little effort.
The only part I did keep around was the ability to make your own fishing portals with water basins and bluespace crystals, they just use the new fishing spot code and an updated loot list.

It's not really intuitive to the players to have two completely different fishing systems, both of which cannot interact with each other at all. Seconding that, I'd also imaging not having two different systems that do the exact same thing (but one is overpowered) is better from a code standpoint.

## Changelog

:cl:
del: Skyrat fishing has been removed in favor of using tg fishing, though fishing portals can still be made by upgrading water basins
/:cl:


press-mark:
bless

Nerev4r:
I'd raise the rewards a little bit, it does take genuinely forever to get what little fish you can get out of the fishing portals (because there aren't really natural water sources around.)

tf-4:
I do think there's a middleground to be found between stock /tg/ fishing and our own. /tg/ fishing is _extremely_ slow to get much fish out of if you're trying to cook for 100 people.

Paxilmaniac:
I think a good solution is I make an upstream pr that lowers the amount of time it takes to actually hook a fish, as I do agree it takes FOREVER to actually get to catch anything.

dawsonkeyes:
on my way to get a single legion core while using lavaland bait and fishing 20 times

for real through this is a cool decision so long as the tg fishing isn't worthless

PR #16676, opened 2022-10-06
[MIRROR] [MDB IGNORE] Shuttle engines part 2: Engines are now machines [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/69793
--------------------
## About The Pull Request

Engines are no longer structures, instead they're machines

Now they (except large/huge ones) have circuit boards, but they can't be printed by default. Instead, this allows the machines to be destroyed and rebuilt, but maybe in the future we can add these as printable machines.
Currently the only way to get more is through Cargo so maybe we can add the boards there instead? This is really just opening the door to anyone who wishes to expand upon this.

I also removed 2 engine types, 'router' and 'platform', beacuse their sprites sucked, they had no unique mechanics, and cutting down on the amount of engines might help make the others unique just by having less to work with. They also weren't engines at all, despite what they're subtypes of.

## Why It's Good For The Game

These are already technically machines since they affect Shuttles, I thought it might be a good idea for them to be turned into machines so maybe in the future, unique functionalities can be added to them.

## Changelog

:cl:
add: Engines are now machines, not structures. If destroyed, they will drop circuit boards so you can rebuild them.
removal: In-wall engines have been remvoed. The Tram now uses full-blown engines.
/:cl:


Paxilmaniac:
oh the misery

Paxilmaniac:
They removed in wall engines I am entirely devastated

PR #16683, opened 2022-10-06
Revert "Adds a pointing cooldown (once per half second)"

Gandalf2k15:
Reverts Skyrat-SS13/Skyrat-tg#14917

This was not a performant change and should be done differently, it's also causing a sporadic CI runtime due to the drug check which can force people to point.

Having a variable on every client just to tell them if the pointing cooldown is up is very inefficient and should be replaced.

Also having a new timer setup each time someone points is also very bad. It should be using cooldowns.


Gandalf2k15:
Merging, this can be redone when it's more performant.

PR #16105, opened 2022-09-06
[MODULAR] Prefs-Enabled Synth, IPC and Cyborg Brains!

RimiNosha:
## About The Pull Request
Oh wow, this one was easy, kinda wished I did this earlier, but here we are, at long last!

Adds in two new IPC brain subtypes, and one new cyborg brain subtype!

- Compact AI Circuit (for both Cyborg and Synth+IPC)
- Compact MMI (just Synth+IPC)

The synth MMI DOES NOT follow standard MMI functionality, it would be an absolute pain to code and balance properly.

This does indeed replace brains in cyborgs, regardless of how they were constructed or spawned, as the proc is fired whenever it is given a new mind. This is intentional, and if this is wanted to be changed, let me know!

#### TODO:
- [x] Figure out why the heck overlays for the synth MMI and cyborg circuitbrain aren't working...
- [x] Figure out why the stupid borg code doesn't move the mind to the new brain

image
image
From top to bottom, left to right:
- Active cyborg Compact AI Circuit
- Inactive cyborg Compact AI Circuit
- Searching cyborg Compact AI Circuit
- IPC+Synth Compact Man-Machine Interface
- IPC+Synth Compact AI Circuit

## How This Contributes To The Skyrat Roleplay Experience
Some synths might want to be flavored to be brain jars, while roundstart borgs might not want to be! I think this will make both groups happy without affecting balance!

## Changelog

:cl:
add: Two new IPC+Synthetic brain types: Circuit and MMI!
add: One new Cyborg brain type: Circuit!
qol: Cyborgs, IPCs and Synthetics can now choose their preferred brain in the lower prefs box!
/:cl:

PR #16697, opened 2022-10-07
[SEMI-MODULAR] Synthetic Quirks Rebalancer & Quirk Hider

Floofies:
## About The Pull Request
This PR contains a set of minor enhancements for character quirks and synthetic healing medicines, and is mainly focused on synthetics/IPCs. The changes are made as modularly as possible.

In-short, some character quirks don't work in a balanced way for synthetics, such as "Blood Deficiency" and "Brain Tumor", which this PR is focused on and includes fixes for. Furthermore, this PR represents a step in the right direction rather than being a comprehensive rebalance, so I hope to later implement even more of the concepts outlined here.

First and foremost, currently all quirks are sent to TGUI simply on the basis of existing in the `/datum/quirk` namespace. This PR adds a new boolean to Skyrat's `/datum/quirk` named `is_hidden`, and a matching conditional which checks for it within the quirks subsystem at TG's `SSquirks.get_quirks()`, making it the only semi-modular change. When set to `TRUE`, `is_hidden` will omit the quirk from the list returned by `SSquirks.get_quirks()` and effectively hide it from the TGUI. Normally `is_hidden` will only ever affect the visibility of quirks in the player character preferences window, as it is the only current executor of that function.

I've also renamed "Brain Tumor" to "Brain Degeneration" to improve its roleplay-ability. The quirk only described a tumor, but there is not a real tumor, making for a lot of awkward roleplay situations.

As a main focus of the PR, I have utilized the quirk hiding mechanic to implement balance/roleplay improving detours for quirks. The quirk detours swap player-selected quirks with hidden quirks upon being added to a mob. The quirk detours detect some criteria such as character species, and may swap the quirk being added with a hidden one which is more appropriate for that species. Detouring of quirks is 100% modular, and occurs only when the quirk is added to the mob. The players will still be able to pick from the same list of quirks as before this PR, but they will find that some of their quirks are swapped to new ones upon spawning into the game.

Finally, as a bonus extra, this PR also corrects non-deterministic results in the calculations of synthetic healing reagents. The fix involves implementing a more consistent usage of `delta_time` in time-bound stats changes and deduplicating calculations.

Here is a granular list of changes:

- Renames "Brain Tumor" quirk to "Brain Degeneration" and makes its flavortext more generic. There never was any tumor anyways, so the old flavortext made roleplay awkward in some situations.
- Adds a hidden `/datum/quirk/blooddeficiency/synth` quirk for synths named "Hydraulic Leak", which is more severe by reducing their blood down to at least `BLOOD_VOLUME_BAD`, which causes fainting and blurred vision but is still survivable.
- Adds a hidden `/datum/quirk/item_quirk/brainproblems/synth` quirk for synths named "Positronic Cascade Anomaly", which gives them liquid solder instead of mannitol upon spawning.
- Adds the required pills/pill bottles containing system cleaner, nanite slurry, liquid solder, and small-dose liquid solder for the brainproblems quirk.
- Makes `liquid_solder` actually a liquid.
- Reduces non-deterministic results in the calculations of synthetic healing reagents.

Several character quirks have balance or roleplay-ability issues when they're selected by players with synthetic/robotic/IPC characters, so this PR intends to begin fixing those issues. Notably, the quirk "Blood Deficiency" does not drain enough blood to harm or even inconvenience synths, and they don't have real blood, so this PR adds a more severe effect lending to the intended balance of the quirk. Also the "Brain Tumor" quirk gives synths medications they can't use, isn't even a tumor, and makes no sense on a synthetic person; it has been renamed to "Brain Degeneration" with more generic flavortext. Furthermore, when a synthetic character selects Brain Degeneration and/or Blood Deficiency as their quirks, they will be swapped out with "Positronic Cascade Anomaly" and "Hydraulic Leak" respectively, providing species-appropriate flavortext in medical roleplay.

braincondition

disabilities

traits

## Changelog

:cl: A.C.M.O.
add: Added hidden quirks which can't be selected in TGUI.
add: Added pills and pill bottles for synthetic healing medications.
balance: Re-balanced Blood Deficiency and Brain Tumor quirks for synthetics.
tweak: Re-labeled Brain Tumor as Brain Degeneration for roleplay-ability.
balance: For synthetic characters, the Blood Deficiency quirk swaps to Hydraulic Leak, optimized for synths.
balance: For synthetic characters, the Brain Degeneration quirk swaps to Positronic Cascade Anomaly, optimized for synths.
fix: Makes liquid solder actually a liquid.
fix: Fixes synthetic healing medications to behave more deterministically.
/:cl:


code/controllers/subsystem/processing/quirks.dm
```diff
@@ -62,6 +62,9 @@ PROCESSING_SUBSYSTEM_DEF(quirks)
 		// SKYRAT EDIT ADDITION START
 		if(initial(quirk_type.erp_quirk) && CONFIG_GET(flag/disable_erp_preferences))
 			continue
+		// Hidden quirks aren't visible to TGUI or the player
+		if (initial(quirk_type.hidden_quirk))
+			continue;
```
Zonespace27:
```suggestion
			continue
```
go back to your C-languages, `;`.

modular_skyrat/master_files/code/datums/quirks/negative.dm
```diff
@@ -21,3 +21,75 @@
...
+	lose_text = "<span class='notice'>You no longer feel glitchy.</span>"
+	medical_record_text = "Patient has a cascading anomaly in their brain that is slowly causing brain death."
+	icon = "bp_synth_brain"
+	mail_goodies = list(/obj/item/storage/pill_bottle/liquid_solder/braintumor)
+	hidden_quirk = TRUE
+
+// If brainproblems is added to a synth, this detours to the brainproblems/synth quirk.
+// TODO: Add more brain-specific detours when PR #16105 is merged
+/datum/quirk/item_quirk/brainproblems/add_to_holder(mob/living/new_holder, quirk_transfer)
+	if(!(is_species(new_holder, /datum/species/robotic) && (src.type == /datum/quirk/item_quirk/brainproblems)))
```
Zonespace27: for the first one i'm fairly sure there's an `isrobotic` macro
for the second use `istype`

modular_skyrat/master_files/code/datums/quirks/negative.dm
```diff
@@ -21,3 +21,75 @@
...
+		// Defer to TG brainproblems if the character isn't robotic.
+		return ..()
+	qdel(src)
```
Zonespace27: qdel at the end

modular_skyrat/master_files/code/datums/quirks/negative.dm
```diff
@@ -21,3 +21,75 @@
...
+	// TODO: Check brain type and detour to appropriate brainproblems quirk
+	var/datum/quirk/item_quirk/brainproblems/synth/_bp_synth = new
```
Zonespace27: why is _ needed at the start of _bp_synth

modular_skyrat/master_files/code/datums/quirks/negative.dm
```diff
@@ -21,3 +21,75 @@
...
+// Does not appear in TGUI or the character preferences window.
+/datum/quirk/blooddeficiency/synth
+	name = "Hydraulic Leak"
+	desc = "Your body's hydraulic fluids are leaking through their seals."
+	medical_record_text = "Patient requires regular treatment for hydraulic fluid loss."
+	icon = "bd_synth_tint"
+	hidden_quirk = TRUE
+
+// If blooddeficiency is added to a synth, this detours to the blooddeficiency/synth quirk.
+/datum/quirk/blooddeficiency/add_to_holder(mob/living/new_holder, quirk_transfer)
```
Zonespace27, line 77: same goes for here

Zonespace27 (changes requested):

code/controllers/subsystem/processing/quirks.dm
Floofies: Oops, good catch of my rookie mistake 😛

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Can you explain why? For now I relocated them above the final returns

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: It is meant to indicate that the variable shouldn't be used for anything else

code/controllers/subsystem/processing/quirks.dm
Floofies: Fixed in commit ccc1de5255e605a62e655a7ed579b757e34b5e98

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Ok, I investigated the effects of your recommendation a bit, and it doesn't make sense to me. By moving the `qdel(src)` to the end, it causes an infinite loop. The object chain is being detoured here, so it will just continue to detour itself. `qdel(src)` must be called before the `bp_synth` etc are created:
```runtime error: Maximum recursion level reached (perhaps there is an infinite loop)
To avoid this safety check, set world.loop_checks=0.
proc name: clear signal refs (/datum/proc/clear_signal_refs)
  usr: the new player (/mob/dead/new_player)
  src: Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth)
  usr.loc: null
  call stack:
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): clear signal refs()
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): Destroy(0)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): Destroy(0)
qdel(Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth), 0)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)
Positronic Cascade Anomaly (/datum/quirk/item_quirk/brainproblems/synth): add to holder(Jamari Bell (/mob/living/carbon/human), null)```

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: I playtested it and it caused infinite loops, and the function starts detouring itself. Looking into this `istype` function, and it's not the same use-case. It does not behave as you would expect in this situation. Using it leads to the conditional failing to work as intended, and causes an infinite detour loop as `istype` fails to differentiate between the different objects sharing the same parent namespace.

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Fixed in commit b4d597862c5e1e7aee0daec7868bdf1f6d080817

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Fixed in commit b4d597862c5e1e7aee0daec7868bdf1f6d080817

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Fixed in commit b4d597862c5e1e7aee0daec7868bdf1f6d080817

modular_skyrat/master_files/code/datums/quirks/negative.dm
Floofies: Changed `is_species` in commit b4d597862c5e1e7aee0daec7868bdf1f6d080817

Floofies (commented):
Implemented the requested changes. LGTM

Zergspower:
I'm a bit confused or I'm misreading, the hydraulic quirk will now cause fainting on synths?

Floofies:
> I'm a bit confused or I'm misreading, the hydraulic quirk will now cause fainting on synths?

@Zergspower Correct! Previously, synths could ignore Blood Deficiency because it didn't give them any negative effects. With this PR they will drain down to at least `BLOOD_VOLUME_BAD` or around 224cl. They intermittently faint as expected at around that amount. They don't gain any other debuffs/damage like an organic does. I play tested it extensively and it's playable, too!

Floofies:
FYI! @GoldenAlpharex wanted to review this next week, but it removed them from the review request. Whoops!

Floofies:
@RimiNosha With PR #16105 coincidentally there is some more brains to choose from, making some of this PR's added flavortext inaccurate. If their PR is merged before mine, which I'd prefer, I can add to my PR to detect both the Compact AI Circuit and IPC+Synth Compact Man-Machine Interfaces. Some other contributors have already pitched some names for brain-specific flavortext, including "Processor Deterioration" which may work for the Compact AI Circuit.

Gandalf2k15 (approved):

Gandalf2k15:
LGTM, nice amount of effort put into the PR, code is up to standard, concept is solid. Good work.

PR #16705, opened 2022-10-07
The return of the PDA ringtone preference

GoldenAlpharex:
## About The Pull Request
It got gutted when the PDA refactor came, and now it's back!

I'm marking this as a fix as I don't believe we ever really wanted to get rid of this small feature, and it was still partly there in the code.

It's ever so slightly more convenience, so you can actually use the ringtone without having to remember to set it up at the beginning of every shift.

## Changelog

:cl: GoldenAlpharex
fix: The PDA ringtone preference is finally back.
/:cl:


RatFromTheJungle:
make it say 'stupid cat' if tf4 spawns with it it's in the lore trust me on this i would never lie

Gandalf2k15:
Definitely upstream

Gandalf2k15:
Attempt upstream then here

PR #16721, opened 2022-10-08
[MIRROR] Infiltrator Suit Rework: MODsuit edition [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70303
--------------------

## About The Pull Request

Changes the Infiltrator suit and the respective parts into a single MODsuit. All the functionality of the old suit comes bundled, including two new features:

A) The wearer cannot be examined while activated, and the modsuit cannot be observed while worn. The 'Unknown'ing effect of the mask now extends to full anonymity while worn. Feel free to wear any ID you want! (You're just very obviously evil because holy shit you're blood red)

B) The suit now demoralizes crew that see it due to _PSYCHIC WAVES OooOOoooOOooh_

edit: C) Forgot to mention it: the suit control unit goes on your belt, not your back. Works basically as a really generous belt for storage. And by generous, I mean small items within the 7 slot capacity.

## Why It's Good For The Game

I really enjoyed adding the suit originally, but it was a little clunky in implementation. Then Modsuits happened, and oh look they do basically all the functionality I want. So I decided I would make the suit into a modsuit. Plain and simple.

The things lost: cool mask, special bashing toolbox and guncase.

The anonymity effect was something someone told me was the suits strongest element, but I always felt like that wasn't an aspect that was made apparent or well implemented. This makes this aspect the suits defining feature, really selling it as a 'sneak suit' in an unorthodox manner. True invisibility tends to be quite potent ~~(*cough*cloaker*cough*)~~, so rather than just make the suit stronger than necessary, it makes the suit quite good at the other kind of stealth. Just flat out not being identifiable as you make your getaway or do your evil mischief. Flip between civilian and agent on the fly! Kinda like an alter ego/supervillain.

The demoralization effect is mostly fluff but a potent driving element to be antagonistic towards anyone wearing the suit. The longer they're around, the worse your sanity gets. Dealing with them should be obvious.

## Changelog

:cl:
balance: The Infiltrator suit has been remade into an Infiltrator MODsuit. Has all the original features, but now also has additional powers: total anonymity, and a demoralization aura. 
/:cl:


modular_skyrat/modules/moretraitoritems/code/infiltrator_balaclava.dm
```diff
@@ -0,0 +1,31 @@
...
+	icon_state = "syndicate_balaclava"
+	inhand_icon_state = "syndicate_balaclava"
+	clothing_flags = MASKINTERNALS
+	flags_inv = HIDEFACE|HIDEHAIR|HIDEFACIALHAIR|HIDESNOUT
+	visor_flags_inv = HIDEFACE|HIDEFACIALHAIR|HIDESNOUT
+	w_class = WEIGHT_CLASS_SMALL
+	armor = list(MELEE = 10, BULLET = 5, LASER = 5,ENERGY = 5, BOMB = 0, BIO = 50, FIRE = 100, ACID = 40)
+	resistance_flags = FIRE_PROOF | ACID_PROOF
+
+	var/voice_unknown = FALSE ///This makes it so that your name shows up as unknown when wearing the mask.
```
necromanceranne: You're going to need to do more than this. This bit of code no longer does anything. You'll have to go back into ```human_say``` to fix it, as I removed the check for this mask's var.

code/modules/mob/living/carbon/human/human_say.dm
```diff
@@ -20,12 +23,7 @@
 				return real_name
 		else
 			return real_name
-	if(istype(wear_mask, /obj/item/clothing/mask/infiltrator))
```
necromanceranne, line 23: This is the bit of code you'll need to have readded in order to have the infiltrator balaclava work again. This is not an alternative to ```TRAIT_UNKNOWN``` either. They do different things.

modular_skyrat/modules/moretraitoritems/code/infiltrator_balaclava.dm
Zonespace27: I'm aware, my first concern with mirrors is getting them to pass checks, then i go and fix the functions after

Zonespace27 (approved):
CI's failing for other reasons

PR #16737, opened 2022-10-09
[MODULAR] Allows taurs to wear more dresses and skirts

Jureiia:
## About The Pull Request
So I noticed when trying to make a taur that a lot of clothes that taurs should theoretically should be able to wear due to them having skirts or being dresses were being cropped so went to add the flag that makes them not be cropped to _most_ of the jumpsuits that are dresses or have skirts.Obviously some of the longer skirts or dresses may still have issues on certain taur types but alas.

Ideally this would just be something you could toggle in the character menu to fix that issue but that's not something I want to touch.

Before
EFSXZ4mOce
After
sUtd06pQtz

Lets taurs have more options whilst still being modest,

## Changelog

:cl:
fix: Makes most of the dresses and skirts no longer get cropped as a taur
/:cl:


Jureiia:
I don't know why all the maps are failing...

Gandalf2k15 (approved):

PR #16755, opened 2022-10-10
Adds an override to the inhand icon CI check

Gandalf2k15:
I don't know why but currently the icon CI check only allows for icons in the inhands directory which forces us to go non-modular and move all of our inhand files to a single directory. This is not how we have operated and for now I am adding an override check to this, in essence the CI check still does it's job as it's still checking for inhand sprites, and passing if they actually exist. It's just not limited to a narrow set of sprites in one TG directory.

Also fixes the missing inhand icons that we have.


Tastyfish:
The blueshift failure was the stupid mouse harddel bug. Otherwise I believe this is ready for review.

Edit: ashwalker and xeno screenshot diffs also vary, probably just need to be updated if they were missing an inhand before, but I'm not home anymore to check.

code/modules/projectiles/guns/ballistic/bow.dm
```diff
@@ -68,7 +68,7 @@
 	name = "arrow"
 	desc = "Stabby Stabman!"
 	icon_state = "arrow"
-	inhand_icon_state = "arrow"
+	inhand_icon_state = null
```
GoldenAlpharex:
```suggestion
	inhand_icon_state = null // SKYRAT EDIT - Fixing CI for inhands.
```

GoldenAlpharex (approved):
Let's go!

PR #16758, opened 2022-10-10
Makes Obj Stairs have the INDESTRUCTIBLE flag - no more blob cheese

Zergspower:
## About The Pull Request

Adds the INDESTRUCTIBLE flag to prebuilt stairs, these cant be replaced by crew and can only be done via Map editing or admins pushing them in manually, response mostly to blobs having a rather cheesy advantage on Multi-Z maps. This will not affect ladders and there's no intent on changing those as they can be rebuilt.

Stairs

Less Cheese more something idk rhyming

## Changelog

:cl:
balance: Added an indestructible flag to stairs prebuilt (not affecting tile stairs)
/:cl:

### DIFF
--- a/code/game/objects/structures/stairs.dm
+++ b/code/game/objects/structures/stairs.dm
@@ -11,7 +11,9 @@
 	icon = 'icons/obj/stairs.dmi'
 	icon_state = "stairs"
 	anchored = TRUE
-	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Prevents blobbos from cheesing and destroying irreplacable stairs.
+//SKYRAT EDIT START
+	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Makes OBJ stairs indestructible to primarily prevent blob cheese.
+//SKYRAT EDIT END
 
 	var/force_open_above = FALSE // replaces the turf above this stair obj with /turf/open/openspace
 	var/terminator_mode = STAIR_TERMINATOR_AUTOMATIC
--- a/code/game/objects/structures/stairs.dm
+++ b/code/game/objects/structures/stairs.dm
@@ -11,6 +11,7 @@
 	icon = 'icons/obj/stairs.dmi'
 	icon_state = "stairs"
 	anchored = TRUE
+	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Prevents blobbos from cheesing and destroying irreplacable stairs.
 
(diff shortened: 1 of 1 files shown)


jjpark-kb:
I dislike indestructible things-- especially on station. A solution I would get behind would be able to actually create stairs.

code/game/objects/structures/stairs.dm
```diff
@@ -11,6 +11,7 @@
 	icon = 'icons/obj/stairs.dmi'
 	icon_state = "stairs"
 	anchored = TRUE
+	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Prevents blobbos from cheesing and destroying irreplacable stairs.
```
jjpark-kb: Non-modular change.

code/game/objects/structures/stairs.dm
```diff
@@ -11,7 +11,9 @@
 	icon = 'icons/obj/stairs.dmi'
 	icon_state = "stairs"
 	anchored = TRUE
-	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Prevents blobbos from cheesing and destroying irreplacable stairs.
+//SKYRAT EDIT START
+	resistance_flags = INDESTRUCTIBLE // Skrat Edit : Makes OBJ stairs indestructible to primarily prevent blob cheese.
+//SKYRAT EDIT END
```
jjpark-kb, line 16: this is still nonmodular-- you'll need to find a place you can put this modularly, or create a dm file for it.

Gandalf2k15 (changes requested):
Instead of this make it so stairs can be constructed. There should be very few objects with the indestructible flag.

Zergspower:
> Instead of this make it so stairs can be constructed. There should be very few objects with the indestructible flag.

Again - the issue is less about them being indestructible and more the cheese factor because multi-z sucks ass for stair visuals

Paxilmaniac:
The root of the problem is not the stairs, our players are simply unable to resist throwing themselves down a pit that is clearly full of blob, and are unable to adapt by finding one of the like 70 ladders around to get down.

Iamgoofball:
skill issues are not bugs and thus don't need fixed

the core issue here is not being able to rebuild z-level transition pieces

Iamgoofball:
this can be closed
https://github.com/tgstation/tgstation/pull/70504 resolves the core issue here

Zonespace27:
https://github.com/tgstation/tgstation/pull/70504

PR #16782, opened 2022-10-11
[MIRROR] The curator's PDA now works in space [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70319
--------------------
## About The Pull Request

The Curator's PDA, while out in space, will have low-signal, rather than none. This means that their apps (Messenger, Newscaster, ect) will still work.

## Why It's Good For The Game

Originally I was going to make the newscaster app work without NTNet, but thought this was the better solution. I also recently removed the modular computer upgrades that allow low-signal connections to NTNet from space, so this is my alternative to that as well.

Curators are many things, but is also the only job we expect to explore space. Because of this, I think it's necessary for the Curator to still be able to talk with crew members, and use their Newscaster app, to hopefully make space exploration feel like you're just on a private singleplayer world.

## Changelog

:cl:
balance: The Curator and Crystal PDA is now upgraded to have long-distance NTnet connections, they are able to use their apps in space.
/:cl:


Tastyfish (approved):
> Removes IC NTNet range control.
> Re-adds it as an immutable var. Hope you don't lose your PDA.

It's fine. Gave vanguard PDA's long range as well.

PR #16802, opened 2022-10-12
Fix hand Augments+ bodymarkings when out of FOV

Tastyfish:
## About The Pull Request

Augment+ body markings on hands specifically were still visible through FOV. Turns out they are handled specially for some reason. Fixed.

my john cena RP

## Changelog

:cl:
fix: Hands are no longer sometimes visible when the player is otherwise invisible from FOV.
/:cl:


GoldenAlpharex (approved):
Good catch!

PR #16791, opened 2022-10-12
[MIRROR] Fixes shields inhands [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70395
--------------------
Fixes #70377

Stobe shield has been broken for.... hell I don't even know how long lmao 

:cl: ShizCalev
fix: Fixed some missing shield inhand icons.
fix: The strobe shield's inhand icon now plays the flashing animation when triggered again.
fix: The strobe shield will no longer blind the person holding it when it deflects a blow.
fix: The light from the strobe shield's flash no longer appears in your held inventory slot.
fix: Fixed the east facing left-handed strobe shield sprite being misaligned by two pixels.
/:cl:

PR #16824, opened 2022-10-12
[MODULAR] Makes the armament list load the preview items correctly

Tastyfish:
## About The Pull Request

The armament entry system was loading the preview items for the thumbnail before `SSatoms`, so they weren't initializing and thus appearances couldn't update, as well as there being a new runtime in #16791.

The armament entry system now has its own subsystem, to make it load **after** SSatoms, but **before** the gun company subsystem. Now the entries get their appearances updated before the thumbnail snapshot.

### Example: Defibrillator with its overlays
Defibrillator with its overlays

Improvement, bugfix.

## Changelog

:cl:
fix: Armaments in both armament vendors and in gun cargo now have a more accurate thumbnail, with complete overlays.
/:cl:


GoldenAlpharex (approved):
Nice!

PR #16835, opened 2022-10-13
New xeno balance changes that I told golden I'd make then never did

Paxilmaniac:
## About The Pull Request
The full list of changes are as follows:
The xeno de-evolve ability has been removed, as this made containing xenos and ending the actual threat of them being around much harder.
Runner damage has been lowered from 20-25 to 15-20.
Runners now have a 48 second cooldown (80% of the ability cooldown) after using their evade ability where they will be unable to crawl into events.
Sentinels use 40 plasma instead of 25 per spit.
Sentinel spit now only does 30 stamina damage from 40, and 20 burn from 25 for non-lethal and lethal respectively.
Warriors using their agility ability will do 15-20 damage while they are in agility mode.

Xenos were just a tad too strong and uncontainable because of some of this stuff, most of the suggestions are from JungleRat and the runner changes are Golden's suggestions.

## Changelog

:cl:
balance: Xenos have lost the ability to devolve back into a larva
balance: Runner damage has been lowered to 15-20 (From 20-25)
balance: Runners cannot vent crawl for 48 seconds (80% of the ability cooldown timer) starting from when they activate the evade ability, to prevent cheese
balance: Sentinel spit now uses more plasma per spit (40 from 25) and does a bit less damage (30 stamina from 40, or 20 burn from 25)
balance: Warriors in agility mode will do 15-20 damage, which will return to normal when they stop using agility
/:cl:

### DIFF
--- a/modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/rouny.dm
+++ b/modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/rouny.dm
@@ -11,8 +11,8 @@
 	icon_state = "alienrunner"
 	/// Holds the evade ability to be granted to the runner later
 	var/datum/action/cooldown/alien/skyrat/evade/evade_ability
-	melee_damage_lower = 20
-	melee_damage_upper = 25
+	melee_damage_lower = 15
+	melee_damage_upper = 20
 	next_evolution = /mob/living/carbon/alien/humanoid/skyrat/ravager
 	on_fire_pixel_y = 0
 
@@ -55,6 +55,9 @@
 	addtimer(CALLBACK(src, .proc/evasion_deactivate), evasion_duration)
 	evade_active = TRUE
 	RegisterSignal(owner, COMSIG_PROJECTILE_ON_HIT, .proc/on_projectile_hit)
+	REMOVE_TRAIT(owner, TRAIT_VENTCRAWLER_ALWAYS, INNATE_TRAIT)
+	addtimer(CALLBACK(src, .proc/give_back_ventcrawl), (evasion_duration * 5)) //They cannot ventcrawl until 5x the duration has passed (50 seconds)
+	to_chat(owner, span_warning("We will be unable to crawl through vents for the next [(evasion_duration*5)/10] seconds."))
 	return TRUE
 
 /// Handles deactivation of the xeno evasion ability, mainly unregistering the signal and giving a balloon alert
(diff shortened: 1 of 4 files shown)


modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/warrior.dm
```diff
@@ -82,5 +85,8 @@
 	agility_target.remove_movespeed_modifier(/datum/movespeed_modifier/warrior_agility)
 	agility_target.unable_to_use_abilities = FALSE
 
+	agility_target.melee_damage_lower = 30
```
larentoun: I would suggest to use `initial()` proc, if it is applicable

Paxilmaniac: oh right that exists

modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/rouny.dm
```diff
@@ -55,6 +55,9 @@
 	addtimer(CALLBACK(src, .proc/evasion_deactivate), evasion_duration)
 	evade_active = TRUE
 	RegisterSignal(owner, COMSIG_PROJECTILE_ON_HIT, .proc/on_projectile_hit)
+	REMOVE_TRAIT(owner, TRAIT_VENTCRAWLER_ALWAYS, INNATE_TRAIT)
+	addtimer(CALLBACK(src, .proc/give_back_ventcrawl), (evasion_duration * 5)) //They cannot ventcrawl until 5x the duration has passed (50 seconds)
```
larentoun: What if that timer is called multiple times? Shouldn't it have some flags, so they won't stack?

Paxilmaniac: Maybe, I realized it was dumb to not just go off of the cooldown time so changed it to that, which should significantly reduce the chance this ever happens to begin with (looking at you other admins)

modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/rouny.dm
```diff
@@ -51,10 +51,13 @@
...
 	owner.balloon_alert(owner, "evasive movements began")
 	playsound(owner, 'modular_skyrat/modules/xenos_skyrat_redo/sound/alien_hiss.ogg', 100, TRUE, 8, 0.9)
-	to_chat(owner, span_danger("We take evasive action, making us impossible to hit with projectiles for the next [evasion_duration/10] seconds."))
+	to_chat(owner, span_danger("We take evasive action, making us impossible to hit with projectiles for the next [evasion_duration / 10] seconds."))
 	addtimer(CALLBACK(src, .proc/evasion_deactivate), evasion_duration)
 	evade_active = TRUE
 	RegisterSignal(owner, COMSIG_PROJECTILE_ON_HIT, .proc/on_projectile_hit)
+	REMOVE_TRAIT(owner, TRAIT_VENTCRAWLER_ALWAYS, INNATE_TRAIT)
+	addtimer(CALLBACK(src, .proc/give_back_ventcrawl), (cooldown_time * 0.8)) //They cannot ventcrawl until 80% of the cooldown has passed (48 seconds)
+	to_chat(owner, span_warning("We will be unable to crawl through vents for the next [(cooldown_time * 0.8) / 10] seconds."))
```
GoldenAlpharex: The 0.8 should be a define, to indicate what it means, as it's otherwise not super straight-forward.

GoldenAlpharex (changes requested):

modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/rouny.dm
Paxilmaniac: That is a good idea actually

GoldenAlpharex (approved):

PR #16861, opened 2022-10-14
[MODULAR] make title screen progress bar pacing work a little better

Tastyfish:
## About The Pull Request

So to determine if the game startup progress is unusually ahead or behind, I made the code track the historic timing of each startup message, and then adjust the progress bar shown to users so it's not more than a few seconds off by the end. However, I never accounted for the messages containing the changing time numbers, whoops. Became more obvious when the subsystems started reporting sub-second times so the `startup_message_timings` keys regularly missed.

Just didn't prioritize actually fixing this.

Made it so the internal keys used in the progress cache no longer contain the numbers (or the "seconds" / "s" suffix afterwards). Now it'll work better with the current SS code and the progress cache file will be smaller.

The progress cache file will also reset the first time since I added a version check so they don't contain junk entries.

Better startup progress bar

## Changelog

Technically player-facing but it's pre-lobby so who cares.


GoldenAlpharex (approved):
Nice.

PR #16871, opened 2022-10-15
[Fix] Updates deprecated coloring method for taurs, fixes layering issue on behind view

desertmote:
## About The Pull Request

Updates taur accessory coloring method from 'extra' 'mut' stuff that barely functioned to more commonly used matrix colors.
Updates layering so hair and clothing doesn't cover the behind view of most taurs, as that is not a desired result.
However, it is a desired result for example on Naga taurs, so they keep the same layering.

Maintaining content. Some of this layer stuff wasn't sane at all.

dreamseeker_ICHt15xT7g

## Changelog

:cl:
fix: Fixes coloring issues with taur bodyparts
fix: Fixes layering issues with back-view taur bodyparts
/:cl:


GoldenAlpharex (approved):
I actually forgot to click approve.

PR #16873, opened 2022-10-15
Changes material floor tiles to use a whiter tile as opposed to a grey one that was causing muddy looking tiles

Paxilmaniac:
## About The Pull Request
image

It helps when material datum things are actually the color they're supposed to be

Things made with custom materials will look just that little extra bit better.

## Changelog

:cl:
imageadd: The icon for custom material floor tiles has been lightened up to give things like glass and paper floors the look they were always meant to have
/:cl:


tf-4 (approved):

GoldenAlpharex (changes requested):
Can you just try and fix their shading to fit the rest of the tiles of our tileset, so they properly match?

GoldenAlpharex (approved):
Looks good, thanks!

PR #16881, opened 2022-10-16
Tourist station fixes

Mortisduem:
Minor fixes to the station

- Took away the "bomb" out of the show room due to labels not being read.
- Swapped out pin pointer in the warden office for the correct item
- Keeping the "gorilla bouillon cube" in kitchen, just updating its warning label "A Honk Co. brand gorilla bouillon cube. Now with an extra punch!"
- Fixed HOP shutter access


Mortisduem:
Wrong branch

PR #16882, opened 2022-10-16
[Modular] Oppressive Force Relocation medicell tweaks

softcerv:
## About The Pull Request
This PR does two things:
1.) Fixes the bug that allows the relocation medicell to teleport someone after shooting them twice in a row.
2.) Prevents a target from being teleported if they are buckled.

The ability to instantly teleport people with the relocation medicell was not intended. Without the instant teleport, along with requiring a patient to be unbuckled, should hopefully reduce frequency of medical wordlessly teleporting patients out. 

## Changelog

:cl:
balance: The Oppressive Force Relocation medicell  can no longer teleport mobs that are buckled. 
fix: The Oppressive Force Relocation medicell is no longer able to instantly teleport. 
/:cl:

### DIFF
--- a/modular_skyrat/modules/cellguns/code/medigun_cells.dm
+++ b/modular_skyrat/modules/cellguns/code/medigun_cells.dm
@@ -630,6 +630,9 @@
 	if(teleportee.GetComponent(/datum/component/medigun_relocation))
 		return FALSE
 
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (approved):
Good catch!

PR #16908, opened 2022-10-16
improves some of the ceramic and glassblowing sprites

Paxilmaniac:
## About The Pull Request
image

These were one of the codersprites of all time before, I can only hope these are better than what they were.

## Changelog

:cl:
imageadd: New icons for glassblowing cups, bowls, globes, and plates!
imageadd: New icons for ceramic plates, cups, and bowls!
/:cl:


Paxilmaniac:
why did create and destroy fail

Tastyfish:
> why did create and destroy fail

`/obj/item/reagent_containers` breaks if `MATERIAL_EFFECTS` is set. Just remove that flag from the glass cup and bowl.

Edit: actually just beakers, so this only applies to the glass cup.

Paxilmaniac:
I thought material effects did transparency for glass and whatever, I was wrong and it was actually also on material color

Tastyfish:
> I thought material effects did transparency for glass and whatever, I was wrong and it was actually also on material color

Yep!

jjpark-kb (approved):

PR #16926, opened 2022-10-17
Allows the recently added stone stairs type to be built using (still unobtainable, I gotta work on that) stone

Paxilmaniac:
## About The Pull Request
Simple as the title, adds a single line that makes it build special stairs

the stone stairs are pretty cool

## Changelog

:cl:
add: Stone can now be used to build stone stairs!
/:cl:

### DIFF
--- a/modular_skyrat/modules/stone/code/stone.dm
+++ b/modular_skyrat/modules/stone/code/stone.dm
@@ -7,15 +7,12 @@
 	inhand_icon_state = "sheet-metal"
 	mats_per_unit = list(/datum/material/stone=MINERAL_MATERIAL_AMOUNT)
 	throwforce = 10
-//	flags_1 = null
 	resistance_flags = FIRE_PROOF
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/stone/code/stone.dm
```diff
@@ -18,6 +18,7 @@
 //	cost = 500
```
GoldenAlpharex: Why is there so many lines like these that are commented out in this class declaration?

GoldenAlpharex (dismissed):
It's my only question, if there's no good reason to keep them, remove them and then you can dismiss my request for changes.

modular_skyrat/modules/stone/code/stone.dm
Paxilmaniac: genuinely I could not tell you why these are here or why they are commented out

tf-4 (approved):

PR #16962, opened 2022-10-18
[MIRROR] Seven Billion Hour Wait Time - Or, Expanded Job Config (TOML time) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70199
--------------------
a TOML is just a sexier JSON (bullet flies through my head, killing me instantly)

## About The Pull Request

Hey there!

Twenty-eight days ago, there was a PR up (#69648, which this is an alternative to) that would allow you to define the Minimum Required Playtime (in minutes) or the Minimum Account Age (in days) to unlock a job in config! That was great, but it unfortunately caused too many changes to an existing system, and it staled out on the workbench before it got anywhere. I don't think it's fair to have Cyborgs wait 21 days to Cyborg, especially when it's not that high-impact as a role, so I wrote up a new system.

We're going to use TOML for this new system. This isn't solely because it's new and fancy and I wanted to try out something: there's a few discrete benefits that I wanted to harness.

* Extensibility: In case we want to add even more variables pertaining to jobs in config, we don't have to write up even more contrived RegEx bullshit in order to parse a system, we can just define a string in the TOML format.

* Readability: It's much more readable for me to clearly see what value pertains to what value rather than have to memorize A,B,C,D... etc.

* Verifiability: With this, you should be able to better parse when something's invalid or mis-matches between the code and the config, which isn't something I think would be possible in the old text format. Let me know if that's incorrect.

Anyways, technical considerations. The old jobs.txt format is still supported codeside, and I wrote it such that the TOML file not existing would pass a small message, but still load the old jobs.txt. I think that's what you guys wanted after all that conversation? Sorry if I fucked that up.

Anyways, that should be all? Everything in jobconfig.toml is codebase default values (passed on this PR). Keep in mind that neither jobs.txt nor jobconfig.toml will be used if the LOAD_JOBS_FROM_TXT flag is never enabled in config. Hope that helps.

## Why It's Good For The Game

it breaks my heart every time someone adminhelps asking why they can't play cyborg for another 18 days and all i can do is shake a fist and grumble "coders..." as a tear streaks down my face.

---
The old system working on the failover code (jobs.txt exists, using that instead of jobconfig.json). 50 Captains sounds like a great idea for a proud new Headmin term, doesn't it?:

image

The new TOML loading system working (I lowered the captain's required account age by two days, aren't I a goober?):

image

---

## Changelog

:cl: san7890, with regards to FernandoJ8
config: A new TOML system has been introduced to modify different settings and variables regarding jobs! You can edit the standard stuff, like roundstart positions and overall positions of a job like usual (feature parse), but you can edit two new things! You can change how long an account needs to be "aged" on a server for to unlock a job, as well as the actual number of minutes a player must get to unlock a specific job. Nifty, eh? Do remember, this is entire system won't load without the overall LOAD_JOBS_FROM_TXT flag!

server: Keep in mind that the jobconfig.toml is now the preferred system for updating job-related config, while jobs.txt should be considered deprecated. Support for loading it via jobs.txt will remain indefinitely, but expect to see a small error every time we validate the config.

server: FYI: To use jobconfig.toml, it is strongly advised that instead of downloading the codebase default that you run the Generate Job Configuration verb (in Server tab) in order to import all of your already set variables from jobs.txt. Please check to ensure that all of your wanted settings are present in the new file.
/:cl:

Again, the system of deprecating a config in favor for a new system is based on snippets I've heard of MSO talking on the subject and something mothblocks said once. Let me know if anything is absurdly cringe and I'll change it (do tell me specifically what you want though).


config/config.txt
```diff
@@ -309,8 +309,17 @@ CHECK_RANDOMIZER
...
 ##Toggle for having jobs load up from the .txt
 LOAD_JOBS_FROM_TXT
+=======
+
+## Toggle for having jobs load up from the jobconfig.toml (or jobs.txt) file. Will use codebase defaults if commented out.
+## Jobs have specific "keys" tied to their in-game datums, those should sync up otherwise it will fail to load.
+## Setting Total/Spawn Positions to -1 will open unlimited join slots for it.
+## Playtime Requirements is in minutes, Required Account Age is in days.
+#LOAD_JOBS_FROM_TXT
+>>>>>>> 901662f59b1 (Changes job config format to use TOML, and add tools for migrating existing format (#70199))
```
tf-4:
```suggestion

## Toggle for having jobs load up from the jobconfig.toml (or jobs.txt) file. Will use codebase defaults if commented out.
## Jobs have specific "keys" tied to their in-game datums, those should sync up otherwise it will fail to load.
## Setting Total/Spawn Positions to -1 will open unlimited join slots for it.
## Playtime Requirements is in minutes, Required Account Age is in days.
#LOAD_JOBS_FROM_TXT
```

tf-4:
There's only so much I can do with config stuff

Zonespace27 (approved):

PR #16974, opened 2022-10-19
[MIRROR] Made regenerate_organs() reset external organs as well. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70427
--------------------

## About The Pull Request

As the title says, augments the species.regenerate_organs() proc to regenerate external organs. It appears this was missed when mutant organs were largely replaced with external organs - leading to cases where revival such as admin healing does _not_ currently restore external organs. This has been corrected.

As external organs can be placed on species that do not normally have them, regenerate_organs() will also check all external organs on the target carbon and remove ones that their species doesn't naturally have - properly restoring the "default" appearance of the mob.

Also slightly modifies on_species_gain() so that it does not insert external organs that are already there - which in some cases alongside a regenerate_organs() call, could cause a full extra set of external organs to drop onto the ground.

## Why It's Good For The Game

This proc is supposed to restore _all_ organs to default condition. In the current code, even admin-healing cannot restore external organs such as a moth's wings. Obviously, this is not intentional behavior.

## Changelog

:cl:
fix: made regenerate_organs() apply to external organs as well as internal ones
/:cl:


Tastyfish:
Note that the new version of the monkey screenshot reference has more correct tail layering, not sure how that was broken for us and not upstream before this PR.

GoldenAlpharex (approved):
Yet more changes to the monkey screenshot 😔

PR #16989, opened 2022-10-19
[MIRROR] [MDB IGNORE] Makes only station areas, station areas. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70182
--------------------
## About The Pull Request

GLOB.the_station_areas will now only be populated with ``/area/station``:
1. ``/area/mine`` is also now exclusively the mining station
1.1. ``/area/mine/explored`` was only used once, for Tramstation maintenance, so I made it use Lavaland's explored instead, I also removed FLORA_ALLOWED from it so faunas won't spawn. This path was otherwise entirely unused and usually explored arent supposed to spawn fauna, so I thought this would be best.
1.2. ``/area/mine/unexplored`` was only used once, on Icebox's Last Ash Drake. I replaced it with Lavaland's unexplored instead, since they have their own explored/unexplored subtypes (Icemoon does too, but I didn't want to change mechanics, and this is how it was before)
2. Removes NO_ALARM flag entirely, as it was only used to exclued non-station areas from showing up in alarms, but that's not needed anymore, as we don't have non-station areas in the list anymore. I don't know if this will cause a problem on Tramstation, I'll have to see.
3. Removes shuttles and icemoon from anomaly picker, as it was no longer needed there either.

## Why It's Good For The Game

Now only the station is 'the station', which will hopefully remove many problems.

I won't have an UpdatePath here because there's a pretty important distinction between Icemoon and Lavaland subtypes, so downstreams will have to manually decide which ones they need and use those accordingly.

## Changelog

:cl:
fix: Shuttles are no longer part of the station, as they're god damn moving ships.
/:cl:


Paxilmaniac:
nightmare

Zergspower:
Pain...

Paxilmaniac:
actually not as nightmare as i thought

Paxilmaniac (approved):
remember to not gaslight yourself into thinking a merge mirror is worse than it really is

PR #17040, opened 2022-10-20
[NO GBP] [MODULAR] Fixes Breast Pref Update Almost Always Runtiming

RimiNosha:
## About The Pull Request
Yup. No one, including me caught the fact that I did dumb dumb and caused people's breast size prefs to reset, as well as generating a runtime, which is *problematic* when there's about to be a whole new section added, and a runtime would prevent certain folk's races from not being reset to human.

Less bug good. More bug bad.

## Changelog

:cl:
fix: Breast prefs are no longer reset and throw runtimes on savefile update. Woops.
/:cl:


GoldenAlpharex (approved):
Thank you.

PR #17047, opened 2022-10-20
Upgraded booze dispenser no longer contains juice

YakumoChen:
## About The Pull Request
A longstanding bug in the way we handle upgrades to chem dispensers (why nobody took it upstream is anyone's guess) made it so booze dispensers, a child of soda dispensers, also contained the upgrade chems of soda dispensers at tier 3 and 4 parts.
Also removes the duplicate applejack, for some reason.

bug fix

## Changelog
:cl:
fix: Finally, you should no longer have your booze dispenser bloated with random juices when upgraded.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/chemistry/machinery/chem_dispenser.dm
@@ -665,6 +665,8 @@
 		/datum/reagent/consumable/ethanol/wine,
 	)
 	upgrade_reagents = null
+	upgrade_reagents2 = null //SKYRAT EDIT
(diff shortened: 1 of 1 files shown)


RimiNosha:
If this is an upstream bug/oversight, then shouldn't this be taken upstream?

YakumoChen:
no this is a fix for a bug created by skyrat edits for a feature that probably should have been taken upstream.

RimiNosha:
Ohhh, I see! Nevermind, then!

tf-4 (approved):

PR #17057, opened 2022-10-21
[MIRROR] /mob/living/simple_animal is now frozen. Add a unit test to block all new simple_animal subtypes. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70669
--------------------
@ tralezab has assured me that basic mobs are now intuitive enough to be copied and pasted from even by more novice coders. He is also actively working on a contribution guide on how to transition old simple mobs, and how to create new basic mobs.

With this, we have agreed to completely freezing simple_animal. All new "simple" mobs MUST be basic mobs. Ask maintainers, especially Armhulen, for help in conversion.

As simple_animals get removed, they will be removed from the list (to prevent a compile error). If a non-simple_animal is in the list, likely through a botched find and replace, the test will fail. You need an **extremely** good reason to add to this list. Remember that nobody intends to do the wrong thing, but we will never get rid of these if we make exceptions.


Tastyfish:
Now the hard question of do we
a) Add our custom simple mobs to the whitelist for now
b) Delete them
c) Port all of them to be basic mobs before merging this

tf-4:
> Now the hard question of do we
> a) Add our custom simple mobs to the whitelist for now
> b) Delete them
> c) Port all of them to be basic mobs before merging this

A is the only sane option

Iamgoofball:
C is the correct option as we will be deleting the simple_animal path when we are done with it

Tastyfish:
Probably going to keep this as a draft as a reminder. I'm holding off on converting any of the hostiles, as the subtrees and behaviors for combat stuff is still in heavy flux, and I kinda want to see if tg makes hostile "human" mobs be actual humans with ai controllers, or basic mobs.

Funce:
We follow along, add all our currents to the whitelist, and refactor them as we can (and hopefully not never)

We'll notice when refactors of simple mobs are coming through the mirrors

Funce:
This has already been merged in due to #17215, however we still need our own types to be added.

Tastyfish:
> This has already been merged in due to #17215, however we still need our own types to be added.

Yep, I just didn't merge in the addition to `_unit_tests.dm` so that animals could merge but the test would be circumvented. I'm fine with this though.

Tastyfish (approved):

PR #17075, opened 2022-10-21
Gives First-Gen borers a buff, nerfs every other generation borer.

nevimer:
## About The Pull Request

This pull request gives first generation cortical borers a big health buff to prevent their event, which costs 15, from being a nothingburger. For reference, blob costs 8. **8.** Blob requires a full crew response, and a borer can be ended by two whacks of a toolbox. 

VSCodium_5RNZav5SUL

I've lowered the cost of borer aswell, to 12, to reflect the effort needed to end a borers night. 

I've also added a proc that halves some stats of borer generations after the first one, to prevent their snowballing effect. This should result in less borers being kinda wasted threat, and less insanely powerful borers after the first. Yes, this makes second gen borers have 25 / 2 HP. 

Also I have changed the requirements for borer spawning to be medical-oriented instead of security oriented, as it requires medical aid to prevent borers. 

Borers cost almost 2x as blob for dynamic, yet, they can't summon a full crew response by literally existing (insta-red). I believe this change should prevent "nothing" from happening when a borer spawns due to them dying, while also preventing the round from being dominated by borers. People do help borers anyhow, so this buff for their HP only helps them right in the early game. (the OG spawns screwing up is harder now, every one after is weaker than the last)

## Changelog

:cl:
balance: buffed first-gen borers, nerfed others
balance: lowered borers cost to 12, from 15. 
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer.dm
@@ -199,6 +199,23 @@ GLOBAL_LIST_EMPTY(cortical_borers)
 	/// Multiplier for a borer's negative effects to their host
 	var/host_harm_multiplier = 1
 
+/mob/living/simple_animal/cortical_borer/firstgen
+	maxHealth = 125
+	health = 125
+
+/// When the borers are successful, lets make generations weaker to curb snowballing.
+/mob/living/simple_animal/cortical_borer/proc/generation_loss()
+	if(istype(src, /mob/living/simple_animal/cortical_borer/empowered))
+		return
+	maxHealth = maxHealth / generation
+	health = health / generation
+	health_per_level = health_per_level / generation
+	health_regen_per_level = health_regen_per_level / generation
+	stat_evolution = stat_evolution / generation
+	max_chemical_storage = max_chemical_storage / generation
+	chemical_storage = chemical_storage / generation
+
+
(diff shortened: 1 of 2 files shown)


Zonespace27:
Not a fan tbh, 🔑 
Edit: Looked at this again several hours later, still not a fan of making borers post tier one absolute ass to play

nevimer:
> Not a fan tbh, 🔑 

read through it and think about it, I know you didn't because you responded faster than GitHub actions did its tasks

nevimer:
this is also a draft because it needs discussion

jjpark-kb:
Before I make a comment, let me remind everyone that we need to be tactful. Making rude/harmful comments is not acceptable.

I don't enjoy this change because the difference between 1 to 2 is huge-- its quite literally double. I also fail to see the need for this? I know you have cited snowballing, but that doesn't seem right.

nevimer:
> Before I make a comment, let me remind everyone that we need to be tactful. Making rude/harmful comments is not acceptable.
> 
> I don't enjoy this change because the difference between 1 to 2 is huge-- its quite literally double. I also fail to see the need for this? I know you have cited snowballing, but that doesn't seem right.

It can be less but people complain that borers snowball too hard. Personally, I think 4/5ths as strong is the sweet spot but it's not all numbers that matter for balance.

PR #17102, opened 2022-10-22
[Fix] Feign impairment verb for 'jittering' makes your character shake, instead of your screen

desertmote:
## About The Pull Request

There was a change in impairments so this modular one applied the incorrect impairment.
It was funny but undesired for obvious reasons.

## Proof of Testing

dreamseeker_Z4aI7v3iFL

## Changelog

:cl:
fix: The 'jitter' option for the feign impairment verb applies the correct impairment
/:cl:


tf-4 (approved):

GoldenAlpharex (approved):
Good catch!

PR #17164, opened 2022-10-26
[MIRROR] Removes tablet hard drives entirely (HDD & SSD) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70678
--------------------
## About The Pull Request

This completely removes HDD and SSD as computer parts.

Effects:
- Tablets now store programs on themselves, you can't move the programs over with a hard drive.
- Computers built with metal in-hand now should work from the get-go? Or at the very least, doesn't need a hard drive to get working. I will probably make the cpu a separate thing you print and insert into the PC as a replacement, but that will be in the future.
- Tablets are easier to build now since they have less parts in general, hopefully we can get that number to 0
- Data disks are now their own path, and still works as before. You can upload, download, move over, clone, anything to them. They're basically the only way to move programs from one place to another now.
- This includes the Master system's disk. It's a data disk now. With a lot of space. Beyond more than necessary.
- Virus disks still exist, they are still disks, but they have no space on them to do anything with it.

I also tried some attempts at code improvement where I thought it was necessary, and tried adding clearer comments in places I thought were lacking.

Also, comes with an UpdatePaths, don't worry about it!

## Why It's Good For The Game

Deals with half of https://github.com/tgstation/tgstation/issues/70692- the other half is in a separate PR.

HDD and SSD were the 2 worst parts of Tablets imo, and the only reason it took this long to remove was because I didn't know exactly how to manage it, and because I knew it would be a massive timesink.

We've been removing tablet parts in general, and this brings us one massive step closer to completing that goal, because computer parts has never been a good part of tablet-building. It's too confusing for any SS13 player to bother with, and has very little gain.
It's also beneficial for coders because there's a lot of annoyance around going through the tablet, then checking the part for a hard drive, check if it exists, then checking for the app on that. Instead, you can now just check if the tablet has an app. Much more simple on both ends.

## Changelog

:cl:
balance: Tablet hard drives no longer exist. Applications are now stored directly onto the tablet.
balance: Building a modular computer with metal now should be a proper functioning computer.
balance: The R&D's Master system hard drive is now a data disk, if ever you need one that desperately.
fix: Tablets (not Cyborg ones) now have pens again, like they had before.
/:cl:


modular_skyrat/modules/contractor/code/datums/contractor_tablet.dm
```diff
@@ -115,24 +118,24 @@
...
 	var/list/data = list()
-	var/obj/item/computer_hardware/hard_drive/portable/syndicate/hard_drive = computer.all_components[MC_HDD]
 	var/screen_to_be = null
+	var/obj/item/modular_computer/tablet/syndicate_contract_uplink/preset/uplink/uplink_computer = computer
 
 	data["first_load"] = first_load
 
-	if (hard_drive && hard_drive.opfor_data != null)
-		var/datum/opposing_force/opfor_data = hard_drive.opfor_data
+	if (uplink_computer && uplink_computer.opfor_data)
```
Zonespace27:
```suggestion
	if ( uplink_computer?.opfor_data)
```

modular_skyrat/modules/contractor/code/items/tablet.dm
```diff
@@ -1,17 +1,21 @@
+/obj/item/modular_computer/tablet/syndicate_contract_uplink/preset/uplink
+	starting_programs = list(
+		/datum/computer_file/program/contract_uplink,
+		/datum/computer_file/program/crew_manifest,
+	)
+	device_theme = "syndicate"
+
+	// Contractor uplink system board has the user's data baked directly into it on creation
```
Zonespace27: code doc comment

modular_skyrat/modules/contractor/code/items/tablet.dm
```diff
@@ -1,17 +1,21 @@
...
+	var/datum/opposing_force/opfor_data
```
Zonespace27, line 9: null this in Destroy()

Zonespace27 (changes requested):

modular_skyrat/modules/contractor/code/datums/contractor_tablet.dm
Tastyfish: Fixed

modular_skyrat/modules/contractor/code/items/tablet.dm
Tastyfish: Fixed

modular_skyrat/modules/contractor/code/items/tablet.dm
Tastyfish: Fixed. Occurs to me this most likely leaked before.

modular_skyrat/modules/contractor/code/items/tablet.dm
```diff
@@ -18,5 +18,10 @@
 	install_component(new /obj/item/computer_hardware/battery(src, /obj/item/stock_parts/cell/computer))
 	install_component(new /obj/item/computer_hardware/card_slot)
 
+/obj/item/modular_computer/tablet/syndicate_contract_uplink/preset/uplink/Destroy()
+	. = ..()
+
+	opfor_data = null
+
```
Zonespace27: (I can't suggest for some reason)
```
/obj/item/modular_computer/tablet/syndicate_contract_uplink/preset/uplink/Destroy()
	opfor_data = null
	return ..()
```

Tastyfish: > (I can't suggest for some reason)

strange. but sure, changed.

Tastyfish (approved):

Zonespace27 (approved):
My things lgtm

PR #17192, opened 2022-10-27
[Non-modular] Bluespace RPED only works within a user's view range

desertmote:
## About The Pull Request

Limits the range of the Bluespace RPED to 15 tiles, the furthest a default player can see.
Considered making it variable with view_range, but you cannot hold binoculars and another item at the same time, anyway.

In my view, upgrading the whole station from your chair in R&D is the opposite of what this server's about.
I know being let into a department for a simple goal isn't always offering the most thrilling of roleplay scenarios, but it is up to the players to put some spice behind it. The game is only as fun as you can make it, offering a veto does not help make it fun.

## Changelog

:cl:
balance: Bluespace RPEDs can no longer teleport components across Z-level granted there's a camera connection.
/:cl:

### DIFF
--- a/code/modules/research/stock_parts.dm
+++ b/code/modules/research/stock_parts.dm
@@ -31,7 +31,7 @@ If you create T5+ please take a pass at mech_fabricator.dm. The parts being good
 		if(!attacked_machinery.component_parts)
 			return ..()
 
-		if(works_from_distance && IN_GIVEN_RANGE(user, attacked_machinery, 15)) // SKYRAT EDIT - BS RPED limit - Old code: if(works_from_distance)
+		if(works_from_distance)
 			user.Beam(attacked_machinery, icon_state = "rped_upgrade", time = 5)
 		attacked_machinery.exchange_parts(user, src)
 		return TRUE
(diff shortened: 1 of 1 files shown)


RimiNosha:
I think this'd have more of a place upstream, unless you've asked about it there already?

desertmote:
> I think this'd have more of a place upstream, unless you've asked about it there already?

Nope, pushing here first because of the differences in what Skyrat and TG try to be. I'm not opposed to giving it a shot, though.

YakumoChen:
Removing a clever game interaction in order to preserve the entirely non-verbal roleplay of banging on a door for five minutes while holding the BRPED in order to click on security's protolathe is not something I would consider a worthwhile addition to the Roleplay Experience. 

Science is not obligated to Upgrade The Machines. We should not be making it more tedious for those who do it. 

Furthermore this will just lead to many machines being forgotten entirely by the upgrade pass that already rarely see upgrades, such as the equipment in the prison.

Iamgoofball:
👎 
This is an intentional feature of the BSRPED

source: i coded the BSRPED and implemented it, and discussions about removing this upstream have also been torpedod because upgrading components is not actually good or fun

the actual solution is a refactor of power to handle equipment strength and removal of stock parts as a concept

Zonespace27:
Neutral-leaning negative on this one. I never see any RP past "; medical let me in" *runs around wordlessly upgrading everything before walking out*
As mothblocks said, upgrading parts is a chore, not something that's good for the game or generates much RP

tf-4:
I was in favour of this at first, but having thought on it some as well as seeing the thoughts put forward both here and on the upstream PR, I'm not sure it's something we want.
Upgrades are a chore and there's little RP that will be gained from this change. If you want to run parts in person you're still of course perfectly able to.

PR #16980, opened 2022-10-19
[MODULAR] Re-adds the fancy kimono (new sprite)

YakumoChen:
## About The Pull Request

I went to play my character best known for their fancy kimono and found out it was gone! What the hell? Put that shit back.
Partially resprited by @Ebin-Halcyon 

image
image
image
image

## How This Contributes To The Skyrat Roleplay Experience

It has a unique sprite that the recolourable yukata can't replicate. Also, yukatas aren't kimonos. 

## Changelog

:cl:
add: Readds the fancy-looking kimono that looked different from the others.
/:cl:

PR #17196, opened 2022-10-27
[MODULAR] Resprites and re-adds the fancy kimono

YakumoChen:
## About The Pull Request
Continuation of #16980

I went to play my character best known for their fancy kimono and found out it was gone! What the hell? Put that shit back.
Partially resprited by @Ebin-Halcyon 

image
image
image
image

It has a unique sprite that the recolourable yukata can't replicate. Also, yukatas aren't kimonos. 

## Changelog

:cl:
add: Readds the fancy-looking kimono that looked different from the others.
/:cl:


YakumoChen:
add the [sprites] tag I guess

Sigmar-alkahest:
i completely agree with this, although a recolorable version eventually would be nice, thinking of adding something like that too?

YakumoChen:
> i completely agree with this, although a recolorable version eventually would be nice, thinking of adding something like that too?

I'm not given the struggle I've had to make to get this in, but someone else might after this pr

glitchbow:
> > i completely agree with this, although a recolorable version eventually would be nice, thinking of adding something like that too?
> 
> I'm not given the struggle I've had to make to get this in, but someone else might after this pr

It would be really nice to kill two birds with one stone here, because otherwise later it's just going to be "Fancy Kimono" and "Colorable Kimono" which is the same, but colorable. The white bits would remain, but there would be three options for flower petals color, sash color, and main dress color. Makes it nice so the outfit can even be worn more professionally with various jobs, instead of clashing with department colors. It would also be easier to match reference images with, and probably get more general use from other players. Would be much prettier than having to use the spray can on it, or twinning with others.

YakumoChen:
I am not going to rewrite and resprite this PR for the third time because you people keep moving the goalposts of what you want

Ebin-Halcyon (changes requested):
image

remove the pixels where the red bits are then it's fine, ig.

Ebin-Halcyon (approved):

YakumoChen:
@tf-4

tf-4 (approved):
Ok

PR #17212, opened 2022-10-28
[MIRROR] Merges the Export Scanner, Sales Tagger, and Price Tagger into one item. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70314
--------------------
## About The Pull Request

The Export Scanner, Sales Tagger, and Price Tagger are all niche, cargo adjacent items that each perform a different minor task and if using one, requires usually one of the three extremely rarely. Each one requires printing a new item at the lathe, all for maybe one or two uses before it needs to be replaced.

This PR merges the three into a singular, modular item. Existing export scanners have been reflavored the "universal scanner". When used in-hand, it produces a radial to swap between the three modes, Each mode has identical function to their previous scanner, with the only functional difference being that the sales tagger now prints new tags on secondary attack as opposed to primary.

The other two items lathe recipes have been removed, as their functionality has been fully moved to the universal scanner.

Also flips the export scanner icon state to match the other 2 scanner items that have been lumped together.

## Why It's Good For The Game

Cargo now only needs to swap between 1 of three nearly identical barcode scanner looking items at a time, as opposed to having to print an item that sees functional use once every thousand rounds. By increasing the ease of use, it will also make it easier to use custom vendors, as well as print sales tags on regular rounds, as the universal scanner can still be found in shipping boxes located around the station and in tool storages.

Arguably not a balance change, as the three tools are extremely cheap to print, and don't have a formal impact on cargo balance except for tedium.

## Changelog

:cl:
qol: The Export Scanner, Sales Tagger, and Price Tagger have been merged together into the Universal Scanner.
qol: Using the Universal Scanner in-hand allows for switching between the three tools functionality on the fly.
/:cl:


code/game/objects/structures/crates_lockers/closets/secure/cargo.dm
```diff
@@ -21,7 +21,13 @@
 	new /obj/item/clothing/gloves/fingerless(src)
 	new /obj/item/megaphone/cargo(src)
 	new /obj/item/tank/internals/emergency_oxygen(src)
+<<<<<<< HEAD
 	new /obj/item/export_scanner(src)
+=======
+	new /obj/item/clothing/mask/gas(src)
+	new /obj/item/clothing/head/soft(src)
+	new /obj/item/universal_scanner(src)
+>>>>>>> f3ca3424c9c (Merges the Export Scanner, Sales Tagger, and Price Tagger into one item. (#70314))
```
press-mark:
```suggestion
	new /obj/item/universal_scanner(src)
```

press-mark: conflicts cause we commented out the clothing above (to move it to the command vendor or something idk)

Tastyfish (approved):

PR #17233, opened 2022-10-29
QM can no longer be managed by HOP.

SpaceLoveSs13:
## About The Pull Request

HR Core App(Job management) can no longer open/close/interfere/access QM role like rest of the Head roles.

Like rest of the Head of Staff roles, QM is now also locked from being interfered by the HR console.
p.s. I know beuracratic errors are not always fun, but that would require an entire rework of that event as a whole. This only respects how every head role is used currently.

## Changelog

:cl:SpaceLove
fix: QM can no longer be managed by HR core app.
/:cl:


SpaceLoveSs13:
Just realized QMs are also heads at TG. Putting it on upstream first.

PR #17248, opened 2022-10-30
You can no longer see the icon of languages you don't know

Zonespace27:
## About The Pull Request
Unless you know a language, the icon for it no longer appears before the text

A large portion of the "multicultural" aspect and the mystique about is that you don't know where everyone's from, you don't know what they're all speaking. Knowing what language someone's speaking, regardless of if it's uncommon or Nar'sian, removes a good amount of the effect. Now, a player will actually have to learn what [insert language] sounds like to be able to tell if someone's speaking it, and if they don't, it might lead to some interesting RP of "what language is that, where are you from?".

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
del: You can no longer see the icon of a language that you don't know
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/language/language.dm
+++ b/modular_skyrat/master_files/code/modules/language/language.dm
@@ -1,11 +1,23 @@
+/datum/language
+	flags = LANGUAGE_HIDE_ICON_IF_NOT_UNDERSTOOD
 /datum/language/aphasia
 	secret = TRUE
 
 /datum/language/codespeak
 	secret = TRUE
 
+/datum/language/common
+	flags = TONGUELESS_SPEECH | LANGUAGE_HIDE_ICON_IF_NOT_UNDERSTOOD | LANGUAGE_HIDE_ICON_IF_UNDERSTOOD
+
+/datum/language/draconic
+	flags = TONGUELESS_SPEECH | LANGUAGE_HIDE_ICON_IF_NOT_UNDERSTOOD
+
 /datum/language/drone
 	secret = TRUE
+	flags = NO_STUTTER | LANGUAGE_HIDE_ICON_IF_NOT_UNDERSTOOD
+
+/datum/language/machine
+	flags = NO_STUTTER | LANGUAGE_HIDE_ICON_IF_NOT_UNDERSTOOD
 
 /datum/language/narsie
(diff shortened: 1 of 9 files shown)


Jolly-66:
Maybe include a generic "?" Icon? Just a suggestion.

Zonespace27:
> Maybe include a generic "?" Icon? Just a suggestion.

That removes the fun of figuring out if it's a language at all!

nevimer:
I don't know about this one, anyone who has ooc knowledge will know anyways and this will only serve to confuse newer players. If anyone had any doubt about which language something is, they, I, would just search for the syllables in the code. And this may break hallucinations.

ghost:
Juuuuust to make sure, you tested to see if the icon is still there when you *do* know the language right? Not being able to tell if it's common or one of your languages would suuuck

Zonespace27:
> Juuuuust to make sure, you tested to see if the icon is still there when you _do_ know the language right? Not being able to tell if it's common or one of your languages would suuuck

Yep, only applies to unknown languages

Zonespace27:
> I don't know about this one, anyone who has ooc knowledge will know anyways and this will only serve to confuse newer players. If anyone had any doubt about which language something is, they, I, would just search for the syllables in the code. And this may break hallucinations.

It shouldn't break hallucinations, IK codespeak isn't broken with it (which uses the same flag)

tf-4 (approved):

theselfish:
I agree with the "question mark icon" point. Currently, it adds a level of confusion of "did someone's cat walk over their keyboard or are they speaking one of the few languages left after most of them were removed in the language PR".

GoldenAlpharex (changes requested):
I've had enough time to get a feel for this, and whilst I'm not against it, I too think it should still have some kind of icon to indicate that it was, in fact, in another language, mostly for accessibility reasons. I don't really like the added cognitive load of reading a sentence I don't understand to figure out "yeah I don't know what that says", having an icon allows your brain to have an easier time filtering it out, so to speak.

I'd like to see it tried, at the very least.

YakumoChen:
I noticed ashtongue still had an icon

Gandalf2k15:
implement requested changes and reopen

PR #17289, opened 2022-10-31
[MIRROR] Brimdemons & Lobstrosities drop (slightly) useful organs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70546
--------------------

## About The Pull Request

Goliaths, Legions, Watchers, and (as of recently) Bileworms all drop something vaguely useful when they die.
Brimdemons and Lobstrosities do not. This PR aims to fix that, so that there's at least some vague benefit to hunting them.

In this case it takes the form of organs you get when you butcher them, similar to the regenerative core from Legions.
As they're similar to the regenerative core, I modified the regenerative core to extend from a new common "monster core" typepath which these two new organs also extend.
Like the regenerative core, both of these items do something when used and something slightly different if you go to the effort of having someone implant them into your body. They also decay over time, and you can use stabilising serum to prevent this from happening.

https://user-images.githubusercontent.com/7483112/195967746-55a7d04d-224e-412d-aedc-3a0ec754db3d.mp4

The Rush Gland from the Lobstrosity lets you do a little impression of their charging attack, making you run very fast for a handful of seconds and ignoring slowdown effects. Unlike a lobstrosity you aren't actually built to do this so if you run into a mob you will fall over, and if you are doing this on the space station running into any dense object will also make you fall over (it shouldn't make you _too_ much of a pain for security to catch).
The idea here is that you use this to save time running back and forth from the mining base.

The Brimdust Sac from the Brimdemon covers you in exploding dust. The next three times you take Brute damage some of the dust will explode, dealing damage equal to an unupgraded PKA shot to anything near you (but not you).
If you do this in a space station not only is the damage proportionally lower (still matching the PKA), but it _does_ effect you and also it sets you on fire. You can remove the buff by showering it off.
The idea here is that you use this for minor revenge damage on enemies whose attacks you don't manage to dodge.

https://user-images.githubusercontent.com/7483112/195967811-0b362ba9-2da0-42ac-bd55-3809473cbc74.mp4

If you implant the Rush Gland then you can use it once every 3 minutes without consuming it, and the buff lasts very slightly longer. It will automatically trigger itself if your health gets low, which might be good (helps you escape a rough situation) or bad (didn't want to use it yet).

https://user-images.githubusercontent.com/7483112/195967888-f63f7cbd-60cd-4309-8004-203afc5b2153.mp4

If you implant the Brimdust Sac then you can use it once every 3 minutes to shake off cloud of dust which gives the buff to everyone nearby, if you want to kit out your miner squad. The dust cloud also makes you cough if you stand in it, and it's opaque. If you catch fire with this organ inside you and aren't in mining atmosphere then it will explode inside of your abdomen, which should probably be avoided, resultingly it is very risky to use this on the space station.

## Why It's Good For The Game

There's a small aspect of the Shaft Miner job where you're living off the land and scavenging parts off creatures you have hunted and I think that's pretty cool, so I would like some more of it.
It is also nice to mitigate the slight disappointment of killing a monster which just drops raw meat and maybe a pile of reagents for botany which you can't even pick up unless you brought a beaker with you for some reason.
I don't think people will really regularly be having these surgically implanted into them but maybe someone can think of something funny to do with a cloud of dust which makes people explode when they get punched some day.
Ashwalker content...?

## Changelog

:cl:
add: You can harvest useful organs from Brimdemons and Lobstrosities.
fix: You can no longer 'preserve' regenerative cores which have already expired to repair them.
/:cl:


Zenitheevee:
i can already *sense* the gamers rushing for lobstrosity implants.

Elli-Skala:
This changes the path for legion cores and will likely break a few things skyrat has added in crafting, and or the ritual system

GoldenAlpharex (approved):
Should hopefully be fine!

PR #17299, opened 2022-11-01
[MODULAR] Automapper now supports /turf/template_noop

Tastyfish:
## About The Pull Request

Automapper now supports `/turf/template_noop` for non-rectangular templates.

Note that to be safe, `/area/template_noop` should be used on the same turfs, although you might get away with doing weird things.

Note that a tile is only fully replaced now if the turf is **not** `/turf/template_noop`.
- If you put down objs and mobs on a no-op turf, it'll place those atoms top of the existing content (as you might expect).

Making this work was slightly complicated as there's no official way to inspect template content before it's loaded onto the map, so I documented what I did to read the parsed map cache in a dedicated proc. As part of this, the map is parse cached when automapper initializes now, and auto-purges when actually placed on the map, which is probably preferable anyway.

As an example since this is an exceptionally dangly template, I made the tramstation NT consultant template now just have the hallway and NT consultant room so that botany and primary tool storage don't have to be kept in sync unless the hallway also changes.

I was not able to measure any increase in map loading time as it's minor enough to be lost in the mapgen noise.

Probably useful for mappers.

## Proof of Testing

<details>

### Same as it ever was

NT Consultant office and hallway

### Except in the template

In StrongDMM

</details>

## Changelog

No player-facing changes (hopefully).

### DIFF
--- a/_maps/skyrat/automapper/templates/tramstation/tramstation_ntrep_office.dmm
+++ b/_maps/skyrat/automapper/templates/tramstation/tramstation_ntrep_office.dmm
@@ -1,15 +1,4 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
-"ab" = (
-/obj/machinery/vending/tool,
-/obj/machinery/airalarm/directional/east,
-/obj/effect/turf_decal/trimline/neutral/filled/line{
-	dir = 4
-	},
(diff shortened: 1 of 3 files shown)


Jolly-66:
Thank fucking christ.

Iamgoofball:
thank fuck 👍

GoldenAlpharex (approved):
That seems quite cool and useful, actually.

PR #17306, opened 2022-11-01
[MIRROR] Easy's Super Omega  "unarmed strike based species var moved to limbs" refractor, unarmed strike striking with specific body parts rather than it just being flavor, and brain based attacking limb selection extra chunky edition. And also bodypart traits. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70422
--------------------

## About The Pull Request
Ok first if you don't want to read any further explanation you can watch these two videos and you will understand all of the mechanical depth of this change.

https://user-images.githubusercontent.com/55666666/194788103-8019dad8-7e44-4cc8-bc8f-0a4f8f00a357.mp4

https://user-images.githubusercontent.com/55666666/194788109-345a3146-fb4b-4a2e-9c2a-e2ee786ba97d.mp4

###  **Disclaimer: Effect on gameplay extremely limited and niche read at your own risk** 

Starting at the top, regarding the species datum, the vars attack_type, punchdamagelow, punchdamagehigh, punchstunthreshold, attack_verb, attack_effect, attack_sound, and miss_sound have been removed.

All bodyparts (not just arms) now have corresponding variables for how they should act if utilized in an unarmed attack. The bodyparts vars are correlated to their corresponding species. All arm type bodyparts have been repathed through a common parent for the purposes of keeping variables consistent. The same is true for the legs.

When a carbon begins an unarmed attack, it will check the carbon's brain to see what limb should be used for the attack.
If the brain has no answers  it will default to the arm that corresponds to the active hand of the attacker. Currently in all brains except monkeys, it check to see if the attackers target target is laying down, and if so, call for a kick with the leg that corresponds with the attackers active hand. If the attacker has no useable legs, or the attacker does not have a corresponding leg to the active hand, or the target is not laying down, the brain will simply default for an attack with the active hand.

Monkeys brains (take note the difference between having a monkey brain and being controlled by monkey AI) are the exception, they will simply always choose their heads, and by default an attack with the head is a bite attack.

As an example:
Previously ethereal would make attacks that used the verb "burn" and did burn damage despite what limbs they possessed.
Now anybody with an ethereal limb will make an unarmed attack that does burn damage and has all the same verbs as an ethereal would.

And finally, the chunky finger species trait has been moved over to the species arms. Effect on other sources like insuls remains unchanged.

## Why It's Good For The Game
The main motivation is moving direct mechanical elements off of species to, quoting tralezab here, "Make species like a blueprint."
Opens more opportunities for coders to add specific elements to limbs, perhaps we could see a buff to unarmed strikes from robot arms, or a species that headbutts people to death.
Also undeniably cool, and fixes some weirdness like fully auged ethereals still doing their normal ethereal attack.

## Changelog

:cl: itseasytosee
refactor: Elements of unarmed strikes are now limb dependent instead of species dependent. Go rip off an ethereal arm, sew it onto yourself, and burn some people. 
/:cl:


Jolly-66:
This PR name is a sin.

+1

GoldenAlpharex:
This touches so much that I'd rather just start with a test-merge for like a day first, just so we know what needs fixing (hopefully not a lot!)

Alanii:
Haven't seen anyone mention it. But it looks like this PR conflicts with how the Medical Limb Grower handles creating limbs.

Not sure if the limb grower is a skyrat specific machine so I can't say if its due to this PR with certainty. However if the limb grower is indeed a skyrat only machine, this PR is the most likely cause of the issue noted below.

**Limbs do not print properly when using the limb grower.**

The synthflesh is used, but the limb does not spawn.

GoldenAlpharex:
> Haven't seen anyone mention it. But it looks like this PR conflicts with how the Medical Limb Grower handles creating limbs.
> 
> Not sure if the limb grower is a skyrat specific machine so I can't say if its due to this PR with certainty. However if the limb grower is indeed a skyrat only machine, this PR is the most likely cause of the issue noted below.
> 
> **Limbs do not print properly when using the limb grower.**
> 
> The synthflesh is used, but the limb does not spawn.

Yeah I saw that, I already opened up a PR to fix it upstream, and included the fix in this one. Should be fixed from now on!

GoldenAlpharex (approved):
It should be all ready to go now! Tested it for a while.

PR #17322, opened 2022-11-03
[MODULAR] Fixes Bureaucratic Error Being Spawned by Chaos For Sure!

RimiNosha:
## About The Pull Request
This was supposed to be removed, but due to how chaos works, the original solution ~~only removed it from dynamic~~ didn't work at all due to our archaic system, so here's your fix!

Fixes an oversight that honestly isn't obvious.

## Proof of Testing

I'm not editing every damn picture embed to work with dropdowns on every PR I make, sorry zonespace.

This is what the event viewer looks like when only the Bureaucratic Error event being in the controller list!
image
image
image
image

## Changelog

:cl:
fix: Bureaucratic Error should no longer be enabled.
/:cl:

### DIFF
--- a/config/dynamic.json
+++ b/config/dynamic.json
@@ -120,9 +120,6 @@
 
 		"Spiders": {
 			"weight": 0
-		},
-		"Bureaucratic Error": {
-			"weight": 0
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


tf-4 (approved):

GoldenAlpharex (approved):
I want this in now. It's getting a bit ridiculous lately otherwise.

PR #17325, opened 2022-11-03
Syndicate sprite touchups pt. 2184182: Not actually all syndicate stuff edition

Ebin-Halcyon:
## About The Pull Request

Tin, updates some some Syndicate sprites, mainly the combat gloves, adds a new mask for Syndicate operatives, a neck gaiter for a less conspicuous but still vaguely syndicate enough mask.

For the rest of the bulk, it was a tacticool turtleneck (the station one that's blue) update. Forgot to do this after everything else.

Sprites update. Time chugs on.

## Proof of Testing

<details>
  
Spin around of the new stuff.
Spin around

</details>

## Changelog

:cl:
imageadd: Tacticool turtlenecks have been resprited
imageadd: The Syndicate has sent in some neck gaiters to their agents.
/:cl:


GoldenAlpharex (commented):
I'm going to be entirely honest here, I don't like the belly details on the sprite at all. I feel like it's just... There for the sake of "oh I should add some details to the sprite" when it... Honestly doesn't feel necessary? Like, to me it just feels... Out of place? I don't see what it's meant to achieve, and it's not like if plainer-looking t-shirts were a spriting sin either.

There's also something that feels slightly off with the angle of the arms on the side sprites? Like, they don't follow the natural shape of the arms on the default human sprites, which feels a little... Strange?

Ebin-Halcyon:
> I'm going to be entirely honest here, I don't like the belly details on the sprite at all. I feel like it's just... There for the sake of "oh I should add some details to the sprite" when it... Honestly doesn't feel necessary? Like, to me it just feels... Out of place? I don't see what it's meant to achieve, and it's not like if plainer-looking t-shirts were a spriting sin either.
> 
> There's also something that feels slightly off with the angle of the arms on the side sprites? Like, they don't follow the natural shape of the arms on the default human sprites, which feels a little... Strange?

I've done the chest crease for all the turtlenecks, along with it being there for other turtlenecks, ala the cargo one. It's an easy guide for the eyes and a simple way to break up the repeating pattern of the cableknit.

there's a small divet on the arm there to imply an elbow curve where there can't actually be one, i've also been doing this for all my recently sprited clothes, along with upstream clothing doing much the same in recent times.

With this as an example/tutorial i've followed for my clothing:
image

GoldenAlpharex:
> I've done the chest crease for all the turtlenecks, along with it being there for other turtlenecks, ala the cargo one.

It being done on other sprites doesn't necessarily mean that I think it really looks too great on there either. I wear a lot of shirts (shocking, I know), and they don't have a belly crease, because I'm just a very flat person. I don't think that, like... _Forcing_ depth onto clothes in such an intense way, is a direction I really like seeing being taken.

Ebin-Halcyon:
> > I've done the chest crease for all the turtlenecks, along with it being there for other turtlenecks, ala the cargo one.
> 
> It being done on other sprites doesn't necessarily mean that I think it really looks too great on there either. I wear a lot of shirts (shocking, I know), and they don't have a belly crease, because I'm just a very flat person. I don't think that, like... _Forcing_ depth onto clothes in such an intense way, is a direction I really like seeing being taken.

There needs to be a visible indicator of atleast a bit of depth in an outfit, and unfortunately at 32x32 you need to force it a bit. Upstream does it more with shadows rather than a small divet. With the outfit getting hit with light from the top.
image
image
image

Gandalf2k15 (approved):
ok

PR #17346, opened 2022-11-04
[MIRROR] Allows mobs caught in disposals to escape once they have stopped moving - Take two [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70908
--------------------

This is a re-opening of #69960, after addressing all prior reviews (except one, see below).

## About The Pull Request
- A mob inside a disposalpipe holder object that has stopped can use the "resist" action (hotkey: B by default) to struggle out. This will take 20 seconds (on top of the time spent waiting for their holder to stop), and will play a sound. The pipe they are in will burst (taking the tile with it if applicable), and deal a random 5 to 15 brute damage to all mobs in the holder.
- When merging a moving holder with a stopped holder, pipes will now check if the stopped holder has a mob, and if so, will merge the moving holder into the stopped holder rather than the reverse. Thus, all new mobs will become stuck on a stopped mob.
- When deleted, holders will now drop their contents to the floor first. This is to prevent a very rare bug deleting all holder contents (including mobs) if the holder attempts to move through a pipe currently being destroyed.

## Why It's Good For The Game
Disposal traps are fun, but there shouldn't be a scenario where a player is trapped indefinitely, with no ability to use tools or other items to escape, that can be made for free using round-start items. If you want to make a trap, the disposals pipes should be the delivery system, not the destination.

## Changelog

:cl:
add: You can now resist out of disposals pipes once you have stopped, at the cost of some brute damage.
/:cl:


Tastyfish:
Can't test my modularization of hurtsposals currently because mobs in disposals holders got broke upstream. Will hold until fixed.

Tastyfish (approved):
I also fully modularized the pipe side of hurtsposals.

PR #17432, opened 2022-11-09
[MODULAR] Fixes more bad clothing variation overrides

Tastyfish:
## About The Pull Request

There were a few more non-hat clothing variant overrides that pointed to out of date paths.

There were also a bunch of entries in under.dm that were duplicated in the same file for some reason, so I removed the duplicates.

There were also a few items that just no longer exist at all, more often than not because they were replaced with modsuits (eg ninja helmet, chrono helmet).

Clothing variants actually pointing to real paths and not creating fake unset items.

## Proof of Testing

<details>

I dunno what to put here. Here's the outputs of the scripts I made to find these.

(Note that it just can't find the bee costume because upstream there's a shitty comment in the definition line)
image

No more duplicates anymore

</details>

## Changelog

:cl:
fix: A few clothing items work better with digi characters again.
/:cl:


GoldenAlpharex (approved):
Good catch!

PR #17441, opened 2022-11-11
Removes the firefighter and any ability to make it

Paxilmaniac:
## About The Pull Request
As shrimple as the title, removes the modular firefighter ripley.

The firefighter is both mostly unmaintained and as I've heard from many people, broken most of the time. Looking back the reasoning behind readding it was to have a good mining mech, but that's literally why the clarke exists already, a job that the clarke also does better even.

So, rather than just have 'the clarke but worse and broken most of the time' I've decided its better to just do what tg did and keep it gone.

## Proof of Testing

<details>
But if you close your eyes....

image

Does it almost feel like nothing's changed at all...
</details>

## Changelog

:cl:
del: The 'firefighter' powerloader variant, a design many years out of date by this point, has finally been fully decommissioned by the Nanotrasen robotics division following a series of industrial accidents and subsequent lawsuits. 
/:cl:


Zenitheevee:
As someone who first agree'd to re-adding the firefighter:
It honestly is good to finally put it to rest.
For what it *Was* as a mining mech, It... Kinda failed outside of very niche conditions. Its tankiness was rarely as useful as Clarke's speed, And if it is of any need for something to *be* tanky... The Ripley is definitely a bit of a better option.
Also like
Its creation required a bit of an unconventional item to use at one point and i think we should appreciate removing the one mech that had a clothing item requirement at one point in its existance

Zonespace27 (approved):

githubuser4141:
> As someone who first agree'd to re-adding the firefighter: It honestly is good to finally put it to rest. For what it _Was_ as a mining mech, It... Kinda failed outside of very niche conditions. Its tankiness was rarely as useful as Clarke's speed, And if it is of any need for something to _be_ tanky... The Ripley is definitely a bit of a better option. Also like Its creation required a bit of an unconventional item to use at one point and i think we should appreciate removing the one mech that had a clothing item requirement at one point in its existance

This post doesn't make any sense. At all. There's literally no different between the FF'er and a (MKII) Ripley other then the FF'er being significantly tankier in all respects - The FF'er is lava- and fire-proof, the FF'er has 150 extra HP - the FF'er also has 30% extra laser resist. All buffs the MKII doesn't have - so there's no situation where a MKII would be superior to the FF'er in regards to tankiness or any aspect. (I'm ignoring the MKI because it's different.)

Comapred to the Clarke, well, you'd be hard-pressed to kill megafauna with a Clarke (which isn't it's purpose, I get that.) OTOH it's roughly comparable to a speedy  mining Durand w/ extra laser armour on Lavaland which can be upgraded with goliath plates and holds two people. So if you're only interested in mining, a Clarke is superior but a FF'er can kill megafauna too.

tl;dr I'm really scratching my head trying to think of a reason a mech with less hitpoints, armour, no lavaproofing and the *same speed* would be in any situation tankier. It's exactly the same but -weaker-. Like how. What.

Also.. correct me if I'm wrong, but from what it looks like, the FF'er in this rendition didn't require a piece of clothing to be made? That's how it was built on older codebases (like Bee or old /TG), but here it was supposdly upgraded from a MKI Ripley via a exosuit fab module? Even if I'm incorrect it wouldn't be difficult to remove the article of clothing requirement.

tl;dr take 2, honestly this should be reworded as a 'robotics/mech nerf'. The FF'er with an armor booster and 3 goliath plates would've had 400 HP - lavaproof - 75 melee, 55 bullet, 85 laser - that's a whopping level of armour for a stationbound mech, especially a civvie one that *runs at Phazon-speed in thin atmosphere* The only thing it'd lose from a Durand/Gygax is it's ability to punch stuff.

PR #17447, opened 2022-11-12
Cuts the completely unused event voter system from the codebase for the sin of being a THIRD random event spawner that normally runs and for evading toggling random events off

Paxilmaniac:
## About The Pull Request
So event voter was added however long ago as a way for admins and I think sometimes players to vote on what event they wanted to roll next. Cool, cool concept I guess, it worked fine until we disabled the actual voting part of it for a few reasons. So, why remove it you might ask?

Here's a funny story, it still rolls events independent of both dynamic and the normal tg random event system, making it one of three ways a random event could roll. Due to the fact that the system ran with basically no supervision from dynamic, and the fact that it ran on completely different timers as well as being independent from the normal event system, it would roll events more often than I reckon was intended.

Even MORE on top of that was the system outright ignoring 'toggle random events' in the server tab. Yep, that's right! For all you event runners out there, if you've ever pressed the toggle random events button and wondered why random events still rolled, you have your answer.

As you should have read above, leaving the event voter system in when we really aren't using it has caused such a long list of headaches. I'd maybe do something to actually fix the system, if it wasn't you know like completely unused and did nothing but cause us to wonder what's making events roll when we don't want them to.

## Proof of Testing

<details>
  I'm not sure if there's a lot I can do here aside from showing that you can still trigger events and that they'll still roll randomly, so here we go:

image

Here is live footage of and event triggering when the event subsystem schedules it to:

image

And here is that same event actually triggering after a full minute has passed:

image

</details>

## Changelog

:cl:
del: The now unused and pretty problem causing event voter system has been removed
/:cl:


Paxilmaniac:
how did codeowner fail

press-mark:
didn't know this was even a thing lmao

Iamgoofball:
didn't we just put this in recently? why did we just disable it instead of removing it the first time lmao

Zonespace27:
> didn't we just put this in recently? why did we just disable it instead of removing it the first time lmao

"recently" was like six months ago

RimiNosha:
Oh, nice, someone beat me to it!

Gandalf2k15:
This will be reworked by me at some point.

PR #17464, opened 2022-11-15
Add support for running tests on alternate versions (#71177)

Zonespace27:
https://github.com/tgstation/tgstation/pull/71177


GoldenAlpharex (approved):
I think it'll have to be me that merges this, and ensures that the workflows work properly afterwards.

GoldenAlpharex:
Okay, let's go.

PR #17466, opened 2022-11-15
Makes the event voter low chaos event system obey the toggle random events off button, also changes event voter permissions to be the same as literally everything else event related

Paxilmaniac:
## About The Pull Request

The event voter 'low' chaos event system completely bypassed the 

image

button. This is because it runs events completely outside of the normal event system and thus did not have a check for if this button was actually pressed. Extremely annoying for admins who want to turn that off!

OH but you could just edit when the next low chaos event happens! WRONG buddy! Why? Because everything in the event panel requires R_PERMISSIONS to use. What is that? That is the permissions level that I believe staff managers and above only have that gives them permissions to edit admin ranks and other related super high level activities.

Why? No goddamn clue.

So, I've taken to fixing both of these issues by adding a check to the low chaos system for the toggle random events button, and by changing the permissions required to use the menu to R_FUN, which is what we use for literally everything else event related.

The system completely ignoring the toggle random events button was annoying enough, annoying that I've had several staff who tried to run events using it make a complaint that it doesn't work. That was easy enough to fix. Secondly the whole R_PERMISSIONS thing. I really don't understand why we felt the need to lock it behind the perms required to edit admin ranks, rather than the perms used for everything else event related. If the concern was people editing it to do things they shouldn't, guess what we've got something else in there to prevent that already, every edit made is admin logged and admin announced!

## Proof of Testing

<details>

There's not like much I could do to show these changes because one makes something NOT happen and the other, well I have every permission as a localhost so.

Here, have it compiling.

image
  
</details>

## Changelog

None of these changes are player facing.

### DIFF
--- a/modular_skyrat/modules/event_vote/code/event_vote.dm
+++ b/modular_skyrat/modules/event_vote/code/event_vote.dm
@@ -440,17 +440,17 @@
 			cancel_vote(usr)
 			return
 		if("start_vote_admin")
-			if(!check_rights(R_FUN))
+			if(!check_rights(R_PERMISSIONS))
 				return
 			start_vote_admin()
 			return
 		if("start_vote_admin_chaos")
-			if(!check_rights(R_FUN))
+			if(!check_rights(R_PERMISSIONS))
 				return
 			start_vote_admin_chaos()
 			return
 		if("start_player_vote")
-			if(!check_rights(R_FUN))
+			if(!check_rights(R_PERMISSIONS))
 				return
 			var/alert_vote = tgui_alert(usr, "Do you want to show the vote outcome?", "Vote outcome", list("Yes", "No"))
 			if(!alert_vote)
@@ -461,7 +461,7 @@
 			start_player_vote(public_vote)
 			return
 		if("start_player_vote_chaos")
-			if(!check_rights(R_FUN))
+			if(!check_rights(R_PERMISSIONS))
 				return
 			var/alert_vote = tgui_alert(usr, "Do you want to show the vote outcome?", "Vote outcome", list("Yes", "No"))
 			if(!alert_vote)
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/event_vote/code/event_vote.dm
```diff
@@ -52,7 +52,9 @@
 	low_chaos_needs_reset = FALSE
 
 /// Triggers a random low chaos event
-/datum/controller/subsystem/events/proc/triger_low_chaos_event()
+/datum/controller/subsystem/events/proc/trigger_low_chaos_event()
+	if(!CONFIG_GET(flag/allow_random_events)) // If random events are disabled we shouldn't be running these
+		return
```
GoldenAlpharex, line 57:
```suggestion
		return

```

modular_skyrat/modules/event_vote/code/event_vote.dm
```diff
@@ -487,7 +489,7 @@
 			var/time
 			if(alert == "Yes")
 				time = tgui_input_number(usr, "Input custom time in seconds", "Custom time", 60, 6000, 1) * 10
-			if(!check_rights(R_PERMISSIONS))
+			if(!check_rights(R_FUN))
 				return
```
GoldenAlpharex: This should be moved right after the `if("reschedule_low_chaos")` line, to be in line with the other ones (and to not allow for more access to those that shouldn't have access to it).

Also just this portion as a whole feels like it should be slightly re-written to make more sense, that code looks _really_ weird to me.

modular_skyrat/modules/event_vote/code/event_vote.dm
```diff
@@ -427,27 +429,27 @@
...
 		if("start_vote_admin_chaos")
-			if(!check_rights(R_PERMISSIONS))
+			if(!check_rights(R_FUN))
 				return
 			start_vote_admin_chaos()
 			return
 		if("start_player_vote")
-			if(!check_rights(R_PERMISSIONS))
+			if(!check_rights(R_FUN))
 				return
```
GoldenAlpharex: At the very least, I want these four (so including the `start_player_vote_chaos`) to remain based on `R_PERMISSIONS`, as we specifically requested for those to be usable by Staff Managers and above only.

If you want that to be based on a config or whatever, sure, that's fine by me, but as it stands for Skyrat we want those permissions to be for SM+, as they shouldn't really ever be ran in the current state of things anyway.

GoldenAlpharex (changes requested):

modular_skyrat/modules/event_vote/code/event_vote.dm
Paxilmaniac: Nah these make sense to lock behind PERMISSIONS now that you say it, just that most of the other ones didn't

GoldenAlpharex (approved):
Looks good now.

PR #17498, opened 2022-11-16
[MIRROR] Removes tablet cell parts [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71078
--------------------
## About The Pull Request

Removes cell parts and cell part cells, now tablets directly stole a power cell in them, and uses regular stock cells like every other machine in the game. This also makes it less confusing because people are more used to stock cells over computer cells. Because cells generally hold more power than computer ones, I bumped up the active power usage from 50 to 75.

## Why It's Good For The Game

We are nearly finished removing all parts, holy cr*p

## Changelog

:cl:
balance: Tablets now use regular cells instead of computer cells.
balance: Tablets no longer need a power cell component to hold power cells.
/:cl:


GoldenAlpharex (approved):
Should be good to go now.

Zonespace27 (approved):
contractor stuff's good

PR #17499, opened 2022-11-16
[MIRROR] Refactors Rabbits to be a Basic Mob [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71205
--------------------

## About The Pull Request

Back in #64175, I reworked rabbits such that their base behavior was just a cute fluffy snuggle monster, and not have the "easter" variant be the default. Now that we're transitioning everything from simple_animal to basic, I figured now was the time to shift that over too.

Pretty much everything should be the same as it was before, I even took some time to add behavior to some elements to allow it to work (let me know if I should handle it a different way) but rabbits as a simple_animal and rabbits as a basic mob should now not be very distinguishable (beyond the fact that they only speak via subtrees).

I also got rid of the single-letter icon_states in the DMI and accomodated the code to fix because I finally got irritated enough to do something about that.
## Why It's Good For The Game

Although I didn't really have any pressing urge to add more complex AI behavior to rabbits than just pretty much re-implementing what they had as a simple_animal, this is an excellent first-step to allowing much more extensible behaviors to these fuzzy creatures.

Also, it takes three more mobs off "the frozen list". Whoopie!
## Changelog
:cl:
fix: Dead Black Space Rabbits should now properly have a sprite.
/:cl:

The UpdatePaths is useless for the maps we have on our repository (holodecks use a spawner code-side), but I'm going to be nice to downstreams who need it.


GoldenAlpharex (approved):
Should be good to go now!

Tastyfish (approved):

PR #17504, opened 2022-11-16
[MIRROR] mech bustin update 2022 [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70891
--------------------

## About The Pull Request

Adds a huge ass crowbar to robotics (the mech removal tool), it deals 5 damage unwielded, or 19 wielded. (should be fine, considering robotics also has the easiest access to the materials needed for a chainsaw)
You can use it while wielded on mechs to break the occupants out. This takes 5 seconds (or 3 in an unenclosed mech like a ripley)
When you die in a mech you no longer automatically get ejected.
refactors fire axe cabinets to support more items than the fireaxe
makes some vehicle code better
closes #70845 (you can still enter a mech without limbs, i think thats fine because you can use it to protect yourself from death in a dangerous situation or something until someone breaks you out with the really large crowbar)
video: https://streamable.com/x4gom2

## Why It's Good For The Game

robotics having a giant ass crowbar to break people out of mechs seems like a fun idea
you currently cant exit a mech if youre incapacitated inside it unless you DIE

## Changelog

:cl: Fikou, sprites by Halcyon
refactor: fire axe cabinets support items that aren't fire axes
balance: mechs no longer eject you when you die in them
add: Adds a giant crowbar to robotics, it can break open mechs to eject their pilots.
/:cl:


GoldenAlpharex (approved):
Looks good, although I think the map changes will need to be looked at just to be sure they're not like the wrong ones.

Zonespace27:
Needs void raptor

Jolly-66 (approved):
Maps checkout.

PR #17622, opened 2022-11-21
[MIRROR] Coffee Shop DLC - done right this time edition [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70991
--------------------
## About The Pull Request

this is a re-attempt at PR #70725 that was practically ready but was ruined in the course of my hardships with git. Mistakes were made.

In this re-edition I also addressed a few suggestions from the comments of the original pr. There is a shaker added in the meta station cafe and the pill bottle is moved out of sight to the fridge (I still left it cause it contained antidepressants for the especially overworked NT employees). Additionally, the naming of  `/cup/glass/coffee` has been handled differently this time to minimize the need of changing the code in multiple places.

Please refer to the original PR for all details concerning the content, below I add just a rough line-out for the sake of coherency.

image

The content of the pr extends to:
- a new coffee bean driven coffeemaker
- syrup bottles
- a coffee condiment display box
- almost complete make-over of the meta station cafe
- adding the new coffeemaker in a few break rooms on delta and tram (2 machines per station)

## Why It's Good For The Game

Please refer to #70725

## Changelog
:cl:
add: After a massive success of the Modello 3 series, Piccionaia Home Appliances rolls out a completely new coffeemaker model and renovates the meta station cafe for free in a promotional campaign!
add: Syrup bottles, condiment displays, and more, to make the spess coffee experience even better
/:cl:

PR #17626, opened 2022-11-21
[MIRROR] [MDB Ignore] Unit Tests for Invalid Space Turfs (Area Bullshit Edition) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/70967
--------------------
## About The Pull Request

So, there's some bullshit with the map loader(?) sometimes where it'll let space turfs spawn in spots where we REALLY don't want space turfs. Or, it could also just be a mapper screwing up. Anyways, we might miss these, so let's set up a broad Unit Test that checks and verifies that these round-ruining snagglers do _not_ exist.

In order to help me to do this, I standardized and fixed the nomenclature such that `/area/ruin/space` is default for any map file in `_maps/RandomRuins/SpaceRuins`, as well as it's subtypes. I also touched up how we handle shuttle areas in these scenarios. This got a lot of Unit Test noise filtered out, and is crucial for its functioning. It should also be how we did it from the start anyways. I added in an UpdatePaths for any compatible change, but it was completely non-workable for some of the area type updates.

I also fixed any organic bugs that didn't require an areas type update. Cool.

Placing space turfs on IceBox:
image

Organically found issues:
image

I also added a `planetary` variable to `/datum/map_config` because I didn't like the hack I was using to see if we had a planetary map, and I'd rather it just be an explicit variable in the map's JSON.

## Why It's Good For The Game

The less times we get Space Turfs showing up on IceBoxStation, the better. It also standardizes areas a bit more, which I like (we were using some incorrect ones in the wrong spots, so those were touched up in this PR as well). Like, if it's a space ruin, we don't need to use the lengthy `/area/ruin/unpowered/no_grav` when `/area/ruin/space` does the same thing.
## Changelog
Nothing in here should concern a player (unless I broke something)

Expect a few commits as I spam unit tests a few times and play whack-a-mole with bugs.


Jolly-66 (dismissed):
This needs #17622 merged first so it can pass CI.

SOURCE: CI and I'm doing mirrors for my server as well.

Paxilmaniac:
The pr jolly mentioned has been merged

Paxilmaniac (approved):
Fixed up the mapping related issues, none of the test failures should be because of this pr anymore

Jolly-66:
What Paxil said.

PR #17651, opened 2022-11-22
[MIRROR] Biogenerator tweaks, leather makes more belts and clothing [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71175
--------------------
### Revamped the biogenerator UI:

https://user-images.githubusercontent.com/3625094/200973283-b703f21b-c747-493e-98d9-043eef86d410.mp4

### Changed biogenerator icon to use layers and see the biomass level:

https://user-images.githubusercontent.com/3625094/201396065-caeaa412-6676-46f6-875e-efa2dca34985.mp4

### Biogenerator rebalance:

- Now you don't need the beaker to print solid products.
- Biogenerator now accepts all food, not just plants.
- Biogenerator now treats all nutriment subtypes as nutriments, so vitamins and proteins also turn into biomass.
- Biomass now has the same units as other reagents (you get 5 biomass from 5 nutrient with tier 1 parts).
- Doubled the cost of all items and reagents. (biomass generation reduced by 10 and prices - by 5)
- Chemicals output amounts are now in units and you can select how much you want to output exactly. It will not let you specify more than the size of container or above 50 units with one button click.
- Reduced the amount of stored items and introduced a limit to the biomass, both tied to the matter bin tier.

### Recipes changes:

Made biogenerator more dumb by moving the clothing out from the biogenerator designs, and extending leather recipes instead.

The biogenerator is a grinder/recycler style machine so it doesn't make sense that it outputs clothing.
Also you need to make leather to craft the toolbelt, while you can't do the same to craft job-specific belts.
Now you can print leather in biogenerator and craft the leather clothing by using the leather in-hand.
And the rice hat is now crafted from bamboo, instead of biogenerator.

Also added paper to the biogenerator recipes as it makes stuff from cellulose and barely anyone knows that you can craft paper from 1 log and 50 water. And paper is needed in large quantities to craft some items, like paper frames.

And it doesn't output a pack of rolling paper. It's dumb now. It prints the rolling paper sheets instead.

## Why It's Good For The Game

Biogenerator had terrible UX and backend logic. I didn't improve much on BE though, but now it should be less frustrating to use.

Also I hate how biogenerator is superior to all other means of obtaining its products. It doesn't make sense to grow and grind wheat, for instance, when you can just throw shit into biogenerator and get the flour fast. And the costs are ridiculous - you can get a couple of bottles of fertilizers just from one medium potato. 

It honestly begs for more nerfing, at least to make the nutriment - chemicals exchange rate 1:1.

The reason for the biomass cap is because people use it as a sink for veggies and generate infinite biomass. Maybe the limit will make them care more about the part upgrade and offload some of the veggies to the fridge for the Cook.

Also it was weird that biogenerator could tailor some things, while others have to be crafted in-hand. Now you can print leather and craft all types of belts and leather clothing.

## Changelog
:cl:
refactor: biogenerator UI revamped
qol: biogenerator no longer requires beaker for materials, monkey cubes and nori
balance: biogenerator accepts all food, not just plants
balance: biogenerator treats all nutriment subtypes as nutriments (vitamins, protein, etc.)
balance: biogenerator product prices doubled
balance: biogenerator biomass storage is limited depending on the level of matter bins
balance: cowboy boots recipe moved from crafting to leather recipes
balance: leather clothing & belt recipes moved from biogenerator to leather recipes
balance: rice hat recipe moved from biogenerator to bamboo recipes
balance: biogenerator now outputs rolling paper sheets instead of a pack
add: biogenerator can now print paper
imageadd: biogenerator icons now use overlays, have emissive layer and indicate the biomass volume
/:cl:


Tastyfish (approved):
Tested recipe changes.

PR #17697, opened 2022-11-23
[MIRROR] Fishing Odds Code Improvements and Rescue Hooks [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71415
--------------------
## About The Pull Request
I wanted to try and implement an easier way for people to fish out corpses from chasms, as I heard many tales of people trying to fish others out of chasms and it taking over one IRL hour, with some cases where it would take over two hours. Obviously, that's not really interesting gameplay, and it doesn't really give people an incentive to fish, it just turns it into an annoyance that people won't want to do for fun. Now, we don't want that, do we?

As such, I've created the rescue hook, a special fishing hook that can only be used in chasms (as that's currently the only place you can find people into), which will only be able to fish out duds, skeleton corpses, any mob that's fallen into a chasm and hasn't been rescued yet, or rarely, a hostile monster lurking below. It has, at the time of writing this, a weight of 5 (50 without bait, lower with bait) for duds and a weight of 30 for chasm detritus, which themselves have a 50% chance to be a random skeleton corpse, or a lobstrosity, and the remaining 50% chance of fishing out a mob that's fallen into a chasm. I'm open to tweaking these values if we think it's too easy or too hard, but it's still a rather expensive item, so I'd consider it quite fine the way it is myself, as it's still not risk-free.

It's currently only obtainable through buying it from cargo in the goodies section, at a default price of 600 credits (making it SIGNIFICANTLY more expensive than the rest of the fishing content, and making it something that assistants will have to put some elbow grease into if they want to be able to afford it).

As it stands currently, it can't be used to recover the fallen's belongings that weren't on their person (i.e., their crusher if they were holding it in hands), ~*but* I'm down to make that easier to fish out using, for instance, the magnet hook, while also making it incompatible with fishing out bodies, which would make it a nice way to recover those lost items without spending over an hour fishing for them, if that's something that maintainers would want.~ Maintainers did want it, and as such...

The Magnetic hook is now the go-to hook to retrieve objects from chasms! Not only does it inherently do a much better job at fishing out non-fishes, it also has a lesser chance of retrieving random junk from chasms, and an even lower chance of fishing out lobstrosities!

I also improved the code for the fishing weights calculation so that the hooks and the rods can have an effect on the odds of certain types of rewards more easily, with the option of offloading a more of what's currently being calculated on `fishing_challenge` over on the rods or even the hooks themselves.

I finished by fixing a handful of capitalization and punctuation issues in various fishing items, as that bugged me when I was testing my changes.

## Why It's Good For The Game
Corpses being recoverable from chasms was a great idea, however making it so people would have to sink a major portion of their shift for a chance at recovering a corpse doesn't create a particularly interesting gameplay loop. However, being able to spend your hard-earned funds in order to streamline that process without really being able to use that to cheese other mechanics sounds like a great deal to me.

## Changelog

:cl: GoldenAlpharex
add: Added a Rescue Hook, that will allow the fishing rod it's attached onto to become a lot more proficient at recovering corpses from chasms, at the expense of making it unusable for more traditional fishing. It isn't entirely lobstrosity-proof, however...
balance: The magnetic hook can no longer fish out corpses from chasms, but will fish out items much more efficiently than any other hooks, while also being much less attractive to lobstrosities. Some still fall for it regardless, however.
spellcheck: Fixed the capitalization and punctuation in the description of multiple fishing accessories.
code: Improved the code for fishing weights, to allow for different hooks to have some more noticeable results on the weights without having to add to an already massive proc.
/:cl:


GoldenAlpharex (approved):
Very nice PR 😎

PR #17229, opened 2022-10-29
[MODULAR] Complete DS2 remap + 2 new syndicate spawners

zydras:
## About The Pull Request

Congratulations, Operatives. DS2 has been selected for an upgrade and has now been turned into a fully operational spaceship. 
Syndicate Command has found fitting to hire more personnel to your cryobanks, and you will now see Mining Officers and Corporate Liaisons staff your station. 

<details>
 <summary>The full map</summary>

image
</details>

<details>
 <summary>New SecHUD icons & IDs</summary>
 
Screenshot (1650)
Screenshot (1649)
</details>

## How This Contributes To The Skyrat Roleplay Experience

* DS2 had issues with being sprawled out and without any proper objective or game loop compared to Interdyne. This led to DS2 oftentimes being deserted in favor of everyone being on Interdyne. A complete remap ought to fix that.
* The addition of 2 new spawner types (Mining Officer & Corp. Liaison) will further increase the ability DS2 has to roleplay. Mining Officer is simply the Syndicate version of a Shaft Miner while Corporate Liaison is akin to a Syndicate Head of Personnel, now with a fitting quarters.

## Proof of Testing

<details>
 <summary>They said 'we need proof of testing', they're gonna get proof of testing.</summary>

Screenshot (1614)
Screenshot (1615)
Screenshot (1617)
Screenshot (1618)
Screenshot (1619)
Screenshot (1620)
Screenshot (1621)
Screenshot (1631)
Screenshot (1623)
Screenshot (1624)
Screenshot (1625)
Screenshot (1626)
Screenshot (1627)
Screenshot (1629)
Screenshot (1630)
</details>

## Changelog

:cl:
add: DS-2 has been fully retrofitted into a working spaceship and have had their ID trims fully updated.
add: DS-2 is now staffed with two Mining Officers and a Corporate Liaison.
add: Donk Co. has provided DS-2 with all new outfits for most of its members.
balance: Many of the DS-2 gear the ship starts with has been either cut, modified or rebalanced. No more loot pinata. 
del: DS-2 has had its pool decommissioned and was told to "visit the space hotel". Rude!
/:cl:

PR #17774, opened 2022-11-26
[Fixes Broken Mirror: #14965] Greyscales ID Trims

unit0016:
## About The Pull Request
Yeah yeah, haven't played in forever, promised this when I noticed that it wasn't mirrored properly in #17229 (not their fault) - now *every* ID trim can be recolored - you get it.
I'll freely admit I probably scuffed something as it's a nightmare to navigate through skyrat's modular content even with VSC, but in live any differences shouldn't be too noticeable and it's an easy fix if they do crop up. Do note this requires some uber-light maintenance to keep three icon states up-to-date, but it'll be fineee.

Additionally, gets rid of a few unused trims and trim icon states, and redoes the DS-2 Miner icon to be a different take on a pickaxe instead of just the funny redtider icon again.

...Also gets rid of a meanyface code comment of mine from a year ago. Honk.

Having a bunch of ID cards non-recolorable while others are... isn't intuitive at all.

## Proof of Testing

<details>

image

image

image

  

</details>

## Changelog

:cl:
fix: Skyrat's ID trims are now recolorable, as they are upstream. XOXO.
/:cl:


Ebin-Halcyon (approved):
Hi Ruby :]

unit0016:
> Hi Ruby :]

🦈

PR #17807, opened 2022-11-28
Adds a subtype of that coat I added that's slightly shorter

Paxilmaniac:
## About The Pull Request
As the title might imply, adds a colorable jacket that is very similar to the other one I added, except this one is missing the bottom hem greyscale group so it can be shorter.

I realized that the original was a bit too long for some looks, and that people would probably want something that's about the same size as some of the shorter jackets we have.

## Proof of Testing

<details>

OUHHH!!!!

image

  
</details>

## Changelog

:cl:
add: The tailored jacket now has a shorter variant that doesn't go as far down as the original does.
/:cl:


Ebin-Halcyon (approved):
cooltext424589430608330

Tastyfish (approved):
Lordy, that's a lot of GAGS overlay layers. I guess the longcoat already does that so it is what it is.

PR #17819, opened 2022-11-29
Undoes the disaster that is having two different 9mm and 10mm rounds when you can achieve the same exact thing with a single var on the gun

Paxilmaniac:
## About The Pull Request
I find it silly that we need that have two different versions of rounds when the only difference between them is damage. We have a var on guns called projectile damage multiplier I can invalidate the existence of 9x19mm by putting that var with a value of 0.5 on a gun.

I also just removed the 'advanced' subtype of bullets because literally all it did was set an icon and those icons weren't great especially compared to what the tg projectiles folder has for bullets.

I've seen a large number of players get a gun that takes a 9mm or 10mm ammo, then go to get ammo for it, only to be confused as the specific one they found doesn't actually go in that gun.

## Proof of Testing

<details>

Just to show that everything still works its not all missing textures and all.

image
  
</details>

## Changelog

:cl:
qol: There is now once again only one type of 9mm, and one type of 10mm, no more playing the does this ammo type go in my gun game with one of the several rounds with the same exact naming scheme.
balance: Coincidentally, all guns that use a 9mm now use the same ammo, similarly with 10mm firearms, though the damage of the rounds they fire should be very similar if not the exact same as it was per gun before this pr.
/:cl:


code/modules/projectiles/ammunition/ballistic/pistol.dm
```diff
@@ -3,30 +3,30 @@
 /obj/item/ammo_casing/c10mm
 //	name = "10mm bullet casing"						// SKYRAT EDIT: Original
 //	desc = "A 10mm bullet casing."					// SKYRAT EDIT: Original
-	name = "10mm Magnum bullet casing"				// SKYRAT EDIT
-	desc = "A 10mm Magnum bullet casing."			// SKYRAT EDIT
+	name = "10mm Auto bullet casing"				// SKYRAT EDIT
+	desc = "A 10mm Auto bullet casing."			// SKYRAT EDIT
```
tf-4: do we need "Auto" in there?

Paxilmaniac: Not really its just that every other round has a funny name so

tf-4 (approved):

PR #17852, opened 2022-12-01
[MIRROR] The Mining vendor now works like the Chef produce console (has to go through Cargo) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71023
--------------------
## About The Pull Request

Now comes with a Hackmd: https://hackmd.io/ImTe5FLeTgmI7spTWKBaFQ?view

In-game screenshots:

image
image

image

Removes the old Mining vendor console and all its bad code. Instead, the chef produce console was generalized and a NEW mining vendor is a subtype of it. If they try to Express this console, it will be 1.5x the mining points (compared to 2x for the Chef produce console atm), so it is technically possible, but it is still better to order it through Cargo.
Different to the Kitchen crate, this one is a private order by the Shaft Miner, using mining points instead of Credits. Cargo CAN emitter it open, but I think that's an acceptable risk with all crates. As shown in the screenshot, Cargo will immediately know who ordered the items so knows who to call to pick it up when needed.

This also means Shaft Miner's vendor is now categorized somewhat. I tried my best to make sense out of it but some items really don't make sense (laser pointer, soap...)

I split the different sections of orderable items into different files for better management, and de-hardcoded it and its TGUI to make it easier for anyone who wants to add more to it. 
I also made the produce console use paths and added ways 'categories' to produce consoles, which indicates which sections you should and shouldn't be allowed to see. https://github.com/tgstation/tgstation/pull/71007 already did part of this but it isn't merged yet so :/

Free golems are mostly unaffected by this. Their console only works in express mode and doesn't increase the prices for it. The only downside is the cooldown.

I still have some things to finish on this PR so it'll be left as draft until at least tomorrow.

## Why It's Good For The Game

https://hackmd.io/ImTe5FLeTgmI7spTWKBaFQ?view

1. A large problem currently with Miners is that they don't interact with the station, this will at least help integrate them more into their own department, by making the cost of their equipment cheaper if they bother to actually go through Cargo for their gear.

2. It also means that a non functional Cargo would affect Shaft Miners too, and as they have access to the shuttle, maybe we can expect some Miners to pick up the slack if needed.

3. The old mining vendor was the ONLY vendor in the game that had infinite stock. It doesn't need a refill or anything like any other vendor, and every other vendor uses credits, mining points is just shaft miner credits. Why are they an exception? At least being ordered through the shuttle makes sense.

4. It opens the QM being able to see easier what Miners are doing, and prevents miners from hiding on Lavaland to do nothing but hunt fauna if they were meant to be demoted or something by the QM. Basically, gives the QM more control over the people working in their department.

## Changelog

:cl:
add: Shaft Miner's equipment vendor now orders their equipment through the Cargo shuttle, though you can spend 1.5x the points to express it, making it a Mining version of the Chef's produce console, with a weakened express tax.
/:cl:


Stalkeros2:
Based??????

Zonespace27:
miners malding

GoldenAlpharex:
Something tells me the miners are just going to rush for that LTSRBT machine as soon as possible, and basically nothing of great value will come out of this change, just more tedium.

Paxilmaniac:
I think miners will rush the LTSRBT machine every single round for the first few weeks this is added then still complain until people forget about it and we never hear about it being a problem again.

Tastyfish:
Interdyne will need a LTSRBT machine that specifically works with their console, or something or else this hard halts Interdyne's game loop. Not sure.

CliffracerX:
"Wait and complain at the QM until you finish rushing the machine that lets you stop waiting and complaining at the QM" is not going to add much, especially for the cost of breaking ghostroles & frequent NTOS errors in live environments, per comments on the upstream PR.

MidoriWroth:
yeah this really ain't it chief, i see this only causing complications and unnecesary additions to interdyne and such. shouldnt have been merged upstream in the first place, its very silly.

Paxilmaniac:
Interdyne will be literally unaffected by this once I finish fixing our maps you may stop using that as an argument now.

MidoriWroth:
as golden said, it seems this only exists to cause more tedium and make people wait for something because... 'miners need more station interaction' whatever that means. id say thats a pretty good argument. its a system that should just be left well enough alone.

Paxilmaniac (approved):
Modular maps have been updated with the new vendors.

Ghost roles that had a mining vendor (Tarkon, Interdyne, so on) will see zero impact on how their vendors work because there's vars you can change to easily make a vendor independent of the cargo shuttle.

modular_skyrat/modules/mining_vendor_additions/code/order_mining.dm
```diff
@@ -0,0 +1,7 @@
+/datum/orderable_item/mining/survival_bodybag
+	item_path = /obj/item/bodybag/environmental
+	cost_per_order = 500
+
+/datum/orderable_item/mining/survival_bodybag
```
Zonespace27:
```suggestion
/datum/orderable_item/mining/suit_voucher
```

Paxilmaniac: oop forgor thank you

CliffracerX:
Even with ghostroles now left well enough alone, it still doesn't change the fact this is adding tedium for tedium's sake that people are just going to try and powergame away by speedrunning the LTSRBT - and the fact that per ongoing complaints upstream, it's not even _well-implemented_ tedium thanks to UI bugs and existing problems with TG Cargo sometimes eating private orders.

Zonespace27 (approved):
Modular stuff looks good, this is gonna cause a ton of conflicts if not merged soon (now)

PR #17881, opened 2022-12-02
[MIRROR] Atmospheric machineries now interact with each other directly, rather than going through a radio layer -- saves about 0.4s of init time [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71232
--------------------
More than a decade ago, these were all coded to use a weird dynamic list radio broadcasting system to communicate with each other. If there was any depth that they were planning on creating with this, it didn't come to fruition, and it instead just wasted a lot of init time. 

Removing `post_signal` saves 198.41ms, fired 588 times from lots of different machinery. Its self cost was 81.44ms.

`broadcast_status`, also removed, was taking 218ms.

I'm pretty sure I'm done with this, but it's hard to tell given the nature of old radio signal code.

A small self cost of 34.9ms was added in the form of /obj/Initialize checking id_tag to set in a global list. This could be optimized away by tagging everything that does use id_tag, but it's a loooot and I think this is just a useful mechanism to have. Not worth it IMO.

The "Check Atmos Chamber Devices" verb has been removed. Everything it did *should* be replicated by runtimes on Initialize, which is both more obvious to mappers and shows up in unit tests since we spawn every ruin.


Tastyfish:
Note that I changed the tarkon atmos machine defs to have unique chamber ID's since they relied on filtering by radio frequency (in a strange way). 

Regardless, I or someone else should probably double check that Tarkon atmos still actually works.

Tastyfish (approved):
Making Tarkon work was painful but appears to work as expected.

PR #17885, opened 2022-12-02
Deduplicates the normal meteor wave event

### LINKED ISSUE #17884: 3 Instances of Normal Meteor Wave in admin event panel
## Round ID:
Private Test Server
Checked and non-existent on TG

## Testmerges:
N/A

## Reproduction:
Be Admin
Open Event window
Type Meteor

Peep the horror

image

GuiltyNeko:
## About The Pull Request
In our modular edits, we had a "/datum/round_event_control/meteor_wave/major_dust", but #16211 moved that to be a subtype of space_dust, which we didn't mirror in our modular edits. 
This resulted in the creation of a second identical event with a different chaos level set but otherwise identical.
Fixes #17884, though I only had two instances, and an admin I asked in-round in dchat also reported two instances of the event.

Obviously this also makes the assumption that the major dust event should have the same chaos level as before, but it's 5 AM and I don't even know what event chaos does, somebody who does can decide if that should be kept or not.

Twice as many of an event as expected is bad. Less bad equals good.

## Proof of Testing

<details>
  
image
Behold, the one-line change actually works as advertised. 

</details>

## Changelog

:cl:
fix: The normal meteor wave event has been de-duplicated
/:cl:

### DIFF
--- a/modular_skyrat/modules/event_vote/code/event_chaos_system.dm
+++ b/modular_skyrat/modules/event_vote/code/event_chaos_system.dm
@@ -110,7 +110,7 @@
 /datum/round_event_control/ion_storm
 	chaos_level = EVENT_CHAOS_LOW
 
-/datum/round_event_control/meteor_wave/major_dust
+/datum/round_event_control/space_dust/major_dust
 	chaos_level = EVENT_CHAOS_LOW
 
 /datum/round_event_control/market_crash


hugmeorelse (approved):

hugmeorelse:
@GoldenAlpharex We should TM ASAP.

tf-4 (approved):

lessthnthree:
Thank you, I was confused why we had three instances of Meteor Wave Normal in our admin panel.

PR #17960, opened 2022-12-06
Void Phase won't hurt the caster if cast by a non-heretic (#70310)

tf-4:
https://github.com/tgstation/tgstation/pull/70310

🆑
fix: If an admin gives you Void Phase, you won't be hurt by your own spell whenever you cast it.
/🆑


Zonespace27:
(ignore my fuckery with the commits) As far as I can tell we already got this PR when we merged something else first

PR #17600, opened 2022-11-20
Adds a toggle for BSRPD remote piping

strandsofivy:
## About The Pull Request

Adds an Alt-Click toggle for the bluespace rapid pipe dispenser's remote piping feature. Simple as.

## How This Contributes To The Skyrat Roleplay Experience

The BSRPD can be annoying to deal with when you want to click on the screen while holding it for any reason. This helps a lot with quality of life surrounding this item, at least from my perspective.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>

https://user-images.githubusercontent.com/47710522/202921564-f715b92e-d902-46ff-a5a3-0bd5aae209ee.mp4

</details>

## Changelog

:cl:
qol: The Chief Engineer's BSRPD now has a toggle feature for remote piping.
/:cl:

PR #17980, opened 2022-12-07
Fix: BSRPD Alt-Click / Remote Piping Safety

Floofies:
## About The Pull Request

PR #17600 added an Alt-Click to the Bluespace RPD which should toggle "remote piping" functionality, however it does not work as expected. The bug was caused initially by a minor developer oversight that then caused some of the code to become unreachable.

The bug is caused by the conditional logic on line 53 in `/obj/item/pipe_dispenser/bluespace/AltClick()` which checks the return value of `/obj/AltClick` and its other chained parent procs. The conditional expects a "truthy" value to be returned which is an erroneous assumption, because AltClick is *not* implemented to ever return a truthy value, it will instead return `FALSE` upon failure *or* `null` upon success.

This PR includes a 1-line fix to the aforementioned conditional which changes it to strict-equals `FALSE` rather than just check for falsy value.

This PR fixes the newly-added Bluespace RPD's Alt-Click functionality to toggle the "remote piping" feature, which isn't working as expected currently.

## Proof of Testing

<details>

- [Link to video I posted in #development-code](https://cdn.discordapp.com/attachments/640760727489085450/1050104203630092328/dreamseeker_OHmWIT2eKH.mp4)
- [Link to message in #development-code because it might force you to download the above video](https://discord.com/channels/596783386295795713/640760727489085450/1050104203864985682)
</details>

## Changelog

:cl: A.C.M.O.
fix: Fixed the Bluespace RPD's Alt-click for its "remote piping safety".
/:cl:

### DIFF
--- a/modular_skyrat/modules/bsrpd/code/bsrpd.dm
+++ b/modular_skyrat/modules/bsrpd/code/bsrpd.dm
@@ -50,7 +50,7 @@
 
 /obj/item/pipe_dispenser/bluespace/AltClick(mob/user)
 	. = ..()
-	if(!.)
+	if(. == FALSE)
 		return // too far away
 	remote_piping_toggle = !remote_piping_toggle
 	balloon_alert(user, "remote piping [remote_piping_toggle ? "on" : "off"]")


tf-4 (approved):

strandsofivy:
thanks for the help, click code is really weird ._.

PR #17981, opened 2022-12-07
[MIRROR] Delimbs Stamina [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71663
--------------------

## About The Pull Request
This changes stamina damage from a damage that affects limbs (like brute and burn) to a damage that affects the body at whole (such as tox or oxy)
Stamina being run like brute or burn is rooted all the way back to when it was halloss, and used exclusively for holodeck items and hallucinations.
Most coders probably didn't even know that stamina worked like this, and most implimentatiosn of stmaina mechanics use a generic "adjust stamina" or a "apply_damage(type = stamina zone - body_zone_chest)" to avoid the issues that come when stamina is evenly distributed among limbs (and some limbs only count as a fraction to stamina damage total)

The only thing that is truly lost here, is disabling limbs (arms and legs) with stamina damage, by aiming specifically at them with disablers (batons always apply their stamina damage to the chest, no matter where they hit) our through smashing limbs on tables.
The idea of making an opponent drop a weapon out of pain is fun, but stamina doesn't support it, or make sense in context.

A depth filled limb-based pain system that has interactions with painkillers and replaces damage based move slowdown would be fun, and maybe something that could be implemented in the future.
## Why It's Good For The Game
The biggest draw here, is that it will make stamina easier to maintain 
## Changelog
:cl: itseasytosee
code: Stamina no longer affects individual limbs. 
/:cl:


Tastyfish (approved):
I think I've fixed all of the SR code, but there are so many API changes that I think it warrants a second pair of eyes.

PR #17987, opened 2022-12-08
Space heater grenades

Stonetear:
## About The Pull Request

spaceheater_testing_pic

This creates heater grenades that explode into tiny space heaters!  The space heaters use different sprites and have less health, marginalizing their combat potential.

When deployed, they will explode into a space heater that starts anchored and powered with a normal cell, set to room temperature.  They have 50 health and take about 4 hits with a medical saw to destroy.  When deconstructed or destroyed, they don't drop a machine frame or metal, only a sheet of cardboard plus the usual space heater parts.  (Machine board, T1 laser, coil, ect...)  Trying to detonate two in the same tile will cause it to explode into useless debris. 

Atmos can be quite annoying to respond to when you're the only engineer working on it, especially when there's a bunch of cold air causing firelocks to shut, which ends up trapping the cold air in place.  Rather than having to carry the parts for a space heater around in your backpack all the time, now you can merely carry one of these grenades in your bag and always be ready to respond to cold air problems.

This is not intended to replace the act of _finding_ heaters for normal crew, only constructing them in engineering, hence why it's in the engi-vend machine.  The intended use case is when you are a beginner engineer spending your shift trying to fix arrivals by yourself and crying because you have to make 3 trips to engineering because you keep forgetting parts and everyone else is working on their pet projects and not helping you.  Now all you need is to buy a grenade or two from the vendor (They are not printable.) and fix it in one trip.
## Proof of Testing
<details>
<summary>  I tested the following:
</summary>

- Deconstruction and destruction plus hits to destory
- Heating and cooling functionality
- Spawning and throwing
- The secret space heater clusterbang, shh!
</details>

## Changelog

:cl:
add: Adds Space heater grenades.  You can find them in the engi-vend or cargo emergency category.
/:cl:


Paxilmaniac (commented):
I don't think 'grenade that fixes your problems the same way doing it normally would' is a good solution for anything, much less a solution for atmos heating.

The idea of a grenade that turns into a space heater also seems pretty silly from an IC standpoint but that's not as much of a concern.

Finally, if you don't want them to have combat potential either make them not block movement or give them like 30 integrity, because 75 is still pretty high to manage for anything that's not a gun coded by gandalf.

Stonetear:
skyratdev_channel

> I don't think 'grenade that fixes your problems the same way doing it normally would' is a good solution for anything, much less a solution for atmos heating.

They're only intended to be an alternative to having to manually construct heaters by hand at your destination.  I mean, it's everyone's complaint that cargo never has anything worth buying, it's an alternative option.

Making them not block movement seems like an interesting option.  I'll try it.

Edit: In my testing, I tried making them not block movement, but it felt too weird.

Stonetear:
Alright, I ended up lowering the health to 50.  Still makes it interesting for ghetto combat, but not useful against security or antagonists.

Stonetear:
@Paxilmaniac  I assume you're my reviewer?  Would love some feedback or a mark as complete.

Zonespace27:
Eh. Not a fan of the concept (this looks like a solution without a problem to me, honestly), if you need a room heated space heaters are not hard to find. If you want this re-opened, find a maintainer to want the idea in the game

PR #17998, opened 2022-12-08
[NON-MODULAR] Increases the Sprite Overlay Cap

RimiNosha:
## About The Pull Request

Through a combination of shitcode that I tried to fix, and lost my sanity trying to fix, along with just the sheer amount of markings people like to put on their rainbow dogs, we're now at the point where this is now actually an issue (and has been for a while, now)!

Less runtime, less lag, yes.

Can't say the same about this new cap not negatively impacting performance indirectly though.

## Proof of Testing

This is a number define edit that caps how many sprites are rendered, there's no runtimes in my logs using a save I know is proven to be bad, so that's good enough, yes?

## Changelog

:cl:
qol: Sprite overlay cap has been raised to 150 from 100. Hopefully your overly intricate rainbow dogs no longer break or look wrong when they have 50+ markings and other such things.
/:cl:

### DIFF
--- a/code/_compile_options.dm
+++ b/code/_compile_options.dm
@@ -120,7 +120,7 @@
 
 // A reasonable number of maximum overlays an object needs
 // If you think you need more, rethink it
-#define MAX_ATOM_OVERLAYS 100
+#define MAX_ATOM_OVERLAYS 150 // SKYRAT EDIT - Hello, we're Skyrat, and we like sprite overlays very very much, and 100 just isn't cutting it anymore - ORIGINAL: #define MAX_ATOM_OVERLAYS 100
 
 #if !defined(CBT) && !defined(SPACEMAN_DMM)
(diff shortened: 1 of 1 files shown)


tf-4:
hnnnnnnnnnnnnnnnnnnnnnnnnnnngggggggggggg I don't like this at all
this will need some discussion

RimiNosha:
Then fix the code that I tried to unfuck, or be stuck with multiple MB runtime logs. Testmerge this meanwhile if you have to.

desertmote:
what specifically makes you hit this cap? i hit it too when making my modlight PR, but it turned out i wad just coding something wrong. i also only needed to put it at 115 to work shoddy code besides

Tastyfish:
This is an engine-imposed cap to my knowledge. Exceeding it will corrupt the atom's appearance.

tf-4:
See above

PR #18015, opened 2022-12-09
Fix: Self-Actualization Device Doubling Damage

Floofies:
## About The Pull Request

This PR fixes a bug in the Self-Actualization Device which causes it to double the occupant's existing damage. The bug is caused by the usage of `adjustBruteLoss` and `adjustFireLoss` which are intended to "re-apply damage". The procs work as expected on humans but not on synthetic species, and I don't know why that is.

The quick solution was to switch the procs to `setBruteLoss` and `setFireLoss`. I also switched `adjustOrganLoss` to `setOrganLoss` to avoid the same effect taking place there.

This PR will fix the bug wherein the Self-Actualization Device doubles a synthetic character's brute/burn/organ damage, often brutally killing them. The fix will allow the device to work as expected when a synthetic character uses it.

## Proof of Testing

<details>

  [Video I posted in the #development-contributor channel](https://cdn.discordapp.com/attachments/610319137012580383/1050659084795387924/dreamseeker_RKivcZwqvP.mp4)
  [Link to the message in case the above link forces you to download it](https://discord.com/channels/596783386295795713/610319137012580383/1050659085994971136)
</details>

## Changelog

:cl: A.C.M.O.
fix: Fixed the Self-Actualization Device, which was sometimes doubling damage and killing players.
/:cl:

### DIFF
--- a/modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
+++ b/modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
@@ -182,8 +182,8 @@
 		patient_brain.traumas = trauma_list
 
 	//Re-Applies Damage
-	patient.setBruteLoss(brute_damage, required_status = FALSE)
-	patient.setFireLoss(burn_damage, required_status = FALSE)
+	patient.setBruteLoss(brute_damage)
+	patient.setFireLoss(burn_damage)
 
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (approved):
Looks good, thank you!

PR #18035, opened 2022-12-09
fix revitalizing cores for ashies

jjpark-kb:
## About The Pull Request
re-adds (read: fixes) the ability for ashwalkers to revitalize cores that have gone inert. The intent when I made the original PR was so that ashwalkers have the ability to stabilize cores without needing a serum from a miner; additionally, being able to revitalize cores allowed ashwalkers to trade them, or even use them.

Ashwalkers are already bound greatly by policy and even certain gameplay mechanics. Giving them some freedoms and less worries is a good thing for them to focus on other actual meaningful things.

## Proof of Testing

<details>
  
image
image

</details>

## Changelog

:cl:
fix: the necropolis tendril can revitalize cores again (ashwalkers only)
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/buildings/ash_tendril.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/ash_tendril.dm
@@ -10,7 +10,10 @@
 		balloon_alert(user, "must be an ashwalker!")
 		return
 
-	var/obj/item/organ/internal/monster_core/regenerative_core/regen_core = attacking_item
+	qdel(attacking_item)
+
+	var/obj/item/organ/internal/monster_core/regenerative_core/regen_core = new(get_turf(src))
+	user.put_in_active_hand(regen_core)
 
 	if(!regen_core.preserve())
 		balloon_alert(user, "organ decayed!")


Nerev4r:
This isn't really a fix, it was a balance change that legion cores can't be stabilized anymore.

jjpark-kb:
> This isn't really a fix, it was a balance change that legion cores can't be stabilized anymore.

oh but it is a fix because the other PR mislabeled it as a fix.

MitztheKat:
to Nerev4r:
Ashies need economy, ashies get economy from regen cores!
ashies need break from nerfs.
ashies tiiired of being steamrolled.
ashies tiired of existence...
ashies don't wake up now. :(
maybe PR fix ashie problem?
(translated: fuck off minermain, give us SOME bonus for our problems YOU have caused!)

Stalkeros2:
I do like this, adds some incentive for the miners to cooperate with the ashies.

tf-4:
I'm not really sure why they would need the ability to do this, and I don't think they do. They can just use regular, non-decayed cores like everyone else. They already last a while as it is, it's not like they'll need to rush back constantly.

PR #18049, opened 2022-12-10
[fix] fixes det revolver icons

RatFromTheJungle:
## About The Pull Request

I Love Changes To Icon States Pushed Through In A Different PR That I have To Find In Some Vague Modular File That Wasn't Broke Before My Other PR And Only Broke After But I can't Tell Where Because it's Not In Changes For The File

make gun the see

## Proof of Testing

this  
image
to this
image

</details>

## Changelog

:cl:

fix: fixes the revolvers icon state that apparently just changed

/:cl:


GoldenAlpharex (approved):
Well, looks like it'll work. You tested it, so here goes nothing I guess.

Issue #17939, opened 2022-12-06
IPC's / Synths cant be force healed by any means


Zergspower:
Reporting client version: 514.1589

## Round ID:
7150

Pretty sure this PR broke it as for some reason we didnt put back in the synthetic code in living.dm

![image](https://user-images.githubusercontent.com/22140677/205851300-a45b9b70-2cb4-482b-bb7e-e6841a81cb6f.png)
https://github.com/Skyrat-SS13/Skyrat-tg/pull/17535

## Testmerges:
- [MOD Hardlight (Refactors how sprite accessories handle MODs)](https://github.com/Skyrat-SS13/Skyrat-tg/pull/17372)
- [Moves modular uplink items into a module rather than master files + Removes some of the goofy ahh items we had available to traitors](https://github.com/Skyrat-SS13/Skyrat-tg/pull/17813)
- [Fairly large assault operatives changes](https://github.com/Skyrat-SS13/Skyrat-tg/pull/17455)

## Reproduction:

![image](https://user-images.githubusercontent.com/22140677/205850942-9d6a39f6-941d-40d3-ac97-7cf5825ecf63.png)

Zergspower:
This also Extends to the SAD, currently reviving synths and IPC's in a non original body isnt possible

Floofies:
Caused by tgstation/tgstation#71123

Floofies:
I have now fixed this in #18091

Floofies:
By fixing this I exposed more severe bugs such as reagents healing robotic bodyparts.
I've now fixed the additional bugs in #18132 and it should be merged / compiled with #18091

PR #18091, opened 2022-12-11
[MIRROR] Fix: Robotic/Non-Organic Bodyparts not Healing/Damaging [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71864
--------------------
# About The Pull Request
Around 4 years ago an overriding proc in carbons, `take_overall_damage`, dropped a positional argument for stamina, but the calls to it were not adjusted. This PR fixes a 4 year old bug wherein the `adjustBruteLoss` and `adjustFireLoss` functions were not working as expected for robotic bodyparts, and in most cases leads to robotic limbs being invulnerable to spread damage/healing. The bug also affected the Var Viewer and any admin-heals. In downstream repositories such as Skyrat-tg, where there are fully robotic characters, some severe impact was seen such as the inability to revive/heal players without fully respawning them.

The bug was caused by out-of-order positional arguments in the calls to `/mob/living/carbon/take_overall_damage` and `/obj/item/bodypart/proc/receive_damage` within `adjustBruteLoss` and `adjustFireLoss`, which caused its `updating_health` argument to wind up in the wrong position, occupying the same position as `required_status` which is then interpreted by the proc to mean bitflag `BODYTYPE_ORGANIC`. Furthermore, the default argument for `required_status` and `heal_overall_damage` is currently also `BODYTYPE_ORGANIC`, which would cause the issue to persist and prevent most forms of spread-healing and spread-damage to robotic/surplus limbs.

I also found that, after fixing the bug, robotic limbs were suddenly vulnerable to space/pressure damage. @ Fikou has requested that I re-add this mechanic, as it is now a feature of the game. To implement it, I made a small change to `/datum/species/proc/handle_environment_pressure` so pressure damage only damages organic bodyparts.

I also found that any limbs which do not set a `max_damage` simply do not show up in the list returned by `get_damageable_bodyparts`, which could stop them from being damaged/healed, which I think may be intentional. I didn't think this was a bug, but I can change this too if it's asked of me.

## Why It's Good For The Game
The PR includes a bugfix which allows non-organic bodyparts to be damaged by spread damage again. An out-of-order argument to the ubiquitous `take_overall_damage` and `receive_damage` functions was causing an argument to shift one position over to the right, and those functions receive a default `required_status` argument of `BODYTYPE_ORGANIC`; both issues combined caused the functions to almost completely ignore non-organic bodyparts, and only deal/heal damage to organic bodyparts.

The PR should allow more admin-heals to work as expected, and more sources of damage will now work as expected for non-organic bodyparts.

## Changelog

:cl: A.C.M.O.
fix: Fixed robotic/surplus limbs not taking damage/healing as expected; including via the Var Viewer, Player Panel, and from sources of spread-damage.
/:cl:


Floofies:
Please ensure that this is merged along with its follow-up PR #18132

Floofies:
Fixes #17939

Funce (approved):

PR #18138, opened 2022-12-14
Adds antagonist datums for DS2 and Interdyne

YakumoChen:
## About The Pull Request

Adds antagonist datums for the DS2 crew and all Interdyne roles, respectively. 
The main purpose of this is to make both of these roles visible in the observe panel for ghosts, so that observers can see when other DS-2/Interdyne players are in the round.

Makes DS2 and Interdyne more friendly to observers by not mixing them in with the general crew pop on observe.

## Proof of Testing
image
image

## Changelog

:cl:
qol: You can now see who is now staffing DS2 and Interdyne more easily when observing as a ghost.
/:cl:

### DIFF
--- a/code/modules/mob_spawn/ghost_roles/mining_roles.dm
+++ b/code/modules/mob_spawn/ghost_roles/mining_roles.dm
@@ -291,6 +291,7 @@
 /obj/effect/mob_spawn/ghost_role/human/lavaland_syndicate/special(mob/living/new_spawn)
 	. = ..()
 	new_spawn.grant_language(/datum/language/codespeak, TRUE, TRUE, LANGUAGE_MIND)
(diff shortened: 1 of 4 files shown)


GoldenAlpharex:
I feel like the proper solution should be to add a "Ghost Roles" tab to the orbit menu, below the crew, rather than taking up even more space above the crew list. Most people won't actually care to see the interdyne/DS2 folks in there.

tf-4:
agreed

YakumoChen:
That's beyond me to know how to do, however. I had to hack this together with ashwalker code to figure out how it worked. I could say the same for ashwalkers, but they're treated as 'antagonists' in the code like this feature does.

RimiNosha:
TG ashwalkers are antagonists, so that's why they have an antagonist datum.

Zonespace27:
I think i can safely say that we don't really want the non-antag ghost role to be antags code-wise

PR #18161, opened 2022-12-15
Makes department guard gear consistent

### LINKED ISSUE #18117: Departmental Guards Only Get One Gun / Cable Restraints Despite There Being Two.
## Round ID: 
N/A

## Reproduction:
Departmental guards only recieve one locker for two guards to share. This locker is only equipped with one peppergun and one cable tie, meaning only one of the two guards will be able to properly gear up for the job - one will get the gun and cable tie, and the other must either lack the full gear a department guard should have, or must do several bounties or waste cargo budget on ordering equipment. This second guard is generally a latejoin.

Jobs should not be unable to get the equipment they are meant to have, and this should not be designed into the game. If the gun and cable ties are not needed for the job, then they should not get any, and neither guard has a gun or cable ties. If it is meant for the job, then both should get it, and the locker should include two of each just as it includes two berets, two jackets, two uniforms, and two of everything else.

EDIT: There is also only one whistle. The above stated applies to the whistle as well.

YakumoChen:
## About The Pull Request

Fixes #18117 
Department guards now spawn with all of their gear:
- Guards get handcuffs coloured to their department in their pocket.
- Flash was moved from their bag to their other pocket
- Guards now spawn with a pepperball gun box
- Lockers now contain an extra flash, order spares from cargo.

Other minor changes:
- Service guard now has appropriately coloured cable cuffs instead of full handcuffs
- Engineering locker changed to require only ACCESS_ENGINEERING instead of ACCESS_ENGINE_EQUIP (meaning basic engineering access) for consistency with the other lockers.

Locker contents and what guards spawned with was inconsistent and goofy, and there wasn't enough gear for everyone, for some reason. Instead of giving a chance for one greedy guard to take two sets of stuff in the locker, now he has everything he needs at the start of the round.

Lockers holding a spare set of gear is consistent with pretty much every other job locker we have in the game.

## Proof of Testing

image
(not pictured: they have cuffs in their other pocket now)

## Changelog

:cl:
qol: Department guards now spawn with all the gear they need to police their department
qol: Department guard lockers contain a full, spare set of gear, for people getting a job change.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/security.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/security.dm
@@ -132,7 +132,6 @@
 /obj/structure/closet/secure_closet/security/cargo/PopulateContents()
 	new /obj/item/storage/bag/garment/customs_agent(src)
 	new /obj/item/restraints/handcuffs/cable/orange(src)
-	new /obj/item/melee/baton/security/loaded/departmental/cargo(src)
 	new /obj/item/assembly/flash/handheld(src)
 
 /obj/structure/closet/secure_closet/security/engine
@@ -144,7 +143,6 @@
 /obj/structure/closet/secure_closet/security/engine/PopulateContents()
 	new /obj/item/restraints/handcuffs/cable/yellow(src)
 	new /obj/item/storage/bag/garment/engineering_guard(src)
-	new /obj/item/melee/baton/security/loaded/departmental/engineering(src)
 	new /obj/item/assembly/flash/handheld(src)
 
 /obj/structure/closet/secure_closet/security/science
@@ -156,7 +154,6 @@
 /obj/structure/closet/secure_closet/security/science/PopulateContents()
 	new /obj/item/restraints/handcuffs/cable/pink(src)
 	new /obj/item/storage/bag/garment/science_guard(src)
-	new /obj/item/melee/baton/security/loaded/departmental/science(src)
 	new /obj/item/assembly/flash/handheld(src)
 
 /obj/structure/closet/secure_closet/security/med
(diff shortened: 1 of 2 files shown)


tf-4 (changes requested):
People getting a job change can source their own gear, they don't need three guns total between two people.

Iamgoofball:
@tf-4 There should be enough gear roundstart for the default slots, so 2 sets. Remember, job changes also include "there was nobody in this position", not always "we're hiring an extra because 2 isn't enough", which is where a cargo crate or something should come into play.

YakumoChen:
@tf-4 done

tf-4:
The spare batongs as well sorry not just the guns. The spare flashes are meh, remove them or don't idm

Iamgoofball:
@tf-4 You know, the proper solution to this is we just put the job gear in the lockers like normal, have two lockers, and call it a day, but you guys denied the correct fix to this problem so you're stuck in a situation where
1. Job transfers into the position are up shit creek if there's no existing guards
2. Anything that interferes with equipment spawning on the job themselves will cause job equipment to get lost to the void

Maybe next time, consider merging the correct actual fix for a problem so that we don't end up in situations like this.

YakumoChen:
@tf-4 problem is the batons are a unique item you can't get from cargo, not real hot on that and having two batons doesn't make much more difference than having one tbqh

I like the idea of someone being able to steal the locker for a shittier stunbaton anyway.

BardTheBard:
As Goofball said, two lockers would be the ideal solution, but that did get denied. Likewise, on our own map, Voidraptor, https://github.com/Skyrat-SS13/Skyrat-tg/pull/17761 even removed the second locker, causing this problem even on maps we have control of.

tf-4:
add them to cargo then

YakumoChen:
Finally fixed it

tf-4 (approved):

PR #18185, opened 2022-12-17
[MIRROR] Fixes podperson hair colors, adds individual layer colors [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71683
--------------------
This was broken completely by a certain scaly coder

Now it works again! Previously the hair was the inverted color of the body, but it should've been the same and only have the front layer be inverted

Also makes it possible to render layers with different colors

image

Removed changelog


Tastyfish (approved):
Pod person hair doesn't work with our mutant parts system at all and I don't really care enough to entirely refactor it since you can use the full hair system instead already.

PR #18186, opened 2022-12-17
Monkey cube boxes are no longer invisible

lessthnthree:
## About The Pull Request
NFI why, but the monkey cube box icon is missing from our storage.dmi.

You can see monkey cube boxes.

## Changelog

:cl: LT3
fix: Monkey cube boxes are no longer invisible.
/:cl:


GoldenAlpharex (changes requested):
I'll be honest, that new model really doesn't look up to par compared to other box sprites.

Also, I'd prefer that we followed upstream's new file structure for boxes and the like, as this is far from being the only box that was affected (see: donk pockets)

PR #18225, opened 2022-12-19
[Modular] Buffs the 'no guns' quirk point-count

desertmote:
## About The Pull Request
No guns granted -4, now it grants -6.

I noticed that both stormtrooper aim and no guns both grant an equal amount of points. Obviously this makes no sense, as one of the two still lets you use guns; and you would be surprised how well you can become accustomed to its downsides.

Nerfing stormtrooper aim instead of buffing no guns is a possibility too, however the way this PR does it seems more logical to me, as both quirks still have a hefty gameplay penalty.

## Changelog
:cl:
balance: The 'no guns' quirk now grant 6 points to use instead of 4
/:cl:


tf-4 (approved):

tf-4:
Makes sense

PR #18242, opened 2022-12-20
[MIRROR] Fixes vendor showing weird glass icons [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71929
--------------------

## About The Pull Request
Fixes vendor showing weird glass icons
## Why It's Good For The Game
bug fix good
## Changelog
:cl:
fix: vendor no longer shows glass icons in a specific condition
/:cl:


Funce:
I couldn't use bitflag adding, as that refuses to overwrite

Funce:
wait in that case would a...
```
skyrat_products |= products
products = skyrat_products
```
work easier? Interesting thought

EvilDragonfiend:
I believe you don't have to change skyrat product variables since `target_items |= each` will take items exclusively and it doesn't try spritesheet duplication.

Funce:
Issue is, specifically the engi goggles existing in normal products, and also in Skyrat products.

Adding the entire list together introduces the dupe in products,

And `|=` on a list does seem to compare from the initial state of nothing. Therefore the duplicate is added to the spritesheet.

EvilDragonfiend:
not really? the same items are added again to the list of course, but `target_items |= each` accepts the path when it doesn't have it  inside of its list.
the issue you said would happen if the code was written as `target_items += each`

Funce:
That's what I thought too! Then I noticed that `each` is the list, and it was doing something _really wonky_.

I quickly tested this on the coderbus DM bot.

```
var/products = list(
  /obj/thing = 1,
  /obj/thing = 5
)

var/target = list()

target |= products
```
`target` is now
```
list(
  /obj/thing = 5,
  /obj/thing = 5
)
```

Which means haha BYOND strikes again.

EvilDragonfiend:
????? I think the issue you're addressing is not related with the sprite dupe. It will never dup sprite.
Also the code you used is just a kind of copy paste into a list as the whole, and the code I fixed uses `for(each in produc)` which means having dupe items in the list doesn't really matter when building up sprites.

Funce:
That would be true if was `each in products` but it's `each in list(products, premium, contraband)` a list of lists.

The reason I tell you with certainty, is because it _was_ duping the sprite insert, even in your last CI run of your PR, and additionally in this one before my last commit.

EvilDragonfiend:
For example
```
Var/L = list()
L += "a"
L += "a"
Var/another = list()
another |= L
world.log << length(another)
```
Will be 2 of course

But
```
Var/L = list()
L += "a"
L += "a"
Var/another = list()
for(var/each in L)
 another |= each
world.log << length(another)
```
This will be 1

EvilDragonfiend:
> The reason I tell you with certainty, is because it was duping the sprite insert, even in your last CI run of your PR, and additionally in this one before my last commit.

Ah then did I miss the CI failure? I couldn't notice the line

Funce (approved):

code/modules/asset_cache/assets/vending.dm
```diff
@@ -2,7 +2,16 @@
 	name = "vending"
 
 /datum/asset/spritesheet/vending/create_spritesheets()
-	for (var/k in GLOB.vending_products)
+	// initialising the list of items we need
+	var/target_items = list()
+	for(var/obj/machinery/vending/vendor as anything in typesof(/obj/machinery/vending))
+		vendor = new vendor() // It seems `initial(list var)` has nothing. need to make a type.
+		for(var/each in list(vendor.products, vendor.premium, vendor.contraband))
```
Funce, line 9: It's operating on the first example. These are the lists I'm referring to.

Funce (commented):
And it was an asset subsystem runtime, popped up on the maps. But was unique to how Skyrat fucks with products.

Tastyfish (approved):

EvilDragonfiend:
Aight, i got the reason. It is not because of the multiple same items. The reason was because the items that shares the same icons, so actually my fix wasn't that perfect.
I will make another fix from the upstream.

EvilDragonfiend:
> Aight, i got the reason. It is not because of the multiple same items. The reason was because the items that shares the same icons, so actually my fix wasn't that perfect. I will make another fix from the upstream.

Actually nevermind. I was sleepy and I was confused how icons in asset are made. I think it seems right way to fix. Sorry bothering you

Funce:
No, it was the same item. Icon key was based on type so it needed to be.

Engineering scanner goggles specifically on the engivend! It was meant to overwrite with a new amount of goggles (2 to 5) but instead just added together.

I think it works well as an impromptu unit test, can't have duplicate items in the vendor.

Funce:
> Actually nevermind. I was sleepy and I was confused how icons in asset are made. I think it seems right way to fix. Sorry bothering you

Not at all, pleased to discuss it!

PR #18271, opened 2022-12-22
[MIRROR] Gas canisters correctly update their appearance [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72124
--------------------
## About The Pull Request
Fixes issue #71712 & #72140 Don't know how the author wasn't able to reproduce it even though it happens all the time on all maps

https://user-images.githubusercontent.com/110812394/208730887-7ad4b02d-9dcc-4d71-b54f-489c3bd616f6.mp4

## The Problem
The current code only updates the appearance of the canister when its valve is opened & if it takes damage . It doesn't deal with the case when its wrenched normally to a port without any problems

## The Solution
Update the appearance during processing but only when its pressure changes significantly for the indicator light to be notified about it. This way we don't have to care if
- the valve is open
- its wrenched to a port
- its leaking & taking damage

Regardless of what the cause is only if the pressure inside the canister changes significantly from its current value do we update the appearance. 

This is so much more optimal than the previous code as` update_appearance()` will be called a maximum of 8 times during its full cycle of losing & refilling gas and won't be called at all if there is no pressure change

## Changelog
:cl:
fix: correctly update canister appearance
/:cl:


Tastyfish:
Depends on #18243 which is a doozy

Tastyfish (approved):

PR #18280, opened 2022-12-22
[MIRROR] Revamps tram windows and layout [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72047
--------------------
## About The Pull Request
-Tram windows are now flush with the tram walls instead of sticking out.
-Shuttle seats are replaced with benches.
-Moves the icon state of the tram controls from the map file to the tram controls subtype, to avoid mapper pain if they make changes in the future.
-Makes tram tiles a subtype instead of editing their vars on the map.

image

## Why It's Good For The Game
Tram looks better, don't want mappers to accidentally break tram because they didn't see the icon override.
## Changelog
:cl: LT3
fix: Tram windows are now properly flush with the tram walls.
imageadd: Tram shuttle seats have been replaced with benches so people can sit too close to each other for that real metro experience.
code: The icon for tram controls is now set in the tram controls subtype instead of the map itself.
code: The tram floor tiles are now in a tram floor subtype instead of the map itself.
/:cl:


Paxilmaniac (approved):
ok

PR #18356, opened 2022-12-27
[MIRROR] Makes `flags_inv` better at actually hiding things when used on non-suit clothing [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72107
--------------------
## About The Pull Request

`flags_inv` contains a series of flags which dictate which items can be seen / which items are rendered when it is worn. This lets you hide things that shouldn't be visible, stops weird bugs

Right now it really doesn't work to hot. Most of the flags aren't checked so a lot of silly things happen.

Here's an example, which was broken for 2 reasons

image

1. it was using the wrong var
2. nowhere checked that the shoes were hidden

So, I went through and made the `update_x` procs consistently check when the slot is actually hidden.

image

Much better?

~~It's actually still broken. It's hiding the arms for some reason. But that's for another PR?~~

## Why It's Good For The Game

Flags that exist should work as expected.

## Changelog

:cl: Melbert
fix: Wedding dresses no longer render their shoes over the dress
fix: Things which have inv flags set should more consistently actually hide the things they mean to hide
/:cl:


Tastyfish (approved):
Everything seems to work. Going to get @GoldenAlpharex's approval before merge.

GoldenAlpharex (approved):
Looks good to me!

PR #18361, opened 2022-12-27
[MIRROR] Experiment with holding hard references to objects being qdeleted in 515 (saves 1.1+ seconds on init times, more on prod) [MDB IGNORE]

### LINKED ISSUE #10: Migrate SSgarbage's queue from using \ref to using direct references
## What sucks

\ref is real slow, at least in the scale of qdel. This is primarily because adding a getting ref requires inserting a new string into the string tree, and we bloat that sucker to hell.
We use refs for SSgarbage's queue, because it lets us assert that byond's garbage collector will do its work if there are no other refs remaining 

## How fix

515 adds a var that gives us the amount of references an object has remaining. We could leverage this alongside some information we know about how many refs an object "should" have by that point to save the cost of all those \refs

That works out to about 1500ms

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72033
--------------------

## About The Pull Request
Adds `EXPERIMENT_515_QDEL_HARD_REFERENCE`, which will queue to the GC subsystem using hard references rather than `\ref`.  This is only possible in 515 because of the new `refcount` proc. `\ref` is very very slow and has some nasty knock on effects, so removing its usages where possible is good.

This is an explicit opt in define because I want to give us the ability to test 515 on live while only testing 515 itself, not our experimental changes. We have a few more of these we want to do so I made a separate file for them. They're auto-defined in unit tests so we see them with the alternate test runner. In a perfect world we'd test both on and off, but eh.

Closes https://github.com/tgstation/dev-cycles-initiative/issues/10


Tastyfish:
It looks like there's a few upstream atoms that are leaking references. Leaving as draft until fixed.

Tastyfish:
I pre-merged the the upstream fix, https://github.com/tgstation/tgstation/pull/72669 but it should be coming down soon.

Tastyfish:
Oh and for obvious reasons, don't merge this unless Alternative Tests passes.

Tastyfish (approved):

PR #18398, opened 2022-12-29
[MIRROR] Fixes emissives, starlight, and a filterrific bug [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72326
--------------------
## About The Pull Request
[Fixes emissive lighting, we never actually used the thing's layer, should have just been using float layers anyhow](https://github.com/tgstation/tgstation/commit/2af035d48742eb4337dbfabed699a9acef67d883)

[Fixes starlight not disabling fullbright](https://github.com/tgstation/tgstation/commit/7a4e3f0c9abacaf944a4be801d2d99daa72e9c50)

It used to, but tivi messed it up with area lighting. makes maps like kilo look WAY worse then they should (I stole this slightly from kapu)

[Fixes filterrific erroring when editing something with a color matrix](https://github.com/tgstation/tgstation/commit/72dfd4fd8fe4755858568ab7d80e31e146f49ea5)

It's not supported right now, mostly becuase it would require reworking
how filterrifc works out what window type to use, but at least this way
I can mess with the emissive plane without needing to recompile or do it
manually

## Why It's Good For The Game

Makes starlight actually pretty instead of just fullbright + the windows glow a bit, emissives working is good actually, and filterrific not being annoying makes me happy.

## Changelog
:cl:
fix: Starlight will actually show now, instead of just being replaced by fullbright most of the time
fix: Emissive lighting will once again work. Sorry lads
/:cl:


code/__HELPERS/lighting.dm
```diff
@@ -4,6 +4,27 @@
...
+	blocker.appearance_flags |= make_blocker.appearance_flags | EMISSIVE_APPEARANCE_FLAGS
+	blocker.dir = make_blocker.dir
+	blocker.color = GLOB.em_block_color
+
+	// Note, we are ok with null turfs, that's not an error condition we'll just default to 0, the error would be
+	// Not passing ANYTHING in, key difference
+	SET_PLANE_EXPLICIT(blocker, EMISSIVE_PLANE, make_blocker)
+	return blocker
+
+>>>>>>> e874c2116a4 (Fixes emissives, starlight, and a filterrific bug (#72326))
```
ghost:
```suggestion
// This is a semi hot proc, so we micro it. saves maybe 150ms
// sorry :)
/proc/fast_emissive_blocker(atom/make_blocker)
	// Note: alpha doesn't "do" anything, since it's overriden by the color set shortly after
	// Consider removing it someday?
	var/mutable_appearance/blocker = new()
	blocker.icon = make_blocker.icon
	blocker.icon_state = make_blocker.icon_state
	// blocker.layer = FLOAT_LAYER // Implied, FLOAT_LAYER is default for appearances
	blocker.appearance_flags |= make_blocker.appearance_flags | EMISSIVE_APPEARANCE_FLAGS
	blocker.dir = make_blocker.dir
	blocker.color = GLOB.em_block_color

	// Note, we are ok with null turfs, that's not an error condition we'll just default to 0, the error would be
	// Not passing ANYTHING in, key difference
	SET_PLANE_EXPLICIT(blocker, EMISSIVE_PLANE, make_blocker)
	return blocker

```
Do these Mirror branches fall out of date or something? This was added in #18243 but this PR is acting like it's not here...

ghost (commented):
Single conflict that shouldn't even be here. Odd.

code/__HELPERS/lighting.dm
Tastyfish: That mirror was merged after this one was opened. No automatic merging occurs because that would make a mess (and create a cascading amount of automated jobs to run that would break everything).

Tastyfish (approved):
Ah thank 🦅, emissives work correctly again in testing.

PR #18402, opened 2022-12-29
[MIRROR] Add lints for idiomatic balloon alert usage [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72280
--------------------
Adds lints for `balloon_alert(span_xxx(...))` (which is always wrong), and balloon alert where the first letter is a capital (which is usually wrong). Fixes everything that failed them. As a reminder, abbreviations like "AI" and "GPS" shouldn't be capitalized in a balloon alert.

In cases where this is intentional for flavor (there was one case), you can `UNLINT` like so:

```dm
balloon_alert(user, UNLINT("OMGWTFBBQ!!!"))
```

:cl:
qol: Several balloon alerts have been changed to be more consistent.
/:cl:


code/game/objects/structures/crates_lockers/crates/syndicrate.dm
```diff
@@ -0,0 +1,88 @@
+/obj/structure/closet/crate/syndicrate
```
ghost, line 1: This mirror's branch seems out of date, this file was added in #18209 while this PR acts as if it doesn't exist.

ghost (commented):
Cannot review proper until this file's re-ticked, though it seems there's no conflicts. We'll just have to abide by the new lints.

PR #17886, opened 2022-12-02
Makes sensitive snout a neutral trait

YakumoChen:
## About The Pull Request
Title
Value -2 > 0

## How This Contributes To The Skyrat Roleplay Experience
This trait has never once had mechanical impact on a round. As-is it is free points for what amounts to little more than flavour. It can stay as the cute thing you find out by accident by being worth nothing in quirks.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
  
</details>

## Changelog

:cl:
balance: Sensitive Snout trait no longer a negative trait, now neutral
/:cl:

PR #18405, opened 2022-12-29
Revert "Makes sensitive snout a neutral trait"

nevimer:
Reverts Skyrat-SS13/Skyrat-tg#17886

This reverts commit 5a2e65932fd486a114f734c7c1bb79e086e362b7.

## About The Pull Request
Since sensitive snout was changed, my characters have been broken and unable to edit traits. Additionally, the combat potential of this trait is very large. Honestly, we should be combat logging this as we're running it blind with no fact checking, and the cost edit was a wildly unpopular change that very much was made in bad faith to begin with. 

Free traits should not have massive debuffs. No other free trait, other than DNR, can fuck you over. We know stun batons are busted, but this is way worse. 

## Proof of Testing

<details>
  It's the inverse of the prior. 
</details>

## Changelog

:cl:
balance: sensitive snout now costs 2 points again, as it can paralyze you helplessly. 
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/datums/traits/negative.dm
+++ b/modular_skyrat/master_files/code/datums/traits/negative.dm
@@ -27,6 +27,16 @@
 	user.physiology.brute_mod /= 1.25
 	user.physiology.burn_mod /= 1.2
 
+/datum/quirk/sensitivesnout
+	name = "Sensitive Snout"
+	desc = "Your face has always been sensitive, and it really hurts when someone pokes it!"
+	gain_text = span_notice("Your face is awfully sensitive.")
+	lose_text = span_notice("Your face feels numb.")
+	medical_record_text = "Patient's nose seems to have a cluster of nerves in the tip, would advise against direct contact."
+	value = -2
+	mob_trait = TRAIT_SENSITIVESNOUT
+	icon = "fingerprint"
+
 /datum/quirk/monophobia
 	name = "Monophobia"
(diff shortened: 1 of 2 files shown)


YakumoChen:
> key notes here: 
-They can't be wearing a mask 
-You have to know in advance they have the trait, and there is no tell for it 
-Finding the knowledge for it requires esoteric means or just low RP running around and booping people en masse.
> 
> There is simply not enough here to make the trait viable, and the amount of times it actually comes into play is exceedingly low, given the above.
> 
> Being able to have a free 2 trait points just by wearing a mask all the time seems like not much of a downside if you ask me.
> 
> Compare to, say, nearsighted, where you can avoid it by wearing glasses sure, but it has to be a specific eye-slot item that severely limits your options in eyewear until research is done (and even then, you can't get sunglasses a large selection of other options)

The trait is, frankly, badly tuned, and despite being a "debilitating" trait, has been made as difficult as reasonably possible to exploit. The trait point nerf is well justified until better changes are made.

>  my characters have been broken and unable to edit traits

This isn't a reason to change it, consider going to a keyholder to have your traits reset if this is an issue.

nevimer:
Attempts to tune it have been shut down. Additionally, pretending it doesn't exist leaves you in the best place, as now theirs absolutely no reason to take it. Why would you take it?

nevimer:
I have a better solution, actually.

YakumoChen:
> Why would you take it?

Why would you take any quirk lmao
you're literally admitting you're taking it for the free points as it in this state

PR #18437, opened 2022-12-31
Probably saves init time by not loading an edited centcom shuttle dock that we don't use anymore

Paxilmaniac:
## About The Pull Request

Its as shrimple as that, I remade the dock when we still used centcom as the destination for the evac shuttle, but we don't anymore and jolly never fixed it.

All it does now is eliminate the shuttle docking port because I love the interlink :)) I love my corn syrup :)))

The server might load like a picosecond faster because we're not doing automapper loading on this single template.

## Proof of Testing

I can't connect to my own local for some reason so proof of testing might be a while :(

<details>
  
</details>

## Changelog

:cl:
del: The fancy CC dock won't load anymore because the shuttle doesn't actually dock there anymore
/:cl:


Jolly-66 (commented):
I'm on mobile so take this with a grain of salt.

Make sure the shuttle dock marker isn't there.

Paxilmaniac:
Please don't tell me that the entirety of the interlink working relied on the automapper template deleting it.

Jolly-66:
I don't know I never tested the otherwise.

Paxilmaniac:
Bruhhhh It fuckin is

Jolly-66 (approved):
image

PR #18458, opened 2023-01-01
[MIRROR] Completely Culls req_access_txt/req_one_access_txt [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72281
--------------------
## About The Pull Request

Hey there,

Now that every instance of `req_access` and `req_one_access` is a list of strings, there is absolutely no reason for req_access_txt/req_access_one_txt to exist. In fact, any instance where they were still used in the codebase was very convoluted and was very broken! Don't worry, I fixed it all out, and life is good. 

I also dmdoc the surviving access variables, because those were missing. I went on the side of caution and made a more verbose documentation with an example just to have people really grasp this (it took me a while to actually get it)
## Why It's Good For The Game

Fixes #72102.

I believe that we changed _everything_ over to the req_access/req_one_access system earlier this year in VV, but the problem is that _new mappers don't understand the difference between the two systems_. In fact, the "txt" system is completely redundant since all it does is transition stuff to the "base" system. So, let's just completely cull the one that's all but deprecated and ensure this confusion no longer arises. The whole purpose of "txt" seemed to be to convert the access, but it's all pointless now that we can just read the list directly.

I'm also 99% certain that the "access check" on vending machines broke (and didn't seem to have correct logic in the first place? I legitimately couldn't find a case where it could fail in testing, so I changed that up), and that's fixed up now. Let me know if I was clueless there. I know it's short-circuiting now as opposed to "all must be true", but it just didn't work.
## Changelog
:cl:
fix: Those with kitchen access should now once again be the only one who are allowed to vending-machine-restock individual snacking items via their trusty tray.
/:cl:


Tastyfish:
It was actually broken upstream, fixed by https://github.com/tgstation/tgstation/pull/72458. I've cherry picked the fix in now, so it should work.

Paxilmaniac (approved):
Not failing every single check???? No way????

PR #18460, opened 2023-01-01
[modular] alot of Company Requisitions price changes + return of the rave-visor and anomaly-modules

RatFromTheJungle:
## About The Pull Request

changes the prices on alot, like, alot, alot, mostly focusing on the medical, tool, and modsuit companies, but also featuring the energy gun one while I was doing it, also, adds the scav module (rave) to hacked consoles (modsuit company) and makes the teleporter module [illegal] while raising the price, and causing them to once-more, come with cores in them per: (price  negotiable, I felt 3k-3.6k is fair enough, though)
image

Left TODO: make sure this isn't broke, make a new subtype of crate that has a significantly lower export value, for the items to come in

alot of stuff is cheaper, making it more affordable; meaning you'll see people opting to use it more often, and alot of prices will be more consistent

## Proof of Testing

image
I hope this explains my point

(I tested it, I just forgot to get a screenshot)
## Changelog

:cl:
balance: a large majority of items in Company Requisitions have had their prices radically adjusted
add: readds the rave module, and the preloaded version of the anomaly cores
/:cl:


YakumoChen:
3k seems really low for modules that are supposed to be locked behind limited availability (anomaly cores have a hard cap) and stringent research (toxins)
Not to mention the cost of the raw cores themselves not really factored in.
Why not up it to like 5k or something

RatFromTheJungle:
> 3k seems really low for modules that are supposed to be locked behind limited availability (anomaly cores have a hard cap) and stringent research (toxins) Not to mention the cost of the raw cores themselves not really factored in. Why not up it to like 5k or something

not necessary; that's why, if someone wants money, it's not particularly hard to amount a lot, 3k is above the pay for any job passively, and around 2-3 of the better paying bounties (usually, including getting stock), which is an ideal level without forcing tedium for the sake of tedium.


and the hardcap doesn't 'really' come into play with my decision, because toxins never makes enough (8) of any to hit the caps, and i'd rather people get fun things to mess around with, then throw in more arbitrary numbers for the sake of doing it.

YakumoChen:
Its of secondary importance to note anomaly modules are _very_ powerful. 
Antigrav + jetpack works as ultra noslips and unrestricted movement on multiz and turf hazards.
The kinetic one let's you move and throw people and anchored machines around (though I would have to test how useful it is in combat)

RatFromTheJungle:
> Its of secondary importance to note anomaly modules are _very_ powerful. Antigrav + jetpack works as ultra noslips and unrestricted movement on multiz and turf hazards. The kinetic one let's you move and throw people and anchored machines around (though I would have to test how useful it is in combat)

for the former, maybe that's a pickle, but, at the same time, I don't really think of slips as the most engaging in-depth mechanic to be protected anyway.

for the latter; the standard one /cannot/ pick up people, thats kinesis+, which is the admin one, it(company reqs one) can just fling machines and objects in a method much clunkier then just getting TK and spam clicking

RatFromTheJungle:
i'll fix it if it broke more over the week or so, but fixing the fact (all the private order crates broke) is going to take alot longer


it infact, didn't break, but crates absolutely did

UntoldTactics:
based, maybe now medical and people can buy med supplies and more importantly the advanced mesh and medicated sutures that are never otherwise made because botany/chemistry don't doith, without spending thousands upon thousands...

Zonespace27 (approved):
I've had this TMed a few days, all seems fine
I trust you've handled the balance on this pretty decently besides

PR #18467, opened 2023-01-02
[MIRROR] Minor improvements to mining order console [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72239
--------------------
## About The Pull Request

Mining order consoles are now called 'order console' instead of 'vendor'

Cargo orders through the mining vendor can no longer be cancelled ("clear" on the ntos cargo app no longer removes department orders too)

Cargo orders made with mining points now state 'mp' (for 'mining points') instead of 'cr' to the Cargo console.

The order consoles now have a '+' icon that will instantly add +1 item to your cart, so you don't have to manually type 1 each time.

The order console's UI name is now named after the machine, rather than being hardcoded to 'Produce Orders'

Also makes the UI resilient to breaking if the wearer doesn't have an ID equipped.

## Why It's Good For The Game

Better player-feedback and makes a better distinction between mining order console and normal vendors, as well as fixes some problems like Cargo just deleting people's mining points by deleting their unrefundable orders.

The + button was a suggestion in the comments here and thought it would just be a nice harmless inclusion.

The departmental order cancelling on the NTos cargo app and UI not breaking if you don't have an ID were just bonuses that I fixed while dealing with everything else.

## Changelog

:cl:
qol: Mining order consoles are now named 'order console'.
qol: Orders from said mining order consoles can now no longer be cancelled, and are listed as 'mp' (mining points) in the Cargo console instead of 'cr'.
qol: Order consoles now have a + button that will instantly add 1 of the item.
fix: Departmental orders can no longer be cancelled through the NTos cargo application.
fix: Wearing an ID will no longer break order consoles.
/:cl:


Tastyfish:
Manually mirrored https://github.com/tgstation/tgstation/pull/71676

Tastyfish (approved):

PR #18475, opened 2023-01-02
[MIRROR] Keeps gc_destroyed from getting updated on every step thru the gc queue. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72401
--------------------
Keeps gc_destroyed from getting updated on every step thru the gc queue.

Fixes logic that assumed gc_destroyed is the time the object first qdel'ed. it used to get updated on each stage of the garbage controller and there are 3 stages.

Added list index defines for the inner gc item list.


Tastyfish:
@Zonespace27 This depends on #18361 which for no coherent reason, the indestructable door button leaks its circuit board even though both types are stock to tg. Cherry picking this into that PR doesn't fix it, so I dunno.

Tastyfish:
Ended up cherry picking it into https://github.com/Skyrat-SS13/Skyrat-tg/pull/18361 to make it work correctly.

PR #18488, opened 2023-01-03
ERT smartgun rebalance

Zonespace27:
## About The Pull Request
The smartgun is now far more accurate to the source material, firing bullets at a delay of 0.15 instead of 0.5, while having all AP  and wound bonus removed (and damage per bullet halved). Due to the increased fire rate, Marine engineers will need to be more careful to not run their suit out of power.

The smartgun currently kills through an inordinate number of wounds, which isn't what it should be doing. The smartgun should be a peashooter whose use is suppressing enemies and allowing your team to push up.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
balance: ERT smartgun now fires faster while doing less damage and having nearly all wounding potential removed
/:cl:

### DIFF
--- a/modular_skyrat/modules/marines/code/smartgun.dm
+++ b/modular_skyrat/modules/marines/code/smartgun.dm
@@ -29,6 +29,7 @@
 	mag_display_ammo = FALSE
 	mag_display = FALSE
 	company_flag = COMPANY_NANOTRASEN
+	fire_sound_volume = 30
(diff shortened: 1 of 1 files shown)


dawsonkeyes:
good pr. the real question is, will this make marines any more used?

Zonespace27:
> good pr. the real question is, will this make marines any more used?

no

tf-4 (approved):

PR #18502, opened 2023-01-04
[Modular] Replaces loadout headphones with the modular one, adds it to fun vendor too

desertmote:
## About The Pull Request
For the loadout it replaces `/obj/item/instrument/piano_synth/headphones` and you get `/obj/item/clothing/ears/headphones` instead, which can be worn in 3 different slots and has a nice 'on' sprite.

For the fun vendor it is added in addition... it may be hard to differentiate the two, though.

We have a headphones item that is more usable for accessorizing. I noticed it was hard/impossible to obtain, so decided to revive it like this.

## Proof of Testing
<details>
  I tested it, I promise :)
</details>

## Changelog

:cl:
qol: the more customizable headphones item is available again in the loadout and fun vendor
/:cl:


tf-4 (changes requested):
needs proof of testing

PR #18515, opened 2023-01-04
Remove hurtsposals

YakumoChen:
## About The Pull Request

Removes modular hurtsposals
Based this PR on comments from an upstream PR (which may partly return this behavior to the code if its merged anyway)
Todo:

- [ ] Remove the trashman trait from the cargo loader

Hurtsposals limits the creativity of our players and needlessly punishes them for using disposals for creative escapes or projects, among other things. You can't use it to kick people out of places they shouldn't be, make disposals for fast travel or rides, and a myriad of other things, because it tends to crit or kill people who ride down them.

## Proof of Testing

<details>
  
</details>

## Changelog

:cl:
del: Disposals is safe to ride again. 
/:cl:


dawsonkeyes:
Now, I'm not one to talk, but I do think being able to jump into a pressurized trash chute at any time without consequence solely for a laugh is a bit wonky. You wanna ride disposals, you pack meds!

RatFromTheJungle:
jumping down disposals isn't a 'creative escape;' you could theoretically use it for projects, it does 5 brute on a prob 20 (even if you shouldn't).

And, most importantly, you absolutely should not be using the /trash system/ to 'kick people out of departments', or fast travel, or rides, it is layers of exactly what we wouldn't want on an RP server for people to be flinging themselves down the trash willy-nilly (or because 'it's funny.'), it should be a last-resort situation. Infact; that's why they were were made to do damage in the first-place.

YakumoChen:
Too many maintainer -1s

PR #18530, opened 2023-01-06
DS-2 Firing Range

Kepteyn:
## About The Pull Request
Adds a firing range on DS-2 where the fitness area used to be. The firing range has two practice laser guns, along with a crate of shooting targets, an unbuilt ammo workbench, two weapon chargers and two earmuffs. The fitness gym and showers have been relocated to be connected to dorms. Location was picked to be as unintrusive as posslbe to the rest of the map as well as to make the ship symmetrical. Additionally adds cut AI wires to the prison cells, so the station AI can't lock prisoners in.
DS-2 is a forward operating base for the Syndicate. A shooting range for its crew to train at thus makes logical sense for an FOB to have. It also opens up new roleplaying opportunities, such as two operatives challenging each other to see who's got better aim, or an operative training a new recruit on how to use guns.
## Proof of Testing
<details>

Old version without firing range
normds2

New version with firing range
newds2l

</details>

## Changelog

:cl:
add: Adds a firing range to DS-2 (Interdyne FOB) along with a new area, DS-2 Firing Range.
fix: adds cut AI wire helpers to DS-2 prison cells so AI can't mess with them.
/:cl:


tf-4 (approved):
Code changes are utterly minor but fine

Paxilmaniac (changes requested):
We discussed before this PR was made that I didn't like where you put the firing range. Integrate it nicely into the area next to security where a firing range should be and I'll accept this.

Kepteyn:
> We discussed before this PR was made that I didn't like where you put the firing range. Integrate it nicely into the area next to security where a firing range should be and I'll accept this.

changes

Swapped dorms and the firing range around so the firing range is next to security. I also spotted that the DS-2 prison cells lacked the cut AI wire helper like every other door had, so I added them.

Paxilmaniac:
When I said that I meant more like putting the firing range near or next to the armory, not just next to a security area.

Kepteyn:
> When I said that I meant more like putting the firing range near or next to the armory, not just next to a security area.

Part of the reason I have located it where it is is that it could be integrated into that spot with minimal disturbance to the rest of DS-2. While I could probably relocate it to the armoury, the confines of new DS-2 means that it would likely cause quite a bit of change in addition.

Kepteyn:
Closing while I think of a better way to integrate this.

PR #18546, opened 2023-01-06
[MIRROR] RTD Rapid Tiling Device [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72256
--------------------
## About The Pull Request
Interior decor at its finest & fastest available in engineering Proto lathe after research 
- Left click on a tile to infer its type. If the tile is supported by the RTD then a balloon alert will be displayed showing "Tile changed to XXX". If not supported, it will tell you that
- Right click on tile to convert it back to plating. It can only deconstruct floor types which are supported
- Left click on plating to place your floor tile
- Only iron & glass floor types are supported because you can only feed iron & glass to the RTD to keep it consistent

Also, this is my first time ever making sprites so expect less.

https://user-images.githubusercontent.com/110812394/209545438-6a51e7bf-163a-4a18-9102-7b77107eb1b7.mp4

Cleaned up some code in the RCD file as a bonus some notables ones are
- moved `update_overlays()` from rcd to `obj/item/construction `so any class extending from this can use it
- removed `do_after()` when changing turf types cause its unessassary & doesn't play well with the RTD

## Why It's Good For The Game
- Replace damaged or dirty tiles without the need of a janitor, crowbar & removes manual labour by hand
- Give your room some style quickly maybe replace the whole station floor with glass tiles I don't know you decide
- Maybe lay floor tiles quickly in maintenance to stop rats from chewing on those cables

## Changelog
:cl:
add: RTD Rapid Tiling Device
imageadd: Sprite for RTD
code: moved update_overlays() from rcd to obj/item/construction so anyone can use it 
del: cooldown of 3 seconds when changing turf type
qol: Faster Turf decoration
/:cl:


code/modules/research/designs/tool_designs.dm
```diff
@@ -77,6 +77,7 @@
 	)
 	departmental_flags = DEPARTMENT_BITFLAG_ENGINEERING
 
+<<<<<<< HEAD
 //SKYRAT EDIT ADDITION//
 /datum/design/engi_plumbing
```
Zonespace27, line 82: what the fuck why

Zonespace27 (approved):

PR #18585, opened 2023-01-08
Makes the wall mounted nanomed vendors use the new sprite tg added rather than ours

Paxilmaniac:
## About The Pull Request

A comparison between the two, here is the one we currently override tg's with:

image

And below is the new tg version that was added with a past mirror:

image

I don't know about you but I do not think we should override sprites that already have a perfectly fine tg equivalent. In the past I know tg's used to be awful (white square with a plus on it) but the new one is already set up to be compatible with wallening when that happens. It will just be less work on our side.

## Proof of Testing

The second screenshot in the about the pull request section is an in game screenshot of my locally testing the change.

## Changelog

:cl:
imagedel: The skyrat sprite override for wall mounted medical vendors will no longer be
/:cl:


Ebin-Halcyon (approved):
Aesthetic module my beloathed.

ghost:
While you're touching wall mounts, could you include the bounty board (Newscaster) sprite as well? TG's is really nice and has a little keyboard and everything, but ours is still just a flat panel and :(

tf-4 (approved):

PR #18651, opened 2023-01-12
Mirror #72195

Zonespace27:
https://github.com/tgstation/tgstation/pull/72195
Closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/18409

🆑 SkyratBot
fix: Blob minions are now sorted into biohazards in orbit UI.
add: Blobs now get antag info screens.
add: Blob minions get objectives (protect the core!).
add: Adds blob style text for victories, etc.
refactor: Refactors some blob & blob minion code.
/🆑


Tastyfish:
@Zonespace27 Tempted to merge this anyway since we so rarely run blob, but does all of this work with our blob controller overrides?

Tastyfish (approved):

PR #18664, opened 2023-01-12
[MISSED MIRROR] Replace nearly every map grep with a YML file that actually respects structure, and fixes the massive amount of failures that were never caught (#72372)

Tastyfish:
Mirrors https://github.com/tgstation/tgstation/pull/72372

[Documentation here](https://github.com/Mothblocks/tgstation/blob/maplint/tools/maplint/README.md)

We should not be using greps to the capacity that we currently are. If you are not smarter than a parser, then you should not try to beat one. DMM files should NOT be treated as text files that can be parsed with any old Unix tool. They are a structured language. Because of our abuse of greps, check_greps is full of hard to read, hard to maintain checks that do not consistently work, because they all make very specific assumptions about how TGM works, which are provably untrue.

This format is mostly straightforward for the lints people write, and easily extensible to the ones people will want to write.

:cl:
fix: Fixes a bunch of cases of windows not being where they were supposed to be, tables/chairs stacking on each other, and other very small stuff you've never noticed before.
/:cl:


Tastyfish:
Some of our maps have problems.

Tastyfish:
@Paxilmaniac 's map fixes are merged in.

Paxilmaniac:
lets see if i got everything

Paxilmaniac (approved):
lints didnt fail so i approve this message

Paxilmaniac:
This is like mostly mapping I'm gonna merge it because I hate map merge conflicts. Bite me.

PR #18671, opened 2023-01-13
[MIRROR] Removes cargo culting from human mob icon update [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72641
--------------------
These if checks did nothing, `icon_exists` would always return false in these calls because the `icon_file` var was defined to null the line before.


Zonespace27:
That feeling when every edit (sans 1) conflicts

GoldenAlpharex (approved):
👍  Golden, Add This! 👍

PR #18692, opened 2023-01-13
[MODULAR] Fixes Cryopods not removing players from Joined Player List

softcerv:
## About The Pull Request
When a player is cryo'd they are now removed from the global `joined_player_list`. 

Without having cryo'd players removed from the global `joined_players_list` , persistent scars ends up runtiming because it looks at the joined_players_list and then tries to save the data of a human in the list that does not exist.

Runtimes are bad for RP.

## Proof of Testing

<details>

image
  
</details>

## Changelog

:cl:
fix: Cryopods now remove players from the global list of joined players, stopping the persistent scars system from runtiming
/:cl:

### DIFF
--- a/modular_skyrat/modules/cryosleep/code/cryopod.dm
+++ b/modular_skyrat/modules/cryosleep/code/cryopod.dm
@@ -161,7 +161,7 @@ GLOBAL_LIST_EMPTY(valid_cryopods)
 	var/tucked = FALSE
 
 	/// What was the ckey of the client that entered the cryopod?
-	var/stored_ckey = FALSE
+	var/stored_ckey = null
 
 /obj/machinery/cryopod/quiet
 	quiet = TRUE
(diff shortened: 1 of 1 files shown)


Zonespace27:
This might fix some harddels that cryo's been causing

modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -160,6 +160,9 @@ GLOBAL_LIST_EMPTY(valid_cryopods)
 	/// Has the occupant been tucked in?
 	var/tucked = FALSE
 
+	/// What was the ckey of the client that entered the cryopod?
+	var/stored_ckey = FALSE
```
Zonespace27: just have this be an empty string or `null` instead of a bool

softcerv: done

Zonespace27 (approved):

PR #18735, opened 2023-01-15
[MIRROR] Ties research servers to Techwebs and de-hardcodes research point gen [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72513
--------------------
## About The Pull Request

Research point generation is now done on all techwebs that have ``should_generate_points`` set to TRUE, which by default is only the Science techweb.

I also entirely replaced the old 'servers' and 'master_servers' lists on SSresearch with the individual 'techweb_servers' (that already existed prior to this PR, the title is a little misleading) which are all stored on the individual techwebs, so we no longer store these in 2-3 different lists. Funnily enough, master server's list was unused as their behavior was refactored a good while ago.

Sabotaging a master server now only affects the techweb the master server is connected to, rather than cut everyone's by half

I also removed ``calculate_server_coefficient`` instead of reworked it to work with techweb servers, because it was entirely unused.

## Why It's Good For The Game

Better consistency for several techwebs, as sabotaging one techweb won't sabotage all of them.
This was also at the request of one of our contributors who wanted to make use of this system but didn't know that non-science techwebs didn't generate points, so here you go.

## Changelog

:cl:
fix: If there are several techwebs, sabotaging the master server of one of them will no longer cut research point generation of the rest.
/:cl:


Zonespace27 (approved):

Stalkeros2:
WOOOOOOOOO

GoldenAlpharex (approved):
Should be ready to go now!

PR #18760, opened 2023-01-16
[MIRROR] Converts blindness and nearsightedness to status effects, scratches some VERY dumb blindness handling that resulted in mobs becoming "incurably" blind [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72267
--------------------
## About The Pull Request

- Nearsighted is now a grouped status effect.
- Blindness is now a grouped status effect.
   - Eye handling of blindness has improved. 
      - When eyes are removed, they now cause you to become blind, rather than handling it in `update_tint`.
   - Being ahealed no longer blinds you for one tick, meaning that black overlay on aheal is gone.
- Temporary Blindness is now a status effect.
- Both Nearsightedness and Blindness have been exorcised from mob vars and life chains. This means that we've finally cut 2 procs from life, `handle_status_effect` and `handle_traits`, and moved both to event based processing. Wooo optimizations. 
- Swapped pacifism status effect to use apply and set helpers. 
- Removed an unused admin toggle that disabled welding helmet tint but also tint from every clothing item and also blindness from losing your eyes. 
- Clothes now generally all blind their mob more consistently.
- Oculine, eye surgery, and sensory restoration are now no longer the only way to fix blindness from eye damage. If your eyes are healed through any other means, it will also heal your blindness. 
   - Some things that made you blind, such as ling blind sting, no longer just flat made you blind from eye damage forever. They now cause eye damage directly, which in turn makes you blind from eye damage, as expected. 
- Pacifists can't eyestab anymore. Eyestabs now have a limit on the amount of blur applied. 
- Refactored some `is_x_covered` procs to accept flags rather than have a lot of arguments for some silly reason. 
- Unit tests for blindness. 

## Why It's Good For The Game

Blindness was exceptionally poorly handled prior, primarily due to the fact that it was tied to the mob instead of separated out

On top of that the system put a LOT of faith in proper handling of blindness on the coder's end which was misplaced evidently. Many places didn't update or handle blindness correctly, or just let people perma-blind. 

Deferring it to a status effect improves this a lot

## Changelog

:cl: Melbert
refactor: Refactored blindness and nearsightedness. Important to note is that all mobs are naturally blind until their eyes are actually created. 
refactor: Refactored "is covered" procs
fix: Less sources of blindness now cause permanent blindness. Includes the "Blind" Spell and "Blind Sting" from changelings.
admin: Ahealing someone no longer flashes the blind overlay for 1 tick.
admin: I removed an unused (sort of) inaccessible admin verb that allowed you to toggle the tint from all welding helmets (and clothing) (and lack of eyes) in existence, let me know if you want similar back 
balance: Changeling "Blind Sting" now causes eye damage (enough to blind) rather than arbitrarily forcing blindness.
balance: Visionloss virus symptom now causes eye damage (enough to blind) rather than arbitrarily forcing blindness.
balance: Oculine has been reworked slightly. Prior, Oculine arbitrarily healed blindness and nearsightedness from eye damage reagrdless of how damaged the eyes were, and applied blur on success. Now, Oculine just heals eye damage, and blindness / nearsightedness is restored in the process. There is now a probability every tick that eye blur is applied based on how pure the oculine is while healing very damaged eyes. 
balance: Pacifists can no longer eyestab.
balance: Any clothing item that covers your eyes contributes to getting the bonus while sleeping, and to removing temporary blindness faster
/:cl:


GoldenAlpharex (approved):
Okay, it should be good to go now.

GoldenAlpharex (approved):
I lied, now it's good.

GoldenAlpharex (approved):
I swear, if I have to update this AGAIN.

PR #18809, opened 2023-01-18
[MIRROR] Adds maintenance tablet apps (Health analyzer is one now too) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/71738
--------------------
## About The Pull Request

**Sprites were generously made by** https://github.com/Tramzz

This was a branch I've been putting off for months because my hatred for tablet apps brought me to removing computer parts, tablets, and ntnet, however I've now returned (after a lot more procrastination).

Adds Maintenance tablet applications to the game, so far there's only three of them, for proof of concept;

1. The health analyzer app (the chem analyzer part is removed entirely) move away from being given to medical/geneticists/detectives/RD.

There is on exception, which is the CMO, who gets to have the application on their tablet roundstart. Maybe it could be given to their role disk as well?

2. A camera application:

image
image

3. MODsuit control application:

image
image

image

Maintenance applications stand out from normal tablet apps because they can't be downloaded off the App store, and instead can only be found in maintenance, with a one-use download, cloning the application from a disk to a computer, or vice versa, will delete the old one, meaning you can only have one application at once.

## Why It's Good For The Game

This is more as a proof of concept for maintenance applications, but I also think that the analyzer application wasn't really that good as an app, you should either use a health analyzer or the wand in front of medbay, you shouldn't just have one in your tablet at all times because it makes it lame if your analyzer is stolen.

## Changelog

:cl: JohnFulpWillard, sprites by Tramzz
add: Added Maintenance tablet applications, applications that can't be cloned or downloaded from the store, instead you can find one app in maintenance.
balance: The Analyzer tablet application is also a maintenance tablet application now.
/:cl:


GoldenAlpharex (approved):
Willard Moment. 😔

PR #18835, opened 2023-01-20
Mirrors https://github.com/tgstation/tgstation/pull/67678

nevimer:
[…67678)
](https://github.com/tgstation/tgstation/pull/67678)
It occured to me, we didn't have a good way to "see" what turfs were actually being updated Figured I'd fix that

I've also added some debug vars on SSlighting to make testing with/without some checks easier

Speaking of which, I've added a second check to lighting corner updating Basically, if our past and current cached rgb values are the same, there's no point updating

This is possible because static lighting is relative. If you've got a TON of blue, it'll outweight the red and green you have in smaller amounts

We also do some rounding to ensure values look right

Similarly, if you've got roughly the same lighting, and a bit of something you already have a lot of is added, you're not likely to actually enter a new "bracket" of color

Anyway uh, it's hard to profile this, but I've seen it help quite a bit, mostly with things like emergency lighting that updates lighting in small amounts often, and in constricted spaces.

To some extent just comes down to map design

(cherry picked from commit a6d4e180adf47504e16b5e521ed249ba496bb1c4)

# Conflicts:
#	code/modules/lighting/lighting_object.dm

### DIFF
--- a/code/__DEFINES/subsystems.dm
+++ b/code/__DEFINES/subsystems.dm
@@ -283,6 +283,7 @@
 	* * callback the callback to call on timer finish
 	* * wait deciseconds to run the timer for
 	* * flags flags for this timer, see: code\__DEFINES\subsystems.dm
+	* * timer_subsystem the subsystem to insert this timer into
 */
 #define addtimer(args...) _addtimer(args, file = __FILE__, line = __LINE__)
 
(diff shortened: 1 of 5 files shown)


GoldenAlpharex (approved):
Nice catch!

PR #18875, opened 2023-01-22
[MIRROR] Fix signal override merge skew [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72882
--------------------
Caused by https://github.com/tgstation/tgstation/pull/72517, with https://github.com/tgstation/tgstation/pull/72561 raising an error that was being hidden before

This changes equip logic somewhat so that set_wearer was being called twice.

I don't think this should be being run in visualsOnly at all, as it does a ton of stuff like registering signals, updating UI, etc etc etc. Don't have time to investigate further.


Funce:
Waiting on #18884

GoldenAlpharex:
Will be closed by the PR that introduces it as well, so we don't break CI.

PR #18948, opened 2023-01-25
Spawn Curator PDA with ringer ON by default

DannyBoy3642:
## About The Pull Request
Configures the Curator's PDA to spawn with the ringer on by default by removing initialization code that was only needed for muting it. Coincidentally, this probably shaves a few nanoseconds off the initialization of Curator spawns.

While it makes sense flavor-wise for a Curator's PDA to be silent in the library, this should be something the player does by choice. Not all Curators are aware that their PDA will not ring by default - resulting in missed communications - and some that are aware do not like it. This way, the community can vote on whether it should or should not be muted by default. This change was requested by a frequent Curator player.

## Proof of Testing
Yes, it works. Green highlight not included.

[Proof of Functionality](https://user-images.githubusercontent.com/98357253/214572881-c4ee170d-b8d2-47f9-a938-b188c3197f5c.png)

## Changelog

:cl: Danny Boy
qol: Curator PDA now spawns with ringer ON by default
/:cl:

### DIFF
--- a/code/modules/modular_computers/computers/item/role_tablet_presets.dm
+++ b/code/modules/modular_computers/computers/item/role_tablet_presets.dm
@@ -348,6 +348,13 @@
 		/datum/computer_file/program/newscaster,
 	)
 
+/* Mutes the Curator's ringer on spawn
+/obj/item/modular_computer/pda/curator/Initialize(mapload)
+	. = ..()
+	for(var/datum/computer_file/program/messenger/msg in stored_files)
+		msg.ringer_status = FALSE
+*/
+
 /**
  * No Department
  */
--- a/code/modules/modular_computers/computers/item/role_tablet_presets.dm
+++ b/code/modules/modular_computers/computers/item/role_tablet_presets.dm
@@ -348,12 +348,12 @@
 		/datum/computer_file/program/newscaster,
 	)
 
-/* SKYRAT EDIT REMOVAL BEGIN - Mutes the Curator's ringer on spawn
+/* // SKYRAT EDIT REMOVAL BEGIN - Mutes the Curator's ringer on spawn
 /obj/item/modular_computer/pda/curator/Initialize(mapload)
 	. = ..()
 	for(var/datum/computer_file/program/messenger/msg in stored_files)
 		msg.ringer_status = FALSE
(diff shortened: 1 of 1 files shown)


Zonespace27 (dismissed):
Don't remove code entirely, comment it out

DannyBoy3642:
> Don't remove code entirely, comment it out

Commited this as per your request, but it was originally deleted entirely to comply with guidelines for removing unused code rather than commenting it out

DannyBoy3642:
Seeing where my issue is with modularization. Would you recommend a modular override that leaves the code intact or a label for a SKYRAT EDIT? I'm happy to figure out an override module if you think it'll be better, just gotta know what's best. I'm still very new to this codebase.

Zonespace27:
> Seeing where my issue is with modularization. Would you recommend a modular override that leaves the code intact or a label for a SKYRAT EDIT? I'm happy to figure out an override module if you think it'll be better, just gotta know what's best. I'm still very new to this codebase.

Yea, put sr removal comments on both ends

DannyBoy3642:
> Yea, put sr removal comments on both ends

Should be set now. In the future if I figure out a higher quality/safer way to do this I will PR that. Hopefully my other PRs will be of a better quality. I'll set this one as ready once tests are passed

DannyBoy3642:
Alright, looks good. Thanks for the patience. I'll try to pay more attention if I get into core files in the future, though hopefully that won't be needed any time soon. Cheers

Zergspower:
this seems exceptionally odd to have a callout like that for a role, curious as to why

DannyBoy3642:
Several reasons. Requested by a frequent Curator player in suggestions, the Curator can end up being rather important to a round and making any path for communication less likely to be effective doesn't seem the best. On top of that, this makes it about as close as it can be to atomic for something that, while tested as working and checked against other code to verify it would have the intended effect, is still modifying a core file.

It was intended for whoever was around on the PR list to vote freely on it, but I didn't account for the lack of voters. I could theoretically make any role that doesn't need the silence for a reason (stealth?) spawn with the ringer on, but that is outside of the scope of the request I saw

tf-4 (approved):

PR #18962, opened 2023-01-26
[MODULAR] Fixes invisible suit storage item sprites when wearing cloaks

### LINKED ISSUE #18787: Chaplain weapon not visible.
Round ID: 7456
Issue: Chaplain weapon will not be visible if put on the storage suit slot if you have a boatcloak equipped. (As a teshari, in case it works with other races, I haven't tested.)
Reproduction:
So when you put the weapon onto the storage slot while You have a boat cloak on it is simply not going to be visible, if You unequip the boatcloak and put the weapon onto the slot then it is visible even if you equip the boatcloak after as long as you don't get the weapon in hand and put it back in which case the bug occurs again.

vinylspiders:
## About The Pull Request

Fixes #18787 , which was actually an issue across all cloaks. 

Equipping a suit storage slot item before the cloak allowed the item's sprite to display. But if you equipped a cloak first, then the suit storage sprite would be invisible.

To get it to show again you'd have to take off the cloak, put the item into suit storage, and then put the cloak back on. Which was clunky and inconsistent.

---

This makes cloaks no longer hide the suit storage slot by default, unless the HIDESUITSTORAGE flag is turned on for the specific cloak type. This does not appear to have any negative side effects as far as I can tell.

Many suit storage items are big and should not be hideable, and the small ones still get drawn under the cloak anyway and so are basically still hidden/partially hidden with this change. The only thing it really affects is the examine text showing your suit storage slot when wearing a cloak, which it now does at all times. 

Less fiddling around with equip order to adjust for wonky sprite update behavior.

## Proof of Testing

<details>
  
image

image

image

</details>

## Changelog

:cl:
fix: items equipped in the suit storage slot no longer have their sprites hidden by cloaks
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/clothing/suits/cloaks.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/suits/cloaks.dm
@@ -1,2 +1,2 @@
-/obj/item/clothing/neck/cloak/
+/obj/item/clothing/neck/cloak
 	flags_inv = NONE //removes the HIDESUITSTORAGE flag
--- a/modular_skyrat/modules/customization/modules/clothing/suits/cloaks.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/suits/cloaks.dm
@@ -1,4 +1,2 @@
-/obj/item/clothing/neck/cloak/Initialize(mapload)
-	. = ..()
-	AddComponent(/datum/component/surgery_initiator)
-	flags_inv &= ~HIDESUITSTORAGE //removes the HIDESUITSTORAGE flag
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/customization/modules/clothing/suits/cloaks.dm
```diff
@@ -0,0 +1,2 @@
+/obj/item/clothing/neck/cloak/
```
tf-4:
```suggestion
/obj/item/clothing/neck/cloak
```

tf-4 (approved):

PR #18976, opened 2023-01-27
Missed Mirror: Remove NTNet from doors (saves 0.2s of init) -- Fixes windoors not wo

### LINKED ISSUE #3: Remove NTNet
Code that creates NTNet interfaces, broadcasts signals, etc is very slow.

### Estimated Cost
- `/datum/component/ntnet_interface/Initialize` is 125.11ms.
- `/obj/proc/init_network_id` is 114.75ms over 1,076 calls.
    - Most of this is in `log_telecomms` (81.72ms) and is easily fixable. There is a path where it replaces the current network with a new one, which is ran regardless of if the networks are the same, which happens thousands of times in init. Check telecomms.log of any round!

### Solutions
NTNet and the general radio framework is an enormous piece of code that is completely invisible to players, as there is no interesting way to utilize NTNet itself. There exist things like the NTNet circuit components, but they are on their own isolated network and can be trivially replaced.

NTNet and the radio frequencies from atmos stuff can be completely removed. Aside from that, fixing the `log_telecomms` will lower a significant cost.

nevimer:
…rking with remotes, and balloons them (#71735)
https://github.com/tgstation/tgstation/pull/71735
## About The Pull Request
Removes NTNet from doors. NTNet as a whole is getting excised, but this was an enormous cost to initialize. Was only used for door remotes, so just changes them to act on the door directly.

Windoors were supposed to be controllable, but were broken at some point to always display an error message in chat.

Closes https://github.com/tgstation/dev-cycles-initiative/issues/3

## Changelog

:cl: Mothblocks
fix: Door remotes now open windoors again.
qol: Door remotes now use balloon alerts for failures. /:cl:
:/cl:
(cherry picked from commit db2566f44dccc56a4b96808cb1a83b3b4dc4d0a5)

# Conflicts:
#	code/game/machinery/doors/airlock.dm


nevimer:
if you want it go ahead and reopen, im not fighting stalebot.

nevimer:
it's mirror time baby, feature freeze february let's goooo

lessthnthree (approved):

PR #19059, opened 2023-01-31
[MIRROR] Starlight Polish (Space is blue!) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/72886
--------------------

## About The Pull Request

Adds support to underlays to realize_overlays
Ensures decals properly handle plane offsets
Fixes space lighting double applying if it's changeturf'd into. this will be important later
Makes solar vis_contents block emissives as expected
Moves transit tube overlays to update_overlays, adds emissive blockers to them

#### Adds render steps

An expansion on render_target based emissive blockers. 
They allow us to hijack an object's appearance and draw it somewhere else, or even modify it, THEN draw it somewhere else. 
They chain quite nicely

Fixes shuttles deleting z holder objects

#### Makes space emissive, makes walls and floors block emissives
The core idea here goes like this:
We make space glow, and give its overlays some color

This way, the tile and space parallax remain fullbright, along with anything that doesn't block emissives, but anything that does block emissives will instead get shaded the color of starlight

This requires a bit of extra work, see later

This is done automatically with render relays, which now support specifiying layer and color (Need to make an editor for these one of these days)

The emissive blocking floor stuff requires making a second render plate to prevent double scaling

Also adds some new layering defines for lighting, and ensures all turf lights have a layer. We'll get to this soon

#### Makes things in space blue

We color them the same as starlight, by taking advantage of space being emissive
This means that things in space that block emissive will block it correctly and be colored blue by the light overlay, but space itself
will remain fullbright

This does require redefining what always_lit means, but nothing but cordons use that so it's fineee

#### Makes glass above space glow, and some other stuff

Glass tiles that sit above space will now shine light with matching
color to the glasses color. This includes mat tiles.

Glass tiles (not mat because they have no alpha) also only partially block emissives.
Adds a new proc that uses render steps to acomplish this, essentially we're cutting out bits below X alpha and drawing what remains as an emissive.

#### Modifies partial space showing to support glow

Essentially, alongside displaying space as an underlay, we also display a light overlay colored like starlight.
That starlight overlay gets masked to only be visible in bits that do not contain any alpha.

We also mask the turf lighting to not go into bits that have no alpha, to ensure we get the effect we want.
This is done with that lighting layer thing I mentioned earlier.

#### Makes appearance realization's list output ordered

I want it output in order of overlay, sub overlay suboverlay, next overlay
Need to use insert for that

## Why It's Good For The Game

Pretty!
Also having space be emissive is a very very good way to test for fucked emissive blockers (If it's broken why are we even drawing the overlay)
I know for a fact mob blockers on lizards and socks are kinda yorked, I think there's more

<details>
<summary>
Old
</summary>

image
image
image

</details>

<details>
<summary>
New
</summary>

image
image
image

</details>

## Changelog

:cl:
add: Space now makes things in it starlight faintly blue
fix: Glass floors that display space now properly let space shine through them, rather then hiding it in the dark
add: Glass floors above space now glow faintly depending on their glass type
/:cl:


GoldenAlpharex:
You know, I've been looking at this for a moment and realized "man I don't want space to be blue"

Paxilmaniac:
Space should be blue I think

RimiNosha:
It's a little too blue IMO, but otherwise it's fine to me

lessthnthree (approved):

lessthnthree:
The blue looks nice. Perhaps it can be reduced in intensity, but I don't agree with removing it.

PR #19063, opened 2023-01-31
Fix: Quirk Points & Augments+

### LINKED ISSUE #18512: Unable to spend quirk points
## Round ID:
197604 According to sb.atlantaned.space/rounds

## Reproduction:
I was unable to spend my last two quirk points.  I have a single prosthetic arm for -1 and -5 from two other quirks: Heavy Sleeper and Depression, which should give me -6 total.
image
image
image

However, I am unable to spend the last two perk points on my desired perk: Hardened Soles.  I am able to select the perk, and it appears in the Current Quirks tab of the menu, though my quirk balance does not change when I do so.  It only updates based on slotting in Light Step.  Once the menu closes it removes Hardened Soles and I am left with the remaining perks I posted above.

A brief note:  while I am fairly certain this is the correct round ID, this bug has been consistent ever since character creation across multiple logins and two played rounds.

Floofies:
## About The Pull Request

This PR fixes a bug which prevents the usage of Quirk points with the Augments+ page of Character Preferences. The bug was caused by a minor developer oversight wherein some TG procs were missing the required modular edits, causing the client-side data validation to fall out of parity with the server. The impact of the bug on the player is that their Quirks/Augments+ pages show them incorrect data, leading them to believe they have selected a Quirk or Augment when such selection was invalid.

This PR fixes #18512 

There are two independent bugs, and this PR only fixes one. First and foremost TGUI is sending a list of selected Quirks which the server detected was invalid due to insufficient points balance, as Augments' costs were not being added to that balance serverside.

When the player selects a Quirk or Augment that exceeds their point balance due to a failure of clientside validation, and the server detects this and attempts a "static data reset", the subsequent call to `update_static_data` fails to update Quirks and Augments+ pages only. I was not able to fix that bug in this PR, however I did fix our custom client-side prediction so it is now in-parity with the server.

Here is a granular change list:
- Added code for counting augment cost in `/datum/controller/subsystem/processing/quirks/proc/filter_invalid_quirks`, which necessitated the addition of a new argument to include selected augments.
- Added usage of `filter_invalid_quirks` to `/datum/preference_middleware/limbs_and_markings/proc/set_organ_aug` and `set_limb_aug` for gracefully filtering new invalid selections. This use-case is in the context of validation, and it removes some positive quirks if the client sends erroneous data that would cause the point balance to become invalid. Ideally these functions are never used unless we break client-side validation again.

This PR fixes the usage of Quirk points in combination with Augments+ in the Character Preferences window, allowing players to select Quirks and Augments according to their actual Quirk points balance. For quite a while this bug has remained hidden due to the server failing to update the client, a "silent failure" which is normally only detected after spawning or after re-opening the window.

While there is some more work needed to fix the overarching bugs upstream, now our custom system will work properly and players will be able to use Quirks interoperably with Augments as expected.

## Proof of Testing
<details>
  
snip1
snip2
snip3 snip4

</details>

## Changelog
:cl: A.C.M.O.
fix: Fixed Augments+ and Quirk points to interoperate as expected. You can now use your points balance to select augments, and select augments to earn points!
/:cl:


code/controllers/subsystem/processing/quirks.dm
```diff
@@ -157,13 +157,19 @@ PROCESSING_SUBSYSTEM_DEF(quirks)
 /// be valid.
 /// If no changes need to be made, will return the same list.
 /// Expects all quirk names to be unique, but makes no other expectations.
-/datum/controller/subsystem/processing/quirks/proc/filter_invalid_quirks(list/quirks)
+/datum/controller/subsystem/processing/quirks/proc/filter_invalid_quirks(list/quirks, list/augments)
```
tf-4: skyrat edit comment

Floofies: Oopsie, fixed by commit 9f4a5e9e8550b1b0994be29237c273b47ef347b4

tf-4 (approved):

PR #19072, opened 2023-01-31
[MANUAL MIRROR] Lazarus Injected mobs don't fight each other

lessthnthree:
## Original PR: https://github.com/tgstation/tgstation/pull/72440

Fixes #72404
The Lazarus Injector doesn't currently work on basic mobs, but should.

Problem:
The EMPed state of the Lazarus Injector is intended to make a revived mob hostile to everyone except you, including other mobs you have revived wtih an EMPed Lazarus Injector.
This is trivial to achieve for Simple Mobs which essentially all share the same AI, but I could not think of a single workable solution for Basic Mobs which don't, or at least any which didn't come with a tedious requirement to closely consider this niche item when programming any additional AI.

Solution:
Change the default behaviour of the Lazarus Injector so this is not a problem.
If all it does it make the mob loyal to you _and_ friendly to other mobs which are loyal to you, then it's pretty easy because we can just use the existing faction flags.

This is unambiguously a buff to using the item for nefarious purposes as now if you revive four ice drakes and fulton them onto the station they won't kill each other until only one is left, but is the only workable solution I could really think of.
A lot of the very dangerous mining fauna can't be dragged so transporting your army to the station still poses a question.

The alternate solution was just to replace the AI controller of any emp-revived basic mob with a "zombie" AI controller, but this has the problem that
A- It would now make things like cows and dogs into hostile creatures when they previously weren't.
B- It loses any interesting behaviour the mob previously had and for cases like Bileworms doesn't even make any sense (they'd try to walk and just get stuck in place).
This ultimately leads to needing to make bespoke versions for various mobs, which doesn't seem desirable from a maintainability standpoint.

As a side note it's still not a great idea to revive Bileworms _anyway_ as, their ability to move is tied to their ability to attack so once they don't have a target they will just kind of sit there and if they _do_ get a target their attempts to help you fight are difficult to distinguish from attempts to kill you... but at least being able to revive them makes it easier to make one sapient if you really want to trap a player's mind inside a body which is incapable of leaving lavaland.

Additional edit:
At Fikou's suggestion I've also added a sentience comparison proc to `mob/living` and removed some code duplication which dealt with this problem in the sentience/mind transfer potions, as well as added it to the Dominate spell.

This device is meant to revive mobs and it shouldn't be required for players to memorise an arbitrary list of which mobs it does and doesn't work on.
Especially as the goal is eventually that all simple mobs should be basic mobs.
This way of working is more intuitive, even if it is also stronger. I was surprised when I used EMPed injectors and my "new minions" just killed each other.

:cl: Skyratbot
fix: You can now revive 'basic mobs' with a Lazarus Injector, such as dogs, cows, axolotls, or carp.
fix: The same category of mobs can also now be effected by the Runic Golem Dominate spell.
fix: Basic Mobs will switch target if they can no longer attack their current target; meaning that if you become a Bileworm's friend it will stop attacking you.
balance: Mobs injected with the Lazarus Injector while it is EMPed will no longer attack other mobs revived by EMPed Lazarus Injectors.
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/xenobiology.dm
+++ b/code/modules/research/xenobiology/xenobiology.dm
@@ -719,7 +719,7 @@
 		smart_animal.sentience_act()
 	dumb_mob.mind.add_antag_datum(/datum/antagonist/sentient_creature)
 	balloon_alert(user, "success")
-		dumb_mob.copy_languages(user, LANGUAGE_MASTER)
+	dumb_mob.copy_languages(user, LANGUAGE_MASTER)
 	after_success(user, dumb_mob)
 	qdel(src)
 
--- a/code/modules/research/xenobiology/xenobiology.dm
+++ b/code/modules/research/xenobiology/xenobiology.dm
@@ -689,46 +689,39 @@
 	var/sentience_type = SENTIENCE_ORGANIC
 
 /obj/item/slimepotion/slime/sentience/attack(mob/living/dumb_mob, mob/user)
-	if(being_used || !ismob(dumb_mob))
+	if(being_used || !isliving(dumb_mob))
 		return
-	if((!isanimal(dumb_mob) && !isbasicmob(dumb_mob)) || dumb_mob.ckey) //only works on animals that aren't player controlled
-		to_chat(user, span_warning("[dumb_mob] is already too intelligent for this to work!"))
+	if(dumb_mob.ckey) //only works on animals that aren't player controlled
+		balloon_alert(user, "already sentient!")
 		return
 	if(dumb_mob.stat)
-		to_chat(user, span_warning("[dumb_mob] is dead!"))
+		balloon_alert(user, "it's dead!")
 		return
-	if(isanimal(dumb_mob))
-		var/mob/living/simple_animal/dumb_animal = dumb_mob
-		if(dumb_animal.sentience_type != sentience_type)
-			to_chat(user, span_warning("[src] won't work on [dumb_animal]."))
-			return
-	else if(isbasicmob(dumb_mob)) //duplicate shit code until all simple animasls are made into basic mobs. sentience_type is not on living, but it duplicated  on basic and animal
-		var/mob/living/basic/basic_dumb_bitch = dumb_mob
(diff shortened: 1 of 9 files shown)


code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -689,46 +689,39 @@
...
+	var/mob/dead/observer/C = pick(candidates)
+	dumb_mob.key = C.key
+	dumb_mob.mind.enslave_mind_to_creator(user)
+	SEND_SIGNAL(dumb_mob, COMSIG_SIMPLEMOB_SENTIENCEPOTION, user)
+	if(isanimal(dumb_mob))
+		var/mob/living/simple_animal/smart_animal = dumb_mob
+		smart_animal.sentience_act()
+	dumb_mob.mind.add_antag_datum(/datum/antagonist/sentient_creature)
+	balloon_alert(user, "success")
+		dumb_mob.copy_languages(user, LANGUAGE_MASTER)
```
tf-4:
```suggestion
	dumb_mob.copy_languages(user, LANGUAGE_MASTER)
```

code/modules/research/xenobiology/xenobiology.dm
```diff
@@ -689,46 +689,39 @@
...
+	var/mob/dead/observer/C = pick(candidates)
+	dumb_mob.key = C.key
+	dumb_mob.mind.enslave_mind_to_creator(user)
+	SEND_SIGNAL(dumb_mob, COMSIG_SIMPLEMOB_SENTIENCEPOTION, user)
+	if(isanimal(dumb_mob))
+		var/mob/living/simple_animal/smart_animal = dumb_mob
+		smart_animal.sentience_act()
+	dumb_mob.mind.add_antag_datum(/datum/antagonist/sentient_creature)
+	balloon_alert(user, "success")
+	dumb_mob.copy_languages(user, LANGUAGE_MASTER)
```
tf-4:
```suggestion
```

tf-4 (approved):

PR #19025, opened 2023-01-29
[MIRROR] Underwear Update (Sanity edition) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73010
--------------------

## About The Pull Request
Two changes to underwear/undershirts/socks work:

Randomly spawned species will not be given underwear unless they are a human, felinid, dullahan, or vampire.
This does not effect character creation,

Losing your torso will cause you to loose your undershirt and underpants. You can put more on at a dresser if you'd like.
Losing a leg will make you loose your socks (sad)
## Why It's Good For The Game
I tried to spawn in some naked android for an event and they had underwear on and it made me mad. This problem is consistent about just about every race expect for felinids and humans,
Also when a human gets auged, they now get to show off their sexy new robot chest without going to a dresser. I think this kinda stuff should have been in since day one,
## Changelog
:cl: itseasytosee
del: You are less likely to see underwear in places it logically should not be.
/:cl:

PR #19077, opened 2023-01-31
[MIRROR] Revert "Underwear Update (Sanity edition)" [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73074
--------------------
Reverts tgstation/tgstation#73010

This PR caused problems for underwear code that somehow didn't get caught in testing. Discovered so far are:

- Spawning at roundstart, latejoin, or by admin tools as a non-human character removes your underwear #73063
- Taking a head roll with a non-human character results in the human alternative character also having new underwear https://github.com/tgstation/tgstation/pull/73010#issuecomment-1408080782

According to @ NamelessFairy the root problem here is that on_removal() gets called during the process of setting up non-human species, leading to their underwear being removed. The human alternative is presumably a further extension of this same behaviour.

Given this is likely to require some level of refactoring to avoid, I think it would be better for everyone to revert until that can be done and save all non-human players the burden of having to find a dresser and fix their underwear every time they want to wear it.


RimiNosha:
#19025 is safe to merge provided this follows right after.

tf-4 (approved):

PR #19082, opened 2023-02-01
[MANUAL MIRROR] Big Paperwork

lessthnthree:
## Original PR: https://github.com/tgstation/tgstation/pull/70863

Adds paperwork -- an indecipherable mess of papers that only a seasoned department head could hope to decipher. If processed and sent to Central Command via the cargo shuttle, the budget gets a bonus.

Paperwork is distributed through a new Shuttle Loan offer. A nearby station needs some paperwork reviewed. Do you have what it takes to handle BIG PAPERWORK?

Processing paperwork is relatively simple. You can inspect it to see which head of staff it "belongs" to, and you can use their stamp to complete it. You get a lot of these, so the bonuses can add up if you get them all in. Just be careful not to return any unprocessed paperwork.

With big paperwork comes big paperwork fraud, and big fines.

You can photocopy your completed paperwork to double-dip with Central Command and squeeze out a little bit more money. The more paperwork fraud you commit, the higher the chance you'll be caught and fined.

On top of that:

Some say that in the long forgotten halls of maintenance, there lies _ancient paperwork_. Documents whose importance has long since passed, but are still important for bookkeeping at Central Command. With a keen eye, you might spot one of these while scouring through maintenance. Make sure it gets forwarded to the right person, cargo will thank you.

🆑 Skyratbot
add: Paperwork! Huge paperwork. Gigantic. You can stamp it and send it to Centcom via cargo for BIG MONEY.
add: New shuttle loan event variant -- You get a bunch of paperwork!
add: Ancient paperwork has appeared in the maintenance tunnels!
/🆑


GoldenAlpharex (approved):
Looks good, thanks!

PR #19120, opened 2023-02-02
ssDecay doesn't destroy tram turfs

lessthnthree:
## About The Pull Request
Exempts tram tiles from being destroyed during map load by ssDecay

The tram is built by [magic, tape, and bubblegum] so technically only the bottom left tile of the tram has actual floor tile. (/obj/structure/industrial_lift/tram/subfloor)

What this means for ssDecay is that it puts dirt/destruction on the tiles below the tram, and due to the other tram tiles being 'fake' the dirt shows up on the tram instead even though it's on the floor layer underneath.

This means any ssDecay applied to tram tiles can't be cleaned, such as these four:

image

You don't want a tram missing rails or unscrubbable dirt

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl: LT3
fix: ssDecay no longer applies on tram turfs
/:cl:


RimiNosha:
~~Depending on how confident you are with messing with SSdecay, it might be better to add a flag for SSdecay that says "you may do anything but destroy this turf".~~

Disregard, apparently SSdecay hates tram tiles in general.

Paxilmaniac:
Shoutout to ssdecay being one of the worst goddamn subsystems ever made

tf-4 (approved):

PR #19161, opened 2023-02-04
Gives assault ops a proper antag prefs icon, one that isn't just the goldeneye disk

Paxilmaniac:
## About The Pull Request

image

Look at these cool dudes

The antag icon being just the disks they need to collect falls behind with other icons in that it shows very little of what playing the role will be like. At a glance you can tell nukeops will work as a team of people because there's multiple people in the preview, and so on.

Since assops are a team, and should work together as such, it only makes sense to show them as a team on the preference.

## Proof of Testing

Look above

## Changelog

:cl:
imageadd: The antag preferences image for assault operatives has been upgraded from just being the goldeneye disk
/:cl:


tf-4 (approved):
nice

PR #19178, opened 2023-02-05
[DNM] [TM ONLY] Revert "[MIRROR] fixes fov removal f12 exploit [MDB IGNORE]"

Paxilmaniac:
Hey so you know that bug that's causing people's screens to get caught in the void VERY OFTEN?

image

Something in the exploit fix pr caused a significantly worse issue.

Yeah, we should uhhh testmerge this until the bug is fixed.


Paxilmaniac:
This is not needed anymore as of #19329

PR #19192, opened 2023-02-06
[Non-modular] Raises maximum character age from 85 to 300

desertmote:
## About The Pull Request

What it says on the tin.

The number is a little arbitrary. Sometimes a character happens to be older than 85. 

## Proof of Testing

dreamseeker_iv5sNml9u5

## Changelog
:cl:
qol:  Raises maximum character age from 85 to 300
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -420,7 +420,7 @@
 //MINOR TWEAKS/MISC
 //#define AGE_MIN 17	//youngest a character can be //ORIGINAL
 #define AGE_MIN	18	//youngest a character can be //SKYRAT EDIT CHANGE - age
-#define AGE_MAX 85 //oldest a character can be
+#define AGE_MAX 300 // SKYRAT EDIT CHANGE - OLD: #define AGE_MAX 85 //oldest a character can be
 #define AGE_MINOR 20 //legal age of space drinking and smoking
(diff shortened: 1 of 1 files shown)


Elli-Skala:
How does this effect old timer aging death?

lessthnthree:
Why 300? That's equally as arbitrary.

desertmote:
> Why 300? That's equally as arbitrary.

it is arbitrary but _not_ equally so.
ppl older than 85 exist in real life, first of all. and then furthermore we have races that can be even older like mechanicals or hemos.
the real answer is because the person who requested this PR gave me this number.

bowlofsoup4ik:
duh. what about:
a) raise limit even more;
b) maybe split biological and chronological age, so players can choose, how old character looks like and how old they actually is. or just let ppl choose examine text about age or smth (which like "she's seems young adult", "she's seems middle aged", etc etc)
upd1: yes, i stole it from rimworld
upd2: zolmachi gamers (like me) would be happy btw (⁠｡⁠•̀⁠ᴗ⁠-⁠)⁠✧

theselfish:
300 is a very arbitrary number.
Why not 999, which is the highest three-digit number. Or 9999.

Also, I swear that having age over 90 has a special examine text.

tf-4:
QOL changes come under the freeze

PR #19200, opened 2023-02-07
[MIRROR] Basic Mob Spiders I: Ability refactor [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73201
--------------------
## About The Pull Request

This morning a man presented me with a mission
image
I realised that I can achieve this in merely _three_ pull requests.
...because I'm leaving one that adds new AI until after feature freeze february.

This one pulls all of the ability code out of the giant spider mob file and spruces it up a bit in preparation for making them into basic mobs.
It removes an ancient "todo" about merging the spider and genetics web abilities by, doing that.
Spider abilities no longer extend from a shared and reasonably pointless "innate" root ability and also don't require their owners to be spiders to work properly any more.

Also while I was in there I fixed a bug where the "set directive" ability did not actually pass that into the antag datum of a new spider unless they logged out and back in.

## Why It's Good For The Game

Code should smell nicer now.
Removes an essentially duplicated ability.
Clears up some chat log spam which didn't need to be in there.
You can actually set an objective for spiders you create as a spider brood mother.
I guess you can make spiders-man as an admin by giving spider abilities to other mobs.

## Changelog

:cl:
refactor: Tidied up spider ability code, they should all work the same, but please report if they do not.
refactor: Spiders and Geneticists now use the same ability to make webs rather than two virtually-identical ones, which should also perform just as it did before this change.
fix: Setting a directive as a Spider Midwife should properly pass and announce itself to new spiders.
/:cl:


KathrinBailey (approved):
looks good?

PR #19201, opened 2023-02-07
[MIRROR] Space/Changeturf fixes and optimizations [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73261
--------------------
## About The Pull Request

We've got a few space related things that are busted, and shuttle movement is slow.
I'd like to try to improve these things, if just a bit.

Long list of only tenuously related topics. Sorry for the shotgun blast

#### [Fixes lazyloaded stuff having bad space](https://github.com/tgstation/tgstation/commit/d4de176a63f87d0f820e0cb610cd192750c897d6)

We need to handle area transferring in maploading code under niche cases, and we also need to actually init reservation spaces we create.

It's also redundant and potentially dupe creating to do area lighting handling in changeturf, because it gets touched in turf init anyway. Old me is stupid.

#### [Adds some doc comments, yeets ssmappping/transit](https://github.com/tgstation/tgstation/commit/269717145d70a4a73198791ca50253c708ee3ac1)

We had a reserved space for just shuttles to use, except it wasn't for just shuttles.
So in theory if the space got clogged with other shit, the shuttles could have nowhere to actually use.

It's better to just have the two groups share real estate. More sane

#### [Starlight optimization part two (infer context)](https://github.com/tgstation/tgstation/commit/be94c422ed76aa3f07b43cad4d1dc6b6148f135f)

Starlight was causing each space turf to cause itself and its neighbor to constantly recheck if they had starlight off changeturf.

The exact same effect can be had by taking advantage of some pre-existing information, namely if the space turf is gaining or losing a source of starlight.
Essentially, instead of telling a turf to check all adjacent turfs to see if it's got starlight, we tell the turf if WE are a source of starlight, or if we might be taking something away from it.

There's a bit of wasted cpu here but not much, if it's worth doing a register signal pattern for clearing depends on the case we're working with.

Being intelligent about this makes things much faster, something in the neighborhood of 4 to 3 fold.
I've also made openspace's starlight work better, cause the old pattern was a bit silly.

#### [Micro ops changeturf and turf deletion a bit](https://github.com/tgstation/tgstation/commit/386b3ab7fc2a820a9ffe3d2e39d78f96dc562d64)

Don't do work if the thing you're working on doesn't exist, don't check every adjacent turf for firelocks on turf change (just have thefirelocks manage that), don't check all atoms on the turf for decals on turf change, similar.
Also moves visibility changes from camera code into changeturf, to avoid unneeded work.

Needs some extra work to optimize the guts for this path but I can do that!

#### [Micros camera vis changes](https://github.com/tgstation/tgstation/commit/ebab69e9ea4adffd8787671f309f4ba27756c82e)

We should only update vis when our opacity changes. 
In addition, we don't need all the camera handling fluff if we only want to update our turf's static groups.

Also micros a camera net helper to be less crap for non multiz maps

#### [Ensures space_lit tiles never accidentially inherit lighting objects](https://github.com/tgstation/tgstation/commit/a99ff2265a4d1b157849fb7485adee17a3250df5)

S dumb, and leads to space turfs having two sources of lighting, which looks wrong.
This was invisible when their lighting was fullbright, but it sucks now.

#### [Cleans up a turf helper to not be stupid](https://github.com/tgstation/tgstation/commit/bf4ee6710026e6ca9922d0f1fa49020ebde8cd6f)

WHY ARE YOU USING THE RANGED TURF HELPER IF YOU GO ONE TILE

#### [Moves transit turf signal cleanup to destroy, I named this proc wrong](https://github.com/tgstation/tgstation/commit/c85c2cfc86f3b2dd224cae6b12e2fc428846c30b)

I'm sorry @ Time-Green 

#### [Adds better transit caching to shuttles](https://github.com/tgstation/tgstation/commit/35e85334c4f815da0cadd8172e9908267a01d334)

Adds a max reserved transit size to the shuttle subsystem, to keep things in bounds.
In addition, adds a soft cap under which existing transit space will get hold onto, to make repeated non escape/arrive shuttle movements faster

Hopefully this makes common shuttle moves less bad.
## Why It's Good For The Game

Speed


GoldenAlpharex (approved):
It works!

PR #19212, opened 2023-02-07
[MANUAL MIRROR] Fix a merge skew on armor datums

lessthnthree:
## Original PR: https://github.com/tgstation/tgstation/pull/72909

And also add a check grep to find it in the future


Paxilmaniac (changes requested):
Remove the armor varedits on the two maps linters are saying are causing problems first

Paxilmaniac (approved):

PR #19217, opened 2023-02-07
[MANUAL MIRROR] Adds an Escape Menu

lessthnthree:
## Changelog
🆑 SkyratBot
add: Added a new menu that shows on Escape.
/🆑


Jolly-66 (commented):
I'm not a codetainer, but I did personally speak with Mothblocks about this for my code.
This can be done relatively modularly. Not sure if it would be accepted for here, but you can look and see what we did here:

https://github.com/TaleStation/TaleStation/pull/4268

tf-4 (approved):

PR #19222, opened 2023-02-07
[MIRROR] Changeling Meteors will land near planetary stations instead of at the map edge [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73269
--------------------
## About The Pull Request

This is the actual reason I made #73241 but this fix ended up essentially unrelated to those ones so I atomised it.
When the Changeling Meteor event rolled on Icebox it would spawn a meteor at the edge of the game map and throw it towards the station, but generally speaking this means it would spawn one inside of a wall. This leaves the Changeling surrounded by ice turfs with no idea which direction to walk in, and while they can mine out the walls by hand it takes a really long and boring time.

To fix that, on planetary maps they now get dropped at Carp Spawn landmarks, which Icebox still has and so which I assume are expected to be a standard "spawn point which exists near but outside of the space station".

As a side note, this can often still work out quite badly for the changeling but in a way I wasn't sure I can fix in this PR. 
They spawn using a changeling spacesuit which feeds you Salbutamol instead of providing internals, which is fun and good flavour... for space maps. Icebox has a randomised atmosphere so depending on what chems are in the air it can still kill you anyway (or fill you with BZ metabolytes, which prevent your changeling resource bar from regenerating).
I was thinking of making a new chem to replace salbutamol for the suit but as that would have wider balance effects that felt like overstepping the boundary of Freeze February, so it might have to wait.

Until then you still have a much higher fighting chance of crawling through an airlock, but it might take you a regeneration or two. You're at least likely to be _near_ one now rather than half of a map away from the station.

## Why It's Good For The Game

Spending threat to put a ghost into the body of an antagonist who spawns inside of a wall at the edge of the map and then asphyxiates isn't very fun.

## Changelog

:cl:
fix: Changeling meteors on Icebox will now land close to the station instead of a random map edge tile.
/:cl:


tf-4:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/19154

code/modules/events/ghost_role/changeling_event.dm
```diff
@@ -42,14 +42,24 @@
  * * candidate - The mob (player) to be transformed into a changeling and meteored.
  */
 
-/proc/generate_changeling_meteor(mob/candidate)
-	var/mob/dead/selected = make_body(candidate) //Give the selected player a body, and grab their mind
+/proc/generate_changeling_meteor(mob/dead/selected)
```
tf-4:
```suggestion
/proc/generate_changeling_meteor(mob/candidate)
	var/mob/dead/selected = make_body(candidate) //Give the selected player a body, and grab their mind
```

tf-4 (approved):

PR #19231, opened 2023-02-08
[Void Raptor] Removes the floating Moffucinis Pizzeria

Jolly-66:
## About The Pull Request
Title.

At first, it was fine, I did have my doubts initially, but I let it go.
Having talked with Paxil a little bit, and having let it fester in my mind more, I don't like how this is just, ripped from Ice Box and pasted into the middle of the map sort-to-speak.

If we *really* want to have a unique ruin attached to maps, I believe we can do this with the .json files.

Also, Void Raptor is a modeled cruiser map. It makes NO SENSE for a carved out piece of land, asteroid, WHATEVER, to somehow just exist off the top of the ship. Its annoying, silly, and somehow breaks some immersion.

At least with the chapel, its loosely connected (although could be aesthetically "strengthen"). 

## Proof of Testing

<details>

image

image

</details>

## Changelog

:cl: Jolly
del: On Void Raptor, the Moffucinis Pizzeria was removed.
/:cl:

### DIFF
--- a/_maps/map_files/VoidRaptor/VoidRaptor.dmm
+++ b/_maps/map_files/VoidRaptor/VoidRaptor.dmm
@@ -296,16 +296,6 @@
 /obj/effect/turf_decal/stripes/line,
 /turf/open/floor/pod/dark,
 /area/station/service/chapel/funeral)
-"afE" = (
-/obj/effect/turf_decal/tile/blue/opposingcorners{
-	dir = 1
-	},
-/obj/machinery/atmospherics/pipe/smart/manifold/supply/hidden{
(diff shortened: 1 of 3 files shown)


Paxilmaniac (approved):
Approving and merging a pr from mobile epic style

Kepteyn:
While I cannot disagree with the intention here, the fact that this was made and merged in under an hour without consulting the original creator, who is an active member of the community who poured their soul into this, and not even giving them the courtesy of even being given time to know about or respond to it - resulting in much grief to them - is a massive smack in the face to not only them, but to any contributor.

PR #19232, opened 2023-02-08
[MDB IGNORE] Revert "[Void Raptor] Removes the floating Moffucinis Pizzeria"

RimiNosha:
Reverts Skyrat-SS13/Skyrat-tg#19231

Speedmerge moment. Please for the love of integrity do not make this kinda behavior normal. Do the right thing and reopen the removal PR and give people time to provide feedback.

### DIFF
--- a/_maps/map_files/VoidRaptor/VoidRaptor.dmm
+++ b/_maps/map_files/VoidRaptor/VoidRaptor.dmm
@@ -296,6 +296,16 @@
 /obj/effect/turf_decal/stripes/line,
 /turf/open/floor/pod/dark,
 /area/station/service/chapel/funeral)
+"afE" = (
+/obj/effect/turf_decal/tile/blue/opposingcorners{
+	dir = 1
+	},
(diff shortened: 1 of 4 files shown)


Kepteyn:
Reverting it is probably not best - it was rude, it was sudden, it hurt someone who has done nothing but show love and care to this community because they care about this community, but the change happened, and probably would have happened in some way or another anyway. What's important is to learn from this and handle things better in the future.

Paxilmaniac:
Jolly and myself are both in agreeance that we do not want the floating pizza place as it was implemented, and I'm not reverting on this.

RimiNosha:
So be it. Not like I'm staying here much longer anyways.

Jolly-66 (approved):
The speed merge wasn't warranted.
Discussion can go forward with this, however, my feelings and opinions in my PR are still true, and would personally see this not being baked into the map. 

This is discussion for another time, however.

RimiNosha:
It won't automerge cause MDB errored out.

PR #4584, opened 2021-04-01
Liquids system

Azarak:
## About The Pull Request
This introduces a proper atmos-like system for liquids, including immutable liquids such as the ones you'd use for an ocean map.
The subsystem is very efficient, it's a hybrid of cell and group processing and uses various caching methods to be more efficient.

Spill a beaker over a tile, or rupture a tank to get some liquids on the ground, they'll even out across the surface as a puddle, and slip people. Or use a grenade without foam or smoke, just raw reagents to spill them down
If you get enough liquids in one tile (will require closing a room, or making a pool with metal sheets), depending on the volume they'll behave as you'd expect an ankle-high, or shoulder-high or cant-breath-high water would. High enough water will make falls harmless, and will constantly expose the people inside to the reagents by touch, any slips or falls will cause you to ingest some too. The higher the level, the slower it'll make you too
You can even do reactions in the puddles, throw some potassium into a pool and cause an explosion.
The puddles can be set on fire aswell, provided they're a good accelerant, such as welding fuel, alcohol, clf3 or napalm

Reinforced Tables and Sandbags provide a partial blockage against liquids. If a room is set on fire by a liquid puddle, get on an r table!

Engineers get a plumbing device that can be used to create input and output pumps that interact with on-tile liquids

How do you get rid of them?
Water evaporates by itself, aslong as its in a puddle.
Use a mop, then right click a bucket to squeeze the liquids out.
If a water is not the most shallow puddle possible, you can also scoop it out by a beaker, or other containers
There's also a portable liquid pump you can find, functions similarly to the porta scrubber

## Changelog
:cl:
add: Added a liquids system, which includes a lot of changes. See https://github.com/Skyrat-SS13/Skyrat-tg/pull/4584
/:cl:

PR #19244, opened 2023-02-09
[MODULAR] Fixes some buggy reaction behavior by further modularizing liquid code

### LINKED ISSUE #13168: Ice Melting Unusually (Interferes with Bartending)
Reporting client version: 514.1582

## Round ID:
5185

## Testmerges:
- [Goofsec 2.0 aka "I guess we're going back to bullets in Security": Security Pistol w/ 6 Ammo Types, Warden Dispatch Console](https://github.com/Skyrat-SS13/Skyrat-tg/pull/12949)
- [[MODULAR] Status Indicators with Preference](https://github.com/Skyrat-SS13/Skyrat-tg/pull/12529)
- [Database revision revert](https://github.com/Skyrat-SS13/Skyrat-tg/pull/10823)

(I don't believe it's relevant to these, as it's been happening for a while)

## Reproduction:

So, I've noticed this because it interferes with a recipe I like a lot as a bartender. 

Step 1) Using a glass, add 15u of Sugar and swipe with a lighter to turn the sugar into caramel,
Step 2) Put the glass back into the dispenser and add 15u of Cream, followed by 15u of Ice (in that order)
Step 3) Only get 36.2u of Caramel Shake, and 13.8u of Water

Now, the reason that made me realize this was bugged, is that over on TG, doing that exact recipe gives you the proper 45u of Caramel Shake. However, what's happening here is that the ice melting into water and the caramel shake mixing at the same time fill the glass together, and because it caps out at 50u, you don't get the full drink.

I tested it using a Shaker, and I got the full 45u of Drink, with some extra Water leftover, too. I also tested a 10u glass of Ice, melted to see if the ice melts into more water than there was ice, and got 9.8u Water.

Something fucky is going on with the way our chems mix, and I know it's Skyrat sided because the recipe works perfectly well on TG proper.

vinylspiders:
## About The Pull Request

Fixes #13168

This was a really difficult to track bug that was ultimately caused by the liquids system (#4584) using an existing var (`no_react`) to reduce overhead by delaying the reactions, but as a result of the way they went about it it had the side effect of breaking some things unexpectedly across all reagents.

Removed the offending Skyrat edits and modularized it as much as I could. There are now no more Skyrat edits for this, as none were needed it turned out. 

No logic should be changed as far as liquids go, their non-reactive optimizations still happen where they're supposed to, and it will be less likely to break things in the future now.

No more mysterious incomplete reactions--less pain for bartenders, that is for sure. 

## Proof of Testing

<details>
<summary>About to add ice...</summary>

image

</details>

<details>
<summary>The ice now fully reacts</summary>

image

  
</details>

## Changelog

:cl:
fix: fixed some bugs in reaction code that was causing incomplete reactions, especially related to bartending
/:cl:


code/modules/reagents/chemistry/holder.dm
```diff
@@ -275,18 +275,14 @@
...
-	if(!no_react)
-		handle_reactions()
-	//SKYRAT EDIT CHANGE END
+		add_reagent(r_id, amt, data)
 
 
 /// Remove a specific reagent
-/datum/reagents/proc/remove_reagent(reagent, amount, safety = TRUE, no_react = FALSE)//Added a safety check for the trans_id_to
+/datum/reagents/proc/remove_reagent(reagent, amount, safety = TRUE)//Added a safety check for the trans_id_to
 	if(isnull(amount))
```
vinylspiders, line 286: (this was an unmarked Skyrat edit)

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
```diff
@@ -279,7 +279,7 @@
 /obj/effect/abstract/liquid_turf/proc/take_reagents_flat(flat_amount)
 	var/datum/reagents/tempr = new(10000)
 	if(flat_amount >= total_reagents)
-		tempr.add_reagent_list(reagent_list, no_react = TRUE)
+		tempr.add_noreact_reagent_list(reagent_list)
 		qdel(src, TRUE)
```
vinylspiders, line 283: By making a new proc we don't need to have a skyrat edit here either.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -303,7 +299,7 @@
 			amount = clamp(amount, 0, cached_reagent.volume)
 			cached_reagent.volume -= amount
 			update_total()
-			if(!safety || !no_react)//So it does not handle reactions when it need not to //SKYRAT EDIT CHANGE
+			if(!safety)//So it does not handle reactions when it need not to
 				handle_reactions()
```
vinylspiders, line 303: `no_react` is not needed here, the default behavior for `remove_reagent` is to not call `handle_reactions` because `safety = TRUE` by default.

All this was doing was causing reactions to happen prematurely when `no_react` is false (which it is almost everywhere by default) leading to incomplete chemical recipes as in the initial bug report.

code/modules/reagents/chemistry/holder.dm
```diff
@@ -275,18 +275,14 @@
 	return TRUE
 
 /// Like add_reagent but you can enter a list. Format it like this: list(/datum/reagent/toxin = 10, "beer" = 15)
-/datum/reagents/proc/add_reagent_list(list/list_reagents, list/data=null, no_react = FALSE) //SKYRAT EDIT CHANGE
+/datum/reagents/proc/add_reagent_list(list/list_reagents, list/data=null)
 	for(var/r_id in list_reagents)
```
vinylspiders, line 279: No Skyrat edits are needed in holder.dm for what the original coder was trying to do, so I removed all of the ones that they added in #4584.  They did nothing but cause problems. Thus this fix is now fully 'modular.'

tf-4 (approved):

PR #19272, opened 2023-02-10
Reagent containers AltClick update

Shadow-Quill:
## About The Pull Request

If a reagent container has only two options for its output, it will now just swap between them.
I've also commented out the code in `robot/items/hypo.dm` which was causing borgs, when trying to swap the injection amount, to swap to the opposite of what they chose.

Additionally, the input list for changing containers that _do_ have more than one option is now a TGUI input. UI!

Popups will no longer happen if a reagent container only has two possible outputs. Multiple choice UI looks nicer.

## Proof of Testing

<details>
  
image
image

</details>

## Changelog

:cl:
qol: Reagent containers with only two possible settings (ie borg hyposprays) will now just cycle through the options instead.
qol: Reagent transfer setting lists are now in TGUI! Rejoice for clean UI!
fix: Borg hyposprays are no longer fiddly when setting the injection amount.
spellcheck: Reagent containers that can transfer one unit now properly pluralize units.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/items/hypo.dm
+++ b/code/game/objects/items/robot/items/hypo.dm
@@ -232,9 +232,9 @@
 
 /obj/item/reagent_containers/borghypo/AltClick(mob/living/user)
 	. = ..()
-	if(user.stat == DEAD || user != loc)
-		return //IF YOU CAN HEAR ME SET MY TRANSFER AMOUNT TO 1
-	change_transfer_amount(user)
+	// if(user.stat == DEAD || user != loc) // SKYRAT REMOVAL START - Changing transfer amounts is now handled by the parent proc in modular files.
+	// 	return //IF YOU CAN HEAR ME SET MY TRANSFER AMOUNT TO 1
+	// change_transfer_amount(user)	// SKYRAT REMOVAL END
 
 /// Default Medborg Hypospray
 /obj/item/reagent_containers/borghypo/medical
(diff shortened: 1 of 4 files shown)


code/game/objects/items/robot/items/hypo.dm
```diff
@@ -232,9 +232,9 @@
 
 /obj/item/reagent_containers/borghypo/AltClick(mob/living/user)
 	. = ..()
-	if(user.stat == DEAD || user != loc)
-		return //IF YOU CAN HEAR ME SET MY TRANSFER AMOUNT TO 1
-	change_transfer_amount(user)
+	// if(user.stat == DEAD || user != loc) // SKYRAT REMOVAL START - Changing transfer amounts is now handled by the parent proc in modular files.
+	// 	return //IF YOU CAN HEAR ME SET MY TRANSFER AMOUNT TO 1
+	// change_transfer_amount(user)	// SKYRAT REMOVAL END
```
tf-4:
```suggestion
/* SKYRAT REMOVAL START - Changing transfer amounts is now handled by the parent proc in modular files. 
	 if(user.stat == DEAD || user != loc)
		return //IF YOU CAN HEAR ME SET MY TRANSFER AMOUNT TO 1
	 change_transfer_amount(user)
*/ //SKYRAT REMOVAL END
```
Don't commit this directly cus I did it on mobile so the indentation is probably fucked, but it's a little cleaner to do it this way

tf-4 (approved):

PR #19289, opened 2023-02-11
[MIRROR] refactors emote visuals [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73146
--------------------
## About The Pull Request
So right now you have the *blush and *cry emote which apply an overlay to you after emoting. They both do the same thing but have different code for it. This has been refactored using the new bodypart_overlays. An emote visual can be defined with a /datum/bodypart_overlay/simple/emote subtype. Running an emote can then apply this overlay to the user and delete it after a duration. The drunk mood event also uses one of these overlays now to apply the blush effect. The emote overlays use images instead of mutable_appearances so they should be able to support animations that only play once.

Also the crying visual now appears even if you are eyeless. This restriction didn't make much sense in the first place because the tear ducts aren't in the eyes. Emote overlays are attached to bodyparts. If you emote without the required bodypart the overlay won't be added. And if the bodypart is dismembered _after_ emoting, the overlay will stay on the dismembered bodypart. Once emote duration runs out it disappears from the bodypart. You can't re-emote to reset the duration because the bodypart isn't attached to you anymore.

Oh and also I put the blush and tears sprites in a new file.
## Why It's Good For The Game
- Makes it easier and less scary to add new visuals for emotes.
- Takes blush and cry overlay code out of eye and head code.
- Emote sprites are easier to find.
- human_face.dmi becomes ever so slightly less clogged.
## Changelog
:cl:
refactor: refactored *blush and *cry emote visuals into bodypart_overlays
del: lack of eyes will no longer prevent the *cry emote overlay from displaying
/:cl:


Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/19635

GoldenAlpharex (approved):
Honestly, you love to see it. It works super well, too.

GoldenAlpharex (approved):

PR #19181, opened 2023-02-05
[MIRROR] Netherworld Mobs Refractor [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73086
--------------------

## About The Pull Request

This PR refactors netherworld mobs into basic mobs as best as possible. Also makes some of them run faster when they are getting damaged or deal more damage. Now the mobs might be able to keep up a little with the players.
## Why It's Good For The Game

Makes the mobs have better movement and more dynamic movement. Makes the quality of these mobs better.
## Changelog
:cl:
add: Added new damage buffs for netherworld mobs
refactor: Refactors netherworld mobs into basic mobs
/:cl:

PR #19315, opened 2023-02-13
[MIRROR] Basic Mob Spiders III: Sgt. Araneus is no longer a bat [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73350
--------------------
## About The Pull Request

This PR converts giant spiders into basic mobs and resultingly fixes #37793 
They _should_ have the same behaviour as their simple mob versions although I can't verify that their movement speeds are _exactly_ the same. It should at least be pretty close.
A quirk of spiders is that they had a pretty large `move_to_delay` which made them slow in the hands of AI (because it would just pause for ages between taking steps) and faster in the hands of players, and they often appear in both forms so I had to implement this as a speed modifier based on player control.

Additionally this is the first basic mob which can be set on fire.
This is currently implemented as a var on `mob/living/basic` but I know there was some annoyance at adding the environment tolerances as vars on there so if desired I can try and extract it out, I'm just not sure how easy it will be.

Something else I noticed is that spiders seem to take stamina damage from bug spray... but stamina damage does nothing to either simple _or_ basic mobs. I have left it in for now in case I am missing something, and rebalancing it to do something else would be more like a balance change.

Oh also I killed the `mob/basic/retaliate` folder because that isn't a classification that needs to exist or makes sense.

## Why It's Good For The Game

We don't want to use simple mobs any more.
Sergeant Araneus can finally actually be a spider, instead of being a bat.

## Changelog

:cl:
refactor: Spider code has been refactored and AI-controlled spiders may have slightly different movement or reaction times.
fix: Basic mobs can now be slowed when they take stamina damage, however currently only spiders actually _can_ take stamina damage.
fix: Spiders should now more reliably disable their AI when controlled by a player.
fix: Araneus is no longer considered to be a bat and so cannot fly.
fix: Araneus is no longer considered to be a bat and so is no longer frightening to people who are scared of the supernatural.
/:cl:


Gandalf2k15:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/19181

Gandalf2k15:
closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/19657

Gandalf2k15:
closes https://github.com/Skyrat-SS13/Skyrat-tg/pull/19561

PR #19339, opened 2023-02-14
[MIRROR]  [NO GBP] Hotfixes lungless oxyloss immunity, and lungless plasmamen being able to be healed by salbutanol etc. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73291
--------------------
## About The Pull Request

I had thought I took care of this edge case but I was mistaken.

## Why It's Good For The Game

Fixes the edge case of a removed lung preventing oxyloss damage in carbon mobs. Big derp.

---

Also took care of the case of mobs without lungs being injected with oxyloss healing chems by adding a `mob_respiration_type` var.

This essentially just makes it so that plasmamen without lungs can no longer be healed by salbutanol etc. They are currently the only mob with a different respiration type than the default, so this particular bug only affected them. In the future this will be important as chems that target other respiration types are added.

The `mob_respiration_type` var also allows for simple mobs to have their own respiration types defined. By default everything is `RESPIRATION_OXYGEN,` so there is no difference from how it was before unless you specify a different type for that particular mob.

## Changelog

:cl:
fix: hotfix for lungless oxyloss immunity
fix: hotfix for mobs being able to circumvent reagent respiration_type requirements by removing their lungs
/:cl:


vinylspiders:
prerequisite: the pr that I made this hotfix for, #19206

lessthnthree (approved):

PR #19379, opened 2023-02-16
Radiation Leak Leaks Radiation

lessthnthree:
## About The Pull Request
Oops, I forgot to actually activate it.

:cl: LT3
fix: Radiation leak will now leak radiation
/:cl:


RimiNosha:
Didn't we explicitly disable this?

lessthnthree:
> Didn't we explicitly disable this?

I don't think so, it was recently merged. Are you confusing it with Radiation Storm?

edit: We didn't disable it, but it was never added. But they don't want it.

PR #19384, opened 2023-02-16
[MIRROR] Wraps all instances of invoking Fail() into the TEST_FAIL Macro [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73407
--------------------

## About The Pull Request

We shouldn't really be invoking the proc itself because then we don't pass along the failing line/file to our consumers, let's use the macro in all instances that really need it.
## Why It's Good For The Game

I noticed people were invoking the `Fail()` proc directly rather than using `TEST_FAIL` instead, so they weren't getting those neat annotations on their failing code because we never passed along the failing line/file to actually apply those annotations. That's silly. We don't even return on `TEST_FAIL` either, so there's no reason to not do this (only upsides wahoo).
## Changelog
Nothing that concerns players.


Tastyfish (approved):
thumbs up emoji

PR #19396, opened 2023-02-17
Anomaly Event: Ectoplasm

lessthnthree:
## About The Pull Request
Adds the new anomaly to the random events roster.

Paranormal activity can spawn roleplay.

## Changelog

:cl: LT3
code: Ghosts have been spotted in the corridor...
/:cl:

### DIFF
--- a/modular_skyrat/modules/event_vote/code/event_chaos_system.dm
+++ b/modular_skyrat/modules/event_vote/code/event_chaos_system.dm
@@ -146,6 +146,9 @@
 /datum/round_event_control/anomaly/anomaly_grav/high
 	chaos_level = EVENT_CHAOS_LOW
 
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
I don't know if we should have an event that allows:
image


deadchat to start breaking stuff on the station, considering all the problems we have with them currently

image

lessthnthree:
We could adjust it. The level below 50 makes a mess and some damage, didn't seem too bad when tested.

RatFromTheJungle:
> We could adjust it. The level below 50 makes a mess and some damage, didn't seem too bad when tested.

if we 'adjusted it', we should cut off the deadchat/observer parts entirely; no deadchat control, and no spawning ghosts, as those are the main issues with it

PR #10684, opened 2022-01-14
[NON-MODULAR] You can no longer use Dwarfism and Gigantism if you've got an already smaller/larger body respectively.

Iamgoofball:
## About The Pull Request

You can no longer use Dwarfism and Gigantism if you've got an already smaller/larger body respectively.

## How This Contributes To The Skyrat Roleplay Experience

i sure do love unclickable 80%ers + dwarfism, or excessively large tileblocking oversized users getting even bigger with gigantism

## Changelog

:cl:
balance: The Galactic Gene Therapy Consortium would like to issue a PSA regarding size gene mods and abnormally sized humanoids. It is not advised to use size adjusting gene modifications while abnormally sized, as adverse side effects can occur, up to and including a very painful death.
/:cl:

PR #19418, opened 2023-02-18
Fixes gigantism/dwarfism being able to resize players in cases where it shouldn't

### LINKED ISSUE #11522: You can no longer use Dwarfism and Gigantism if you've got an already smaller/larger body respectively, but small issue.
## Reproduction:
Get gigantism while you sprite size >1. Lose it, get 0.8 modification of your size 
https://github.com/Skyrat-SS13/Skyrat-tg/pull/10684 - couse this bug
dwarfism/on_losing and gigantism/on_losing dont checking if they changed size in on_acquiring

vinylspiders:
## About The Pull Request

Fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/11522

As the issue creator states https://github.com/Skyrat-SS13/Skyrat-tg/pull/10684 made it so you couldn't be resized if you were already big/small via character prefs, but they forgot to consider the case of losing the mutation which also changes your size. So you could still get resized that way, until now.

Bugfix

## Proof of Testing

  
</details>

## Changelog

:cl:
fix: fixes gigantism/dwarfism resizing some players after they lose the mutation when it shouldn't
/:cl:

### DIFF
--- a/code/datums/mutations/body.dm
+++ b/code/datums/mutations/body.dm
@@ -135,6 +135,12 @@
 /datum/mutation/human/dwarfism/on_losing(mob/living/carbon/human/owner)
 	if(..())
 		return
+	// SKYRAT EDIT BEGIN
+	if(owner.dna.features["body_size"] < 1)
+		to_chat(owner, "You feel relief as the pressure building inside you subsides.")
+		return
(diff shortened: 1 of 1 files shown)


tf-4 (approved):
nice

PR #19426, opened 2023-02-18
Fixes Tramstation elevator panels

### LINKED ISSUE #19343: Elevators are initializing with the wrong floor numbers
## Round ID:
7650

image

:psyduck:

lessthnthree:
## About The Pull Request
Semi-modular aesthetic update for Tramstation (and other map) elevators. 
- Updates code to display the correct floor - Fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/19343
- New icons to match our existing aesthetic overrides (and the upstream tram controls making its way here shortly.) https://github.com/tgstation/tgstation/pull/73476
- All items are now 'elevator' for consistency instead of a mix of lift/elevator.

Us having an Interlink z-level screws with the base elevator code. Buttons/indicator/panel redone to match our other aesthetic button overrides. Half the things saying lift and the others elevator is dumb.

## Proof of Testing

https://user-images.githubusercontent.com/83487515/219903380-7f388ea7-175e-47d5-be90-c8a6c409165b.mp4

## Changelog

:cl: LT3
fix: Tramstation elevators now correctly display what floor they're at
imageadd: New elevator sprites
spellcheck: All elevator components are now named elevator
/:cl:


ghost:
The icon changes would likely be accepted upstream, might be worth tossing that part up there

lessthnthree:
> The icon changes would likely be accepted upstream, might be worth tossing that part up there

I asked, they want their 3/4ths one.

Paxilmaniac:
I want their 3/4s one as well.

The existing aesthetics overrides have been thrown out before (look at newscasters, for example) and I really do not see a good reason to override an already really good sprite.



Which brings me to point two: I really don't think we need to non-modularly edit tg code when the solution to the problem could be solved by making the interlink zlevel load after the station rather than before it.

tgui/packages/tgui/interfaces/ElevatorPanel.tsx
```diff
@@ -52,9 +52,9 @@ export const ElevatorPanel = (props, context) => {
    * you should reconsider your map's layout. It's not worth it.
    */
   const calculatedHeight = clamp(all_floor_data.length * 90, 400, 600);
-
+  // SKYRAT EDIT CHANGE: line 57: theme ntos, original retro
```
Paxilmaniac, line 55: And this, I mean really?

lessthnthree:
> Which brings me to point two: I really don't think we need to non-modularly edit tg code when the solution to the problem could be solved by making the interlink zlevel load after the station rather than before it.

Okay, so let's do that.

PR #19469, opened 2023-02-20
[MIRROR] Most actions cannot be used while incapacitated [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73513
--------------------
## About The Pull Request

Fixes #39775
The `TRAIT_INCAPACITATED` trait is intended to block you from acting but does not inherently block actions. Actions only check for "conscious", "has available hands", "immobile", or "lying down". 
Most actions still _can't_ be used while incapacitated. This is because most actions are spells, most spells have invocations, and you can't talk while you are incapacitated. This is silly.

I have resolved this by adding a new flag which blocks actions while incapacitated. This is somewhat redundant with "conscious" because most sources of that also make you incapacitated but not _always_, you might want the specificity.

I have tried to be discerning about where this flag is applied, it is not in all cases (for instance you can still swallow implanted pills while incapacitated and such), but it's not only possible but I can't really avoid implementing this without it being a balance change in _some_ cases,

Actions this effects are things such as:
Xenomorph Tail Sweep, Lesser Magic Missile (cult constucts), Heretic Shadow Cloak, the Smoke spell in general, Conjuring (but not firing) Infinite Guns, Mime spells

Times when these actions will no longer be available but were previously are such times as when the mob is:
Stamina Crit, Stunned, Paralysis, and Time Stopped.

## Why It's Good For The Game

The incapacitated trait is applied by effects which should block acting but still permits several actions which logically would be prevented. 
This fortunately hasn't come up too often due to the high ratio of actions with invocations, but it feels bad to stun someone and then have them still able to perform an action they logically wouldn't be able to take while stunned.
This is especially egregious in the case of Time Stop (the only way to stun a lot of the mobs effected by this) but that's a rare pick on a rare antagonist and would also rarely be used on these mobs, so a bit of an edge case.

## Changelog

:cl:
fix: Many spell-like abilities which could be stunned, paralysed, or frozen in time now cannot be.
/:cl:


Gandalf2k15:
https://github.com/tgstation/tgstation/issues/72918

PR #19489, opened 2023-02-22
Makes CMGs in the armory spawn as gunsets that include both non-lethal and lethal magazines

Paxilmaniac:
## About The Pull Request

It was host requested that CMGs come with lethals of some kind on spawn, so that they aren't useless when lethals are required without making people print their own ammo immediately.

Having the option of both lethals and non-lethals on a security longarm without having to print your own of either right out the box is something we should've had a while ago.

## Proof of Testing

If I feel less lazy I'll give the boxes pixel shifting but you know
image

image

## Changelog

:cl:
balance: CMGs in the armory come in a gunset which includes both a non-lethal and two lethal magazines right out the box, rather than making officers have to print lethal magazines to use the weapon in a lethal role.
/:cl:


Paxilmaniac:
Test merge only until the freeze is over as golden said

UntoldTactics:
Great merge, though frankly we still need to fix that ol' bug (or its probably intended) where only lethal mags let you print the large variety of ammos in a gun, whereas the rubber mags only let you print rubbers (thus making it impossible to get ammo for a CMG from the ammo fabricator and instead requires using the clunky autolathe)

Zonespace27 (approved):

PR #19501, opened 2023-02-23
[PROJECT] Kindle

Zonespace27:
## About The Pull Request
*TF4 said it was alright to make project PRs during the freeze

The clock cult version of the bullshit stun hand, woo.
On a serious note, adds the "kindle" scripture, costing 1 cog to unlock and 125W to cast. Instead of the old way of a 15 second hardstun, Kindle instead acts like the heretic's mansus grasp, except slightly stronger to compensate for its requirement to stand still for 1s to cast it (not to use it, mind you).

Blood cultists and Clock cultists are unable to stun hand eachother.

I'm getting *decently* close to being done with clock cultists, honestly

## Proof of Testing

<details>
  
image

</details>


code/__DEFINES/~skyrat_defines/colors.dm
```diff
@@ -0,0 +1 @@
+#define LIGHT_COLOR_CLOCKWORK "#BE8700"
```
Gandalf2k15, line 1: I swear we have this already, make sure you move all color defines here

Zonespace27: We don't, not from what i've seen

tf-4 (approved):

PR #19536, opened 2023-02-25
fixes ghost cafe sleepers

tf-4:
image

## Changelog
:cl:
fix: Ghost cafe sleepers work again
/:cl:

### DIFF
--- a/code/modules/mob_spawn/mob_spawn.dm
+++ b/code/modules/mob_spawn/mob_spawn.dm
@@ -135,6 +135,8 @@
 
 
 	// SKYRAT EDIT ADDITION
+	/// Does this spawner have infinite uses?
+	var/infinite_use = FALSE
 	/// Do we use a random appearance for this ghost role?
 	var/random_appearance = TRUE
 	/// Can we use our loadout for this role?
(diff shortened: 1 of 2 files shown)


Paxilmaniac:
Weak solution: Skyrat edit the check

Strong solution: Uses = INFINITY

tf-4:
Eh, editing the check rather than having the uses be a big number seems more 'proper' to me, even if less modular, though I can do it either way.

vinylspiders:
> Eh, editing the check rather than having the uses be a big number seems more 'proper' to me, even if less modular, though I can do it either way.

Another modular option would be to change this to some number greater than 1

https://github.com/Skyrat-SS13/Skyrat-tg/blob/9465310dfa666860e0dcd88e9e33cc296f4c0355/modular_skyrat/modules/ghostcafe/code/ghost_role_spawners.dm#L11

https://github.com/Skyrat-SS13/Skyrat-tg/blob/9465310dfa666860e0dcd88e9e33cc296f4c0355/modular_skyrat/modules/ghostcafe/code/ghost_role_spawners.dm#L42

and then override 
```

/obj/effect/mob_spawn/ghost_role/robot/ghostcafe/attack_ghost(mob/dead/observer/user)
. = ..()

uses += 1
```
```

/obj/effect/mob_spawn/ghost_role/human/ghostcafe/attack_ghost(mob/dead/observer/user)
. = ..()

uses += 1
```

so that the uses are just immediately incremented after getting decremented, so 0 is never reached.

That is how I would do it anyway, if I were trying to avoid a Skyrat edit. I don't really see the big deal of having one for this though personally.

GoldenAlpharex (changes requested):
I feel like this could easily be taken upstream, I'm sure that they'd appreciate it, even if only for the admins being able to have a proper way to make spawners infinite.

PR #19560, opened 2023-02-26
[MIRROR] Refactor mob spawner component so that it is independent [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73645
--------------------
## About The Pull Request

Went down a rabbit hole with touching the migo files... I noticed that all of those mobs had a reference to `/datum/component/spawner` on them which looked a bit off.
After investigation it seems like this component is breaking the prinicple of using an ECS system by assigning a reference to itself on every mob it creates? There doesn't seem to be a good reason to do that, as we can just use signals.

This also doesn't work for basic mobs, because most of them don't _have_ this reference to assign to. If we don't want to add it to every basic mob (and why would we?) it would make more and more converted mobs invalid for spawners.
Also it means that it has never been valid to create a Monkey spawner, which seems like a big oversight.

I replaced all of the parts dependent on telling the mob where it was spawned from with signals.

Megafauna seemed to have a reasonable amount of code related to "not straying a certain distance from what spawned them", but as far as I can tell unless someone varedited one onto a map we have never had a spawner which creates megafauna (nor would we want one? That would virtually always cause it to respawn instantly after being killed).

## Why It's Good For The Game

Improves future maintainability
Brings implementation up to current standards.
Makes the code work the way I would have assumed it already worked in the first place.

## Changelog

:cl:
fix: Mob spawners will no longer break if instructed to spawn certain kinds of basic mob, or monkeys.
/:cl:


Cyprex:
CI failures are because this PR depends on #19181

PR #19583, opened 2023-02-28
Removes every instance of openspace used on void raptor, since single zlevel maps should not have such a thing.

Paxilmaniac:
## About The Pull Request

For some reason all of void raptor's rear engines had openspace under them.

This was causing runtimes. PSA to all mappers in the future, do not use openspace on a non-multiz map. It will cause problems.

## Proof of Testing

image
## Changelog

:cl:
fix: Void raptor no longer has openspace under its rear engines, which was causing problems.
/:cl:


GoldenAlpharex (approved):
Looks good, thank you!

PR #19590, opened 2023-02-28
[MIRROR] Faction defines [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73681
--------------------
## About The Pull Request

Quite a lot of mobs had faction defines as a string, which always has a chance for error. For example, the clown mob spawner had their faction written as "clown", when the official faction name was "clowns", and a define existed for it anyways! This PR moves every single string based factions over here. No references or special role factions. Hopefully I didn't miss anything.

I also moved a global define used for picking your uplink provider's flavour to the only file that used it, datum_traitor.dm, and renamed them a bit to avoid confusion.

I have also noticed that the mimic faction was assigned to the petrified player with += instead of |=. |= would ensure no duplicate factions, so I have changed it.

Future improvement:
I have noticed that there is a lot of bloat with factions that contain only one or two entries (examples: gnomes, floating eyeballs, penguins, the pet lizards), and some always appear in pairs (vines and plants, the rare exceptions being killer tomatoes and strange reagent spawned pine trees), but trimming consolidating them is a matter for a different time.

## Why It's Good For The Game

Makes assigning factions a bit less error prone.

I can finally remove the ```/// Later all mob factions will be moved here and their current factions replaced with faction defines.``` comment. Later is NOW.

## Changelog
Nothing player facing


Gandalf2k15:
I am unstoppable.

PR #19610, opened 2023-03-01
[99% NONMODULAR] Adds It/Its Pronouns

Nerev4r:
## About The Pull Request
A port from https://github.com/Artea-Station/Artea-Station-Server/pull/49 (hi Rimi!) which is a port from Orbstation (hi Lexi!). This, simply, adds it/its pronouns to select, and renames the options from "Male", "Female", and "Other" to "He/Him", "She/Her", and "They/Them". To quote VioletN, 'A character who uses he/him pronouns isn't necessarily "male", so the button shouldn't be labeled that.'

Further gender expression is cool as fuck.

## Proof of Testing

<details>

image

image

</details>

## Changelog
:cl:
add: You can now select it/its pronouns in setup.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/PreferencesMenu/MainPage.tsx
+++ b/tgui/packages/tgui/interfaces/PreferencesMenu/MainPage.tsx
@@ -20,629 +20,630 @@ const CLOTHING_SELECTION_WIDTH = 5.4;
 const CLOTHING_SELECTION_MULTIPLIER = 5.2;
 
 const CharacterControls = (props: {
-	handleRotate: () => void;
-	handleOpenSpecies: () => void;
-	handleLoadout: () => void; // SKYRAT EDIT ADDITION
-	gender: Gender;
-	setGender: (gender: Gender) => void;
-	showGender: boolean;
+  handleRotate: () => void;
+  handleOpenSpecies: () => void;
+  handleLoadout: () => void; // SKYRAT EDIT ADDITION
+  gender: Gender;
+  setGender: (gender: Gender) => void;
+  showGender: boolean;
 }) => {
-	return (
-		<Stack>
-			<Stack.Item>
-				<Button
-					onClick={props.handleRotate}
-					fontSize="22px"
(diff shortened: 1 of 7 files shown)


tgui/packages/tgui/interfaces/PreferencesMenu/MainPage.tsx
```diff
@@ -20,627 +20,629 @@ const CLOTHING_SELECTION_WIDTH = 5.4;
 const CLOTHING_SELECTION_MULTIPLIER = 5.2;
 
 const CharacterControls = (props: {
-  handleRotate: () => void;
-  handleOpenSpecies: () => void;
-  handleLoadout: () => void; // SKYRAT EDIT ADDITION
-  gender: Gender;
-  setGender: (gender: Gender) => void;
-  showGender: boolean;
+	handleRotate: () => void;
```
RimiNosha: Install the prettier plugin for VSC and re-save, please

RimiNosha (commented):
Also, hi!

Ebin-Halcyon:
Friendly reminder from me, add an extra new line to your details tab so images show up like so:

image

GoldenAlpharex (changes requested):
As stated on the Discord, I'd much rather have a pronouns pins instead of a solution like this, that would only affect the examine text (and leave you with `they`/`them` everywhere else because BYOND macros moment), as this is both highly non-modular, and relatively restricted in its appeal.

If people want to show their pronouns, so be it, but I would rather have it be more accessible than just hard-coding another option.

Nerev4r:
> As stated on the Discord, I'd much rather have a pronouns pins instead of a solution like this, that would only affect the examine text (and leave you with `they`/`them` everywhere else because BYOND macros moment), as this is both highly non-modular, and relatively restricted in its appeal.
> 
> If people want to show their pronouns, so be it, but I would rather have it be more accessible than just hard-coding another option.

We'll send this upstream and I'll do pronoun pins after that.

PR #19613, opened 2023-03-02
Undoes the override for holopad sprites

Paxilmaniac:
## About The Pull Request

Holopad sprites were redone by tg in https://github.com/tgstation/tgstation/pull/69126

image
image
image
image

I'll be honest I just do not like how our holopad override looks, and the tg one was redone so there's no justification left to keep the override.

## Proof of Testing

Soon

## Changelog

:cl:
imagedel: The sprite override for holopads has been undone, reverting us to to what tg has
/:cl:


Ebin-Halcyon (approved):
Knew I missed something. Fuck the aesthetics folder.

Paxilmaniac:
Fastest pr approval known to mankind

GoldenAlpharex (changes requested):
The fact that they were re-done upstream doesn't mean that they're necessarily better.

I can understand not liking the current ones, but I don't really think upstream's are much better. I'd much rather have our own version of it.

Overrides aren't a bad thing, being visually different is also part of the appeal of playing here.

Paxilmaniac:
I do not think we should override things when we do not have to. This is definitely a "we do not have to".

GoldenAlpharex (approved):
Fuck it I guess.

Gandalf2k15:
Those are objectively worse. Ney.

PR #19616, opened 2023-03-02
Borer fear (no host) works for one extra second

YakumoChen:
## About The Pull Request

Adjust Borer fear on a potential host to be 8 seconds of stun time instead of 7

Boring into someone takes 6 seconds without the upgrade. That's a single second leeway to press a button and potentially navigate a menu to enter a host you've stunned. This isn't very intuitive or fair, since all it takes to interrupt you (certain death) is moving a single tile in any direction. Gives slightly more grace to enter a host you've stunned

## Changelog

:cl:
balance: Borer stun lasts 8 seconds from 7, giving slightly more time to get in a host.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer_abilities.dm
@@ -438,7 +438,7 @@
 /datum/action/cooldown/borer/fear_human/proc/incite_fear(mob/living/carbon/human/singular_fear)
 	var/mob/living/basic/cortical_borer/cortical_owner = owner
 	to_chat(singular_fear, span_warning("Something glares menacingly at you!"))
-	singular_fear.Paralyze(7 SECONDS)
+	singular_fear.Paralyze(8 SECONDS)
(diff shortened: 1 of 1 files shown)


lessthnthree:
Skill issue? Haven't heard this complaint from anyone else.

YakumoChen:
it's easier when the target is alone or not aware of you but if you have to navigate the menu (multiple people around) the grace period becomes untenable.

Only other ways I see to fixing this would be to make boring into a feared target automatic or adjusting it so you auto-unhide when fearing so the boring upgrade always works (this doesn't apply to new borers however, which this PR helps)

RatFromTheJungle:
i don't know if the warningless (almost instant) hardstun needs even more help, being honest

YakumoChen:
It needs to be more intuitive than giving a single second to get through a prompt.

Zergspower:
yeah an even longer hardstun on an already broken mechanic? No thanks

Gandalf2k15:
As much as I hate the number 7(8 is massively superior in many aspects, namely it being an even number), I must say that this change is not required.

PR #19641, opened 2023-03-04
[MIRROR] [no gbp] Removes AB_CHECK_INCAPACITATED from spells [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73744
--------------------
## About The Pull Request

In #73513 I was a little overzealous and now most spells can't be cast when you are batoned even though _silence_ is meant to be the counter to wizards (and heretics).
I took this back off the root of `spell` and then using the principle of "anything with an invocation should be able to be used while stunned" checked which spells didn't have invocations.

It seems like literally the only one of _those_ which was problematic was the one from the original bug, implying that just maybe perhaps I should have used a targeted solution instead of applying my fix to like 80% of actions in the game.

Side effect: Any non-invocation spell can once again be cast while time-stopped. So like, Blink, Smoke, Cult Wall Conversion, and Fire Breath.
Personally I think this is a niche enough interaction that it doesn't matter and I don't think it's worth adding a new trait and check just for this one spell.

## Why It's Good For The Game

Numerous abilities like jaunting and notably Hasty Realignment, the heretic spell which removes stuns from you, were being blocked by stuns. This makes their owners much more vulnerable than they are supposed to be.

## Changelog

:cl:
fix: Most spells can once again be cast even after someone stuns you with a baton.
/:cl:


Zonespace27:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/19469

PR #19678, opened 2023-03-05
[MIRROR] Adds a stack_trace for emissives with invalid icon_states, fixes all that appeared roundstart [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73678
--------------------
## About The Pull Request
So, this spiraled from one missing icon being fixed to an entire check for said icons.
Several icon files no longer use error icons because its assumed that the checks will handle any missing ones, but the checks don't apply to emissives nor overlays at all. This led to the radsuit having an emissive but no icon_state for it - a relic of the old radsuit. This was only noticed because of a downstream with an error icon appearing for it...

I was curious how many were actually having the same issue, so I made a small little stack_trace in the mutable_appearance proc.
There were like, 2k. Lots of them were icons named, like, "transparent" or "blank" too...
I moved that check to the emissives proc because I semi-understand that system so could actually fix it, and it moved to around fourty roundstart. Much more achievable.

(The error usually has more info if you click on it, including the item that caused it. I dunno how to add that to the check itself because of where it's located, though.)
F99cOII1XJ

This fixes all the ones I could find, including...
Nonexistant icons that shouldn't be adding emissives on:
- Empty Barsign
- Radsuit
- Mass Driver Controllers
- Telescreens
- Aux Base Consoles
- PanDEMIC
- Kobayashi computer (holodeck)
- Abductor camera console
- Syndie drop pod
- BSA controller
Entirely missing icons on:
- Pwr Game Vendor (this was just misnamed)
- Generic Soda Vendor
- Engivend
- Security Laptop (proud of this one.,.,) dreamseeker_36PwO4HSLO

There are no doubt more of them hidden about, but I don't really know what I'm doing... If there's a check that'd be better than this, please review telling me what to change <3
## Why It's Good For The Game
Fixes missing icons, fixes attempts to add icons where we don't need them, and adds a check to help fix more of the two issues as they occur.
## Changelog
:cl:
fix: fixed missing emissives on the Engivend, Pwr-Game Soda, and generic Soda vendors. Also fixed the seclaptop having no valid screen icon!
fix: fixed a few items trying to apply emissives when they shouldn't.
code: added a stack_trace for emissives with missing icon states.
/:cl:


ghost:
### Line 25 of `modular_skyrat\modules\cryosleep\code\cryopod.dm` needs the following line added to pass CI:
`	icon_screen = null`

PR #19783, opened 2023-03-12
I now codeown just `/tools/updatepaths` instead of `/tools/updatepaths/scripts`

Jolly-66:
This is more broad as now I can check for SR specific UpdatePaths, not just /tg/ specific one.


GoldenAlpharex (approved):
Sure.

PR #19790, opened 2023-03-12
[MODULAR] The adult helmet is no longer a space helmet along with some other tweaks.

softcerv:
## About The Pull Request
Makes it so the adult helmet is no longer spaceproof along with removing the armor it had and the resistance to peppespray. 

This item is strictly for roleplay, the mechanics attached to this item that affect normal gameplay should be minimal.

## Proof of Testing

<details>

image

image

</details>

## Changelog

:cl:
fix: the sensory helmet is no longer a space helmet with armor values and pepperspray immunity. 
/:cl:


softcerv:
I am marking this as a fix because of these lines in the code:
```
//seems like it kinda SPACE helmet. So probably abusable, but not really.
//If you want to make it less abusable and remove from helmet/space to just /helmet/ - please, add code that removes hair on use. Because we weren't able to do that.
```

In addition, this item isn't meant to be of any mechanical use in the first place.

Gandalf2k15 (approved):

PR #19820, opened 2023-03-13
[TM-ONLY] Prevents the psyker pirate variant from rolling

GoldenAlpharex:
## About The Pull Request
Apparently there's some rather sizeable issues with Psyker pirates where they tend to crash every five minutes on average, and they can't properly control their ship, and overall they need constant admin supervision to work properly. I was asked to disable them

Being able to have antags that work without being babysat by admins just to be able to do the basics of their gimmick is good.

## Changelog

Not needed, it's TM-only.


Gandalf2k15:
these pirates were removed

PR #19842, opened 2023-03-14
[MIRROR] Fixes limbs not storing their external organs like they should be [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73951
--------------------
## About The Pull Request

That's about it, really. It was messing up the bodytypes, but nobody noticed because that's not exactly something /tg/ uses much. Downstream, it's another story, however.

## Why It's Good For The Game
Things going where they're meant to go is definitely a good thing, I think we can all agree on this.

## Changelog

:cl: GoldenAlpharex
fix: Fixes limbs not properly storing their external organs, resulting in issues when calculating the bodytypes.
/:cl:


GoldenAlpharex (approved):
This is some good stuff right there, some good stuff right there indeed.

PR #19845, opened 2023-03-14
[FIX] Arachne Sprite Proper Reshuffling

Nerev4r:
## About The Pull Request
Turns out we didn't use the proper sprites, or proper paths for some of the new Arachne models featured in the other PR that I'm too lazy to link because it's 3:33 AM. This puts the borgs in their proper paths, so there's no more confusion about color or model.

Less confusion is cool.

## Proof of Testing

<details>

image

</details>

## Changelog

:cl:
fix: The new Arachne line of borgs are now properly sorted, and the right sprites have been used.
/:cl:


GoldenAlpharex (changes requested):
I was the one that asked that the sprites on the right would be used, because while the ones with black outlines look better on their own, we don't use black outlines in the aesthetic of our sprites, and thus it looks jarringly out of place. I would like for that decision to be respected.

lessthnthree (approved):

Nerev4r:
> I was the one that asked that the sprites on the right would be used, because while the ones with black outlines look better on their own, we don't use black outlines in the aesthetic of our sprites, and thus it looks jarringly out of place. I would like for that decision to be respected.

yeah ok

PR #19871, opened 2023-03-15
[MIRROR] Replaces internal_organs with organs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73918
--------------------
Internal_organs now also contains external organs, so the naming was incorrect

Requested by @ tralezab in #72734

Also removed some now incorrect 'as anythings' that assumed everything in the internal_organs list was an internal_organ (which is a lie since I put extorgans in there which means runtimes and unintentionakl behaviour

:cl:
fix: fixes deadly harvesting just taking harmless extorgans
code: renames internal_organs to organs now that it can also contain external_organs
/:cl:


GoldenAlpharex (approved):
Alright, this should hopefully work just fine now.

PR #19886, opened 2023-03-15
Masterclass on why we wait for CI to pass - Makes the bolter wrench not cause every map CI to fail

Paxilmaniac:
## About The Pull Request

The bolter wrench was looking for its icon in our modular tool file, which we just so happen to not have a bolter wrench in! Easy fix, now there's a bolter wrench icon in there.
 

I need the green checkmark on my pull requests.

## Proof of Testing

If CI on this passes that's your proof of it working

## Changelog

This isn't player facing


GoldenAlpharex (approved):
Thank you.

PR #19899, opened 2023-03-16
[MODULAR] Fixes a modular glove's attackby not calling its parent

vinylspiders:
## About The Pull Request

Someone in the modular files overrode the attackby proc for _all the gloves_ and did not call the parent, where the crucial `COMSIG_PARENT_ATTACKBY` is sent from. 

Since the armor plating component waits for this `COMSIG_PARENT_ATTACKBY` signal, the armor never gets applied.

This could potentially fix any other issues caused by things that make use of that particular signal for gloves, because it was simply never being sent, ever, due to the improper modular override.

I also saw another glove item that has the same issue so I fixed that. There are probably more out there. `SHOULD_CALL_PARENT(TRUE)` should probably be added to this proc TG side to find them all and prevent issues like this in the future.

Edit: Moves the fix for all gloves to https://github.com/tgstation/tgstation/pull/74048. This now just fixes the one other modular glove override.

Fixes bug

## Changelog

:cl:
fix: fixes plate gloves not being able to be upgraded with goliath/polar bear hides
/:cl:

### DIFF
--- a/code/modules/clothing/gloves/_gloves.dm
+++ b/code/modules/clothing/gloves/_gloves.dm
@@ -55,7 +55,6 @@
 	return TRUE
 
 /obj/item/clothing/gloves/attackby(obj/item/tool, mob/user, params)
-	. = ..()
 	if(tool.tool_behaviour != TOOL_WIRECUTTER && !tool.get_sharpness())
 		return
 	if (!can_cut_with(tool))
--- a/code/modules/clothing/gloves/_gloves.dm
+++ b/code/modules/clothing/gloves/_gloves.dm
@@ -55,6 +55,7 @@
 	return TRUE
 
 /obj/item/clothing/gloves/attackby(obj/item/tool, mob/user, params)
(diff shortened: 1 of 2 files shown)


code/modules/clothing/gloves/_gloves.dm
```diff
@@ -55,6 +55,7 @@
 	return TRUE
 
 /obj/item/clothing/gloves/attackby(obj/item/tool, mob/user, params)
+	. = ..()
```
tf-4: non modular change

vinylspiders: oh you're right...what the heck. I was up late bugfixing, this is a TG issue then technically.

vinylspiders: See https://github.com/tgstation/tgstation/pull/74048. Changing this PR to be just for the other gloves then.

tf-4 (approved):

PR #17321, opened 2022-11-02
Adds polarized windows - one button privacy

Zergspower:
## About The Pull Request

Adds a reinforced window that can be toggled with a button to make it opaque, lending to an alternative to ugly curtains or spraypaint if you want options.

GIF 11-2-2022 12-43-35 PM

Todo:
Add an obj spawner
If possible make the glass non-transparent

## How This Contributes To The Skyrat Roleplay Experience

A good alternative to curtains in general, these can be used to divide areas in a less jarring manner instead of 10 curtains imbedded on top of windows for example.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
  
GIF 11-2-2022 12-43-35 PM

Mapping

image
image

</details>

## Changelog

:cl:
add: Adds a new reinforced window subtype - Polarized 
/:cl:

PR #19907, opened 2023-03-16
Polarized Windows: Component Edition

GoldenAlpharex:
## About The Pull Request
Basically, a total rewrite of https://github.com/Skyrat-SS13/Skyrat-tg/pull/17321, making it so it would work as a component rather than baked into certain window subtypes.

You can enable the polarization feature on any window subtype (windoors are not compatible, it wouldn't be hard to make it possible, but for now I decided against it) by slapping any tier of capacitor onto it. You can remove it by using a multitool on it, which gives you the capacitor back.

To toggle the opacity of the window, you need a `window polarization remote controller`, which is a subtype of the blast door controller assembly (which, fun fact, all other button-inserted controller devices are subtypes of too), and as such functions similarly. You can set the id of the controller by using it in-hand, or using a multitool on it (like other controllers of the sort), and then inserting a number from 0 to 1000. To link it with windows, you need to hit them with it, which allows you to link the id of the controller with the one on the polarization component.

You can print the window polarization remote controller from autolathes and from engineering techfabs at roundstart, with the same ingredients as the ones for a blast door controller (100 iron and 50 glass).

You can then either insert it in a button (which there's already a subtype for mappers, `/obj/machinery/button/polarizer`, where you can var-edit the `id` to whatever string you like, ideally not numbers, so that players can't replicate it in-round), or link it to other assemblies to activate it (like a remote signaler or a proximity sensor, or a temperature sensor, etc.). When it activates, it takes half a second to do an animation to turn it from the color it was at, to black, or the other way around. That also ensures that the polarization won't just remove any color from the window, as it'll return it back to whatever color it was beforehand. Do note that if you color them after tinting them, that's on you, and that color will be lost, most likely. Didn't test that, to be honest.

### For mappers:
- `/obj/machinery/button/polarizer` - Set `id` to the id you want to control with that button.
- `/obj/effect/spawner/structure/window` - This type, and all of its subtypes, now have a `polarizer_id` subtype so that you can easily link them with the `id` of a button. Putting any string in that variable will make it spawn with the polarization component attached, so it can be polarized, without you needing to do anything else.
- `/obj/structure/window` - This type, and all of its subtypes, now have a `polarizer_id_on_spawn` subtype so that you can easily link them with the `id` of a button. Putting any string in that variable will make it spawn with the polarization component attached, so it can be polarized, without you needing to do anything else. *This variable only has an effect when the window is spawned, and won't do anything if VV'd in-round.* This variable is only when you're mapping in windows directly, rather than the spawner described above.

A nicer alternative to curtains, that also makes more sense visually than the curtains over windows do. It also gives more control than the regular tinted windows, which I think mappers will enjoy, on top of players.

## Proof of Testing

<details>
  
Polarization inactive.
image

Polarization active.
image

The look of the little controller remote itself.
image

</details>

## Changelog

:cl: GoldenAlpharex
add: Added an option to polarize windows, so that they can be toggled between tinted and clear at will in-round. To do so, simply slap a capacitor on any kind of window, except windoors.
add: To control the polarized windows, you can use a window polarization remote controller, which can be printed from any autolathe and from engineering techfabs at roundstart. Use it in-hand to set its channel ID, and then slap it against windows with polarization controllers installed to link them to it. Place it in a button, or put it together with other assemblies, in order to have a way to activate it, and toggle all connected windows at once.
add: Added support for mappers to create polarized windows with custom IDs, look at https://github.com/Skyrat-SS13/Skyrat-tg/pull/19907 for more details on how to use them.
/:cl:


Zergspower:
Thanks again, i simply had no path forward once i hit my code knowledge on the build portion hah

modular_skyrat/modules/polarized_windows/capacitor.dm
```diff
@@ -0,0 +1,27 @@
...
+	SIGNAL_HANDLER
+
+	if(!istype(target))
+		return
+
+	. = COMPONENT_CANCEL_ATTACK_CHAIN // Just to reduce the unnecessary repetition at every early return.
+
+	var/datum/component/polarization_controller/window_polarization_controller = target.GetComponent(/datum/component/polarization_controller)
+
+	if(window_polarization_controller)
```
softcerv, line 21: Would it be possible to do this from the if statement without assigning it to a variable. I don't see the variable used anywhere else inside of the proc.

GoldenAlpharex: I guess it would be, yeah.

lessthnthree (approved):

Gandalf2k15 (approved):

PR #19924, opened 2023-03-17
Makes the Medibeam 100k

StrangeWeirdKitten:
## About The Pull Request

This is very simple. It drops the medibeam cost from 1 million to 100 thousand.

I don't think this had much thought put behind it. It is not easy, or typically too much of a bar to hit a 1 million cargo budget. Especially without abusing bugs. Even then, it's far cheaper and easier to print the beam that goes on a mech. The most I've seen cargo have in total was 450 thousand and that's from atmospherics mass selling anti-nobelium and botany abusing bounties. Even if cargo hits 1 million, I don't think they'd buy something that can be placed on a mech. 

This was clearly added to replace the cargo bike in the "something overpowered that no one can afford" range. The medibeam is not overpowered. It's just a nice luxury, and it should be obtainable.

## Proof of Testing

image

## Changelog

:cl: StrangeWeirdKitten
balance: reduces the cost of the medibeam to 100k
/:cl:

### DIFF
--- a/modular_skyrat/modules/cargo/code/packs.dm
+++ b/modular_skyrat/modules/cargo/code/packs.dm
@@ -265,7 +265,7 @@
 /datum/supply_pack/misc/medibeam //Moved to Misc so Medical can't order them for free through department consoles and break the economy.
 	name = "Medical Beam Gun"
 	desc = "Nanotrasen offers you, for an exorbatant fee, the ability to lease one of their ERTs favorite gadgets, the Medical Beam Gun"
-	cost = 1000000 //Special case, we don't want to make this in terms of crates because having bikes be a million credits is the whole meme.
+	cost = 100000 //Special case, we don't want to make this in terms of crates because having bikes be a million credits is the whole meme.
 	contains = list(/obj/item/gun/medbeam)
 	crate_name = "medical beamgun crate"
 


Zonespace27:
👎 The medibeam is pretty powerful (nevermind getting two and exploding 1/2 of cargo inexplicably but that's beside the point), and its purpose is the "streeeeeeeetch" goal that also serves as a dogwhistle for "something's completely broken" when people see this thing being purchased

Elli-Skala:
The bike use to be the thing that costs 1,000,000 to alert coders cargo broke but that became a meme so every shift people would break cargo to get the bike(s) on skyrat. It the medi beam was never intetned to actually be gotten

StrangeWeirdKitten:
> 👎 The medibeam is pretty powerful (nevermind getting two and exploding 1/2 of cargo inexplicably but that's beside the point), and its purpose is the "streeeeeeeetch" goal that also serves as a dogwhistle for "something's completely broken" when people see this thing being purchased

It's still one of the most expensive items on the cargo shuttle. I don't think we need an item that shows something broken when you can look inside the cargo budget and realize something's up. Especially with everyone being able to see what's being sold. Plus 200k is pretty expensive to blow cargo up. If you cross the beams, with a drained budget, then I suppose that's the way god intended.

Medibeams have been placed on mechs all the time with the same power, and for a lot cheaper. I think spending 100k is pretty reasonable for a handheld one. Code wise, it's the same exact thing. Just on a mech.

Doing some testing, and looking at the code. It changes health by all values by -4. It does not stop bleeding, nor heals wounds, and will not stop someone from dying of blood loss. It may get someone out of crit, but it's a highly expensive tool. 100k for the amount of healing power seems fairly reasonable.

tbf, no QM will dump 1 million on a medibeam. If you want a cockatoo to send into the mines, we can have a much better item than a medibeam.  The bike was actually worth the 1 million, but this isn't the point of this PR nor should that ever return.

HoodMilk:
Medical doesn't see too much action here so it doesn't seem too bad having it cheaper. Even in emergency situations like blob healing chemicals are better. You can get them on the mech after like 30 minutes into the round and they still aren't really used.

YakumoChen:
The medibeam and the bike (I'm pretty sure you can still purchase the bicycle) are not meant to be purchased, they're intended as, as Zonespace said, dogwhistles that there exists a cargo exploit that allows these to be purchased.

The medibeam is powerful and you can already buy a lower-power version of it through imports.

If you want medibeams in game, get an Oddy mech

Elli-Skala:
> The medibeam and the bike (I'm pretty sure you can still purchase the bicycle) are not meant to be purchased, they're intended as, as Zonespace said, dogwhistles that there exists a cargo exploit that allows these to be purchased.
> 
> The medibeam is powerful and you can already buy a lower-power version of it through imports.
> 
> If you want medibeams in game, get an Oddy mech

The medi beam was added to replace the bike on skyrat

YakumoChen:
> The medi beam was added to replace the bike on skyrat

it didn't though, the bike crate is still there

Grunnyyy:
The medibeam was added to Cargo to give Cargo more items of use. It was cheaper and then got its price adjusted to 100k, to not make it too cheap for being fairly strong. Someone else came later and increased its price to bike range. Not sure why, but it's original intention with that has been pretty much I validated. A better mvoe would be moving it to T3 Imports, as they are already plenty expensive.

YakumoChen:
>The medibeam is powerful and you can already buy a lower-power version of it through imports.
> If you want medibeams in game, get an Oddy mech

Elli-Skala:
> The medibeam was added to Cargo to give Cargo more items of use. It was cheaper and then got its price adjusted to 100k, to not make it too cheap for being fairly strong. Someone else came later and increased its price to bike range. Not sure why, but it's original intention with that has been pretty much I validated. A better mvoe would be moving it to T3 Imports, as they are already plenty expensive.

#6933 made it the price of the bike because the bike was at the time seen as to over powered due to every other shift it being bought from cargo (due to broken skyrat features) 

The real "bug" here is that the bike was readied to skyrat Cargo likely due to a refactor

YakumoChen:
Would rather have the dogwhistle item be a bike, because its very obvious when someone buys it and by its nature it isn't supposed to be appearing in roundw

The medibeam is not visually distinct on skyrat and can be brought by ERTs anyway, making it a poor choice to signal something is very wrong in cargo

PR #20030, opened 2023-03-23
[SEMI-MODULAR] Latejoin janitors can get access keys, two!

lessthnthree:
## About The Pull Request

Guarantees three sets of janitor access keys on the station given we have an increased number of janitors, and not always a roundstart janitor.

A key ring may end up unreachable for 'reasons.' Sometimes no key ring at all! You can also go find the key rings if they're lost or stolen (and beat up who stole them.)

## Changelog

:cl: LT3
code: Latejoin janitors now also get access keys, to a maximum of 3
/:cl:


modular_skyrat/master_files/code/modules/jobs/job_types/janitor.dm
```diff
@@ -0,0 +1,13 @@
+#define MAX_ACCESS_KEYS 2
+
+/datum/outfit/job/janitor/pre_equip(mob/living/carbon/human/human_equipper, visuals_only)
+	. = ..()
+
+	var/access_keys = 0
+	for(var/obj/item/jani_gear in GLOB.janitor_devices)
+		if(istype(jani_gear, /obj/item/access_key))
```
Zonespace27:
```suggestion
	for(var/obj/item/access_key/jani_gear in GLOB.janitor_devices)
```
Will need to fix line 9 indentation though

Zonespace27 (approved):

PR #20061, opened 2023-03-25
[MODULAR] Removes pillows from Grimoire Purpura

softcerv:
## About The Pull Request
Does as the title says.

The pillow structures generated by the pillows can be used to farm cloth. Until pillows are refactored, it's best that we have them removed from Grimoire Purpura to prevent abuse.

## Proof of Testing

<details>
 
image

</details>

## Changelog

:cl:
fix: removes pillows from Grimoire Purpura, preventing infinite free cloth.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_implants/code/nifsofts/dorms.dm
+++ b/modular_skyrat/modules/modular_implants/code/nifsofts/dorms.dm
@@ -27,7 +27,6 @@
 					/obj/item/spanking_pad,
 					/obj/item/clothing/sextoy/vibrator,
 					/obj/item/restraints/handcuffs/lewd,
-					/obj/item/fancy_pillow,
 
 					// Head / Mask /Neck Items
 					/obj/item/clothing/mask/ballgag,


Zonespace27 (approved):

YakumoChen:
Is it possible to just make a child of the pillow with the deconstructing action (wirecutters?) just doing an early return?

softcerv:
> Is it possible to just make a child of the pillow with the deconstructing action (wirecutters?) just doing an early return?

Even if I did that, I would need to update the code for the prop summoner to account for structures, which is something that I'd rather not deal with. Especially since with the pillow structures, you can put multiple pillows on a pile.

YakumoChen:
Make the pillow structure unable to use the grimoire pillows then

softcerv:
> Make the pillow structure unable to use the grimoire pillows then

More than that would need to be done. I don't feel like adding more onto the organic interface code for something like this to work. If you wish to handle it in a future PR, you are more than welcome to.

tf-4 (approved):

PR #20102, opened 2023-03-26
[MIRROR] Removes bepis Techweb and reworks tech disks in general [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/74147
--------------------
## About The Pull Request

Did you know that there's 2 types of disks Science can print?
One of them creates a whole new techweb and is used to transfer data to and from techwebs (which used to be for Theft objective, but nowadays all it's used for is Admin techweb).
The other one is these disks you can find from Lavaland & for the Limbgrower which holds nodes on them directly, they don't need anything else.

These are both printable and have very similar names which could easily confuse people, especially since they are both printed and used at the same place, being the R&D room and R&D console.

This will hopefully simplify it, by removing the base Tech disks from being printable. The only one that can be printed now is for Limbgrowers, which can't be easily mixed with the other type of Disk. Outside of that, Lavaland disks are staying the same, but I've moved Bepis disks to use this, which allows us to remove Bepis techwebs being made every single time a new bepis disk is created.

Examples of it in-game
image
image
image

## Why It's Good For The Game

There's no reason why every single Bepis node should be making a whole new techweb set with experiments to complete, roundstart nodes researched, being constantly checked if it should gain research points, have a list of papers to publish, and everything else that techwebs do.

Can you guess which disk does what in this screenshot?
image

## Changelog

:cl:
balance: You can no longer print Technology data disks. You can still print Tech disks, which hold techweb information on it, just not the one that holds up to 5 nodes.
balance: ^ Because of this, there's no way to download nodes from an RD console and upload them to an Autolathe to bypass departmental restrictions, you have to go through a Techfab/Circuit imprinter for your needs.
balance: Ones that are found cannot have anything uploaded/deleted off of them either, you can only upload them to the Web.
code: Every individual Bepis disk no longer create an entire techweb
/:cl:


ORCACommander:
ping me when merged, this is breaking Compile Maps CI downstream

PR #20115, opened 2023-03-27
Fix elevator panel/floor indicators

lessthnthree:
## About The Pull Request

Fixes the calculation for what floor elevators are on, since our Z-levels differ from TG.

Elevator displays correct floor, pulls correct data from the map file.

## Changelog

:cl: LT3
fix: Elevator indicator and panel show correct floor
/:cl:


Paxilmaniac (changes requested):
Didn't I already go over with you what the actual solution to this should be? If we loaded skyrat centcom /after/ the station then we wouldn't need to non-modularly edit elevator panel code.

lessthnthree:
Until someone answers _how_ to implement your desired solution, I reopened the one that's tested to work now that it needs less edits.

Gandalf2k15:
In this case I'm going to opt to keep the broken elevators as an incentive for someone to fix the foundational problem of z-level loading offset. This would just somewhat mask the actual problem.

PR #20141, opened 2023-03-28
removes the decals making tramstation arrivals fail linters

Paxilmaniac:
## About The Pull Request

in a recent pr to do with changing linters, it wasn't taken into account that our own maps might not be passing these linters, which has caused a few problems

linters should probably be passing

## Proof of Testing

trust me bro

## Changelog

do players really need to see this


tf-4 (approved):
yep thats a maps

tf-4:
it looks kind of goofy but not goofy enough for me to care more about it than I care about the green ticks of passing checks

Paxilmaniac:
> it looks kind of goofy but not goofy enough for me to care more about it than I care about the green ticks of passing checks

its identical to how tramstation arrivals on the base map looks so what it do

PR #20180, opened 2023-03-30
[MODULAR] Science Prolathe can print Hand Drills

softcerv:
## About The Pull Request
Because of https://github.com/Skyrat-SS13/Skyrat-tg/pull/20102, science is no longer able to print Hand Drills.This PR adds the ability for science to print Hand Drills again.

I will most likely end up making more PRs to address the things that science lost because of this.

Experiments are already a pain to deal with, having the hand drill made a lot of the experiments less of a pain.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
balance: the science prolathe can now print hand drills
/:cl:


tf-4:
not really a huge fan of this for the same reasons I gave about the suggestion of allowing plasma cutters, but I care less about this one to the point I'm not going to block it.
Still not in favour though.

softcerv:
> not really a huge fan of this for the same reasons I gave about the suggestion of allowing plasma cutters, but I care less about this one to the point I'm not going to block it. Still not in favour though.

While I can understand your reasoning, I feel like having the hand drill makes the experiments that science has to handle far less grueling.

burgerenergy:
If science is waddling out to use a hand drill then it means the department they'd interact with in this case has failed them, anyway.

Gandalf2k15:
Giving science even more autonomy is not something that we want to start aiming for, interdepartmental interactions provide more roleplay, so I'm going to say no to this one.

softcerv:
While I do understand your choice, from what I have experienced on Skyrat, I don't think that the interactions generated by the lack of hand drills from science will end up being meaningful.

PR #20212, opened 2023-03-31
[FIX] Removes Duplicate NRI Belt Sprites

Nerev4r:
## About The Pull Request
image
There are many things that need to be erased. I forgot about these in my last fix.

It sure as fuck doesn't, but these were bugging stuff out for the longest; and were misplaced to boot.

## Proof of Testing

<details>

image

</details>


Stalkeros2:
I kid you not I've removed those like three or so times at this point and they keep reappearing.

Zonespace27 (approved):

PR #20213, opened 2023-03-31
[FIX] Snails No Longer Slip

Nerev4r:
## About The Pull Request
This was erroneously taken out in #19874, I'm 99%. Every other source of noslipall was refactored, this one was forgotten; given snails on Skyrat are slightly more important than their rare cousins on /tg/, and the trait was there before and taken out without mention, I think this qualifies as a fix.

It doesn't quite make any sense for them to in the first place.

## Proof of Testing
<details>

image

</details>

## Changelog
:cl:
fix: Snails don't slip anymore.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/snail.dm
+++ b/code/modules/mob/living/carbon/human/species_types/snail.dm
@@ -15,6 +15,7 @@
 		FACEHAIR
 	)
 	inherent_traits = list(
+		TRAIT_NO_SLIP_ALL, //SKYRAT EDIT - This Was Mistakenly Taken Out, I Think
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species_types/snail.dm
```diff
@@ -15,6 +15,7 @@
 		FACEHAIR
 	)
 	inherent_traits = list(
+		TRAIT_NO_SLIP_ALL, //SKYRAT EDIT - This Was Mistakenly Taken Out, I Think
```
Fikou: this isnt a skyrat edit, it exists upstream

Nerev4r: true

PR #19039, opened 2023-01-30
[MODULAR] [MDB IGNORE] Dim Lighting Some More, Removes Egregious Lighting Varedits in Interlink, Dynamically Calculate Night Shift Light Brightness and Color

RimiNosha:
## About The Pull Request
Holy *fuck*, why and how did that get past maptainers?

Anyways, onto the actual topic:
Dimms lights more. Yup. It's been kinda agreed on that we just aren't particularly frontier-like, and this is part of the problem.

Also **severely** kneecaps door lights. Dear god, they're brighter than some of the dim lights.

Also made it so night shift lights automatically get their colors adjusted programmatically, cause I didn't feel like making a custom nightshift light value for every light. Pretty performant cause it uses nothing but simple multiplication and bit operations, with probably the most expensive part being to convert hex to number and back.

## How This Contributes To The Skyrat Roleplay Experience
Supposed to be frontier, and all that cool shit.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>

## Before
image

## Interlink
image

## Station
image
image
image
image
image

## Station Night Shift
image
image
image
image
image

</details>

## Changelog

:cl:
qol: Lights have been dimmed some more.
qol: Doors are no longer the next brightest light after tube lights.
qol: Night shift is darker, and is now dynamically applied to lights.
fix: Fixed just about all randomly brighter night shift lights.
/:cl:

PR #20222, opened 2023-03-31
Revert default light tube color

lessthnthree:
## About The Pull Request

Reverts an undocumented change to the default light tube color made in https://github.com/Skyrat-SS13/Skyrat-tg/pull/19039. We have mapping helpers such as /obj/machinery/light/warm if you want to use different bulb colors.

Light colors are as you'd expect when placing or building them.

## Proof of Testing

<details>
 Before:

image

After:

image

</details>

## Changelog

🆑 LT3
del: Reverts an undocumented change to the default light tube color
/🆑

### DIFF
--- a/modular_skyrat/master_files/code/modules/power/lighting/light.dm
+++ b/modular_skyrat/master_files/code/modules/power/lighting/light.dm
@@ -1,5 +1,4 @@
 // Kneecapping lights.
 /obj/machinery/light
-	bulb_colour = "#ffe6cd"
 	bulb_power = 0.5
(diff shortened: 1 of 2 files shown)


RatFromTheJungle:
the latter looks way-way to bright, almost back to tg brightness which is 'not ideal'

Zonespace27:
👎 Due to how lighting works (both in-game and in real life), color shifting is necessary to make a moodier lighting
(also the old just looks plain better)

lessthnthree:
We can get moodier lighting without having warm lights and making them even warmer on night shift, but then it will be a different PR.

PR #20257, opened 2023-04-01
[FINALLY HERE] Clock Cultist OPFOR bundle

Zonespace27:
## About The Pull Request

**Every part of this has already been reviewed and thoroughly tested.**

This PR (re)adds Clock Cult in a form, specifically as an OPFOR antagonist[^1].

[Design Doc](https://hackmd.io/@mRAdleXgRfmKqh97O8ixSA/H1Xfbbue3)

## General Idea
Clock cultists gained through OPFOR start alone with naught more than a **Clockwork Slab**[^2], **Whirring Convergence**[^3], **Slab Recall**[^4]. At the beginning, you're weak enough to not even have real equipment. So, you need to use one of your few available scriptures, **Integration Cog**, to gain power. Simply insert into an APC (use a screwdriver or the cog itself), then close it and it'll work! Cogged APCs generate 1 **Cog**[^5] on insertion, in addition to passively draining APC power to convert into **Clockwork Power**[^6]. Then, you can unlock things like scriptures and structures, in order to improve your defensive capabilities. There's tons more than just this, but it'd take me a good while to write out every detail about everything.

[^1]: I haven't added it in this PR because it's related to a TM
[^2]: A channel for Clock Cultists to cast spells, called "scriptures". You must have it in-hand to cast any scriptures.
[^3]: A communication spell for clock cultists and their constructs.
[^4]: A form of recall spell that only works with Clockwork Slabs.
[^5]: The currency used to unlock new scriptures. Removing a cog from an APC doesn't remove the cog, but re-inserting a cog won't get you a new one.
[^6]: The more common currency. Used to cast scriptures and to power structures.

A new OPFOR antagonist with depth.

## Screenshots
<details>

image

image

image

</details>

## Changelog

:cl:
add: Clock Cult exists now! Become one in the "Antagonists" tab of the OPFOR equipment menu.
/:cl:


Gandalf2k15 (approved):
Providing this passes CI, it is LGTM as all individual project pieces were reviewed too.

PR #20264, opened 2023-04-02
Engineering Event - Wall Eating Mushrooms!

Gandalf2k15:
## About The Pull Request

In our strive to be ever more roleplay focused, I'm going to be adding more non-combat-orientated random events.

This event simply adds a new wall interaction, wall fungus. This fungus is not very damaging at first, but if left to spread out of control, it can become a problem!

To deal with wall fungus you simply weld the affected wall. 

If you don't get it in time the wall will fall apart.

Also adds in a wall mushroom plant which can be harvested, grown, eaten and planted on walls!

WALL MUSHROOMS
image

Growable!
image

:cl:
add: Wall fungus: A new engineering event that requires minimal involvement, watch out for pesky wall fungus!
add: Wall-eating mushrooms for hydroponics!
/:cl:

there is a fungus amongus


modular_skyrat/modules/wall_fungus/code/wall_fungus_component.dm
```diff
@@ -0,0 +1,141 @@
...
+		progression_step_amount = override_progression_step_amount
+	if(override_spread_chance)
+		spread_chance = override_progression_step_amount
+	if(override_spread_distance)
+		spread_distance = override_spread_distance
+	if(override_drop_chance)
+		drop_chance = override_drop_chance
+
+	if(!iswallturf(parent))
+		return COMPONENT_INCOMPATIBLE
```
softcerv: You might want to move it to the start of the Initialization, that way it doesn't do anything unnecessary before returning.

softcerv:
I'm wondering if at higher levels of infestation, if it might make sense for the mushrooms to leak hallucinogen or some other kind of chemical/gas into the environment around themselves.

modular_skyrat/modules/wall_fungus/code/wall_fungus_component.dm
Gandalf2k15: True

HoodMilk:
Finally the station can have weird unidentifiable mold on the walls like a crackden.

Elli-Skala:
Will this be in the lavaland/random weed fungus pool?

YakumoChen:
The sprites are really ugly and overlay way too much. Would really benefit from a lot of transparency and maybe emphasis on creating cracks in the wall.

Gandalf2k15:
> Will this be in the lavaland/random weed fungus pool?

no pe

ghost:
How hard is it to reach the corners?
One of the biggest issues with walltile-based stuff right now (see: rust) is the unreachable corners between walls where you either have to go around a whole room to reach, or in the case of 4-tile connecting "+" walls just entirely dismantle a wall to reach it.
Is there a way you could make the corner reachable by users or unreachable by the mushrooms?

QuacksQ:
Burning these off the walls does not work without having to weld the wall down to the girder

Welding walls non stop is also not a fun game mechanic for engineering Imo, its nearly identical to fixing emagged doors a ninja left behind.

burgerenergy:
I, random player, am not a fan I have to admit!

The first few times it appeared it was relatively unknown, so it did massive damage. Already awareness is up, so the event prop becomes a scramble through maintenance to find it not unlike the blob or mold, but instead of a hectic fight you... burn things off walls. Often times they're in awkward corners or between points, needing walls to be deconstructed and/or space walks taken to get to them. It's all pretty tedious.

Given it's spicy, I would recommend folding fungi (and maybe mold, for that matter) into the blob/biohaz-7 call, which also makes it more of a mystery what level of spice you're going to get, as well as make it clear it's a serious threat, just not an directly lethal one like the other two.

Gandalf2k15:
> Burning these off the walls does not work without having to weld the wall down to the girder
> 
> Welding walls non stop is also not a fun game mechanic for engineering Imo, its nearly identical to fixing emagged doors a ninja left behind.

This was a bug originally, and they were very overtuned. Now they're extremely slow to spread, and can be dealt with whenever, it's just one more thing that can go wrong on the station.

QuacksQ:
> > Burning these off the walls does not work without having to weld the wall down to the girder
> > Welding walls non stop is also not a fun game mechanic for engineering Imo, its nearly identical to fixing emagged doors a ninja left behind.
> 
> This was a bug originally, and they were very overtuned. Now they're extremely slow to spread, and can be dealt with whenever, it's just one more thing that can go wrong on the station.

based, if thats the case then its good.

Gandalf2k15:
ok

Iamgoofball:
this should go upstream

PR #20322, opened 2023-04-04
Disables hulk + A way to disable mutations

Gandalf2k15:
## About The Pull Request
Removes from the DNA console hulk. This mutation has not really seen much use outside of valid hunting, which is against server rules.

Also adds a way to disable mutations from the DNA console without outright removing them.

## Changelog

:cl:
balance: Hulk has been removed from the possible mutations.
/:cl:

### DIFF
--- a/code/game/machinery/computer/dna_console.dm
+++ b/code/game/machinery/computer/dna_console.dm
@@ -708,6 +708,11 @@
 				var/datum/mutation/human/matched_mutation = null
 				//Go through all sequences for matching gene, and set the mutation
 				for (var/M in subtypesof(/datum/mutation/human))
+					// SKYRAT EDIT ADDITION
+					var/datum/mutation/human/iterating_mutation = M
+					if(iterating_mutation.disabled)
+						continue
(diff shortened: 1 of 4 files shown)


dawsonkeyes:
I don’t see why Hulk ought to be removed if validhunting is already against the rules

Zonespace27:
I disagree, not with the change itself[^1], but with the concept of "throwing the baby out with the bathwater", simply removing something because it *can* be used to validhunt is poor justification, in my eyes.

[^1]: I don't care enough about hulk to argue for keeping it that much

softcerv:
Even if I think the reasoning is questionable, considering that most of the server's population wasn't able to use this because this was human only, I don't consider this to be a great loss by any means.

GoldenAlpharex (approved):
Hulk was always hot garbage, it's never fun to deal with, and only had to be severely baby-proofed so people couldn't just easily grief with it. Honestly? Good freaking riddance.

Iamgoofball:
hulk is in it's weakest form it's ever been, and you guys still removed it?

PR #20342, opened 2023-04-05
[MODULAR] Emissives fix-Skyrat edition. Emissives will no longer layer over worn clothing.

vinylspiders:
## About The Pull Request

Sorry this took a while- I didn't realize it was borked on our end still.

I fixed the emissive blocker issue upstream but Skyrat still needed a tweak as upstream does not have body emissives. 
No more emissives layering over clothing.

There is still one more emissives related issue that exists upstream and that is the one where the lying down emissives for clothing fail to transform. That is on my list of things to fix eventually on TG's side.

Hooray they're back!

dreamseeker_aqXAMs1Of9

## Proof of Testing

See above.

## Changelog

:cl:
fix: body emissives will no longer layer over worn clothing
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/clothing/clothing.dm
+++ b/modular_skyrat/master_files/code/modules/clothing/clothing.dm
@@ -1,6 +1,6 @@
 GLOBAL_LIST_EMPTY(taur_clothing_icons)
 
-/obj/item/clothing/
+/obj/item/clothing
 	blocks_emissive = EMISSIVE_BLOCK_UNIQUE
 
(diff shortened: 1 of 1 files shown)


modular_skyrat/master_files/code/modules/clothing/clothing.dm
```diff
@@ -1,5 +1,8 @@
 GLOBAL_LIST_EMPTY(taur_clothing_icons)
 
+/obj/item/clothing/
```
Zonespace27:
```suggestion
/obj/item/clothing
```

Gandalf2k15 (approved):

PR #20355, opened 2023-04-05
Unapologetic Hemophage Buffs Vol. 1

Nerev4r:
## About The Pull Request
Special thanks to @Hatterhat and @vinylspiders.
This does three things.
Satiated Thirst now lasts 30 minutes.
Brute regen upped to a whole 1, up from 0.75. All other values the same; burns should indeed take a bit longer.
Sucking blood from someone with a ckey (gotta actually drain them, this doesn't apply to blood in a cup) now gives you double the blood for the same amount drained from the victim (so that we don't kill them in two slurps.)
From literal spawning in at roundstart (meaning these numbers are subject to 'you often have less than this',)
558 > 608 > 657 > 706 > 754 NORMAL
559 > 658 >  758 > 857 > 957 CKEY

I was told last night that playing a Hemophage is moderately annoying from the constant resource management, and a lot of people have either stopped playing it or switched to being a 'hemophage in custom species name only.' The amount of time spent properly roleplaying out a feed does not necessarily equate to the amount of time the blood really lasts, and healing in my experience is a bit lackluster, being so slow for something that actually requires resources that you have to essentially harm other players for. I would like for draining someone to be worthwhile, but we have to give the carrot, not the stick.

## Proof of Testing
<details>

image
  
image

</details>

## Changelog
:cl:
balance: Hemophages now get 1.5x the blood when bite-draining people with ckeys
balance: The Satiated Thirst effect now lasts 30 minutes, up from 20.
/:cl:


Hatterhat:
note: my only contribution to this pr was the addition to line 558 about blood returns from feed

specifically the new conditional

`        hemophage.blood_volume = clamp(hemophage.blood_volume + (drained_blood * (is_target_human_with_client ? 2 : 1)), 0, BLOOD_VOLUME_MAXIMUM)`

Nerev4r:
I made one more buff, their stomachs now have a really good metabolism efficiency to help with drinking blood in liquid form. Let me know if this is too much, we can dial it back to a T2 level (0.07) if so.

vinylspiders:
> I made one more buff, their stomachs now have a really good metabolism efficiency to help with drinking blood in liquid form. Let me know if this is too much, we can dial it back to a T2 level (0.07) if so.


What you could do instead of increasing the metabolism rate for all reagents, is to override `on_life` add a type check for blood. 

See skeleton stomachs for a template. 

https://github.com/Skyrat-SS13/Skyrat-tg/blob/e4b4d4d3c081a9a609527e85a8a858ad4508c151/code/modules/surgery/organs/stomach/_stomach.dm#L279-L291

That way it wouldn't affect any reagents besides blood, and you can do whatever you want with it from there.

Nerev4r:
> > I made one more buff, their stomachs now have a really good metabolism efficiency to help with drinking blood in liquid form. Let me know if this is too much, we can dial it back to a T2 level (0.07) if so.
> 
> What you could do instead of increasing the metabolism rate for all reagents, is to override `on_life` add a type check for blood.
> 
> See skeleton stomachs for a template.
> 
> https://github.com/Skyrat-SS13/Skyrat-tg/blob/e4b4d4d3c081a9a609527e85a8a858ad4508c151/code/modules/surgery/organs/stomach/_stomach.dm#L279-L291
> 
> That way it wouldn't affect any reagents besides blood, and you can do whatever you want with it from there.

True. Might look into this later.

Nerev4r:
Hemophages now only specifically process blood really well when they drink it.

tf-4:
1. GBP
2. I am very unconvinced that hemophages need any kind of buff, let alone more free healing and gaining more blood than they drain which doesn't really make sense.

Closing when this reaches 24h open.

Nerev4r:
> 1. GBP
> 2. I am very unconvinced that hemophages need any kind of buff, let alone more free healing and gaining more blood than they drain which doesn't really make sense.
> 
> Closing when this reaches 24h open.

1) I'm already working on bugfixes and some of my newly-gained GBP got wiped in the branch being deleted.
2) I can walk back the buff on the healing, but it has been demonstrably aggravating for people to deal with the resource management.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage.dm
```diff
@@ -472,6 +473,15 @@
 	body.vomit(lost_nutrition = 0, stun = FALSE, distance = 1, force = TRUE, purge_ratio = HEMOPHAGE_VOMIT_PURGE_RATIO)
 	return ..()
 
+/obj/item/organ/internal/stomach/hemophage/on_life(delta_time, times_fired)
+	var/datum/reagent/blood/blood = reagents.get_reagent(/datum/reagent/blood)
+	if(blood)
+		. = ...() // yes this is cursed, but we don't want to call /obj/item/organ/internal/stomach 's on_life here
```
GoldenAlpharex: I never knew this worked, that's really funny.

GoldenAlpharex: That being said, isn't blood actually instant when you drink it? I thought it was.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage.dm
```diff
@@ -555,7 +565,7 @@
 	var/drained_blood = min(victim.blood_volume, HEMOPHAGE_DRAIN_AMOUNT, blood_volume_difference)
 
 	victim.blood_volume = clamp(victim.blood_volume - drained_blood, 0, BLOOD_VOLUME_MAXIMUM)
-	hemophage.blood_volume = clamp(hemophage.blood_volume + drained_blood, 0, BLOOD_VOLUME_MAXIMUM)
+	hemophage.blood_volume = clamp(hemophage.blood_volume + (drained_blood * (is_target_human_with_client ? 2 : 1)), 0, BLOOD_VOLUME_MAXIMUM)
```
GoldenAlpharex: I feel like the ternary you added could be in its own variable, and using a define to say how much of a modifier it is.

GoldenAlpharex (approved):
Overall, I'm for this PR. I probably would've done something similar had I had time on my hands in the recent months.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage.dm
vinylspiders: Ah yeah about that...I was helping with this-- the issue was that when you ingest blood, it gets processed by the stomach first which does this:

https://github.com/Skyrat-SS13/Skyrat-tg/blob/e4b4d4d3c081a9a609527e85a8a858ad4508c151/code/modules/surgery/organs/stomach/_stomach.dm#L74-L82

tf-4: It's pretty bad and it almost always is the wrong way to go about something. Why don't you want to call parent here?

Nerev4r: We don't want to call the parent stomach because otherwise organ failure and such won't happen if it does; @vinylspiders worked w/ me on this bit.

vinylspiders: You can also do `....()` and so on technically. But if you feel the need to use more than three dots at that point you should probably consider refactoring everything. : ' )

vinylspiders:
> > 1. GBP
> > 2. I am very unconvinced that hemophages need any kind of buff, let alone more free healing and gaining more blood than they drain which doesn't really make sense.
> > 
> > Closing when this reaches 24h open.
> 
>     1. I'm already working on bugfixes and some of my newly-gained GBP got wiped in the branch being deleted.
> 
>     2. I can walk back the buff on the healing, but it has been demonstrably aggravating for people to deal with the resource management.

Mmm, if anything, you could most certainly label this as a fix PR. Perhaps some of the buffs can be scaled down a bit if they are deemed excessive BUT this fixes a legitimate issue that was preventing consumed items containing blood from transferring blood in some cases.

> 1. GBP
> 
>     2. I am very unconvinced that hemophages need any kind of buff, let alone more free healing and gaining more blood than they drain which doesn't really make sense.
> 
> 
> Closing when this reaches 24h open.

While the buffs are up for debate, this can be considered a fix PR.

<details>

<summary>From eating blood noodles without the stomach code change</summary>

image

</details>

Blood does not get transferred to the mob at all, which doesn't seem to be the intent with hemophage food and probably wasn't always the case until TG made some changes. Edit: okay, to clarify, _normally_ it does transfer but not always. This just guarantees that it will happen, and happen faster than usual.

Nerev4r:
ty @Hatterhat 
Jackpot Devil May Cry GIF - Jackpot Devil May Cry Gun - Discover   Share GIFs

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage.dm
vinylspiders: Yeah, on second thought I think removing it is the right call. It's not super important to call the parent in this case, missing a tick or two of organ damage while blood is in your stomach is not a big deal and even feels on brand for hemophages tbh.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage.dm
Nerev4r: that last commit should do the trick

tf-4:
The fixes can be atomised out in that case

Nerev4r:
> The fixes can be atomised out in that case

It's a side effect of the 'stomach metabolism' thing. I can if it's really needed.

tf-4:
Still not even slightly convinced that hemophages need any buffs and the fix was atomised out

GoldenAlpharex:
I do think that the healing buff was unnecessary, but the changes to blood intake from other people was also something that's been on my radar for a bit. While I think it's not necessarily the _best_ solution to that problem, I couldn't really give you a better one either, and I've been thinking about it for a little while now.

For reference, this isn't so much of a buff as it is simply making the blood draining become actually relevant, instead of being unable to compare to other means of increasing your blood volume.

GoldenAlpharex (approved):
All good with me. Thanks!

PR #20466, opened 2023-04-10
[wip-ish] making wallrot less a ballrot

Hatterhat:
# dont merge this yet im making this pr to remind myself later
## About The Pull Request
- makes wallrot only hop to adjacent walls
- (the wip part) making it only able to hop to walls that aren't surrounded
- possible todo: maybe nerfing its base spread speed? will contemplate
- maybe it should also announce where it is too

image
## Proof of Testing
it's wip i'll get to it when i get to it
<details>

</details>

## Changelog

:cl:
balance: Wallrot can only jump to adjacent walls.
/:cl:

### DIFF
--- a/modular_skyrat/modules/wall_fungus/code/wall_fungus_component.dm
+++ b/modular_skyrat/modules/wall_fungus/code/wall_fungus_component.dm
@@ -96,9 +96,10 @@
 /datum/component/wall_fungus/proc/spread_to_nearby_wall()
 	var/turf/closed/wall/parent_wall = parent
 	var/list/walls_to_pick_from = list()
-	for(var/turf/closed/wall/iterating_wall in RANGE_TURFS(3, parent_wall))
+	for(var/turf/closed/wall/iterating_wall in RANGE_TURFS(1, parent_wall))
 		if(iterating_wall.GetComponent(/datum/component/wall_fungus))
 			continue
+		// todo: accessibility check (alt title: "man i love being able to burn fungus w/out having to rcd down a wall")
 
 		walls_to_pick_from += iterating_wall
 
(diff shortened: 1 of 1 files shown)


softcerv:
You may want to convert your PR into a draft until you are done with it.

Hatterhat:
> You may want to convert your PR into a draft until you are done with it.

completely forgot that was a thing i could do thanks

YakumoChen:
Ehhh, this ~~kinda~~ completely neuters an event that already usually gets cleaned up quickly outside of rare circumstances before real damage happens (kind of the point of most events) 

Being unable to go through rock makes sense but I disagree in that it should be able to go through airgapped walls. It's a mushroom. It has spores.

This will in most cases completely contain it to one section of wall - 4way manifold walls are fairly common and will block spread, unless you've coded they can jump diagonally as well. Doors and windows will also ensure these never spread anywhere and make the event largely a waste of a roll. Doubly if it announces location. We don't announce where blobs are, after all.

It's meant to be a dangerous event if left unchecked. Sometimes it does catastrophic damage. This is how most events are intended to go, it makes a story, not a brief annoyance.

Gandalf2k15:
reference to the other comment, the situation in which this was made was during an admin event round where everything was going to hell. The extent of damage was also blown out of proportion.

Hatterhat:
i still think that it jumping to inaccessible corners is dogwater but alright

chemistrymain2:
> reference to the other comment, the situation in which this was made was during an admin event round where everything was going to hell. The extent of damage was also blown out of proportion.

The extent of the damage was not blown out of proportion on a relatively recent round, where it destroyed half the station and the shuttle had to be called because comms were down when it spawned and it spawned far enough away that nobody caught it.

It also jumps corners and has bad pathing.

YakumoChen:
1 anomalous round does not the whole make.

PR #20479, opened 2023-04-11
Changes plasma thrower to an energy gun and lowers its blob damage

tf-4:
## About The Pull Request

Title, makes the gun behave like an energy gun rather than like an energy-ballistic hybrid kinda thing.
In practice, this just means you recharge it directly rather than having power packs that you can swap in and out. 

Also drops its damage against blobs by 1/4.

This gun is *pretty good*, making a BIG dent in blobs and the likes. Unfortunately, however, the replaceable (and rechargeable) energy-magazines are a bit of a balance problem, meaning with just a couple you can have a nigh-on infinite supply of shots.
This way, it's held to the same balance standards as other energy guns - i.e. you have to back up and recharge it directly every once in a while (20 shots).

## Proof of Testing
<details>
  
image

</details>

## Changelog
:cl:
balance: The plasma thrower is now an energy gun, meaning it has to be recharged in a charger rather than reloaded with magazines.
/:cl:


RatFromTheJungle:
it already is only 10 damage; you should really pick one-nerf or the other, both at once makes it's notably pointless against blobs with a slow reload, and 1/4th(?) the lasergun damage

Gandalf2k15 (approved):
HYAAAAAAAAAA BALLISTIC ENERGY GUN 😨

Gandalf2k15 (changes requested):
fix CI tho

tf-4:
CI is no longer my problem

Gandalf2k15 (approved):

PR #20486, opened 2023-04-11
Changes the sec department jacket to hold security items (+slight armour buff) 

tf-4:
## About The Pull Request

Adds `security_vest_allowed` to the dep jacket, in line with other security suit options, making it actually not CBT to wear cus you can now store a disabler and stuff.

Buffs the armour of it SLIGHTLY - now it's somewhere in between a sec winter coat (which is worse) and a sec armour vest (which is better). It covers the arms, so I made it a bit worse than the vest, but it does not cover the head so I made it a bit better than the winter coat. Mostly tends towards the winter coat, though, the armour values are a little conservative. It's for fashion anyway, right?

Makes a fashion choice more viable.

## Proof of Testing
<details>
  
image

</details>

## Changelog
:cl:
qol: The security jacket can now hold the same items as an armour vest in its suit storage slot.
balance: The security jacket has had its armour values slightly raised - though it's still a bit worse than a regular armour vest.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/clothing/under/utility_port/suits_port.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/under/utility_port/suits_port.dm
@@ -72,11 +72,20 @@
 	name = "security jacket"
 	desc = "A comfortable jacket in security blue. Probably against uniform regulations."
 	icon_state = "sec_dep_jacket"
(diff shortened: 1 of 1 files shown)


Gandalf2k15 (approved):
ok

PR #20507, opened 2023-04-12
[MIRROR] Renames delta time to be a more obvious name [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/74654
--------------------
This tracks the seconds per tick of a subsystem, however note that it is not completely accurate, as subsystems can be delayed, however it's useful to have this number as a multiplier or ratio, so that if in future someone changes the subsystem wait time code correctly adjusts how fast it applies effects

regexes used

git grep --files-with-matches --name-only 'DT_PROB'  | xargs -l sed -i 's/DT_PROB/SPT_PROB/g'
git grep --files-with-matches --name-only 'delta_time'  | xargs -l sed -i 's/delta_time/seconds_per_tick/g'


Gandalf2k15:
merging to prevent constant merge conflicts

PR #20697, opened 2023-04-22
[MODULAR] Fixes ghost cafe visitors receiving the wrong names after re-ghosting

### LINKED ISSUE #20690: People ghosting out of the ghost cafe are called "a cafe visitor" instead of their character's name

## Round ID:
8148

## Testmerges:
- [Antag Revamp*: The Nuclear-ish Option](https://github.com/Skyrat-SS13/Skyrat-tg/pull/18110)
- [[TM-ONLY] Prevents the psyker pirate variant from rolling](https://github.com/Skyrat-SS13/Skyrat-tg/pull/19820)
- [[MOSTLY MODULAR] Time Clock](https://github.com/Skyrat-SS13/Skyrat-tg/pull/18448)
- [[TM ONLY] Disables Sideplane, Makes the Game Playable](https://github.com/Skyrat-SS13/Skyrat-tg/pull/19308)

## Reproduction:
Just ghost out / cryo out from the ghost cafe, type in deadchat, notice that it's not your character name, cry.
image

vinylspiders:
## About The Pull Request

Fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/20690

There are two separate causes of this issue. 

### Problem 1:

Ghost cafe roles were set up with the `mob_name` var which is intended to be used as an override for the mob's actual name. This is meant to be used for when you have nonhuman/nonsilicon mobs generally, because the naming for humans/silicons already gets handled during the creation process.

What this caused to happen was, the mob was initialized with the overridden name (which caused their mind datum's `name` var to be called 'a cafe visitor' etc).

Then as part of the creation process the mob gets their randomized (or pref-given) name. But the mind datum still has the first name that it was initialized with, and when the mind is transferred into the ghost that is the one that it gets.

The solution is to not use the mob_name var for these roles, as they shouldn't have been used in the first place. This fixes the issue for visitors who do not use prefs.

### Problem 2:

But that doesn't completely fix the issue. Because prefs get applied after a mob is created, even if we don't override the name it's still being initialized with the random name that was given upon creating the mob. That means that the mind will have that randomized name and not the ones from the prefs.

So that has to be accounted for as well.

Bugfix

## Proof of Testing

<details>
  
dreamseeker_9Xed1Adghs

image

</details>

## Changelog

:cl:
fix: cafe roles will now retain their name after ghosting instead of just being called 'a cafe visitor'
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/mob_spawn/mob_spawn.dm
+++ b/modular_skyrat/master_files/code/modules/mob_spawn/mob_spawn.dm
@@ -20,6 +20,8 @@
 
 	spawned_human?.client?.prefs?.safe_transfer_prefs_to(spawned_human)
 	spawned_human.dna.update_dna_identity()
+	if(spawned_human.mind)
+		spawned_human.mind.name = spawned_human.real_name // the mind gets initialized with the random name given as a result of the parent create() so we need to readjust it
 	spawned_human.dna.species.give_important_for_life(spawned_human) // make sure they get plasmaman/vox internals etc before anything else
 
 	if(quirks_enabled)
--- a/modular_skyrat/modules/ghostcafe/code/ghost_role_spawners.dm
+++ b/modular_skyrat/modules/ghostcafe/code/ghost_role_spawners.dm
@@ -12,7 +12,6 @@
 	deletes_on_zero_uses_left = FALSE
 	icon = 'modular_skyrat/modules/ghostcafe/icons/robot_storage.dmi'
 	icon_state = "robostorage"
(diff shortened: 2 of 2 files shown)


GoldenAlpharex (approved):
Good fix, thank you. I'm happy I didn't have to do it myself :)

vinylspiders:
> Good fix, thank you. I'm happy I didn't have to do it myself :)

No worries--I had been mucking about in these files a lot for other things so I pretty much knew the issue immediately.

PR #20718, opened 2023-04-23
[MIRROR] Wintercoats can now be zipped and unzipped through alt-click and separates the hood sprites from the jacket sprites [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/74886
--------------------
## About The Pull Request
The title says it all, really.

~~Initially, I was only going to do it for all wintercoats, but then I figured I might as well bring it down to all of `/hooded`, just so other suits could benefit from it, since that behavior came from there anyway. Does that mean that it does nothing for some of them? Yes, it does. Does that justify having another variable to tell whether or not that should be possible? In my humble opinion, not really, but I'm not against it if it's requested.~~

~~That functionality was intentionally removed from the Void Cloak, as there would be balance implications (since bringing up the hood makes the whole cloak invisible, which you could skirt by just "zipping" it, which also makes it invisible.~~

~~The sprites were already there, so this change was very simple to do. Simply unties the zipped up look from the fact that the hood is up. However, toggling the hood forces the zipping/unzipping, just so there's no balance implications involved. It's just simpler that way.~~

So, I ended up going back and changing the sprites so that the hoods would no longer be baked into the jacket's sprites, so that they could be done as overlays instead, which ended up solving my problem with hoods not being there on zipped-up versions.

For now, it's been made on winter coats only, but it shouldn't be that difficult to bring it back down to the `/hooded` level. I just didn't want to bother touching up the sprites down there, as it already took me like 2-3 hours touching up the sprites of the winter coats alone.

I also took the decision to make it so EVA winter coats used the regular winter coat's sprites, because they had special ones that just looked like worse versions of the original, without anything special going on for them. It was just a straight downgrade compared to the base sprite, in my opinion.

There's still issues with the custom winter coat, in that the hood isn't made into an overlay for it yet (and that'll require an extra bit of logic to make it work, too), but it was already an issue before, the hood is always present on the current version of the custom winter coat.

There's still a handful (sadly, most) of the winter coats that don't properly reflect on their obj sprites when they're opened versus when they're closed, but that's due to an initial spriter oversight, and not to my doing. The open versions were just left as closed on many of them, and I simply don't have the patience nor the appropriate skills to edit that many coats that way.

## Why It's Good For The Game
Now you can be stylish with or without the hoodie!
image
image
image

According to ChatGPT, with one small tweak (thanks Opera GX for the suggestion):

> Zipped and unzipped through alt-click, winter coats can now be. Hmm, stylishly warm, you shall be. Feel like a Spaceman, you will. Use the Force, to zip and unzip, you must. Look cool, you will. Yes, hmmm.

## Changelog

:cl: GoldenAlpharex, ChatGPT for the first changelog entry (slightly edited)
qol: Zipped and unzipped through alt-click, winter coats can now be. Hmm, stylishly warm, you shall be. Feel like a Spaceman, you will. Use the Force, to zip and unzip, you must. Look cool, you will. Yes, hmmm.
image: Winter coats no longer have their hood baked into their jacket's sprite, both in item form and when worn.
fix: Updated the Icebox EVA winter coats (the Endotherm winter coats) to use the same sprites as the regular winter coats.
/:cl:


GoldenAlpharex (changes requested):
Will require changes to our own winter coats.

Ebin-Halcyon:
Give me a minute.

GoldenAlpharex (approved):
Nice work fixing it @theOOZ!

desertmote (approved):

PR #20734, opened 2023-04-24
Shotgun revolver changes

Gandalf2k15:
## About The Pull Request

The shotgun revolver now acts like a double barrel shotgun in that it has screenshake and has an accuracy debuff. It also requires you to hold it in both hands to shoot.

## Changelog

:cl:
balance: The shotgun revolver has been balanced: It now shakes the screen and has an accuracy debuff. It also requires you to hold it in both hands to shoot.
/:cl:

### DIFF
--- a/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
+++ b/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
@@ -72,10 +72,10 @@
 	mag_type = /obj/item/ammo_box/magazine/internal/cylinder/rev12ga
 	recoil = SAWN_OFF_RECOIL
 	weapon_weight = WEAPON_HEAVY
-	sawn_off = TRUE
 	icon = 'modular_skyrat/modules/novaya_ert/icons/surplus_guns/guns_32.dmi'
 	icon_state = "shawty_revolver"
 	fire_sound = 'modular_skyrat/modules/sec_haul/sound/revolver_fire.ogg'
+	spread = SAWN_OFF_ACC_PENALTY
 
 /obj/item/gun/ballistic/revolver/cin_shotgun_revolver/give_manufacturer_examine()
 	AddComponent(/datum/component/manufacturer_examine, COMPANY_TKACH)
--- a/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
+++ b/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
@@ -70,8 +70,9 @@
 	name = "\improper Tkach 'Ya-Sui' 12 GA revolver"
 	desc = "An outdated sidearm rarely seen in use by some members of the CIN. A revolver type design with a three shell cylinder. That's right, shell, this one shoots twelve guage."
 	mag_type = /obj/item/ammo_box/magazine/internal/cylinder/rev12ga
-	w_class = WEIGHT_CLASS_BULKY
+	recoil = SAWN_OFF_RECOIL
 	weapon_weight = WEAPON_HEAVY
+	sawn_off = TRUE
 	icon = 'modular_skyrat/modules/novaya_ert/icons/surplus_guns/guns_32.dmi'
 	icon_state = "shawty_revolver"
 	fire_sound = 'modular_skyrat/modules/sec_haul/sound/revolver_fire.ogg'
--- a/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
+++ b/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
@@ -70,13 +70,13 @@
 	name = "\improper Tkach 'Ya-Sui' 12 GA revolver"
 	desc = "An outdated sidearm rarely seen in use by some members of the CIN. A revolver type design with a three shell cylinder. That's right, shell, this one shoots twelve guage."
 	mag_type = /obj/item/ammo_box/magazine/internal/cylinder/rev12ga
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_BULKY
+	weapon_weight = WEAPON_HEAVY
 	icon = 'modular_skyrat/modules/novaya_ert/icons/surplus_guns/guns_32.dmi'
 	icon_state = "shawty_revolver"
 	fire_sound = 'modular_skyrat/modules/sec_haul/sound/revolver_fire.ogg'
 
 /obj/item/gun/ballistic/revolver/cin_shotgun_revolver/give_manufacturer_examine()
-
 	AddComponent(/datum/component/manufacturer_examine, COMPANY_TKACH)
 
 /obj/item/gun/ballistic/revolver/cin_shotgun_revolver/examine_more(mob/user)


RatFromTheJungle:
it'd probably be better to make it equivalent to the sawn-off shotgun (double barrel, or peacekeeper)
image
but a bulky two-handed pistol would make it worse then every other shotgun, instead of equal; whereas a sawn off would be better in everyway (and fit in your backpack)

Gandalf2k15:
> it'd probably be better to make it equivalent to the sawn-off shotgun (double barrel, or peacekeeper) image but a bulky two-handed pistol would make it worse then every other shotgun, instead of equal; whereas a sawn off would be better in everyway (and fit in your backpack)

well, these are better, as you can buy them and they can hold 3 shells. If you could mass produce sawn offs then maybe a bit better, but still, the point is not to add an objectively better gun.

Krypandenej:
Agreed with the above, and maybe give it some screenshake when firing, kinda like the Golden Revolver but not as bad. Someone decided to put 12 gauge in a **pistol**, after all...

RatFromTheJungle:
> well, these are better, as you can buy them and they can hold 3 shells. If you could mass produce sawn offs then maybe a bit better, but still, the point is not to add an objectively better gun.

you can buy peacekeepers to saw-off as well, it wouldn't be objectively better then them, as sawn-offs get 7 shells, it would be better then the bartenders shotgun (if they saw it off), but worse if they don't (because the barkeeps would have the same fire-rate, but much better accuracy when not sawn off), which would put this firmly inbetween them

(peacekeepers being [this]( https://github.com/Skyrat-SS13/Skyrat-tg/blob/af7f047798e6e7ecbd444de1ebe611a5c4cce3b0/modular_skyrat/modules/company_imports/code/armament_datums/bolt_nanotrasen_firearms.dm#L73) which can be sawn-off)

Gandalf2k15:
> > well, these are better, as you can buy them and they can hold 3 shells. If you could mass produce sawn offs then maybe a bit better, but still, the point is not to add an objectively better gun.
> 
> you can buy peacekeepers to saw-off as well, it wouldn't be objectively better then them, as sawn-offs get 7 shells, it would be better then the bartenders shotgun (if they saw it off), but worse if they don't (because the barkeeps would have the same fire-rate, but much better accuracy when not sawn off), which would put this firmly inbetween them
> 
> (peacekeepers being [this](https://github.com/Skyrat-SS13/Skyrat-tg/blob/af7f047798e6e7ecbd444de1ebe611a5c4cce3b0/modular_skyrat/modules/company_imports/code/armament_datums/bolt_nanotrasen_firearms.dm#L73) which can be sawn-off)

These still have a faster fire rate than riot shotguns, doing this balances them in an acceptable way.

SpaceVampire:
I think the idea of making it a two-handed pistol is pretty funny.
Having to deal with the kickback of putting shotguns shells in there.
Great mental image as well.
But making it bulky would be a pretty weird design choice.
Given that every other pistol in the game can fit in your backpack.
It isn't like the current sprite makes it look very large.
There is probably a way to balance it that doesn't break immersion.

Gandalf2k15:
> I think the idea of making it a two-handed pistol is pretty funny. Having to deal with the kickback of putting shotguns shells in there. Great mental image as well. But making it bulky would be a pretty weird design choice. Given that every other pistol in the game can fit in your backpack. It isn't like the current sprite makes it look very large. There is probably a way to balance it that doesn't break immersion.

I mean, when the backpack nerf came around that defense was made. This is still a very very powerful weapon, and should be treated as such. I disagree with riots fitting in backpacks too but that's not relevant to this specific gun. You can't balance the actual damage output of these without changing the ammo type which defeats the purpose. The main thing is that you can also buy loads of these from cargo which sec do a lot. If it was a special weapon? Then sure. You wouldn't see people carrying 7 of em around in their backpack since there would only be 1 instance.

RatFromTheJungle:
> These still have a faster fire rate than riot shotguns, doing this balances them in an acceptable way.

https://youtu.be/4lMkKf5kKgE

the fire-rate difference is like, give-or-take 0.3s in a 3 shot span, so I don't think the fire-rate is that much different if-at-all (considering I did it on live)

Gandalf2k15:
> > These still have a faster fire rate than riot shotguns, doing this balances them in an acceptable way.
> 
> https://youtu.be/4lMkKf5kKgE
> 
> the fire-rate difference is like, give-or-take 0.3s in a 3 shot span, so I don't think the fire-rate is that much different if-at-all (considering I did it on live)

I already stated my disagreement with riots fitting in backpacks. And there is still a difference.

RatFromTheJungle:
> > > These still have a faster fire rate than riot shotguns, doing this balances them in an acceptable way.
> > 
> > 
> > https://youtu.be/4lMkKf5kKgE
> > the fire-rate difference is like, give-or-take 0.3s in a 3 shot span, so I don't think the fire-rate is that much different if-at-all (considering I did it on live)
> 
> I already stated my disagreement with riots fitting in backpacks. And there is still a difference.

there isn't really a difference; it just makes the revolver shotgun worse then all of the others, both being only barely capable of critting if you hit all three, on someone with no armor, but also having all the downsides of a combat shotgun which gets 4(?) more shells

requiring two hands, sure,
downsides of the other-sawn-offs, sure
but making it the one pistol that's inexplicably bulky is just out of place, and feels more like 'it was the first thing that came to mind' over something that would actually balance it

Gandalf2k15:
> > > > These still have a faster fire rate than riot shotguns, doing this balances them in an acceptable way.
> > > 
> > > 
> > > https://youtu.be/4lMkKf5kKgE
> > > the fire-rate difference is like, give-or-take 0.3s in a 3 shot span, so I don't think the fire-rate is that much different if-at-all (considering I did it on live)
> > 
> > 
> > I already stated my disagreement with riots fitting in backpacks. And there is still a difference.
> 
> there isn't really a difference; it just makes the revolver shotgun worse then all of the others, both being only barely capable of critting if you hit all three, on someone with no armor, but also having all the downsides of a combat shotgun which gets 4(?) more shells

The point being I don't want to just have another thing adding to the fire. It was this or remove it from cargo. Or make it a special weapon.

RatFromTheJungle:
all of which-issues, would be solved by making it have the 25 accuracy-debuff,, a-lot of screenshake (preventing, like any functional use of the fast fire-rate), and then, even if you carried however-many you can, or how-ever-much-ammo you can, it'd be functionally impossible to spam-fire them, making full shotguns better for actual fights, and cementing it as a solid self-defense gun.

where-as making it bulky, just makes it a flatly better bartender shotgun, while also being worse then just buying a peacekeeper (which does everything it does, for a supposed, if unnoticable difference in fire-rate)

Gandalf2k15:
> all of which-issues, would be solved by making it have the 25 accuracy-debuff,, a-lot of screenshake (preventing, like any functional use of the fast fire-rate), and then, even if you carried however-many you can, or how-ever-much-ammo you can, it'd be functionally impossible to spam-fire them, making full shotguns better for actual fights, and cementing it as a solid self-defense gun.
> 
> where-as making it bulky, just makes it a flatly better bartender shotgun, while also being worse then just buying a peacekeeper (which does everything it does, for a supposed, if unnoticable difference in fire-rate)

ok, updated

SpaceVampire:
It might be smart to update the desc to reflect this change.
I'm pretty sure this is the only two-handed revolver so having a heads-up would be nice for people.

lessthnthree (approved):

PR #20810, opened 2023-04-27
[MIRROR] Adds a macro for checking map job changes, makes captain outfit more runtime resilient [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/74993
--------------------
## About The Pull Request

- Makes captain outfit more runtime resilient
    -  The initial reason for this PR was that the captain's outfit was not very resistant to the case that something else changes the outfit's left hand before `post_equip`. It would runtime due to assuming the left hand is always the unique banner, and also cause no banner to be spawned. 

- Adds a macro for accessing the map config job changes, to reduce boilerplate. 
    - While I was fixing the above bug, I noticed that we had pretty bad boilerplate copy pasted around for accessing job config changes. This code is not ran so often that the job config needed to be cached locally, nor was all this safety very necessary. So I added a wrapper to handle it for the end user. 

## Why It's Good For The Game

More runtime resistance is always nice, allows for future coders (or me) to tread the area without worry. 

## Changelog

:cl: Melbert
fix: Fixes rare occasions in which the captain can fail to spawn with their banner on certain maps if their left hand was holding something else. 
/:cl:


GoldenAlpharex (approved):
Should be good now!

PR #20819, opened 2023-04-28
[MODULAR] Refactors adult status effects

softcerv:
## About The Pull Request
This PR refactors adult status effects, making it so that they do not run every single life tick on human mobs. Additionally, this PR breaks up the status effect for organ fluid generation into multiple different status effects.

These two status effects are two additional things in fast process for every single human when not every human is going to be able to use them. This PR makes it so that they are only applied when actually being used.

## Proof of Testing

<details>
 
image

</details>

## Changelog

:cl:
refactor: refactors organic interface status effects so that they no longer run on every human.
/:cl:


RatFromTheJungle:
Help me, Gordon! My [adult fluid] generation is off the charts!  Urrggh!

PR #20854, opened 2023-04-30
[MIRROR] Removal of Kilostation [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75039
--------------------

## About The Pull Request
Removes Kilostation from the game.
Since I know this is bound to be a touchy subject (the map has its fans, after all), I feel I need to explain the situation and why this decision has been made.

While working on mapping for TGStation, there's an inherent struggle between what's good for mappers, and what's good for coders. On the mapping side, we want as many maps as we reasonably can have in the repo. On the code side, every map added introduces a new layer of complexity for adding new features. While we're attempting to make strides in improving cooperation between the two teams so we can handle both many maps and lots of new features, in the here and now this means that we find ourselves limited in how many maps we can maintain at once.

With this in mind, we find ourselves in a conundrum when new maps come along. While it would be nice to accept everything new without needing to remove anything else, the above problems force us to stick to a limit. Via a fair amount of discussion on the maintainer team, this was set to 6 recently, which we reached when Northstar was merged, and meaning that a removal was required for any new maps.

As for why the decision was made to take Birdshot over Kilo, I believe it is better for us to have a map that has room to grow and an active creator. As much as I was personally a Kilo enjoyer, the map has consistently proven to be problematic for new code additions due to its scale, and it receives little maintenance outside of the bare essentials as a result. With all this in mind, it comes out as the obvious choice for removal in this case. Of course, nothing stops it from coming back in future if any of the above changes.
## Why It's Good For The Game
TL:DR: In order to keep the maintenance budget down, we need to remove a map for the sake of new blood, and in this case Kilo was the simplest choice.
## Changelog
:cl:
del: Kilostation has went back to its home planet. It served us faithfully for nearly 3 and a half years. May it rest in peace.
/:cl:


Zergspower:
I like Kilo but it makes sense, sadly

PR #20861, opened 2023-04-30
[does commenting count as modular] style meter users can now flip

Hatterhat:
## About The Pull Request
title

style meter feature (flipping as point multiplier) actually usable when active

## Proof of Testing
<details>

image  
</details>

## Changelog
:cl:
fix: People with style meters affixed to their glasses can now do cool flips.
/:cl:

### DIFF
--- a/code/datums/components/style/style.dm
+++ b/code/datums/components/style/style.dm
@@ -98,6 +98,9 @@
 
 	RegisterSignal(src, COMSIG_ATOM_TOOL_ACT(TOOL_MULTITOOL), PROC_REF(on_parent_multitool))
 
+	// skyrat edit: allows style meter chads to do flips
+	ADD_TRAIT(mob_parent, TRAIT_STYLISH, src)
+
 /datum/component/style/RegisterWithParent()
 	RegisterSignal(parent, COMSIG_MOB_ITEM_AFTERATTACK, PROC_REF(hotswap))
 	RegisterSignal(parent, COMSIG_MOB_MINED, PROC_REF(on_mine))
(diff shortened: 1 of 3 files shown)


code/datums/components/style/style.dm
```diff
@@ -98,6 +98,9 @@
 
 	RegisterSignal(src, COMSIG_ATOM_TOOL_ACT(TOOL_MULTITOOL), PROC_REF(on_parent_multitool))
 
+	// skyrat edit: allows style meter chads to do flips
+	ADD_TRAIT(mob_parent, TRAIT_STYLISH, src)
```
tf-4:
```suggestion
	ADD_TRAIT(mob_parent, TRAIT_STYLISH, src) // SKYRAT EDIT ADD - allows style meter chads to do flips
```

code/datums/components/style/style.dm
```diff
@@ -149,6 +152,8 @@
 	if(mob_parent.hud_used)
 		mob_parent.hud_used.static_inventory -= meter
 		mob_parent.hud_used.show_hud(mob_parent.hud_used.hud_version)
+	// skyrat edit: allows style meter chads to do flips
+	REMOVE_TRAIT(mob_parent, TRAIT_STYLISH, src)
```
tf-4:
```suggestion

	REMOVE_TRAIT(mob_parent, TRAIT_STYLISH, src) // SKYRAT EDIT ADD - allows style meter chads to do flips
```

tf-4 (approved):

SpaceVampire:
Would you think about also adjusting the amount of points you get from it, due to the Skyrat emote cooldown limit?
If you can only do it every five seconds, it could probably give more points.

Hatterhat:
> Would you think about also adjusting the amount of points you get from it, due to the Skyrat emote cooldown limit? If you can only do it every five seconds, it could probably give more points.

balance change, let's see if we can just get flipping in first and then worry about it

tf-4:
yeah one bit at a time

PR #20872, opened 2023-05-01
Fixes the AMR fitting anywhere except on your back (YEAH ITS A BUG NOT A FEATURE)

Paxilmaniac:
## About The Pull Request

I foolishly assumed that armor vests would be unable to hold anything with WEIGHT_CLASS_BULKY, this turns out to not be the case. Good news however, there's an even easier solution, bumping the weight class up higher.

The entire idea of this weapon is that you can really hurt someone with it, but you wouldn't be able to carry it anywhere easily. Armor vests being able to fit the amr was something I tried to make NOT happen, but guns fitting on armor vests worked slightly different to what I thought.

## Proof of Testing

image

## Changelog

:cl:
fix: the cargo orderable AMR should no longer fit in suit storage slots, only your back, as was originally intended
/:cl:

### DIFF
--- a/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
+++ b/modular_skyrat/modules/novaya_ert/code/surplus_weapons.dm
@@ -119,7 +119,7 @@
 	icon_state = "amr"
 	inhand_icon_state = "amr"
 	worn_icon_state = "amr"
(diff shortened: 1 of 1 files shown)


tf-4 (approved):

kittysmooch:
will the modsuit mag harness still snap it after this

PR #20879, opened 2023-05-01
[MIRROR] Bows no longer randomly become unusable [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75058
--------------------

## About The Pull Request

Fixes bows randomly becoming unusable.

On firing bows were leaving behind a reference to their casing in the internal magazine. As arrows are technically caseless ammo there was no way to remove this casing, and eventually it would turn into a null, rendering the magazine unable to load any new arrows.

This is fixed by adding a check for whether the ammo is cased to all ballistic guns in `process_fire()`. Caseless ammo has its casing removed from the magazine's `stored_ammo` list to prevent it rotting away to a null down the line.

While I was here renamed and autodocced the var `heavy_metal` which appears to be exactly the var for having a casing that I needed.

## Why It's Good For The Game

Bows becoming randomly useless because of leftover vars is bad and silly, and so are mystery meat vars with undescriptive names and no autodoc.

## Changelog
:cl:
fix: Bows will no longer randomly stop taking new arrows.
/:cl:


code/modules/projectiles/ammunition/energy/_energy.dm
```diff
@@ -8,7 +8,11 @@
...
 	fire_sound = 'sound/weapons/laser.ogg'
 	firing_effect_type = /obj/effect/temp_visual/dir_setting/firing_effect/energy
+<<<<<<< HEAD
 	heavy_metal = FALSE
 	//SKYRAT EDIT CELL LOADED GUNS//
 	var/select_color = FALSE //This is the color that shows up when selecting an ammo type. Disabled by default
 	//SKYRAT EDIT END//
+=======
+	is_cased_ammo = FALSE
+>>>>>>> 461ef29f18e (Bows no longer randomly become unusable (#75058))
```
tf-4:
```suggestion
	heavy_metal = FALSE
	is_cased_ammo = FALSE
	
	//SKYRAT EDIT ADD - CELL LOADED GUNS
	var/select_color = FALSE //This is the color that shows up when selecting an ammo type. Disabled by default
	//SKYRAT EDIT ADD END
```

code/modules/projectiles/ammunition/energy/_energy.dm
```diff
@@ -9,6 +9,8 @@
 	fire_sound = 'sound/weapons/laser.ogg'
 	firing_effect_type = /obj/effect/temp_visual/dir_setting/firing_effect/energy
 	heavy_metal = FALSE
```
tf-4:
```suggestion
```

PR #20953, opened 2023-05-05
Security record console for orderlies on Void Raptor

Shadow-Quill:
## About The Pull Request

Replaces the crew monitoring console in the guard room for Medbay on Void Raptor with a security records console.

:cl:
add: Added a security records console to the medical guard room on Void Raptor.
del: Removed the crew monitoring console from the medical guard room on Void Raptor.
/:cl:


Jolly-66:
Orderlies are not sec, they're a medical job.

This is the same for the other department guards, and as such, do not need access to security equipment (like the sec console)

PR #16192, opened 2022-09-13
Removes chemlab and autolathe from ghost cafe

YakumoChen:
## About The Pull Request
Removes the autolathe and chemistry lab from the ghost cafe
Wardrobe vendors and lockers in the ghost cafe have had their radios removed. 

## How This Contributes To The Skyrat Roleplay Experience

Removes hazardous methods of exploiting the ghost cafe from interacting with the round at large, or being used to escape the ghost cafe, or otherwise cause major unwanted damage to the cafe area.

The chemistry lab was previously removed from the cafe, along with other things in #3820. It was readded despite this, for reasons directly opposing #3820 and should not have been merged. #3820 is the current precedent for ghost cafe additions (it is a hangout space ONLY, not a place to tutorial jobs or experiment)

I have also removed the autolathe, purely because it can be used to print radios and communicate with the station, and I don't think its worth the time or effort to code an entirely new autolathe without radios being printable. All wardrobe vendors that contain radios have had their stock of such removed. 
## Changelog

:cl:
del: Ghost Cafe has had its chemistry lab, autolathe, and all radios removed so as to prevent interaction with other sectors, or damage to the ghost cafe (which chemlabs often do)
/:cl:

PR #21058, opened 2023-05-10
[MODULAR] Ghost Cafe Autolathe

softcerv:
## About The Pull Request
Adds the autolathe back to the ghost cafe, this time with a special variant that completely lacks any kind of radio.

The autolathe was removed from the ghost cafe in https://github.com/Skyrat-SS13/Skyrat-tg/pull/16192 because it could be used to print radios and the author of the PR didn't want to create a separate autolathe type to address this issue. Due to having a variant of the autolathe that is no longer able to print radios, it should no longer be problematic for the ghost cafe to have this.

Additionally, having an autolathe in the ghost cafe provides a way to teach newer players about game mechanics that would otherwise be stressed out by learning under the pressure of a normal round.

## Proof of Testing

<details>

image

image

</details>

## Changelog

:cl:
add: the ghost cafe now has their autolathe back, this time without the ability to print radios.
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom_skyrat_z2.dmm
+++ b/_maps/map_files/generic/CentCom_skyrat_z2.dmm
@@ -9034,6 +9034,10 @@
 	},
 /turf/open/floor/iron/white,
 /area/centcom/holding/cafe)
+"mQL" = (
+/obj/machinery/autolathe/ghost_cafe,
+/turf/open/indestructible/hotelwood,
+/area/centcom/holding/cafe)
 "mSs" = (
 /turf/open/floor/iron/freezer,
 /area/centcom/interlink)
@@ -61064,7 +61068,7 @@ aqf
 vxh
 aXN
 tag
-aUo
+mQL
 aQK
 aQK
 aqf
--- a/modular_skyrat/modules/mapping/code/autolathe.dm
+++ b/modular_skyrat/modules/mapping/code/autolathe.dm
@@ -10,7 +10,7 @@
 	stored_research = GLOB.autounlock_techwebs[/datum/techweb/autounlocking/autolathe/ghost_cafe]
 
 /obj/item/circuitboard/machine/autolathe/ghost_cafe
-	"Ancient Autolathe"
+	name = "Ancient Autolathe"
 	build_path = /obj/machinery/autolathe/ghost_cafe
 
 /datum/techweb/autounlocking/autolathe/ghost_cafe/New()
--- a/modular_skyrat/modules/mapping/code/autolathe.dm
+++ b/modular_skyrat/modules/mapping/code/autolathe.dm
@@ -1,0 +1,19 @@
+/obj/machinery/autolathe/ghost_cafe
+	name = "ancient autolathe"
+	hacked = TRUE
+	circuit = /obj/item/circuitboard/machine/autolathe/ghost_cafe
+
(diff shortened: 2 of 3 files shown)


YakumoChen:
Explain how your 'why this is good for roleplay' doesn't conflict with #3280 ? Ghost Cafe is explicitly not for 'learning the game', it is a hangout zone.

I'm not even sure what the autolathe meaningfully provides that isn't already somewhere in the same room?

Jolly-66 (changes requested):
I'm not entirely sold on giving the ghost Cafe an ahtolathe even if radios are taken out.

There should be sufficient tools and materials around if people want to build something.

Also, the ghost Cafe isn't the place to learn about things. Its a place for people to relax, enjoy and RP without the fears of the ongoing round interrupting them.

softcerv:
>I'm not entirely sold on giving the ghost Cafe an ahtolathe even if radios are taken out.

The PR that removed the autolathe was merged under the premise that it was problematic by granting ghost-cafe users access to radio. Nowhere inside of the discussion of the PR were the other (de)merits of having the autolathe discuessed.

>There should be sufficient tools and materials around if people want to build something.

I'm sure that there are things that are present in the autolathe that aren't present inside of the ghost cafe in it's current state. If you'd like I'm sure I could compile a list for you to look over.

> Also, the ghost Cafe isn't the place to learn about things. Its a place for people to relax, enjoy and RP without the fears of the ongoing round interrupting them.

Not everyone has the knowledge to make a test server, this is the closest thing we have to an ooc way to teach people in-game.  
Could you please explain to me what we gain from forbidding this use case of the ghost cafe?

YakumoChen:
I did remove the autolathe yes but it wasn't really needed in the first place, you can see in the PR that every effort made to justify the machine being in there was pointed out to already be present in the room in other places.

I can't really think of anything you would want to print from there that isn't probably harmful to the cafe, or already found, aside from maybe a chisel or something?

softcerv:
> I can't really think of anything you would want to print from there that isn't probably harmful to the cafe, or already found, aside from maybe a chisel or something?

Spray can, Airlock painter, stock parts, construction parts, circuit tiles, extra silverware and cutlery, fishing rods, various plastic toys, and a bunch of miscellaneous stuff.

((While some of these are already present, they are in limited quantity.))

Additionally, we already have power-tools and RCDs buyable, I am dubious that the things added by this are going to be more problematic then the things already present.

YakumoChen:
Silverware is the only thing I can think of from that list, maybe spray cans that would be useful (and you can just add a few spray cans to the room, ez)

But silverware should be in the kitchen vendor, so that shouldn't be an issue.

Most of the rest of the items are minor, and you can add a couple to the room if you need to without an autolathe. As jolly and I said, most of what you need is present in the room as it is.

hugmeorelse:
While I understand this comes with good intentions, the server is not meant to be a "testing ground" for staff and players alike. In fact, I have already begun discouraging staff from experimenting on the server, and I am looking to setup a separate server for staff (and possibly players) for that. Abuse of resources on Ghost Cafes in the past have caused major disruptions to on-going rounds before. While, yes, this can be solved in a reactionary way by staff, they have enough on their plate to begin with.

Jolly-66:
None of the other maintainers have interest and neither do I.

PR #21155, opened 2023-05-14
Sets all ERP Item breakout times to 5 Seconds

ghost:
## About The Pull Request
There is absolutely zero reason or excuse for these items to have breakout times rivaling, or even being insanely higher than, existing restraints in the game.
They are entirely for the purpose of flavor in roleplay scenarios, and pref-checks or not, **anybody** should have the option to put their foot down and stop an uncomfortable situation. 

Also removes the Reinforced variants, as they were created from existing restraints in the game to simply apply higher timers to the ERP items. Which, as explained, are no longer at all above 5s

Consent is not a set Yes or No. Consent can be stripped away if the situation gets out of hand. Rather than leave a player bound for up to **several minutes** in a situation they no longer want to be part of, they should be able to step aside and stop the roleplay and LOOC what went wrong.

## Proof of Testing

<details>

Tested all changed ones. I literally had to look at the code to understand how to even use these things, and the Shibari rope stand physically cannot be tested with one person due to prefs checking.
Regardless, note every breakout time being listed as 5 seconds or less.
I couldn't fit them all in one screenshot. I tested them all. The two sleeping bag ones are because there's two seperate "modes" on that bag.
image

</details>

## Changelog

:cl:
balance: all ERP restraints now have a short enough breakout time to stop any uncomfortable situations.
/:cl:


desertmote:
i dont see eye to eye with the reasoning here. yes its important, but this type of stuff should be solved with LOOC, not mechanics.
the handcuffs are useful for prison escapes.

60 seconds or 5 seconds doesnt matter. if you looc "stop" you are to be freed instantly anyway.

ghost:
> i dont see eye to eye with the reasoning here. yes its important, but this type of stuff should be solved with LOOC, not mechanics. the handcuffs are useful for prison escapes.

They can LOOC after stopping the uncomfortable situation. There is **zero** excuse for leaving them bound to it when they no longer want to be.
And using ERP items for prison escapes is something that's not meant to be done (and iirc against the rules); they are flavor for ERP and not items that should have a place in game balance.

softcerv:
> using ERP items for prison escapes is something that's not meant to be done (and iirc against the rules); they are flavor for ERP and not items that should have a place in game balance.

I've been contemplating adding better admin logging for the ERP items so that the people that do abuse them for mechanical advantage can get banned easier. Although I have a feeling that most of the abuse has died down.

softcerv:
Bringing this up here too, but it would be better to have a safeword verb that would remove all organic interface gear from the user. That way people would still be able to escape if needed, but those that enjoy and use the longer breakout times would still be able to use them.

ghost:
To be honest, I can see the argument. This is ERP gear. Though, I personally have seen maybe one player abuse ERP gear when sex 2 was first merged. This seems like something that can be solved with a quick LOOC escalated to an ahelp if there’s a very situational issue. 

If prisoners are seriously using reinforced kink items to handcuff officers, may I suggest just changing the item pathing to accept ONLY the metal handcuffs secoff uses? Currently it can take all forms of cuffs. Fuzzy cuffs on fuzzy cuffs. Cable cuffs on gloves. Then, they can only do that if they ever get access to cuffs.

tf-4:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/21287

ghost:
any possible alternative rather than just not making the erp items overpowered.

PR #21196, opened 2023-05-15
Removes some unnecessary non-modular icon overrides

tf-4:
## About The Pull Request

removes the `fixing_missing_icons` folder, as it overrode icons to... the same thing as the original. Not sure about that one. Anyway, it's fixed now.

also changes the BSA computer to have a null keyboard overlay instead of using empty overlays (💀)

## Proof of Testing

<details>
  
image

</details>


desertmote (approved):
Thank you!!

PR #21242, opened 2023-05-16
[you cannot imagine how upset I am right now] the incredible three line fix to create and destroy failing on every pr

Paxilmaniac:
## About The Pull Request

As melbert so kindly pointed out to me over at tg, I forgot to override the original custom materials of these items and thus they would get applied twice. This for several reasons is not optimal.

It won't make the red x go away because we still have nuke cinematic failing but it'd put us back down to one ci failure rather than two.

## Proof of Testing

If create and destroy doesn't fail on this pr then it works.

## Changelog

:cl:
fix: fixes blown glassware making the codebase fail ci
/:cl:

### DIFF
--- a/modular_skyrat/modules/primitive_production/code/glassblowing.dm
+++ b/modular_skyrat/modules/primitive_production/code/glassblowing.dm
@@ -39,6 +39,7 @@
 	desc = "A glass bowl that is capable of carrying things."
 	icon = 'modular_skyrat/modules/primitive_production/icons/prim_fun.dmi'
 	icon_state = "glass_bowl"
+	custom_materials = null
 	material_flags = MATERIAL_EFFECTS | MATERIAL_COLOR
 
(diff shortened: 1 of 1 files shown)


tf-4 (approved):

modular_skyrat/modules/primitive_production/code/glassblowing.dm
```diff
@@ -61,6 +63,7 @@
 	desc = "A glass plate that is capable of carrying things."
 	icon = 'modular_skyrat/modules/primitive_production/icons/prim_fun.dmi'
 	icon_state = "glass_plate"
+	custom_materials = null
```
RatFromTheJungle, line 66: pls add a gun here or else we can't approve the pr sorry it's just the rules! 🇺🇸

Paxilmaniac: ah my bad my bad let me add a the gun material datum, turns anything its applied to into a gun

PR #21287, opened 2023-05-18
[MODULAR] Safeword Verb

softcerv:
## About The Pull Request
Introduces a safeword verb under the ooc tab that removes any and all pieces of organic interface gear from a human. 

Additionally, it is logged when the verb is used along with any and all gear taken off as a result of this, this way it is easier to track down abuse from an admin point of view. 

This PR aims to be an alternative to https://github.com/Skyrat-SS13/Skyrat-tg/pull/21155 , accomplishing the same goal of allowing players to easily OOCly escape from an uncomfortable situation, but while maintaining the escape timers for those that want to use them.

From a code perspective, having this done through a list, instead of through changing individual items, will make it easier to change what items we want affected by this.

## Proof of Testing

<details>
  
Before
image

After
image

image

image

</details>

## Changelog

:cl:
qol: adds a safeword verb under the ooc tab that can be used to remove all equipped organic interface items from someone. 
/:cl:


modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
```diff
@@ -16,3 +16,18 @@
 	. = ..()
 	if(CONFIG_GET(flag/disable_erp_preferences))
 		verbs -= /mob/living/carbon/human/verb/climax_verb
+
+/mob/living/carbon/human/verb/safeword()
+	set name = "Remove Lewd Items"
+	set category = "OOC"
+	set desc = "Removes any and all lewd items from you."
+
+	log_message("[key_name(src)] used the Remove Lewd Items verb.")
```
GoldenAlpharex: I'm not certain if it does it by default, that'd have to be checked, but it could probably be useful to also add the coordinates and area macro to this, so the logging is even more useful.

GoldenAlpharex (changes requested):
Other than that, it's all good with me, this is a good idea.

modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
softcerv: Sure thing, let me see what I can do.

softcerv: image

GoldenAlpharex (approved):
Should be all good to go now!

lessthnthree (approved):

PR #21327, opened 2023-05-21
[Modular] Escape pods now stay in dock if alert level is too low

Iajret:
## About The Pull Request
Adds a config to set minimum level at which escape pods will fly towards centcom. Set as 0 by default (which means green alert will suffice, it uses numbers from `security_alerts.dm`). Emagging pod controls will turn off its alert checking sensor and it will fly away even if alert level's too low

It always felt kinda stupid when station conducting its mandatory crew transfer decides to launch all of its escape pods even if threat is eliminated or haven't existed at all.

## Proof of Testing

<details>

  regular pod still stands in its dock
image

hacked pod flew away
image

</details>

## Changelog

:cl:
add: Escape pods will not evacuate people at low alert level (disabled by default)
add: Escape pods can be emagged to evacuate even if alert is to low
config: added config string to set minimum alert level for pods to work.
/:cl:

### DIFF
--- a/modular_skyrat/modules/pod_locking/pod_locking.dm
+++ b/modular_skyrat/modules/pod_locking/pod_locking.dm
@@ -1,6 +1,3 @@
-///obj/docking_port/mobile/pod
-//		launch_status = NOLAUNCH
-
 /obj/docking_port/mobile/pod/register()
 	. = ..()
 	if(CONFIG_GET(number/minimum_alert_for_pods) != 0)
--- a/modular_skyrat/modules/pod_locking/pod_locking.dm
+++ b/modular_skyrat/modules/pod_locking/pod_locking.dm
@@ -6,7 +6,8 @@
 
 /**
   * Signal handler for [COMSIG_SECURITY_LEVEL_CHANGED].
-  * Pulls minimum required alert level from modular config and sets pod's [var/launch_status] accordingly. Will not do anything of pod already launched.
+  * Pulls minimum required alert level from modular config and sets pod's [var/launch_status] as [NOLAUNCH] or [UNLAUNCHED] accordingly.
+  * Will not do anything if pod already launched.
   */
 /obj/docking_port/mobile/pod/proc/check_for_evac(datum/source, new_level)
 	SIGNAL_HANDLER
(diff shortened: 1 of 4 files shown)


modular_skyrat/modules/pod_locking/pod_locking.dm
```diff
@@ -0,0 +1,22 @@
+/obj/docking_port/mobile/pod/register()
+	. = ..()
+	if(CONFIG_GET(number/minimum_alert_for_pods) != 0)
+		launch_status = NOLAUNCH
+		RegisterSignal(SSsecurity_level, COMSIG_SECURITY_LEVEL_CHANGED, PROC_REF(check_for_evac))
+
+/obj/docking_port/mobile/pod/proc/check_for_evac(datum/source, new_level)
```
Zonespace27, line 12: code doc pls

Zonespace27 (changes requested):

modular_skyrat/modules/pod_locking/pod_locking.dm
Iajret: done. Not sure how this links work though

Zonespace27 (approved):
lgtm

modular_skyrat/modules/pod_locking/pod_locking.dm
```diff
@@ -0,0 +1,27 @@
...
+  */
+/obj/docking_port/mobile/pod/proc/check_for_evac(datum/source, new_level)
+	SIGNAL_HANDLER
+
+	var/min_level = CONFIG_GET(number/minimum_alert_for_pods)
+	if(launch_status > UNLAUNCHED)
+		return
+	launch_status = (new_level >= min_level) ? UNLAUNCHED : NOLAUNCH
+
+/obj/machinery/computer/shuttle/pod/emag_act(mob/user)
```
ghost, line 20: Rather than locking this behind an emag, could it simply send a message about the pod launching manually or require some extra confirmation? It'd let antags or opfors bail without having to also get le hacking card

Iajret: I think i could do something like shuttle hijacking so any antag would be able to hack it. Also im not sure if it should even announce such act or not. Or well, send any message at all. Didn't quite got what you mean by that tbh.

lessthnthree (approved):

PR #21329, opened 2023-05-22
[MIRROR] JSON Logging Take Two [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/73604
--------------------
## About The Pull Request

Converts all logging, excluding perf and investigate, to json.
I focused on making the system as easy to use and as easy to add new categories as possible.

Due to issues related to logging to world at global creation logger is now a byond real, which is created directly before Master

Log categories support versioning, secret flagging, and sub-category filtering. Although all of this is entirely optional for coders.
If you ever want to add a new category and use it, all you need to do is make the barebones category datum and the define.
I've kept existing procs such as log_game, and simply turned them into a wrapper for Logger.Log(xxx, ...)
## Why It's Good For The Game

Makes processing and filtering logs much easier in the future, while only minimally downgrading log crawling experience.
I am also working on a log viewer frontend for admin usage however that will take a little bit longer to finish up.
Also makes special logging and data tracking much easier thanks to a data list processing implementation and handling
## Changelog
:cl:
server: All logs are now formatted in json, excluding perf and investigations
/:cl:


GoldenAlpharex (approved):
I'll have to test-merge this and to do the config changes on the server first before actually going through with this one, just to be safe.

PR #21360, opened 2023-05-22
[TM] (Maybe) fixes enormous RSC bloat

AffectedArc07:
## About The Pull Request
image

Fixes this shit

The above is very undesirable

## Proof of Testing
None, but the DM ref says it
```
Number of days items that are not in use will be saved in the resource cache (.rsc file). Files uploaded by players are stored in the world's .rsc file for future use. If the file is not used for the specified amount of time, it will be removed to save space.

Setting this value to 0 causes items to be saved for the current session only. This is used by the CGI library, because web browsers cannot make use of server-side caches when uploading files anyway.
```

## Changelog

:cl: AffectedArc07
fix: Fixes TTS making the rsc go to like 500 megabytes
/:cl:

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -17,6 +17,7 @@
 	hub_password = "kMZy3U5jJHSiBQjr"
 	name = "/tg/ Station 13"
 	fps = 20
+	cache_lifespan = 0 // SKYRAT EDIT - Makes sure TTS shit isnt kept past rounds
 	//map_format = SIDE_MAP // SKYRAT EDIT - TODO: WALLENING - REMOVE THIS (hopefully the visual z-fighting issues will have been ironed out by then)
 #ifdef FIND_REF_NO_CHECK_TICK
 	loop_checks = FALSE


AffectedArc07:
> You currently have a negative Fix/Feature pull request delta of -6. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

Go away

Iamgoofball:
good catch, we use the CDN upstream so we didn't have this occuring

hugmeorelse:
@Useroth papa, i desire the cdn

AffectedArc07:
Seems to be fixed
image

code/world.dm
```diff
@@ -17,6 +17,7 @@
 	hub_password = "kMZy3U5jJHSiBQjr"
 	name = "/tg/ Station 13"
 	fps = 20
+	cache_lifespan = 0 // SKYRAT EDIT - Makes sure TTS shit isnt kept past rounds
```
AffectedArc07, line 20: Is this sufficient for nonmodular markage.

Zonespace27: yep, for simple things like this it's generally preferred to be a modular override of `/world` in the `modular_skyrat` folder, but it's not a massive deal for a one-liner like this

lessthnthree (approved):

GoldenAlpharex (approved):
Don't know why I haven't merged this yet.

PR #21382, opened 2023-05-23
Blacksmiths across the galaxy in shambles :: You can only forge with metals

Paxilmaniac:
## About The Pull Request

Adds a whitelist of materials that can be used in forging, which includes every metal and metal-like material we have. Why do this you ask? Because people forging with meat, wood, paper, whatever else you can think of that's **not** a metal seems rather quite silly to me.

People won't be able to forge weapons and armor out of meat anymore.

## Proof of Testing

image

## Changelog

:cl:
fix: Forging can only be done with metals, and not bone/wood/meat/paper/diamonds/whatever else
/:cl:

### DIFF
--- a/modular_skyrat/modules/reagent_forging/code/forge_items.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge_items.dm
@@ -15,6 +15,10 @@ GLOBAL_LIST_INIT(allowed_forging_materials, list(
 	/datum/material/alloy/plasteel,
 	/datum/material/alloy/plastitanium,
 	/datum/material/alloy/alien,
+	/datum/material/cobolterium,
+	/datum/material/copporcitite,
+	/datum/material/tinumium,
+	/datum/material/brussite,
 ))
 
 /obj/item/forging
--- a/modular_skyrat/modules/reagent_forging/code/forge_items.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge_items.dm
@@ -1,3 +1,22 @@
+GLOBAL_LIST_INIT(allowed_forging_materials, list(
+	/datum/material/iron,
+	/datum/material/silver,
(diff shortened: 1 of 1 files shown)


tf-4 (approved):
No more diamond sword 😭

Paxilmaniac:
> No more diamond sword 😭

Its joever

Hatterhat:
its so joever

kittysmooch:
speedclose. repository ban. dispatch hunter-killers droids

Samarai:
I can understand removing silly materials like paper, cardboard, and meat. But this also lumps in bone, plasma, plasmaglass, glass, bluespace crystals, and other less silly materials for forging that look rather pretty when used.

tf-4:
actually yeah add bone, that makes sense

softcerv:
It would be amusing if later down the line, the ability to forge with things like paper and cardboard were reintroduced and ended up giving the items comedic effects. Using cloth to make a plush sword that does 0 damage would be funny.

softcerv:
I think having sandstone included in this list would make sense.

Paxilmaniac:
> I think having sandstone included in this list would make sense.

No, no I don't think so.

HoodMilk:
the death of pizza sword

modular_skyrat/modules/reagent_forging/code/forge_items.dm
```diff
@@ -1,3 +1,26 @@
+GLOBAL_LIST_INIT(allowed_forging_materials, list(
```
ghost, line 1: should this be alphabetized or otherwise sorted?

Paxilmaniac: It is sorted just in "oh right I need to add that material too" order

PR #21403, opened 2023-05-24
Adjust default exploitable text to be more informative

Funce:
## About The Pull Request
Small update to default exploitable text

No caps, and better explanation

## Proof of Testing
Surely I don't need that :^)

## Changelog

:cl: Foxtrot (Funce)
config: Default Exploitable Text has been adjusted to make more sense.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/flavor_defines.dm
+++ b/code/__DEFINES/~skyrat_defines/flavor_defines.dm
@@ -1,7 +1,7 @@
 /// How many characters will be displayed in the flavor text preview before we cut it off?
 #define FLAVOR_PREVIEW_LIMIT 110
 /// The default value that will go in any new player's exploitables.
-#define EXPLOITABLE_DEFAULT_TEXT "Used by antagonists. DO NOT PUT SEXUAL THINGS IN HERE. This is where you put flaws that can be exploited in any way. This will be viewable by antagonists if you modify this string, but only if there's anything at all in this box."
+#define EXPLOITABLE_DEFAULT_TEXT "Put information here that is valuable for an antagonist to exploit. This is for them to gain a leverage in order to get access to what they need among other things. While you can put sexual elements in here, that is not the main focus."
 /// The length of records at which they will not show up, to prevent empty records from appearing.
 #define RECORDS_INVISIBLE_THRESHOLD 0
 /// The message displayed when someone received the View Crew Exploitables verb.


code/__DEFINES/~skyrat_defines/flavor_defines.dm
```diff
@@ -1,7 +1,7 @@
 /// How many characters will be displayed in the flavor text preview before we cut it off?
 #define FLAVOR_PREVIEW_LIMIT 110
 /// The default value that will go in any new player's exploitables.
-#define EXPLOITABLE_DEFAULT_TEXT "Used by antagonists. DO NOT PUT SEXUAL THINGS IN HERE. This is where you put flaws that can be exploited in any way. This will be viewable by antagonists if you modify this string, but only if there's anything at all in this box."
+#define EXPLOITABLE_DEFAULT_TEXT "Put information here that is valuable for an antagonist to exploit. This is for them to gain a leverage in order to get access to what they need among other things. While you can put sexual elements in here, that is not the main focus."
```
ghost, line 4: Why is it changing from "Do not put sexual things in here" to "You can put sexual things in here"? There's really zero reason for exploitables to have any sexual content

tf-4:
Is this host request

Funce:
Yes

code/__DEFINES/~skyrat_defines/flavor_defines.dm
CliffracerX, line 4: This update makes it less clear where and when the information is shown, to speak nothing of changing the focus towards kink content that not every antag is going to want to engage with- not to mention adding pressure to fill it out, since it no longer mentions that the text is only displayed if its contents have been changed.

> Put information here that could be learnt of & exploited by knowledgeable antagonists for leverage, such as personal flaws or traumas in your character's medical history.  Only include information you that you are okay with not having to reveal through roleplay.  The contents of this field will only be shown to antagonists if modified & non-empty.  Sexual elements should not be the main focus, and if included, must follow policy.

nikothedude:
god howd i not see this

the way exploitables check to see if it should be hidden is basically seeing if it is eithe rempty or has the default string, by changing the default string, people with already made characters and default exploitables will suddenly appear as if they have exploitables, only for people to see the old default text

it may be worth adding some code to automatically replace the old text when exploitables are laoded

yes, i coded it weirdly, sue me, when i made the pr i was still baby new to coding

code/__DEFINES/~skyrat_defines/flavor_defines.dm
hugmeorelse: > Why is it changing from "Do not put sexual things in here" to "You can put sexual things in here"? There's really zero reason for exploitables to have any sexual content

Because it was never meant like that.

Funce:
Ah good to know

PR #21385, opened 2023-05-24
Voidraptor chemistry access fix & chemistry re-shuffling

KathrinBailey:
## About The Pull Request
• Fixes Voidraptor's little chemistry smoke room. This thing is annoying as hell, and I'd rather just yeet it entirely, but I'll settle for just fixing the door's access instead.

• Replaces reinforced tables in the chemical storage closet with glass tables. This closet is almost always used as a chemical press room, this makes it consistent with the rest of medbay, which uses glass tables, and removes some busy-work from every single round when the chemist comes to remove the table for space for their chem presses.

• Moves all of the west tables to the east side of the chemlab, granting an extra column of tiles space for chemists to work with. I've personally found Voidraptor to be very cramped for chemlab set-ups, so this helps a bunch.

## How This Contributes To The Skyrat Roleplay Experience
Effective utilisation of space means less busy-work chores for the chemist that don't involve doing their job before they can do their job. Bonus: less chemists spear-chucking at that windoor to get to those six tiles of extra space where the smoke machine is.

## Proof of Testing
image
image
image
(it used to be 'chemistry'!)

<details>
<summary>Screenshots/Videos</summary>
 
image
</details>

## Changelog

:cl:
fix: Voidraptor smoke machine room in the chemistry lab can now be opened by chemists/CMO.
qol: Voidraptor chemlab reorganised for better accessibility and reduced roundstart chores, so chemists can get to doing chemistry faster.
/:cl:

PR #21404, opened 2023-05-24
Voidraptor Optimised Kitchen

KathrinBailey:
## About The Pull Request
Re-organises the kitchen to prioritise placing the most often used cooking equipment together, whilst the least important cooking equipment is moved to the side.

Fire cabinet moved in the hallway to allow space on the wall for the requests console.

Makes the kitchen probably the least frustrating of all the kitchens in the map pool. Less frustration means more fun and less time wasted running around.

## Kitchen Heatmap
**Before**
image
**Before**
Green area is prime real estate, the most used kitchen equipment should be here, so you can use them without even having to move.
Yellow is reachable through moving a single tile. Important things that don't fit in green should be here.
Orange is the last few tiles we can reach without having to walk around the kitchen island.
Red is the death-zone. If it's in here, it really isn't important. It is smelly, British, and has poor hygiene. Home of the deep fryer and the two vendors.
**After**
image
**After**

## Proof of Testing

<details>
 I only kinda shuffled things around, I moved the APC and it's definitely powered. The requests console also had no helpers, so...
</details>

May conflict with #21385 ?

## Changelog

:cl:
qol: Voidraptor kitchen re-organised to make all of the service mains out there shed a tear of happiness.
/:cl:


Jolly-66 (changes requested):
Could you perchaps find another spot for the microwave? It being in the corner like this aesthetically doesn't look nice.

It functions without issue, yes, but it just looks bad. I'd recommend removing the grill, honestly.

image

Jolly-66 (approved):

PR #21447, opened 2023-05-26
[MIRROR] Fixes stacking machine and adds a drone dispenser to Tram as well as the required map items list [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75653
--------------------
## About The Pull Request
Fixes #75611
Fixes #75609

This makes a small change to a few tiles outside of Science. There were three turfs that were never actually linked to any of the modular maintenance areas, so they just sat there as dirt turfs with nothing on them. I changed them to metal so they'll fit with the surroundings in each template and put the fab there, since we want it to show up in all three of them. I can go back and add that area to each of the templates instead if MMMiracles thinks that's a better choice, but I feel like this works fine.

Also adds the drone dispenser to required map items so unit tests will fail if they're not present.

## Why It's Good For The Game
Map consistency and bug fixes are good.

## Changelog
:cl:
fix: Adds a drone dispenser to Tramstation.
fix: The stacking machine on Tramstation no longer has incorrect input and output directions at roundstart.
config: Drone dispensers are now listed as required map items for passing unit tests. If one is not present on your map, it will fail the test.
/:cl:


Jolly-66:
I didn't mean to fix this ignore it

Jolly-66 (approved):

PR #21458, opened 2023-05-26
ezdb - A one click script to quickly setting up a development database (#75053)

tf-4:
# Original PR: https://github.com/tgstation/tgstation/pull/75053

---

https://user-images.githubusercontent.com/35135081/235344815-8e825ba9-52cf-44e8-b8e2-a2aeb5d47276.mp4

- Downloads a portable MariaDB (doesn't pollute your main system)
- Sets up a database with a random password on port 1338 (configurable)
- Installs the initial schema
- Every time after, will run updates

Major versions right now explicitly escape hatch, because those historically come with something like a Python script, and I do not want it to pretend to work.


Useroth:
Could you by any chance fix the versioning to ours in the database changelog? We've had some of our own mods on our end

PR #21534, opened 2023-05-30
moves advanced shrapnel mines (are these even used anywhere?) from having their own entire folder to just being in the mapping folder

Paxilmaniac:
## About The Pull Request

Advanced shrapnel mines (mines that only go off if a human steps on them) were just sitting in a module folder called "planet" which I can only assume we just do not use anymore.

Well, yeah, we don't use it anymore, because with the moving of the only file that was in there the folder has just been deleted.

It gets rid of a module folder we quite literally were not using anymore.

## Proof of Testing

No behavior was changed at all.

## Changelog

:cl:
code: The code for advanced shrapnel mines now sits in the mapping folder, not its own folder named "planet" for whatever reason
/:cl:


ghost:
This is indeed entirely unused, you could likely just remove it

Paxilmaniac:
> This is indeed entirely unused, you could likely just remove it

sorry babe removals hurt gbp and I NEED positive again so I can make more gun code

Zonespace27 (approved):
cool beans

PR #21564, opened 2023-05-31
[MIRROR] Abductors can now use batons in general. (not just their specialized baton) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75561
--------------------
## About The Pull Request
Abductors can now use any baton instead of just their special baton.
## Why It's Good For The Game
Abductors are a bit fucked if they lose their baton and are unable to buy one from their shop. Sure they can just use flashes or maint-fu but those methods are hard and or unreliable. (if they manage to lose their baton, they probably aren't that robust anyways). This PR gives more leeway if you fuck up as an abductor.
## Changelog
:cl:
balance: The mothership which abductors came from has worked very hard to train their field agents how to use batons from their specimen's habitat
/:cl:


GoldenAlpharex (approved):
Should be good to go!

PR #21621, opened 2023-06-05
[SEMI MODULAR] Removes subtler and replaces it with a verb to toggle ghost visibilty

softcerv:
## About The Pull Request
Adds in a verb that allows the parent mob to toggle whether or not their subtle and whisper messages are visible to ghosts.
Additionally, the functionality from subtler has been moved into subtle. This verb is automatically disabled if the user of the verb moves out of the area they are currently in.

This makes it easier to hide messages relating to adult roleplay, which is beneficial for both ghosts and participants in adult roleplay. 

While this does somewhat fill the same niche as subtler, subtler requires the user to do additional typing to mimic a whisper. This also should make it less likely for people to send a message with subtle that they thought would be hidden from ghosts.

## Proof of Testing

<details>
  
The verb disabled:

User PoV:
image

Ghost PoV:
image

The verb enabled:

User PoV:
image

Ghost PoV:
image

Leaving an area resets the verb:
image

</details>

## Changelog

:cl:
qol: Introduces a verb that toggles whether or not whisper and subtle messages are sent to dead chat.
del: removes the subtler verb, moving it's functionality into subtle.
/:cl:


SomeRandomOwl:
Adding a extra button for this and keeping the subtler button seems kinda redundant, this should honestly remove the subtler button if it is going to add this button that makes subtle do the exact same thing as subtler otherwise its just a redundancy

softcerv:
> Adding a extra button for this and keeping the subtler button seems kinda redundant, this should honestly remove the subtler button if it is going to add this button that makes subtle do the exact same thing as subtler otherwise its just a redundancy

Subtler does not do the same thing as this button. Subtler allows for the user to choose a specific target or range to send their message to in addition to not having it sent to deadchat. Subtle doesn't provide those same options.

SpaceVampire:
> Subtler does not do the same thing as this button. Subtler allows for the user to choose a specific target or range to send their message to in addition to not having it sent to deadchat. Subtle doesn't provide those same options.

Then you should just move both of those options to subtle.
Because this makes the whole "Anti-Ghost" part of it redundant.
That seems like the best way to handle it.

YakumoChen:
I'm worried that preventing ghosts from hearing whispers will have implications with a much farther reach than is realised, considering whispers are almost as common of speech as regular say (given every radio message is whispered)

I don't really know if this will be maintainable either, I don't think it's worth the risk that isn't solved by just using subtler and typing an extra word and quotation marks. Say code really should avoid non-modular edits unless absolutely necessary. 

This isn't beneficial to ghosts either, they already have preference options to mute most of this if they choose.

SomeRandomOwl:
actually chen has a decent point, whats preventing people from using this on radios? and such?

softcerv:
> actually chen has a decent point, whats preventing people from using this on radios? and such?

Using a radio while whispering does this:
image

I don't think this is a tangible issue.

softcerv:
> I'm worried that preventing ghosts from hearing whispers will have implications with a much farther reach than is realised, considering whispers are almost as common of speech as regular say (given every radio message is whispered)

Admins are now able to hear ghost muted whispers. This is now fine from an admin perspective.

YakumoChen:
> > actually chen has a decent point, whats preventing people from using this on radios? and such?
> 
> Using a radio while whispering does this: image
> 
> I don't think this is a tangible issue.

If you ;say into radios it forces your character to whisper, is the difference

softcerv:
> > > actually chen has a decent point, whats preventing people from using this on radios? and such?
> > 
> > 
> > Using a radio while whispering does this: image
> > I don't think this is a tangible issue.
> 
> If you ;say into radios it forces your character to whisper, is the difference

But if you do that, it'll also say the message over radio, which ghosts can listen into. In the rare case that nothing ends up going through, I don't think anything of value would be lost.

lessthnthree:
If we're going this path I would like to see the subtler's directional function be folded into subtle, and remove the subtler verb.

lessthnthree (approved):

tf-4 (changes requested):
I don't think whispers should be hideable

YakumoChen:
I still stand by "non modular say code is not worth the maintainability cost"

I also think there should be a clear and defined separation between gameplay and ERP, that subtler subtly provided, perhaps the toggle should make a visible indicator for "bedroom combat"

SomeRandomOwl:
CI required to ERP, as funny as that sounds it's not practical

SpaceVampire:
So we are going to remove the whisper part here, right?
It seems better then adding a bunch of non-modular code and weird edge cases.
So that people don't have to press the subtle button sometimes.
The toggle verb option is a great idea, but the rest is overkill for Skyrat.

softcerv:
>So we are going to remove the whisper part here, right?
>It seems better then adding a bunch of non-modular code and weird edge cases.
>So that people don't have to press the subtle button sometimes.
>The toggle verb option is a great idea, but the rest is overkill for Skyrat.

**No, I am not going to remove that part.** I am willing to modify it to ensure that people don't leave it on by accident 24/7, but I feel like this is something that is worthwhile to have in the game.

code/modules/mob/living/living_say.dm
```diff
@@ -371,6 +371,11 @@ GLOBAL_LIST_INIT(message_modes_stat_limits, list(
 
 	var/list/listened = list()
 	for(var/atom/movable/listening_movable as anything in listening)
+		// SKYRAT EDIT DISABLE GHOST VIEWING
+		var/mob/listener = listening_movable
+		if(is_speaker_whispering && isobserver(listening_movable) && !listener?.client?.holder && HAS_TRAIT(src, TRAIT_NO_GHOST_MESSAGES))
```
GoldenAlpharex, line 376: This should probably be moved up, where there's already an `is_speaker_whispering` check, as that would make this change far less intrusive and potentially hot.

modular_skyrat/modules/verbs/code/disable_ghost_viewing.dm
```diff
@@ -0,0 +1,33 @@
...
+	set desc = "Toggles whether or not whispers and subtles done by you are visible to dead chat."
+	set category = "IC"
+
+	if(GetComponent(/datum/component/no_ghost_messages))
+		qdel(GetComponent(/datum/component/no_ghost_messages))
+		to_chat(src, span_notice("You will now send whipsers and subtles to dead chat."))
+		return TRUE
+
+	AddComponent(/datum/component/no_ghost_messages)
+	to_chat(src, span_notice("You will no longer send whipsers and subtles to dead chat."))
```
GoldenAlpharex:
```suggestion
		to_chat(src, span_notice("You will now send whispers and subtles to dead chat."))
		return TRUE

	AddComponent(/datum/component/no_ghost_messages)
	to_chat(src, span_notice("You will no longer send whispers and subtles to dead chat."))
```

modular_skyrat/modules/verbs/code/disable_ghost_viewing.dm
```diff
@@ -0,0 +1,33 @@
...
+	if(!ishuman(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	RegisterSignal(parent, COMSIG_EXIT_AREA, .proc/remove_self)
+	ADD_TRAIT(parent, TRAIT_NO_GHOST_MESSAGES, INNATE_TRAIT)
+
+/// Removes the component from the parent mob while sending them a message. This is called whenever the parent leaves an area.
+/datum/component/no_ghost_messages/proc/remove_self()
+	var/mob/living/parent_mob = parent
+	to_chat(parent_mob, span_boldwarning("Due to leaving an area, your whispers and subtles will be sent to dead-chat. To re-disable sending, use the Toggle Ghost Viewing verb."))
```
GoldenAlpharex:
```suggestion
	to_chat(parent_mob, span_boldwarning("Due to leaving an area, your whispers and subtles will once again be sent to dead-chat. To re-disable sending, use the Toggle Ghost Viewing verb."))
```

GoldenAlpharex (changes requested):
I'm also still not 100% convinced on the whisper part of this. Whispers are still only usable with the same effective range as subtles, it's just that rather than a hard cutoff, it's all garbled. I'm not sure that a slight convenience, with the opportunity for administrative headaches (no matter how we try to avoid them, they'll still be present) and some more overhead on an otherwise rather hot part of the code (due to the sheer amount of people talking at once on the server).

I know that the main reason for whispers to be included was their usage in TTS, but for now I don't believe that the overhead is worth the potential of using it for a feature that we currently can't use due to its atrocious performance. We can find a better, modular way of enabling that later if that's still something we want to do.

SomeRandomOwl:
I still 100% disagree with the hide whispers part, it seems like an unnecessary part that has the potential to cause issues from the administrative side of the things.

tf-4 (changes requested):
There are no maintainers interested in whispers being hidden. Please remove that part.

softcerv:
> There are no maintainers interested in whispers being hidden. Please remove that part.

This is something that people very clearly can't easily keep on 24/7 because it automatically disables itself when people change their area. The actual effects on normal gameplay would be negligible at best.

Anyways, if whispers aren't hidden, there isn't really a point in having this PR open. Change your mind or talk to me about something that would make the whispering part of this viable and I'll open this back up again.

PR #21636, opened 2023-06-05
[MODULAR] Fix hardcoded MCC rack max_batteries

out-of-phaze:
## About The Pull Request
Replaces a hardcoded length check with the already-existing `max_batteries` var. It changes nothing at the moment because `max_batteries` is currently always 4.

Cleaning up the code in the modular Skyrat folder.

## Proof of Testing
Prove `max_batteries = hardcoded length check`.
1. Let max_batteries = 4 and the hardcoded length check = 4.
2. Substitute: `4 = 4`.
3. `4 = 4` is true, therefore `max_batteries = hardcoded length check`.

Oh, did you not mean a mathematical proof? I mean... it's pretty obvious, isn't it?

### DIFF
--- a/modular_skyrat/modules/multicellcharger/code/multi_cell_charger.dm
+++ b/modular_skyrat/modules/multicellcharger/code/multi_cell_charger.dm
@@ -63,7 +63,7 @@
 		if(inserting_cell.chargerate <= 0)
 			to_chat(user, span_warning("[inserting_cell] cannot be recharged!"))
 			return
-		if(charging_batteries.len >= 4)
+		if(charging_batteries.len >= max_batteries)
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/multicellcharger/code/multi_cell_charger.dm
```diff
@@ -63,7 +63,7 @@
 		if(inserting_cell.chargerate <= 0)
 			to_chat(user, span_warning("[inserting_cell] cannot be recharged!"))
 			return
-		if(charging_batteries.len >= 4)
+		if(charging_batteries.len >= max_batteries)
```
Zonespace27:
```suggestion
		if(length(charging_batteries) >= max_batteries)
```

Zonespace27 (approved):

PR #21647, opened 2023-06-05
[MODULAR] Clean up unnecessary `as anything` loop

out-of-phaze:
## About The Pull Request
`as anything` does nothing for ***untyped*** loops and is unnecessary here. However, we can save a line by using `as anything` with a typed for loop of the type we want to do `initial()` on.

Cleaning up the code in the modular Skyrat folder.

## Proof of Testing
Semantically and practically equivalent but one line shorter.


Zonespace27 (approved):
I cannot wait to just be able to use :: instead of this initial junk

PR #21649, opened 2023-06-05
[MODULAR] Remove multi_type var in favor of istype

out-of-phaze:
## About The Pull Request
Removes the multi_type magazine var and replaces it with just an istype checking if it's the type that defined that var to begin with. Much cleaner!

Cleaning up the code in the modular Skyrat folder. I'm serious here, this kind of code should be outright forbidden. `foo:bar` and `in vars` are both terrible patterns.

## Proof of Testing
`multi_type` was only ever defined/set on the one type anyway.

### DIFF
--- a/modular_skyrat/modules/ammo_workbench/code/ammo_workbench.dm
+++ b/modular_skyrat/modules/ammo_workbench/code/ammo_workbench.dm
@@ -121,7 +121,7 @@
 	var/ammo_caliber = initial(ammo_type.caliber)
 	var/obj/item/ammo_casing/ammo_parent_type = type2parent(ammo_type)
 
-	if(istype(loaded_magazine, /obj/item/ammo_box/magazine/multi_sprite) && ammo_caliber == initial(ammo_parent_type.caliber) && ammo_caliber != null)
+	if(istype(loaded_magazine, /obj/item/ammo_box/magazine/multi_sprite) && ammo_caliber == initial(ammo_parent_type.caliber) && !isnull(ammo_caliber))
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/ammo_workbench/code/ammo_workbench.dm
```diff
@@ -121,9 +121,8 @@
 	var/ammo_caliber = initial(ammo_type.caliber)
 	var/obj/item/ammo_casing/ammo_parent_type = type2parent(ammo_type)
 
-	if("multitype" in loaded_magazine.vars)
-		if(loaded_magazine:multitype && ammo_caliber == initial(ammo_parent_type.caliber) && ammo_caliber != null)
-			ammo_type = ammo_parent_type
+	if(istype(loaded_magazine, /obj/item/ammo_box/magazine/multi_sprite) && ammo_caliber == initial(ammo_parent_type.caliber) && ammo_caliber != null)
```
Zonespace27:
```suggestion
	if(istype(loaded_magazine, /obj/item/ammo_box/magazine/multi_sprite) && ammo_caliber == initial(ammo_parent_type.caliber) && !isnull(ammo_caliber))
```

Zonespace27 (changes requested):

Zonespace27 (approved):

Hatterhat:
tragically there was another really niche usecase in regards to the 7.62 surplus clips but also i agree that this was a Really Bad Pattern

PR #21657, opened 2023-06-05
Job2Icon Tgui now supports Alt titles

Funce:
## About The Pull Request
@OrionTheFox I think I managed to find a way for every interface to work it.

This adds all the alt jobs to the Job2Icon module, which is used in orbiting, records and a lot of tg based uis.

Icons that match the alt jobs rather than wigging out is a good thing

## Proof of Testing

<details>
Research Director Icon

image

image

Orbit Observe Panel
image

</details>

## Changelog

:cl: Foxtrot (Funce)
fix: Most tgui interfaces that use job icons should now correctly show alt-jobs
/:cl:


ghost:
beeautiful
now I just need to find out if there's a way to make searching in records/PDA come up with stuff based off real job name instead of **only** the selected alt title...

lessthnthree (approved):

PR #21708, opened 2023-06-07
restrict imbuing to ashwalkers & icecats

jjpark-kb:
## About The Pull Request
restricts imbuing to only ashwalkers & icecats

## How This Contributes To The Skyrat Roleplay Experience
this should limit the amount of chemical warfare you experience on the station and in space.

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
  
image

</details>

## Changelog

:cl:
balance: Reagent Imbuing (from forging) is strictly for ashwalkers & icecats
balance: Forges can only be fully upgraded (legendary smiths) by ashwalkers & icecats
balance: Some of the forge upgrades have been moved around to accommodate reagent imbuing being ashwalker/icecat only
/:cl:

PR #21709, opened 2023-06-07
Add isashwalker macro

out-of-phaze:
## About The Pull Request
Replaces various ashwalker checks with an `isashwalker` macro that fits the format of the other species is helpers.

Came up in #21708.


softcerv:
I feel like this may be more fitting upstream.

PR #21726, opened 2023-06-08
[MIRROR] Improves mob resizing (and UpdatePaths). [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75892
--------------------
## About The Pull Request
So, basically, whenever an atom is scaled up or down through editing of its `transform` variable, these changes apply in all directions in a way the movable is always centered on its location. While good, this can make mobs look like they're not quite standing on the tile they are, with their legs poking onto the turf below or not quite reaching the edge of the turf they're standing on.

It's a visual peeve, and outside that, it doesn't bear any effect on the gameplay, but I need to do something about it because of something else I'm working on that deals with another visual botherance.

I had thought of just copying whatever Skyrat did for their body size pref at first, but it turned out that translating the `transform` variable resulted in rotation looking a little askew. So I had to take the longer path to better incorporate it with the body pixel offsets that mobs use. Speaking of Skyrat, they should update their body size pref code accordingly after this is merged.

I haven't done a "before" screenshot, but you can see in this one how humies of different sizes are nicely aligned with the turfs they're on now.**
scrnshot1

EDIT: This has also become a PR to fix UpdatePaths after I discovered it doesn't support catching props with any set value and that it also has an undocumented (only docced in the .py script to be precise) parameter that is currently broken (prop_value isn't recognized a string type variable). Documentation for the two has been added to boot.

~~Blame me for the unatomic change but my UpdatePaths scripts needs it.~~

## Why It's Good For The Game
Aligns resized mobs to the tile they're standing on. It looks a bit better this way. This also fixes pulls ignoring body offsets and makes carbon/update_transform use the appropriate variable to calculate the new pixel_y value.

## Changelog

:cl:
fix: pixel offsets from grabbing other mobs now respect the body offset variables.
refactor: Changed some code so that resized mobs no longer look like they're poking over the below tile with their legs or not even reaching the bottom edge of the tile they're on.
/:cl:

**I was too lazy to clear the area to make enough space for two rows of spessmen so I took two screenshots and put them together in paint, then I noticed how I hadn't vertically aligned the two screenshots properly.


Jolly-66 (approved):
Maps (should) be fine, code doesn't compile.

tf-4 (approved):

PR #21762, opened 2023-06-10
Performs some much needed pruning of spacevine.dm

tf-4:
Untangles our edits to spacevine-related things and splits what's left out of master_files and into a spacevines module. We had this whole commenting-out thing for the sake of like 10 non-modular edits:)
Uncomments the /tg/ default files.

I'm so done with this stuff https://github.com/Skyrat-SS13/Skyrat-tg/pull/21752/files

I've preserved the intended balance behaviours as best I can, this is purely a code change.


jjpark-kb (approved):
merging this so we can start working on vines more

PR #21776, opened 2023-06-11
[Not Entirely Modular] Removes bundles from opfor : Removes Microbomb implants from Ninja + Contractor

RatFromTheJungle:
-
-
## About The Pull Request

following a staff meeting, and per host request, removes the surplus bundles from the opfor panel(as they aren't really approvable), and the microbomb implant from both ninja and contractor

## How This Contributes To The Skyrat Roleplay Experience

Antags opting to immediately qdel themselves when they're about to lose, tends to encourage them to only play when winning

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
  
</details>

## Changelog

:cl:
del: Removed Surplus bundles from opfor.
del: Remove microbomb implants from Contractors and Ninjas.
/:cl:

PR #21783, opened 2023-06-12
gives lone infiltrators a microbomb

zeroisthebiggay:
## About The Pull Request

idk why they don't have one already but they're a diet nukie full of loot that shouldn't be dropped when they die probably

edit: the box in their bag also has a big pamphlet about how you have a bomb inside of you and also an atropine medipen despite the fact you physically don't so technically it's a fix

uh it makes lone infiltrators look less like loot bags i guess so maybe they won't be validhunted as much

## Proof of Testing

it's a webedit

<details>
  
</details>

## Changelog

:cl:
add: lone infiltrators get a microbomb
/:cl:

### DIFF
--- a/modular_skyrat/modules/Midroundtraitor/code/event.dm
+++ b/modular_skyrat/modules/Midroundtraitor/code/event.dm
@@ -56,6 +56,7 @@
 /datum/outfit/syndicateinfiltrator
 	name = "Syndicate Operative - Infiltrator"
 
+	implants = list(/obj/item/implant/explosive)
 	uniform = /obj/item/clothing/under/syndicate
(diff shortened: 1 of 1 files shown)


GoldenAlpharex:
We're already removing microbomb implants from other antags, we're not going to add one to this one, sorry.
(https://github.com/Skyrat-SS13/Skyrat-tg/pull/21776)

Quink-Fix:
IMO, it makes more sense for Lone Ops to have a microbomb than Contractors (Lone Ops are all-or-nothing suicide operatives, Contractors are not.) I don't see an issue with this change in particular.

PR #21784, opened 2023-06-12
[MIRROR] Experimental Mineral Balance PR. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75437
--------------------
## About The Pull Request

Good morning fokes, this PR is going to attempt to tweak around some of the values of mineral costs in-game in the hopes of making the mineral economy more engaging to the base gameplay loop, in the hopes that by making it more self-realized, it would allow for better integration with the rest of the gameplay loop as well as so that it doesn't completely overshadow the credit economy. That said, **this pr only makes changes to mined and station minerals**.

So, the base change is as follows: items with designated costs of minerals below 1000 minerals, as designated in #75052 as `SMALL_MATERIAL_AMOUNT`, have been left alone. Any item above 1000 or 2000, from `HALF_SHEET_MATERIAL_AMOUNT` and `SHEET_MATERIAL_AMOUNT` have been adjusted, as sheets will now contain **100 minerals each**.

For example, if an object costed 4000 iron, 1000 glass, and 500 diamond with the old numbers, with the changes to these defines, that item would cost 2000 iron, 500 glass, and 500 diamond.

## Why It's Good For The Game

Arcane has been doing some statistics now that he's back in the land of the living. So, I broke down a 40 round aggregate of silo logs, excluding rounds where the silo was left untouched (as, I can't imagine a round where nothing was both lathed or mined would be statistically relevant). These rounds had their minerals obtained tallied, and the amount spent and obtained were compared, giving a net total. The amount consumed as a percentage was then obtained as an average of all rounds, as a measurement of average "health" of that mineral within the material economy.

The results were as follows:
image

In the average round, iron and glass being the most consumed mineral makes sense, since the most useful and most bountiful items are typically iron/glass exclusives. Not to mention, if a tool has to be obtained later in the shift, it's either research locked or it has to be very, VERY good to warrant waiting for it. So that value makes sense. Iron is also typically the most in-demand mineral, as well as the most requested material as a result of having to fix hull breaches. Glass by comparison has issues on some maps like icebox, but overall it's useful enough that it has a much MUCH higher quantity obtained than other materials.

Other mined minerals typically don't break over 50 sheets (100,000 in this metric), and are also consumed less than 30% at maximum, and 14% at the lowest. This is for a number of reasons, but the big two are as follows:
* "Rare" resources are rarely taken in quantities greater than the previous SMALL_MATERIAL_AMOUNT on their most valuable parts, which means that despite the quantity that they come in at, there's no real scarcity of the resource, which makes the bulk of rare material acquisition meaningless after the first load of ores comes in.
* "Rare" resources don't have the same level of usefulness as iron/glass, as they are unique, those unique properties don't help the station actually function more effectively. Gold can't patch a hole the size of medbay, and silver cannot help to refill oxygen to a station doomed to recall. This is hyperbole, I know, but there's nothing that really drives needing specific rare materials due to them not having a truly unique identity within the game. IE: If high cost silver items were a driving force in making sterile workshops in medical, seeing an empty silver reserve becomes an indicator of medical department's state.

So, where am I going with all this?

This PR should serve as an experiment in order to see if by pushing the lower end of mineral consumption, usually taken up by "rare" minerals, we see a better percentage of mineral usage causing trade-offs in gameplay after the first 15 minutes of a shift. By normalizing regular sheets to a standard 1000 units, it also means that finally one unit of a sheet can be a "millisheet", as we've all been demanding for years (this is sarcasm but the uniformity is rather nice).

DNM until it's been testmerged and we can compare some logged rounds silo data, both from an LRP server as well as from MRP. (MRP rounds, being longer in general, gave much higher values while also showing a higher level of surplus, where LRP would be much shorter logs by far, but could have a much wider range of values due to the volatile nature of rounds ending earlier than expected).

## Changelog

:cl:
balance: In general, mineral recipe costs for items in larger amounts has been decreased by 2x.
balance: One sheet now contains a total of 100 units of material, down from 2000. Proportionally the costs of items should remain the same, except for the smaller costs of certain recipes. 
code: Added a python script to the tools folder, letting you see a mineral breakdown of one/several rounds using silo logs.
/:cl:


GoldenAlpharex:
For reference, the PR body isn't totally accurate anymore, if you find this and you're curious, go look at the original PR, linked at the top of this one's body, and read through the comments.

Hatterhat:
yall are gonna want to look over your modular stuff so it uses the new material defines before merging this

lessthnthree (approved):

PR #21787, opened 2023-06-12
[MODULAR][QoL] Action Buttons for leaving and entering Soulcatchers

softcerv:
## About The Pull Request
This PR adds two actions, one granted to ghosts so that they can join soulcatchers, the other granted to soulcatcher souls so that they can leave soulcatchers.

This makes the soulcatcher options that were previously hidden in the verb panel more visible, therefore making soulcatchers easier to use.

## Proof of Testing

<details>
  
Entering:
image

Leaving:
image

Action hidden with pref disabled:
image

</details>

## Changelog

:cl:
qol: entering and leaving soulcatchers now have dedicated action buttons.
/:cl:


tf-4 (dismissed):
I don't think this should be an action button for all ghosts. I don't like the amount of action button clutter we have already, and even though ghosts don't have that many others it's not a pattern I want to encourage particularly, especially for a feature most won't engage with regularly enough to justify it.

A button for leaving it is fine though.

softcerv:
> I don't think this should be an action button for all ghosts. I don't like the amount of action button clutter we have already, and even though ghosts don't have that many others 

If ghosts don't really have that many, if any at all, action buttons, then how would this be contributing to a clutter that doesn't exist for them? While I am in favor of reducing the amount of action icons, I do think this is a justifiable use for them. Having this be on the same bottom bar as the other options would create non-modular issues.

Would it help at all if ghosts could disable having this action in their prefs?

>it's not a pattern I want to encourage particularly, especially for a feature most won't engage with regularly enough to justify it.
> 
Your reasoning here creates a cycle though:
- Soulcatchers aren't used often
- Because soulcatchers aren't used often, we can't add features to make them more visible
- Due to a lack of visibility, soulcatchers aren't used.

softcerv:
A option to toggle this for ghosts has been added to prefs

SomeRandomOwl:
> A option to toggle this for ghosts has been added to prefs

honestly i think what could be a good compromise is if the join soulcatcher option showed up on orbiting someone who has a soulcatcher

softcerv:
> honestly i think what could be a good compromise is if the join soulcatcher option showed up on orbiting someone who has a soulcatcher

TF4 said that they were fine with the pref toggle as a solution.

tf-4 (approved):

PR #21835, opened 2023-06-14
[MODULAR] Removes the ORGAN_EDIBLE flag from adult organs.

Iamgoofball:
since taking even a single bite out of an organ makes it unusable, and these organs can't be replaced by the crew without the SAD, it is best to not let them be force-fed to be people by accident

:cl: Iamgoofball
fix: You can no longer eat adult organs.
/:cl:


YakumoChen:
We still need to fix disemboweling affecting adult organs, they shouldn't be dropping or removable via surgery at all

Iamgoofball:
frankly, they shouldn't be organs to begin with, but that's a large refactor

Zergspower:
> We still need to fix disemboweling affecting adult organs, they shouldn't be dropping or removable via surgery at all


yeah frankly this is the best route - for many of reasoning

tf-4 (approved):

PR #21841, opened 2023-06-14
[MIRROR] Fixes various bow issues and includes new sounds for the bow [MDB IGNORE]

### LINKED ISSUE #75803: Bows have the low ammo clicking sound when fired
## Reproduction:

Bows play the 'low ammo' clicking sound every time they fire, which isn't really realistic for simple ranged weapon.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75828
--------------------
## About The Pull Request

fixes https://github.com/tgstation/tgstation/issues/75803

Fixes various bow issues;
Bows no longer can be loaded while in a slot other than your inactive hand. 

Dropping bows or putting them into a slot OTHER than your hand now drops the arrow, as intended. No longer hardcoded to do this only if it goes on your back.

Divine bows can fit in the chaplain suit storage, like most other bulky null rods. As a result, has a suit storage sprite.

Drawing, loading and firing the bow now plays sounds instead of being entirely silent. New sounds adapted from here:
https://freesound.org/people/SonoFxAudio/sounds/649337/ <- bow_draw.ogg
https://freesound.org/people/SonoFxAudio/sounds/649334/ <- bow_fire.ogg

cleans up the code a bit as well.

## Why It's Good For The Game

*thunk*

Also because I think the bow being completely silent is utterly jarring and it had some really weird hardcoded issues.

## Changelog
:cl:
fix: Bows properly update their ammo when stored, dropped and stowed. You cannot load a bow without holding it in your inactive hand. You have to hold it to keep the bow loaded.
qol: You can remove an arrow from the bow with alt+click, much like other ballistic weapons.
fix: The divine bow can be properly stored in the chaplain armor's suit storage, much like all other bulky null rods.
sound: Bows have sounds for drawing, firing and loading.
/:cl:


code/modules/projectiles/guns/ballistic.dm
```diff
@@ -396,7 +402,16 @@
...
+<<<<<<< HEAD
 */ // SKYRAT EDIT END
+=======
+
+/obj/item/gun/ballistic/proc/check_if_held(mob/user)
+	if(src != user.get_inactive_held_item())
+		return FALSE
+	return TRUE
+
+>>>>>>> ed0ac95c98b (Fixes various bow issues and includes new sounds for the bow (#75828))
```
tf-4:
```suggestion
*/ // SKYRAT EDIT END

/obj/item/gun/ballistic/proc/check_if_held(mob/user)
	if(src != user.get_inactive_held_item())
		return FALSE
	return TRUE

```

tf-4 (approved):

PR #21931, opened 2023-06-19
[MIRROR] Fixes some stupid airlock sleeps [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75961
--------------------

## About The Pull Request

[A common problem with explosions is an overabundance of sleeping](https://github.com/tgstation/tgstation/commit/6499077a09469ff401c224c0510bc184c663b118)

In an attempt to solve this issue, let's not continue to sleep and do work in door closing if the door is already deleted

(This is caused by firelocks activating due to other adjacent objects deleting, triggering an atmos update, and closing the firelocks before they get bombed. I don't have a elegant way of resolving that core problem, so let's just minimize the impact)

[Nukes a stupid sleep loop in airlock code](https://github.com/tgstation/tgstation/commit/5b16360520526d6f5aa3b511049456b74f33f430)

When an airlock was depowered, it would enter a sleep loop, decrementing its delay by 1 second every well, one second, so long as it had the right wires flipped
This is very stupid

Instead, let's use signals off wire changes and a combo of timer and remaining time var to do this with JUST a timer

Most of the changes here are just swapping over wires to a setter to make signal registration work\

## Why It's Good For The Game

Less sleeping around explosions means less dropped ticks after a bomb goes off. Good just in general
Also this excises dumb boomer code and adds some hooks for other devs to use (we should use wires more man)


GoldenAlpharex (approved):
It works, nice.

PR #21984, opened 2023-06-22
[MIRROR] Fixes a runtime in atom init management [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76241
--------------------

## About The Pull Request

The issue was map verification calling build_cache, which uses the define which enables/disables init values on sleep. We avoid this by using a var on map datums and using that to enable the init value modification only when we are actually loading stuff.

Also fixes a bug in clear_tracked_initialize() where it being called with no values lead to bad values/potentially overriding initialized on accident.

Also also I forgot how for loops worked so this would not have worked regardless

## Why It's Good For The Game

Code should like, function


LemonInTheDark:
critical code, https://github.com/tgstation/tgstation/pull/76179 broke shit bad merge as soon as possible

GoldenAlpharex (approved):
Should be good to go now.

PR #22001, opened 2023-06-23
Makes height turfs not block movement if the mover is either flying or floating

Paxilmaniac:
## About The Pull Request

Height turfs (for example, being in a pool tile and moving onto a normal tile) will no longer block your movement if you are either flying or floating. Some examples of this would be that you can now just float over the edge of a pool tile if gravity is off, or that carp can just fly over the edge of pools without being trapped forever.

There are a lot of things that those who can both fly and float are able to traverse over easily (railings, for example), being able to float or fly over the edge of pool tiles without having to climb over them makes enough sense to me.

## Proof of Testing

<details>

image

And look at that, freedom with no climbing required
image

  
</details>

## Changelog

:cl:
qol: If you are either flying or floating, you can now traverse differently heighted tiles (for example, pool tiles onto a normal floor) without having to climb.
fix: Pool tiles are no longer forever jail if the gravity system shuts off.
/:cl:


tf-4:
nice

tf-4 (approved):

lessthnthree (approved):

PR #22006, opened 2023-06-23
[MIRROR] Emergency Profile Dumps [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75924
--------------------
## About The Pull Request

Adds some hooks to the MC that detect if something ate a ton of real time last tick, and reacts by dumping our current profile into a file

It's really frustrating to see a spike in td in our performance logs, but see no reason in the profile because it's only taken every 5 minutes. This resolves that

I'm throwing this up so mso can give it a look over, not sure if I want to use defines or configs here, taking suggestions

🆑
server: Adds a system to emergency dump profiles if too much time passes between ticks
config: Added configs that control how often emergency profiles are allowed to dump, alongside the threshold for what counts as too much time between ticks
/🆑


vinylspiders (dismissed):

vinylspiders:
See https://github.com/tgstation/tgstation/issues/76286

vinylspiders:
Although this doesn't seem to be the actual source of the linked issue, it greatly exacerbates it and leads to more CI failures.

I think we should still hold off on merging this until it's more under control.

vinylspiders (approved):

PR #22053, opened 2023-06-24
gunset use check

tf-4:
This was missing a check because in the parent it relies on it being reskinnable to run the check.

:cl:
fix: gunsets can't be toggled open at range
/:cl:


ghost:
oops

PR #22116, opened 2023-06-28
[MIRROR] Makes bodytype a carbon mob variable, not a species variable [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76210
--------------------
## About The Pull Request

Having a variable in species that constantly gets modified by outside sources is very hacky, and I don't want that.

Bodytypes are handled by bodyparts in the synchronize_bodyparts() proc, no matter the species. Simply put, they are every bodytype gotten from every bodypart, cached.
As such, bodytypes is pretty much a convenience variable (so we don't have to constantly loop through the bodyparts list), and instead of putting it on species, it'd be better to put it on /mob/living/carbon.

## Why It's Good For The Game

Makes the bodytype of human mobs easier to access by making it direct, instead of having to go through dna and species.

## Changelog

nah


GoldenAlpharex (approved):
Hopefully it actually works now.

PR #22119, opened 2023-06-28
Makes the dna's body size feature use `update_transform()`

Ghommie:
## About The Pull Request
`update_transform()` already prevents the mob from spilling over the southern bounds now when resized. The only difference is that when the mob is lying on the floor now, it goes out of both horizontal bounds, rather than one side or the other but twice as much.

Stops the feature from using its own version of something the upstream now has (except for the maptext height offset, whoopsies).

## Proof of Testing
I already have tested `update_tranform()` more than a few times when working on it. There are a couple things yet to be done like the aforementioned meptext offset, or replacing the pixel y offset macro with a proc that uses the icon height instead of a hardcoded value of 32, or bringing all code of it to `living_update_icons.dm`, both of which are already being PR'd downstream (except it's going slow because it's a bit unatomic), but it's no biggie

## Changelog
N/A


GoldenAlpharex (commented):
Seems to be the case based on what I read from the code, but I take it this doesn't affect maptext height, correct?

Ghommie:
> Seems to be the case based on what I read from the code, but I take it this doesn't affect maptext height, correct?

Mhm! For now update_transform() doesn't do that (it kinda passed over my head), which is why I've left line 186/180 intact.

lessthnthree (approved):

lessthnthree:
@Ghommie are you putting the maptext changes in this PR, or is it ready to go?

Ghommie:
> @Ghommie are you putting the maptext changes in this PR, or is it ready to go?

Actually, I have to mark this as a draft at least until tgstation/tgstation/pull/77065 is merged. Furthermore you've got superfluous code that may override the maptext height of the mob and that's something I've to fix.

tgstation.dme
```diff
@@ -5523,13 +5523,13 @@
 #include "interface\interface.dm"
 #include "interface\menu.dm"
 #include "interface\stylesheet.dm"
+#include "interface\skin.dmf"
```
GoldenAlpharex: Yeah you'll have to fix this so the linters don't cry. I think you just have to move it up one.

code/modules/mob/living/living_update_icons.dm
```diff
@@ -20,7 +20,7 @@
 		current_size *= resize
 		//Update the height of the maptext according to the size of the mob so they don't overlap.
 		var/old_maptext_offset = body_maptext_height_offset
-		body_maptext_height_offset = initial(maptext_height) * (current_size - 1)
+		body_maptext_height_offset = initial(maptext_height) * (current_size - 1) * 0.5 //SKYRATE EDIT: Until it's fixed upstream
```
GoldenAlpharex: I take it it's still not fixed upstream?

GoldenAlpharex (changes requested):

code/modules/mob/living/living_update_icons.dm
Ghommie: Yup, A PR is up but I would rather not delay this another time just because of one line.

GoldenAlpharex (approved):
Cheers, let's go.

Samf3232:
This has broken the .84 size. My character suddenly does not have eyes and this is the only thing to do with scaling since last night. 
image

Samf3232 (commented):
This has broken the .84 size. My character suddenly does not have eyes and this is the only thing to do with scaling since last night.

image

Ghommie:
Curious, I've never thought offsetting the pixel position versus translating the icon would have caused that issue. Does the issue apply to the game world too, or is only found in the character creation menu? 🤔

Samf3232:
it does affect in game as well

Samf3232:
actually, all my non standard sprites are messed up in their own weird ways.

Samf3232:
seems to be fixed now

Samf3232:
and now its doing... whatever this is (look at the white cat) 

https://github.com/Skyrat-SS13/Skyrat-tg/assets/130403276/1eeed363-d0d4-4a0b-8db2-5770ba42ab19

Ghommie:
I'm having a hard time seeing what's wrong with the, uh... white-haired catperson, especially when you're moving your mob around. Would you be kind enough to tell me what the problem even is?

Samf3232:
if you look closely, their eye is changing size depending on where i stand in relation to them

GoldenAlpharex:
> if you look closely, their eye is changing size depending on where i stand in relation to them

That might be because of the Pixel Sizing Method you have set in your prefs, and I doubt it's related to this.

GoldenAlpharex (changes requested):
As was brought up already, this brings up some issues with sprite scaling on humans, where things that didn't use to disappear from sprite rescaling, now do.

Ghommie:
> As was brought up already, this brings up some issues with sprite scaling on humans, where things that didn't use to disappear from sprite rescaling, now do.

Well, paint me green and call me a cucumber, Perhaps the reason lies behind translating transform values allowing decimals while pixel offsetting round (i think down, not up) to the nearest integer. If I'm correct, Translating the icon by the remainder decimals MIGHT fix the issue...
Otherwise I'll have to throw in the towel. The visual renderer is one tough beast.

Ghommie:
Ok, I stand (mostly) true on what I said last week, after quick-testing it earlier at the thunderdome (I couldn't care to take a screenshoot). I've a PR up upstream to solve the issue. Until it's merged, this'll go back to being a draft. Then, I suppose you'll TM this again to see if there are still complaints.

Ghommie:
Let us see if this is still an issue now.

Ghommie:
@GoldenAlpharex Have you testmerged this again now that I've dealed (I believe) with the "slightly weird-looking" issue?

GoldenAlpharex:
I'll test-merge it again today, I hadn't seen that there had been updates on this until now. Feel free to ping me on Discord if I end up forgetting about a PR of yours again.

Ghommie:
> I'll test-merge it again today, I hadn't seen that there had been updates on this until now. Feel free to ping me on Discord if I end up forgetting about a PR of yours again.

The updates were done upstream, that's why.

Namwty:
Encountering issue where oversized characters cannot pixel shift upwards more than one pixel, I assume this is related to this test merge.

Laying on the floor does enable pixel shifting up tho, only appears when standing up and hitting the "second tile's roof" if that makes sense...

Ghommie:
Duly noted, I'll get to it soon. Perhaps it's time pixel shifting uses pixel w/z, mhm?

Ghommie:
Ok, I've updated the pixel shift component to take into consideration default body position. Please update the testmerge.

Samf3232:
not sure if this is local or not but when holding B and moving, my character stretches weirdly

https://cdn.discordapp.com/attachments/1115912177128329260/1150292435038109807/20230910-0450-54.7241953.mp4

https://discord.com/channels/@me/1115912177128329260/1150292435289772192

Samf3232:
additional issue: this makes my eyes disappear 
image

Ghommie:
I understand it might look a little wobbly while pixel shifting, but I believe it's a fleeting nit with interpolation at this point, and frankly not that much noticeable unless you've got a big monitor perhaps.

> additional issue: this makes my eyes disappear

Oh great... I suppose I should give up at this point.

Namwty:
Most likely related to this aswell

Noticed that some actions cause oversized character's to shift some pixels downwards... it seems that grabbing and dragging people around causes it ( really rarely, to the point where it's hard to pin point what causes )

but other cause which happens consistently, is fireman carrying and giving people piggyback, as soon as you do either of those as oversized character, your sprite shifts.. and remains like that untill you pixel shift. ( See image of me character stuck shifted south and clipping into the wall ). Not a massive issue, but generally makes you look odd when rescuing people.

Hope you enjoy kindle kicks, I got them specifically for this pic.
Untitled

Ghommie:
umhm, that's probably because the pixel y is being reset somewhere in the code without taking into account the body position offset, which might be a potential issue upstream too.

PR #22129, opened 2023-06-29
[MIRROR] Makes hoods into a component [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75977
--------------------
## About The Pull Request

Refactors the behaviour of "one clothing item deploying another clothing item" from `/obj/item/clothing/suit/hooded` and makes it into a component.
This allows you to make hooded items which are not part of that typepath. It also means you could make (for instance) a hat which can deploy a pair of sunglasses into the eye slot or a jumpsuit with deployable clown shoes or something. 

I need to pass in an assload of callbacks because we have a bunch of special hoodies that want to do things when you raise and lower the hood, but for a normal item you would not need these.

## Why It's Good For The Game

Frees people from the tyrrany of typepaths, mostly.
Plausibly you could use it to do something fun we don't currently do.

## Changelog

Not player facing, hopefully. As long as I did this all right.


dragomagol:
This is currently broken on tg, might want to hold off on merging.

tf-4:
https://github.com/tgstation/tgstation/pull/76408

vinylspiders:
Just letting whoever is going to merge this know: merge this right after, it's a hotfix  @tf-4 

https://github.com/Skyrat-SS13/Skyrat-tg/pull/22154

PR #22147, opened 2023-06-30
[Modular] Nerfs pirate cheese (stop destroying the consoles)

YakumoChen:
## About The Pull Request
Changes the pirate navigation consoles to no longer be able to be conventionally destroyed by the crew

Many players consider the act of destroying the pirate's flight consoles to be "extremely lame". It bypasses most of the threat the pirates pose, neuters their ship, and it's usually done without the pirates knowing or while they're roleplaying, making it extremely cheesy. This forces the crew to actually fight or interact with the pirates, instead of just disabling their ship and making negotiations one-sided.

## Proof of Testing

<details>

image

</details>

## Changelog

:cl:
balance: Local pirates have scrounged up the funds to reinforce their vital ship consoles, making them extremely difficult to destroy.
/:cl:


HoodMilk:
You can still just use the consoles to steal the ship though??? Like move it some place the pirates will never find using the consoles.

BubbleTrroubles:
> You can still just use the consoles to steal the ship though??? Like move it some place the pirates will never find using the consoles.

You are right on this oversight. Pirate vessel controls should be ID locked to the ghost role spawned pirates only. This would prevent what this PR is going for is it not?

YakumoChen:
> You can still just use the consoles to steal the ship though??? Like move it some place the pirates will never find using the consoles.

That's a different issue that would be harder to resolve modularly. More to the point, console destruction happens often while the pirates are still on their own ship, and you wouldn't have much luck trying to steal a ship the pirates are still aboard of.

It also doesn't completely ground the ship since it can still be found. (And there's a limited number of places it can go)

Stonetear:
Perhaps out of scope, but making the boards printable would be more clever.

YakumoChen:
> Perhaps out of scope, but making the boards printable would be more clever.

There aren't boards, the consoles can't be constructed or deconstructed. 
Where would they print the boards from, anyway...? If pirates did have a printer, the crew would just target that instead.

RatFromTheJungle:
pirates aren't supposed to be an extended threat; so making their ship theoretically impossible to be disabled (as if you can't just bomb the APC instead, to the same results) is a very-questionable change, and it will do very-little then artificially extend their presence past it's intended limits of (show up, cause a ruckus, die), and likely just lead to them getting disabled after it will (very quickly) become a routine problem (as they are already, opting to just speedrun mechanics)

OzzyMcPseudonym:
Honestly, if I couldn't destroy the console, I'd just get a plumbing RCD and spam 1/3/7 N2O grenades instead (with relevant command approval, of course). Instead of allowing the ship to be essentially non-lethally disarmed, it's just "Nuke the thing, I guess." Maxcaps aren't really a thing here, but I could plausibly or reasonably see a future where admins allow people to maxcap pirates just to make the round no longer be held hostage by a random midround.

YakumoChen:
> 

Players are already opting to skip/bypass roleplay by disabling the pirate ship while pirates are putting in the effort to roleplay, nothing stops you from "ending the threat" by just killing the pirates, or by negotiating, but ruining the ship with sneaky methods while they're making the attempt is bad for roleplay.

Gyrandola:
I like this change, perhaps make the console pirate ID locked too but in another PR like someone suggested

GoldenAlpharex (changes requested):
You'll have to save the dme again in order for it to be passing the linters, or you might need to move your added line down one manually.

KathrinBailey:
Doesn't solve the issue, you can just bomb the ship from the outside still and it'll blow up the cargo pad, making it impossible for the pirates to achieve their objective.

I'm pretty sure they're grounded on the station Z level until they get 50k anyway?

YakumoChen:
I'm not super concerned about bombing. If things are at the point pirates are getting bombed, they're probably hostile or someone is getting bwoinked

GoldenAlpharex (approved):
I'm fine with giving this a shot, worst case scenario we can just revert this if we see that it's causing more issues elsewhere.

PR #22161, opened 2023-07-01
[MIRROR] Fix organ damage being uncapped [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76438
--------------------
## About The Pull Request

`maximum` being null caused it to default to `maxHealth` which was a sensible value

instead it was passing infinite to apply organ damage which then clamped damage to infinity instead of max health

unit test pending in a follow up pr

Fixes #76435

## Changelog

:cl: Melbert
fix: Fix organ damage being uncapped
/:cl:


GoldenAlpharex (approved):
Looks good, thanks @vinylspiders for the fixing of the PR!

PR #22176, opened 2023-07-01
[MANUAL MIRROR] Ensures that all Unit Tests are ticked in `_unit_tests.dm` (#76404)

lessthnthree:
https://github.com/tgstation/tgstation/pull/76404

## About The Pull Request

Ensures we don't get a repeat of #76345 (unit test that wasn't ticked in the `_unit_tests.dm` file, fixed in
596ca8b6d4cc49cd69fc104b53b7f4973497a2e5). Basically, we leverage the code that was already being used in the DME Validator but then expand it a bunch via using JSON Schemas that correspond to the type of scan we want to run. Even though sorting unit tests alphabetically is a bit different than sorting the tgstation DME, it's good to leverage the already existing framework rather than create a copy-pasta "lesser" code runner. This went through strenous testing on my end, so let me know if anything seems off.

While in the area, I added some other niceties that I've found work really well in GitHub Runners environments, as well as local testing in case you really like doing that before you make a PR for some reason. ## Why It's Good For The Game

image

This is what it looks like pre-596ca8b6d4cc49cd69fc104b53b7f4973497a2e5 (this is now merged, so it will pass CI)

De-hardcodes some stuff and allows for some neater flexibility, less cringe unit tests being coded and not being ticked in the file, etc. etc.
## Changelog
Nothing for players to care about.

Let me know if you have a better idea than the schemas, I couldn't think of one that could be really extensible and flexible in the same way this is.
# Conflicts:
#	code/modules/unit_tests/_unit_tests.dm
#	tools/validate_dme.py


lessthnthree:
Doesn't account for our modular folder, need adjustment.

vinylspiders:
This will also fix the stupid skin.dmf stuff.

vinylspiders (approved):

PR #22227, opened 2023-07-04
[MIRROR] Makes cargo orders and req forms easier to differentiate [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76499
--------------------
## About The Pull Request
This PR makes two very simple changes. First, requisition forms printed off by the order consoles in cargo now print with a timestamp on the name, so that it's a bit easier to differentiate between forms when there's a bunch of them. Second, crates that come on the cargo shuttle now have the order ID they are associated with appended to their name.
The req forms:
image

As well as two crates that would normally be impossible to differentiate
image
image

## Why It's Good For The Game
When there are lots of requisition forms laying on the ground and you need to refer back to one of them (to see who ordered what or something like that), it can be a pain to find the right one since they all have the same name. This is even worse if someone has tried to be helpful and put them all in a folder, since you now have to take them out and look at each one. Adding the timestamp helps to remedy this.

For the crates, when someone orders something that doesn't come with a named crate (for example, some medical orders just have the name "medical crate"), or when multiples of the same things are ordered, it can be hard to match up the shipping manifest with the crate again, leading to unwanted fines, especially when people don't return all their crates at once. Adding the order ID to the name of the crate lets you easily identify which manifest goes where.
## Changelog
:cl:
qol: Crates that come in the cargo shuttle now have the order ID in their name
qol: Requisition forms are now blue, and named with the timestamp they were created on
/:cl:


code/modules/shuttle/supply.dm
```diff
@@ -163,9 +163,15 @@ GLOBAL_LIST_INIT(blacklisted_cargo_types, typecacheof(list(
...
 
+<<<<<<< HEAD
 		spawning_order.on_spawn() //SKYRAT EDIT
 		if(!spawning_order.pack.goody && !(spawning_order?.paying_account in forced_briefcases)) //we handle goody crates below //SKYRAT EDIT
 			spawning_order.generate(pick_n_take(empty_turfs))
+=======
+		if(!spawning_order.pack.goody) //we handle goody crates below
+			var/obj/structure/closet/crate = spawning_order.generate(pick_n_take(empty_turfs))
+			crate.name += " - #[spawning_order.id]"
+>>>>>>> 2333e8b0c1f (Makes cargo orders and req forms easier to differentiate (#76499))
```
softcerv, line 174:
```suggestion
		spawning_order.on_spawn() //SKYRAT EDIT
		if(!spawning_order.pack.goody && !(spawning_order?.paying_account in forced_briefcases)) //we handle goody crates below //SKYRAT EDIT
			var/obj/structure/closet/crate = spawning_order.generate(pick_n_take(empty_turfs))
			crate.name += " - #[spawning_order.id]"
```

softcerv:
```suggestion
		if(!spawning_order.pack.goody && !(spawning_order?.paying_account in forced_briefcases)) //we handle goody crates below //SKYRAT EDIT
			var/obj/structure/closet/crate = spawning_order.generate(pick_n_take(empty_turfs))
			crate.name += " - #[spawning_order.id]"
```

softcerv: I tested the PR with this change and it seemed to work fine.

vinylspiders (approved):

PR #22241, opened 2023-07-04
[MIRROR] fixes abductors unable to get points [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76505
--------------------
fixes #76504

:cl:
fix: abductors can get points again
/:cl:

Not really sure what even happened here. istype(something) always returns true for any instance, so this would always fail
EDIT: #76101 probably mixed up ishuman with istype


code/modules/antagonists/abductor/machinery/experiment.dm
```diff
@@ -116,49 +116,57 @@
  * * type The type of experiment to be performed
  * * user The mob starting the experiment
  */
-/obj/machinery/abductor/experiment/proc/experiment(mob/occupant, type, mob/user)
+/obj/machinery/abductor/experiment/proc/experiment(mob/living/carbon/human/occupant, type, mob/user)
 	LAZYINITLIST(history)
-	var/mob/living/carbon/human/H = occupant
 
+<<<<<<< HEAD
```
softcerv:
```suggestion

```

code/modules/antagonists/abductor/machinery/experiment.dm
```diff
@@ -116,49 +116,57 @@
...
+=======
+	if(!istype(occupant))
+		stack_trace("Abductor '[name]' called /proc/experiment with unexpected occupant ([occupant])")
+
+	if(!ishuman(occupant)) //We shouldn't be processing anything other than humans
+		return "Not a humanoid!"
+
+>>>>>>> 24ed837a4a1 (fixes abductors unable to get points (#76505))
```
softcerv:
```suggestion
	if(!istype(occupant))
		stack_trace("Abductor '[name]' called /proc/experiment with unexpected occupant ([occupant])")

	if(!ishuman(occupant)) //We shouldn't be processing anything other than humans
		return "Not a humanoid!"

```

vinylspiders (approved):

PR #22271, opened 2023-07-06
Adds new dresses & skirts

KathrinBailey:
## About The Pull Request
• Adds a new dress that's inspired by practical real-world clothing & fabrics.
• Adds three new skirts that have bras in their alt states.
• Yes, @Ebin-Halcyon there are digitigrade versions that have been specifically tailored on top of the **new** digitigrade legs model, you have permission to sigh in relief now.

We don't have any actual knee-length skirts except for one coder-sprite one. More character customisation, bras are rare.

## Proof of Testing

<details>

image
I don't want to show the alt states with the bras removed because those pixels underneath slam dunk on TOS.
You can kinda imagine.
image
They're in the loadout menu.

</details>

## Changelog

:cl:
add: Three skirts, one dress.
/:cl:


Draganfrukts:
Absolutely love these, I feel like the dress could be a bit more colorful like this:
image
but I understand if it's not what you had in mind
I'll leave the palette here in case you want to use these colors instead
image
image < native

ghost:
modular_skyrat/master_files/icons/mob/clothing/under/skirts_dresses.dmi
modular_skyrat/master_files/icons/mob/clothing/under/skirts_dresses_digi.dmi

Both of these files have a new empty state at the ends; please remove those if you've the time. Otherwise, all looks good sprite-wise! Needs Hal's go-ahead ofc due to the sprite freeze

Ebin-Halcyon (changes requested):
What Orion said with the empty additions, once that's handled you're set to go. Might be worth it to look into Dragonfruit's palette as well but that's up to you.

Thanks for making it nudigi compatible as well.

Ebin-Halcyon (approved):
Mhm.

PR #22296, opened 2023-07-07
[MIRROR] Bilingual can now choose their language [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76609
--------------------
## About The Pull Request

This was one of the tradeoffs for removing other, more consistent sources of languages, and was requested by Melbert among many others.
This does go against my wanted goal of decreasing the risk of eavesdropping by other players through just magically knowing a language, but it is an expensive quirk and it is in their medical records, which makes it better than language encryption keys or silicon just innately knowing them.

This also limits Bilingual to only roundstart languages (+Uncommon), rather than being randomly selected from a list (that had very useless ones like monkey, podpeople, and beachbum). This is mostly just for modularity, I didn't want to make it look terrible code-wise and thought this may be the optimal way to handle it.

This is also me going back on https://github.com/tgstation/tgstation/pull/71773 - which I had closed myself.

## Why It's Good For The Game

If we're gonna keep the Bilingual quirk, it might as well be something players can choose the language of, it's their character and they should be allowed to decide how their character is, and it is my fault that this stupid compromise of "getting a random language" was made in the first place. It never should've happened.
It now actually limits it to roundstart-only languages, so there's no way you can spy on people who prepare in advance through becoming podpeople, or monkeys, etc.

## Changelog

:cl:
balance: Bilingual quirk now lets you choose your language between ones given to roundstart species.
balance: Foreigner and Bilingual are now mutually exclusive languages.
/:cl:


desertmote:
nekomimetic is BACK!

vinylspiders (approved):

PR #22312, opened 2023-07-08
Reverts wall eating mushrooms

Paxilmaniac:
## About The Pull Request

It is as simple as the title, removes the wall eating mushrooms event that we added.

There's a number of problems with the event, though the first and foremost problem is that there is no enhanced roleplay in forcing the crew to go right click a bunch of walls with welders or else the entire station gets vented into space.

But, ignoring the fact the event has been universally complained about by any who have had to go through it, I'll go over its other problems.

1. For an event that has the opportunity to completely destroy the walls of a lot of areas, the fact the event announcement includes no location is problematic. There are many maps where a wall fungus can go completely unnoticed before it becomes too large to effectively stop.

2. Certain maps are just not fit for having wall fungus run. The best example of this is north star, where the exterior walls are all doubled up, meaning the fungus attaches to the exterior and can spread endlessly to the interior until someone goes on eva to remove it all.

3. The fungus spreads into wall corners, which you cannot weld off without destroying an adjacent wall to reach.

I do not think we should have added such an event in the first place, and thus why I am not even going to bother with trying to salvage it in any way. "Event that causes catastrophic damage that would take an entire round to fix if not given constant attention" is not something I believe we should be adding ourselves.

## Proof of Testing

The event doesn't exist anymore I don't know what kinda screenshots I could give you to be honest.

## Changelog

:cl:
del: The wall eating fungus event has been removed.
/:cl:


RatFromTheJungle:
you should replace the mushrooms with guns that grow on the wall then it would be a good event

SpaceVampire:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/20466

I noticed there was an earlier PR that was going to include fixes to those three main problems. But it was closed so it seems like this is the intended roleplay experience of it.

Paxilmaniac:
> you should replace the mushrooms with guns that grow on the wall then it would be a good event

The mushrooms are turrets they shoot at you like plants vs zombies

YakumoChen:
Almost seems like certain maps are just bad and we shouldn't be playing on them

Cyprex:
Why was this closed? We've had to deal with this at Bubber too and whilst more PvE-y stuff is certainly welcome(like blob) this is absolutely not it. Paxilmaniac has outlined it in the OP well.

lessthnthree:
> Why was this closed? We've had to deal with this at Bubber too and whilst more PvE-y stuff is certainly welcome(like blob) this is absolutely not it. Paxilmaniac has outlined it in the OP well.

So disable it in the config?

PR #22361, opened 2023-07-11
[MIRROR] Bots no longer require PAIs to become sapient [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76691
--------------------
## About The Pull Request

We were talking in the coder channel about what the role of a pAI is, with a general conclusion that as the name would suggest they should be _personal assistants_.
This means they should be sticking around their owner, not wandering away as a holochassis or in the body of a bot.
The former is a matter for a future PR, the latter I am addressing here.

What we also discussed is that clearly some people _want_ to respawn as a weird quasi-useless mob which wanders aimlessly around the station. That seems like a fine thing to exist, but it shouldn't be a pAI.

Resultingly: pAI cards can no longer be placed inside bots.
However, you also no longer need to place pAI cards inside bots in order for them to become sapient, it's a simple toggle on the bot control menu. Enabling this option will poll ghosts 
Toggling the "personality matrix" off while a bot is being controlled by a ghost will ghost them again, so if they're annoying they're not that hard to get rid of.

image

Mobs which couldn't have a pAI inserted don't have this option. Specifically securitrons, ED-209, and Hygienebots (for some reason).

Perhaps most controversially, any bots which are present on the station when the map loads will have this setting enabled by default. We will see if players abuse this too much and need their toys taken away, I am hoping they can be trusted.

Additionally, as part of this change, mobs you can possess now appear in the spawners menu.
image
Here is an unusually populated example.

Oh also in the process of doing this I turned the regal rat "click this to become it" behaviour into a component because it seems generally useful.

## Why It's Good For The Game

Minor stuff for dead players to do if they want to interact with living players instead of observe.
Shift pAI back into a more intended role as a personal assistant who hangs around with their owner, rather than just a generic respawn role.

## Changelog

:cl:
add: PAIs can no longer be inserted into Bots
add: Bots can now have their sapience toggled by anyone with access to their settings panel
add: Bots which exist on the map at the start of the round automatically have this setting enabled
qol: Bots, Regal Rats, and Cargorilla now appear in the Spawners menu if you are dead
qol: Bots can be renamed from their maintenance panel
/:cl:


vinylspiders:
Requires https://github.com/Skyrat-SS13/Skyrat-tg/pull/22480

vinylspiders (approved):

PR #22384, opened 2023-07-12
[MANUAL MISSED MIRROR] Replaces lava and chasm's "safeties" and ignoring turf slowdown on catwalks with traits and a new element

vinylspiders:
Original PR https://github.com/tgstation/tgstation/pull/76376 

(#76376)

This adds a new element for movables that grants turfs they're in traits, changes lava and the chasm component to check for traits instead, ditto for turf slowdown. It also implements another trait that prevents wet floor from slipping people, as well as some other changes (feel free to opine on them really):
- Tables and conveyor belts now stop turf slowdown, much like catwalks, as I imagine people walking on them are not really touching the floor. (I'd include protection against lava too... until they melt, but that'd mean finding a way to have these objects burn in the first place, and lava code is still stupid despite a years old refactor I did)
- Tables also stop slippery turfs from slipping (bananas, soaps etc. still apply). I wish there were a way to make some objects slippery by coating them in water vapor or splashing water/lube, but that's outside the scope of this PR.
- Fixed an edge case in which a mob standing on a lava turf would be left permanently visually on fire if the lava is changed to another kind of turf.
- Removed unused code from stone tiles.

I'm going to include these traits in that global list for admin-added traits... tomorrow perhaps. 💤

Replacing some hard-coded mechanics with easier to use traits and an element, which I also need for the submerge element PR.

:cl:
refactor: Replaced hardcoded "safeties" for lava, chasms and ignoring turf slowdowns on catwalks with traits.
balance: much like catwalks, tables and conveyors also disable turf slowdowns.
balance: slippery turfs won't slip you when walking on a table. fix: Fixed an edge case in which a mob standing on a lava turf would be left visually but permanently on fire if the lava is changed to another kind of turf.
/:cl:


vinylspiders:
This needs the 'species nuke' prs.. sigh.

lessthnthree (approved):

PR #22453, opened 2023-07-14
Some thrown liquid logic improvement

### LINKED ISSUE #22429: [Liquid System] Beakers thrown at walls will spill reagents inside walls

## Round ID:
8892

## Testmerges:
- [Antag Revamp*: The Nuclear-ish Option](https://github.com/Skyrat-SS13/Skyrat-tg/pull/18110)
- [[TM] (Maybe) fixes enormous RSC bloat](https://github.com/Skyrat-SS13/Skyrat-tg/pull/21360)
- [restrict imbuing to ashwalkers & icecats](https://github.com/Skyrat-SS13/Skyrat-tg/pull/21708)
- [The Illusive Language Re-Flavor PR ](https://github.com/Skyrat-SS13/Skyrat-tg/pull/22113)

## Reproduction:
When a beaker hits a wall, either by being manually thrown, propelled by an explosion, sent through disposals, etc, the reagents inside will create a puddle 'inside' the wall, rather than spilling the fluids onto the wall as it would if it hit an object (it may still do this, but also creates the puddle inside the wall anyways), or making a puddle on the floor adjacent to the wall, which would make far more sense. This has been a thing with the liquid system for as long as I can remember, so I always thought it was intended behavior, but the more I think about it the less sense it makes. 

Screenshot (3852)

vinylspiders:
## About The Pull Request

Fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/22429

Makes liquids spill on the tile adjacent to a wall when thrown, lets them ricochet off other objects as well. Now the liquids will appear where you would expect them to, rather than inside of walls and such.

Also adds support for throwing beakers full of liquid at any object now, not just turfs. They will no longer mysteriously disappear when hitting something that isn't a turf.

More immersive liquid throwing: liquids now behave more like how you might expect them to when thrown against something.

## Proof of Testing

<details>
<summary>Throwing against various objects</summary>
  
dreamseeker_k1n0HqovbF

</details>

<details>
<summary>Some more throwing</summary>
  
dreamseeker_ihVgWe2EYh

</details>

## Changelog

:cl:
fix: changes to thrown liquid behavior--now they will ricochet off of walls and other dense objects and spill where the beaker lands instead of inside of the wall. 
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers.dm
+++ b/code/modules/reagents/reagent_containers.dm
@@ -225,16 +225,15 @@
 			log_combat(thrown_by, M, "splashed", R)
 		reagents.expose(target, TOUCH, splash_multiplier)
 		reagents.expose(target_turf, TOUCH, (1 - splash_multiplier)) // 1 - splash_multiplier because it's what didn't hit the target
-		target_turf.add_liquid_from_reagents(reagents, reagent_multiplier = (1 - splash_multiplier)) // skyrat edit: liquid spills (molotov buff) (huge)
+		target_turf.add_liquid_from_reagents(reagents, reagent_multiplier = (1 - splash_multiplier), thrown_from = src) // skyrat edit: liquid spills (molotov buff) (huge)
 
 	else if(bartender_check(target) && thrown)
 		visible_message(span_notice("[src] lands onto the [target.name] without spilling a single drop."))
 		return
 
 	else
-		if(isturf(target)) //SKYRAT EDIT CHANGE
-			var/turf/T = target
-			T.add_liquid_from_reagents(reagents)
+		if(target.can_liquid_spill_on_hit()) //SKYRAT EDIT CHANGE
+			target.add_liquid_from_reagents(reagents, thrown_from = src)
 			if(reagents.reagent_list.len && thrown_by)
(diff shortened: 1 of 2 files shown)


YakumoChen:
How will this effect, for example, thermite (which needs thrown against a wall to work and should (probably doesnt) consume the liquid

How will this effect liquid disposal? There aren't a lot of ways to actually get rid of liquids, one of the only convenient methods I've found is splashing beakers on your backpack to empty them, since not everyone can access chemistry machines.

vinylspiders:
> How will this effect, for example, thermite (which needs thrown against a wall to work and should (probably doesnt) consume the liquid
> 
> How will this effect liquid disposal? There aren't a lot of ways to actually get rid of liquids, one of the only convenient methods I've found is splashing beakers on your backpack to empty them, since not everyone can access chemistry machines.

Splashing and throwing are still separate things. You can still get rid of liquids by splashing (which I do with the combat mode> right click on something, I'm not sure if there are better ways).

The only thing this changes is the throwing mechanic.

Edit: as for the thermite, that will still expose the wall that the beaker hits even if the liquid doesn't end up there. It will leave the decal and the excess liquid will spill onto the ground.

lessthnthree (approved):

PR #22363, opened 2023-07-11
[MANUAL MIRROR] Reworks Duffel Bags (Zippers)

softcerv:
## About The Pull Request
Manual mirror of https://github.com/tgstation/tgstation/pull/76313 and https://github.com/tgstation/tgstation/pull/76442

Closes #22136, #22174

## How This Contributes To The Skyrat Roleplay Experience
Upstream mirror good :)

## Proof of Testing

<details>
<summary>Screenshots/Videos</summary>
  
image

I will post photos of things working soon.

</details>

## Changelog

:cl:
balance: Duffelbags will now only make you slow while they are unzipped. As a tradeoff, you now need to stand still and zip/unzip them to access their contents/not move real slow.
 fix: Fixes some fuck with duffelbags, them not holding enough + issues with spawning gear in them (job shit and all)
/:cl:

PR #22520, opened 2023-07-17
[MIRROR] Adds a new positive quirk, "Spacer Born".  [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76809
--------------------
## About The Pull Request

Adds a new 7 point positive quirk, "Spacer Born". Totally not inspired by The Expanse, don't look at the branch name. 

You were born in space, rather than on a planet, so your physiology has adapted differently.
You are more comfortable in space, and way less comfortable on a planet. 

Benefits:
   - You are slightly taller. (No mechanical effect)
   - You take 20% less damage from pressure damage.
   - You take 20% less damage from cold environments. 
   - You move 10% faster while floating (NOT drifting, this is zero gravity movement while beside a wall). 
   - You drift 20% faster (Drifting through zero gravity, untethered to anything)
   - While in space (z-level-wise, not turf wise), you lose some disgust overtime. 
      - While experiencing no-gravity for an extended period of time, you will start regenerating stamina and reduce stuns at a very low rate.  
   - If you are assigned to shaft miner (or the map is Icebox), you are awarded with a 25% wage bonus (hazard pay). 

Downsides:
   - While on a planet (Yes, this includes Icebox and planetary maps), you gain gravity sickness:
      - Passive accrue disgust (slightly lessened on Icebox) (Capped at low levels)
      - Choking, after extended periods (disabled on Icebox)
      - Slower movement 
      - Weaker stamina (disabled on Icebox)
      - Suffocation from extended periods (disabled on Icebox) (Lungs aren't adapted)
      - Mood debuff

(Effects not final)

## Why It's Good For The Game

I'd figure I throw my hat in with the Positive Quirk Curse. 

This is a quirk that improves your ability in a niche circumstance (low gravity / dangerous pressure), with some downsides that are only generally in effect if you play a few roles (or it's Icebox). 

Because of this I think it'll provide an interesting niche, where Spacer Born engineers are slightly better than their counterparts due to their origin (moving faster in space without a jetpack, withstanding pressure). However, at the same time, if the mining outpost sustains damage and needs repairs... suddenly your buff over your cohorts disappears, and you have to brave somewhere hostile to your body. 

Ultimately, the goal of the quirk is to encourage people to approach situations a bit differently. 
Or take it as a challenge and play shaft miner. 

## Changelog

:cl: Melbert
add: Adds a new 7 point positive quirk, "Spacer Born". You were born in space, and as a result your body's adapted to life in artificial gravity, making you much more effective and comfortable in lower gravity. However, travelling planet-side is quite a chore, especially if you're assigned to work there. 
add: Adds a chemical: Ondansetron, created by Oil + Nitrogen + Oxygen + Ethanol catalyst. A powerful Antiemetic (lowers disgust).
/:cl:


vinylspiders:
Needs https://github.com/Skyrat-SS13/Skyrat-tg/pull/22174, which needs https://github.com/Skyrat-SS13/Skyrat-tg/pull/22363

vinylspiders (approved):

Drifter7371:
As late as this is, I think it'd be preferable to keep the sprite-effecting part of the quirk disabled.
Another concern of mine is if this affects certain ruins in space, being that they use the same tiles as Lavaland at times. Is this tile-bound or Z-level bound?

vinylspiders:
> As late as this is, I think it'd be preferable to keep the sprite-effecting part of the quirk disabled. Another concern of mine is if this affects certain ruins in space, being that they use the same tiles as Lavaland at times. Is this tile-bound or Z-level bound?

Any reason in particular? We made it incompatible to oversized to prevent the sprites from looking terrible. 

image

These are two max-size sprites, one on left has Spacer and one on right doesn't. It's a pretty subtle difference.

As for how it works, it looks like it is Z-level based, not tile bound. So it depends on the map config. Mining zones like Lavaland and asteroids are considered planetary. And moons/planets like the syndicate base, and Icemoon would also be considered planetary. Everything else would not.

PR #22279, opened 2023-07-07
[MODULAR] Resets Human, Lizard, and Moth screams to /tg/ stock.

Iamgoofball:
## About The Pull Request

Resets Human, Lizard, and Moth screams to /tg/ stock.

## How This Contributes To The Skyrat Roleplay Experience

These scream sounds are battle tested, known to be functional, and don't hurt the ears. Overall, a good set of screams.

## Proof of Testing

just edited some lists to point at stock /tg/ files

## Changelog

:cl:
del: Resets Human, Lizard, and Moth screams to /tg/ stock.
/:cl:

PR #22521, opened 2023-07-17
Adds back our own screams, adds the /tg/ screams as alternate choices

GoldenAlpharex:
## About The Pull Request
Does https://github.com/Skyrat-SS13/Skyrat-tg/pull/22279 properly, by making the new screams a separate option. I personally find them very grating, but for now, I'll give them a shot.

Removing shit to bring it back to /tg/ standards is not something we're striving for. If we wanted to be /tg/, we wouldn't be a downstream.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl: GoldenAlpharex
qol: Skyrat's voices for humans and moths are back, and the different versions were added back as alternatives.
/:cl:

### DIFF
--- a/modular_skyrat/modules/emotes/code/scream_datums.dm
+++ b/modular_skyrat/modules/emotes/code/scream_datums.dm
@@ -13,6 +13,17 @@ GLOBAL_LIST_EMPTY(scream_types)
 /datum/scream_type/human
 	name = "Human Scream"
 	male_screamsounds = list(
+		'modular_skyrat/modules/emotes/sound/voice/scream_m1.ogg',
+		'modular_skyrat/modules/emotes/sound/voice/scream_m2.ogg',
+	)
+	female_screamsounds = list(
+		'modular_skyrat/modules/emotes/sound/voice/scream_f1.ogg',
+		'modular_skyrat/modules/emotes/sound/voice/scream_f2.ogg',
+	)
+
+/datum/scream_type/human_two
+	name = "Human Scream 2"
+	male_screamsounds = list(
 		'sound/voice/human/malescream_1.ogg',
 		'sound/voice/human/malescream_2.ogg',
 		'sound/voice/human/malescream_3.ogg',
@@ -49,6 +60,11 @@ GLOBAL_LIST_EMPTY(scream_types)
 
 /datum/scream_type/moth
 	name = "Moth Scream"
+	male_screamsounds = list('modular_skyrat/modules/emotes/sound/voice/scream_moth.ogg')
+	female_screamsounds = null
+
+/datum/scream_type/moth_two
+	name = "Moth Scream 2"
 	male_screamsounds = list('sound/voice/moth/scream_moth.ogg')
 	female_screamsounds = null
 


modular_skyrat/modules/emotes/code/scream_datums.dm
```diff
@@ -12,6 +12,17 @@ GLOBAL_LIST_EMPTY(scream_types)
...
+		'modular_skyrat/modules/emotes/sound/voice/scream_m1.ogg',
+		'modular_skyrat/modules/emotes/sound/voice/scream_m2.ogg',
+	)
+	female_screamsounds = list(
+		'modular_skyrat/modules/emotes/sound/voice/scream_f1.ogg',
+		'modular_skyrat/modules/emotes/sound/voice/scream_f2.ogg',
+	)
+
+/datum/scream_type/human_two
+	name = "Human Scream 2"
```
vinylspiders, line 25: Should we include '/tg/' in the name so it's clearer what it is?' Like "Human Scream (TG-Version)"?

vinylspiders (approved):

modular_skyrat/modules/emotes/code/scream_datums.dm
GoldenAlpharex: We could, but I don't feel like it's really necessary, people can figure it out on their own.

softcerv (approved):

Paxilmaniac:
It would be best if there was a scream sound preview so people can tell what the difference is without having to spawn in somehow, scream, then go back to character prefs to change it

softcerv:
> It would be best if there was a scream sound preview so people can tell what the difference is without having to spawn in somehow, scream, then go back to character prefs to change it

This would be nice upstream.

Paxilmaniac:
> > It would be best if there was a scream sound preview so people can tell what the difference is without having to spawn in somehow, scream, then go back to character prefs to change it
> 
> This would be nice upstream.

Maybe if upstream let you choose what your scream was. As this is not the case I do not quite see how it would be nice there.

softcerv:
> > > It would be best if there was a scream sound preview so people can tell what the difference is without having to spawn in somehow, scream, then go back to character prefs to change it
> > 
> > This would be nice upstream.
> 
> Maybe if upstream let you choose what your scream was. As this is not the case I do not quite see how it would be nice there.

That's not an option up there? That's weird. Disregard what I've said then.

glitchbow:
> It would be best if there was a scream sound preview so people can tell what the difference is without having to spawn in somehow, scream, then go back to character prefs to change it

You still sort of have to do this with TTS voices anyhow, because even though TTS has the "This is my voice" or Cyborg sample that can be replayed, it doesn't hit the same as when your actually type your character dialogue in /say. There's not a ridiculous amount of screams to go through, either, like there are with voices. I do think a preview button would be a nice QoL, but we've been getting by without it. Cat scream is gone, although cat scream was very divisive I think; it made a few people uncomfortable.

coffeecoffeecoffee:
I'm not opposed to the idea of improving the screams we had, but I appreciate this because the "new" (tg) screams seem to have noticeably much lower and jarringly varied sound quality. The audio distortion made them sound more like baby animals where the scream we had just sounds like... A scream, "funny" or otherwise. It could be improved, but the tg screams are not an improvement IMO.

PR #22529, opened 2023-07-17
[FIX] Allows some decals spawn

larentoun:
## About The Pull Request

Adds more exception to some decals, removing runtimes.

Less runtimes, easier to find important runtimes

## Proof of Testing

Just compile it a bunch of times, it works

## Changelog

:cl:
fix: Adds some decals to exceptions, allowing them to spawn and removing runtimes
/:cl:


vinylspiders:
Thank you for doing this. We are still catching up with the last of the mirrors so some of these may be fixed by then.

lessthnthree (approved):

PR #22534, opened 2023-07-17
[MISSED MIRROR] Loads Away Missions for Unit Testing (#76245)

vinylspiders:
## ORIGINAL: https://github.com/tgstation/tgstation/pull/76245

Hey there,

A pretty bad bug (#76226) got through, but it was fixed pretty quickly in #76241 (cf92862daf339e97c76b52c91f31d49ba5113bd4). I realized that if we were testing all the away missions, that this could theoretically get caught and not happen again. Regardless, unit testing gateway missions has been on my to-do list for a while now, and I finally got it nailed down.

Basically, we just have a really small "station" map with the bare bones (_teeny_ bit of fluff, maploading is going to take 30 seconds tops anyways let me have my kicks) with a JSON map datum flag that causes it to load all away missions in the codebase (which are all in one folder). Just in case some admins were planning on invoking the proc on `SSmapping`, I also decided to gate a `tgui_alert()` behind it because you never can be too sure of what people think is funny these days (it really does lock up your game for a second or so at a time).

I also alphabetized the maps.txt config because that was annoying me.

Things that break on production could(?) be caught in unit testing? I don't know if the linked issue I mentioned above would have been caught in retrospect, but it's likely to catch more than a few upcoming bugs (like the UO45 atmospherics thing at the very top) and ensure that these gateway missions, which tend to be the most neglected part of mapping, stay bug-free.

This is also helpful in case someone makes a new away mission and wants to see if stuff's broken. Helps out maptainers a bit because very, very technically broken mapping will throw up runtimes. Neato. Nothing that players should be concerned about.

Let me know if there's a better way to approach this, but I really think that having a super-duper light map with the bare basics to load up gateway missions and then all nine-ish gateway missions can sequentially load during init. I can't think of a better way to do it aside from some really ugly `#ifdef` shit. Also also, it has the added benefit of being a map that will always load your away mission without touching a single shred of config (and it's not likely to break if you follow sane practices such as making your own areas)


Jolly-66:
image

image

vinylspiders:
Needs https://github.com/tgstation/tgstation/pull/77585

lessthnthree (approved):

PR #22550, opened 2023-07-18
Makes the malf ai policy tooltip less ambiguous

vinylspiders:
## About The Pull Request

As per request just makes the Malf AI 'Policy' tooltip a little bit less confusing.

Less people thinking the UI is telling them they should ignore objectives and server policy.

## Proof of Testing

<details>

hPe8Zibpec

</details>


tgui/packages/tgui/interfaces/AntagInfoMalf.tsx
```diff
@@ -85,9 +85,8 @@ const FlavorSection = (props, context) => {
           mt={-0.5}
           icon="hammer"
           tooltip={multiline`
-            This is a gameplay suggestion for bored ais.
-            You don't have to follow it, unless you want some
-            ideas for how to spend the round.`}
+            Please refer to the 'Antagonist Policy' section of the wiki
+            if you have any questions.`}
```
softcerv: GitHub won't let me do a code suggestion on this, but instead of having the original text deleted, comment the original tooltip out and make a new tooltip after with the Skyrat text we want. Additionally, put a comment in the code labeling it as a Skyrat edit. (Also if you do this you *probably* shouldn't need to have the Skyrat Edit comment at the top)

vinylspiders: Reason I did it like this is because comments in JSX are very cumbersome

vinylspiders: Updated, thanks for the suggestion!

lessthnthree (approved):

PR #22564, opened 2023-07-18
[REFACTOR] Pixel Shift Component

larentoun:
## About The Pull Request

- Moves Pixel Shift logic to a `Component`, which is shared among all `living`.
- Moves `MAXIMUM PIXEL SHIFT` and `PASSABLE_SHIFT_THRESHOLD` to vars instead of defines for better manipulation in downstreams
- Moves block movement keybind from direct change to master_files

Less direct changes to the code, more modularity

## Proof of Testing

Pixel Shift, try passthrough modifier, pull, reset keybindings, buckle, move

https://github.com/Skyrat-SS13/Skyrat-tg/assets/31931237/75b526bd-995c-4873-8a9b-36eac018dca8

## Changelog

:cl:
refactor: Moves Pixel Shift logic to Component
/:cl:


modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,24 @@
+/mob/Initialize(mapload)
+	. = ..()
+	add_pixel_shift_component()
+
+/mob/proc/add_pixel_shift_component()
+	return
+
+/mob/living/add_pixel_shift_component()
+	AddComponent(/datum/component/pixel_shift)
```
MrMelbert, line 6: you're kinda crazy adding a component to every living mob in init that many won't use rather than only adding it when someone presses the shift hotkey

larentoun: How about I move it to Login, so it guarantees that a client is there and there won't be a need to check it everytime hotkey is pressed?

MrMelbert: Login is better but you really could just add it in `/datum/keybinding/mob/pixel_shift/down` and delete it in `/datum/keybinding/mob/pixel_shift/up`, I don't see any obvious need to keep the component on mobs which aren't shifting (unless I'm missing something?)

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
...
+	var/passable_shift_threshold = 8
+
+/datum/component/pixel_shift/Initialize(...)
+	. = ..()
+	if(!isliving(parent))
+		return COMPONENT_INCOMPATIBLE
+	owner = parent
+
+/datum/component/pixel_shift/RegisterWithParent()
+	. = ..()
```
MrMelbert:
```suggestion
```
Parent call does nothing

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
...
+	RegisterSignal(owner, COMSIG_KB_MOB_PIXEL_SHIFT_DOWN, PROC_REF(change_shifting))
+	RegisterSignal(owner, COMSIG_MOB_UNPIXEL_SHIFT, PROC_REF(unpixel_shift))
+	RegisterSignal(owner, COMSIG_MOB_CLIENT_PRE_LIVING_MOVE, PROC_REF(pre_move_check))
+	RegisterSignal(owner, COMSIG_MOB_PIXEL_SHIFT_CHECK_PASSABLE, PROC_REF(check_passable))
+
+/datum/component/pixel_shift/UnregisterFromParent()
+	. = ..()
```
MrMelbert:
```suggestion
```

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
...
+	UnregisterSignal(owner, COMSIG_KB_MOB_PIXEL_SHIFT_DOWN)
+	UnregisterSignal(owner, COMSIG_MOB_UNPIXEL_SHIFT)
+	UnregisterSignal(owner, COMSIG_MOB_CLIENT_PRE_LIVING_MOVE, PROC_REF(unpixel_shift))
+	UnregisterSignal(owner, COMSIG_MOB_PIXEL_SHIFT_CHECK_PASSABLE)
+
+/datum/component/pixel_shift/proc/pre_move_check(mob/source, new_loc, direct)
+	if(shifting)
```
MrMelbert:
```suggestion
/datum/component/pixel_shift/proc/pre_move_check(mob/source, new_loc, direct)
	SIGNAL_HANDLER
	if(shifting)
```

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
...
+		pixel_shift(source, direct)
+		return COMSIG_MOB_CLIENT_BLOCK_PRE_LIVING_MOVE
+	if(is_shifted)
+		unpixel_shift()
+
+/datum/component/pixel_shift/proc/check_passable(mob/source, border_dir)
+	if(passthroughable & border_dir)
```
MrMelbert:
```suggestion
/datum/component/pixel_shift/proc/check_passable(mob/source, border_dir)
	SIGNAL_HANDLER
	if(passthroughable & border_dir)
```

modular_skyrat/modules/pixel_shift/code/pixel_shift_keybind.dm
```diff
@@ -0,0 +1,21 @@
...
+		return
+	SEND_SIGNAL(user.mob, COMSIG_KB_MOB_PIXEL_SHIFT_DOWN, TRUE)
+	return TRUE
+
+/datum/keybinding/mob/pixel_shift/up(client/user)
+	. = ..()
+	if(.)
+		return
+	SEND_SIGNAL(user.mob, COMSIG_KB_MOB_PIXEL_SHIFT_DOWN, FALSE)
+	return TRUE
```
MrMelbert: This is a bad use of a signal: Signals should represent events that are occurring that you can react to, not a secondary form of calling procs. 

Generally speaking they should not be sent from multiple places unless they represent a similar event - in this case, these are two different events

(Also `COMSIG_KB_MOB_PIXEL_SHIFT_DOWN` should be the `keybind_signal` for the keybind)

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
...
+		return COMPONENT_LIVING_PASSABLE
+
+/datum/component/pixel_shift/proc/change_shifting(mob/source, active)
+	shifting = active
+
+/datum/component/pixel_shift/proc/unpixel_shift()
+	passthroughable = NONE
```
MrMelbert:
```suggestion
/datum/component/pixel_shift/proc/unpixel_shift()
		SIGNAL_HANDLER
		passthroughable = NONE
```

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,26 @@
...
+	SEND_SIGNAL(pull_target, COMSIG_MOB_UNPIXEL_SHIFT)
+	return ..()
+
+/mob/living/reset_pull_offsets(mob/living/pull_target, override)
+	SEND_SIGNAL(pull_target, COMSIG_MOB_UNPIXEL_SHIFT)
+	return ..()
+
+/mob/living/CanAllowThrough(atom/movable/mover, border_dir)
+	// Make sure to not allow projectiles of any kind past where they normally wouldn't.
+	if(!istype(mover, /obj/projectile) && !mover.throwing)
```
MrMelbert:
```suggestion
	if(!isprojectile(mover) && !mover.throwing)
```

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,26 @@
...
+		if(SEND_SIGNAL(src, COMSIG_MOB_PIXEL_SHIFT_CHECK_PASSABLE, border_dir) & COMPONENT_LIVING_PASSABLE)
```
MrMelbert: ~~Really surprised we don't have a signal for checking allowthrough~~

You should generalize this signal a bit - `COMSIG_LIVING_CHECK_ALLOW_THROUGH`, it's nice to future coders
Then you can move the checks for `isprojectile && !throwing` into your signal handler rather than put them here

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,26 @@
...
+	if(!GetComponent(/datum/component/pixel_shift))
+		AddComponent(/datum/component/pixel_shift)
+
+/mob/living/set_pull_offsets(mob/living/pull_target, grab_state)
+	SEND_SIGNAL(pull_target, COMSIG_MOB_UNPIXEL_SHIFT)
+	return ..()
+
+/mob/living/reset_pull_offsets(mob/living/pull_target, override)
+	SEND_SIGNAL(pull_target, COMSIG_MOB_UNPIXEL_SHIFT)
+	return ..()
```
MrMelbert: Likewise you could make these signals something like `COMSIG_LIVING_SET_PULL_OFFSET` and `COMSIG_LIVING_UNSET_PULL_OFFSET` to react to those events rather

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,26 @@
...
+	. = ..()
+	if(.)
+		add_pixel_shift_component()
+
+/mob/proc/add_pixel_shift_component()
+	return
+
+/mob/living/add_pixel_shift_component()
+	if(!GetComponent(/datum/component/pixel_shift))
+		AddComponent(/datum/component/pixel_shift)
```
MrMelbert: You should make your component `dupe_mode = COMPONENT_DUPE_UNIQUE`, this makes it so if someone tries to add a second component to a mob which already has it, it fails - then you don't have to do this `GetComponent` check

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,82 @@
+/datum/component/pixel_shift
+	var/mob/living/owner
```
MrMelbert: Avoid having a second var cast as parent in your comp, may leave hanging references

Just cast `var/mob/living/owner = parent` whenever you need it in your code

MrMelbert (commented):
Might as well do the rest

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
larentoun: > 

Wouldn't it be heavy to add/remove component if the button is spammed?

larentoun: Also, there needs something to check reset state for the pixel_shift. Probably on unpixel_shift

larentoun: And I need to constantly check if user input wants to pixel_shift or to move and unpixel_shift.

I need to keep the component and signals for "activate shifting" and "disable shifting".
Component:
1. Checks if the user is shifted when Moved - then does unpixel_shift()
2. Checks if the user is still holding Pixel Shift keybind - do pixel_shift() if they do OR do move() if the don't and unpixel_shift()

I need to somehow keep track of 2, so qdel-ing on up() doesn't seem good unless I miss something obvious

modular_skyrat/modules/pixel_shift/code/pixel_shift_keybind.dm
larentoun: So, I need to create a new signal for UP button only for this pixel_shift situation? I see that down() has its own signal, but up() doesn't

larentoun:
Thanks a lot for the review, by the way!

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
MrMelbert: > Wouldn't it be heavy to add/remove component if the button is spammed?

Not really. We do worse in other places

> I need to somehow keep track of 2, so qdel-ing on up() doesn't seem good unless I miss something obvious

Good point, you'd only want to delete the component (`qdel(src)`) when you stop shifting (in `unpixel_shift()`) (which you did!)

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,5 @@
+/mob/proc/add_pixel_shift_component()
+	return
+
+/mob/living/add_pixel_shift_component()
+	AddComponent(/datum/component/pixel_shift)
```
MrMelbert, line 6: This proc's not really necessary anymore, you can just do the following in the keybind proc
```dm
if(isliving(user.mob))
    user.mob.AddComponent(/datum/component/pixel_shift)
````

File: modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm [hunk 14]
```diff
@@ -0,0 +1,92 @@
...
+/datum/component/pixel_shift/proc/check_passable(mob/source, atom/movable/mover, border_dir)
+	SIGNAL_HANDLER
+	// Make sure to not allow projectiles of any kind past where they normally wouldn't.
+	if(!isprojectile(mover) && !mover.throwing && passthroughable & border_dir)
+		return COMPONENT_LIVING_PASSABLE
+
+/datum/component/pixel_shift/proc/pixel_shift_down()
+	SIGNAL_HANDLER
+	shifting = TRUE
+
```
MrMelbert:
```suggestion
/datum/component/pixel_shift/proc/pixel_shift_down()
	SIGNAL_HANDLER
	shifting = TRUE
	return COMSIG_KB_ACTIVATED
```
This just tells the keybind that we did something

modular_skyrat/modules/pixel_shift/code/pixel_shift_keybind.dm
```diff
@@ -0,0 +1,15 @@
...
+	hotkey_keys = list("B")
+	name = "pixel_shift"
+	full_name = "Pixel Shift"
+	description = "Shift your characters offset."
+	category = CATEGORY_MOVEMENT
+	keybind_signal = COMSIG_KB_MOB_PIXEL_SHIFT_DOWN
+
+/datum/keybinding/mob/pixel_shift/down(client/user)
+	. = ..()
+	user.mob.add_pixel_shift_component()
```
MrMelbert:
```suggestion
	. = ..()
	if(.)
		return
	user.mob.add_pixel_shift_component()
```
And by returning `COMSIG_KB_ACTIVATED` from the component, we can check here that the keybind was handled, thus the code won't even bother trying to re-add the component

MrMelbert (commented):
This component's looking pretty darn fine now 👌 (though of course be sure to test it, I'd hate to break your code!)

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
larentoun: The proc is there to allow blocking pixelshift on specific mobs. For example, a downstream would prefer to disable pixelshift on `silicon/ai`

modular_skyrat/modules/pixel_shift/code/pixel_shift_keybind.dm
larentoun: That's handy!

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,93 @@
+/datum/component/pixel_shift
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+	/// Whether the mob is pixel shifted or not
+	var/is_shifted = FALSE
+	/// If we are in the shifting setting.
+	var/shifting = TRUE
+	/// Takes the four cardinal direction defines. Any atoms moving into this atom's tile will be allowed to from the added directions.
+	var/passthroughable = NONE
+	var/maximum_pixel_shift = 16
+	var/passable_shift_threshold = 8
```
GoldenAlpharex: Please document these, thanks!

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,93 @@
...
+/datum/component/pixel_shift/UnregisterFromParent()
+	UnregisterSignal(parent, COMSIG_KB_MOB_PIXEL_SHIFT_DOWN)
+	UnregisterSignal(parent, COMSIG_KB_MOB_PIXEL_SHIFT_UP)
+	UnregisterSignal(parent, COMSIG_LIVING_RESET_PULL_OFFSETS)
+	UnregisterSignal(parent, COMSIG_LIVING_SET_PULL_OFFSET)
+	UnregisterSignal(parent, COMSIG_MOVABLE_MOVED)
+	UnregisterSignal(parent, COMSIG_MOB_CLIENT_PRE_LIVING_MOVE)
+	UnregisterSignal(parent, COMSIG_LIVING_CAN_ALLOW_THROUGH)
+
+/datum/component/pixel_shift/proc/pre_move_check(mob/source, new_loc, direct)
```
GoldenAlpharex, line 36: Document this please!

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,93 @@
...
+	SIGNAL_HANDLER
+	if(shifting)
+		pixel_shift(source, direct)
+		return COMSIG_MOB_CLIENT_BLOCK_PRE_LIVING_MOVE
+
+/datum/component/pixel_shift/proc/check_passable(mob/source, atom/movable/mover, border_dir)
```
GoldenAlpharex, line 43: Document this please!

modular_skyrat/modules/pixel_shift/code/pixel_shift_component.dm
```diff
@@ -0,0 +1,93 @@
...
+	if(is_shifted)
+		var/mob/living/owner = parent
+		owner.pixel_x = owner.body_position_pixel_x_offset + owner.base_pixel_x
+		owner.pixel_y = owner.body_position_pixel_y_offset + owner.base_pixel_y
+	qdel(src)
+
+/datum/component/pixel_shift/proc/pixel_shift(mob/source, direct)
+	passthroughable = NONE
+	var/mob/living/owner = parent
+	switch(direct)
```
GoldenAlpharex: Document these four please!

modular_skyrat/modules/pixel_shift/code/pixel_shift_mob.dm
```diff
@@ -0,0 +1,5 @@
+/mob/proc/add_pixel_shift_component()
```
GoldenAlpharex, line 2: Documentation please!

GoldenAlpharex (changes requested):
Neat change, hopefully this works well!

GoldenAlpharex (approved):
Looks good, should be ready for a test-merge just in case it breaks anything.

GoldenAlpharex:
Also please upload some proof of testing, just describing what you've done isn't sufficient. Screenshots, gifs and videos (of you testing it) are valid proofs of testing.

larentoun:
Added. 10MB limit is so bad

GoldenAlpharex:
It's been working, let's goooooo!

PR #22586, opened 2023-07-19
[MIRROR] balloon alerts for storage failures [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76779
--------------------
## About The Pull Request

Adds some balloon alerts for when you fail to put something in storage items (when there is not enough room, item is too big, etc.)
Also adds some balloon alerts for when you fail with a duffelbag (if it's zipped it will say "closed!", and failure alerts if you move when zipping/unzipping)

## Why It's Good For The Game

Balloon alerts are immediately obvious and leave no residue, which is good for failing to put something into a bag, as you probably don't really care to look back on it in the future, and want to immediately know why you cant put it in. I chose not to add balloon alerts for putting things in storage successfully for a couple reasons. For balloon alerts, they would be too long, but they hold fairly useful information so I don't want to cut down the messages. Also just having something like "placed" looked really bad in my opinion. I also can see what you or others put in their bag theoretically being useful to look back on. I think the compromise of storage failures and successes being held in different places (balloon alerts vs in chat) is fine, but if others disagree I can change it or just close this.

## Changelog

:cl: Seven
qol: Added some balloon alerts for failing to place items in storage containers
/:cl:


vinylspiders:
needs duffelbags probably

vinylspiders (approved):

PR #22606, opened 2023-07-21
[MIRROR] The Leaning Update (and Density Refractor) [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76704
--------------------

## About The Pull Request
WHAT HAS CHANGED MECHANICALLY
You can now lean up against walls.

https://github.com/tgstation/tgstation/assets/55666666/bf81b7b5-6cab-4fc3-9887-075351511505

To lean against a wall, simply face opposite to it and drag your sprite onto it.
Doing so makes you non-dense, just like if you were laying down. This means people can walk through you, but you are still susceptible to melee and ranged attacks. Leaning up against a wall also mitigates your FOV loss by 30 degrees, as you can have a better look at your surrounding when you put more of your surroundings infront of you.
Because it seemed thematic to lean up against the wall while smoking and then flick a cigarette away, cigarettes will now say they where "flicked" instead of thrown when you toss them, I also took the time to add a bit of variation into the throw text.

A few bugs where you could become non dense and then run straight through people has been patched.

NOT PLAYER FACING
So basically I've implemented a system that keeps effects that manage a mob's density consistent with eachother.
An example of some of the situations that could occur
Effect A would render a spaceman undense and turn the player dense again once it was concluded
Effect B would render a spaceman undense and then after a timer revert the spaceman to whatever state the spaceman was in before effect B started.
Thus if you enabled effect A and then Effect B, setting your previous state of denseness to undense, and then concluded effect A, when Effect B would finish it would put you back into the state of density you were in when you started. This would render the spaceman permanently undense.

To solve this, the system has been updated so that all instances of density adjustment to mobs are handled by traits from unique sources (with the exception undensity gained by laying down due to its prevalence.) All effects that handle density will no longer step on each others toes and can now be rain simultaneously without fear.
## Why It's Good For The Game
Leaning is cool. Bugs are bad.
## Changelog
:cl: itseasytosee
add: You can now lean against walls! Simply turn your back to the wall and clickdrag yourself onto it.
fix: There should no longer be any instances of spacemen being able to run straight through eachother as if they weren't even there.
spellcheck: Added more variance to item throwing text.
refactor: Mob density has been refactored
/:cl:


vinylspiders:
Bugged for now, wait for https://github.com/tgstation/tgstation/pull/77002

vinylspiders (approved):

PR #22623, opened 2023-07-21
[MIRROR] Adds an element to water turfs that makes things... look like they're actually immersed in water. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76208
--------------------
## About The Pull Request
This PR adds an element, for now added to water turfs, that gives passing movables an overlay to make them look immersed in that turf.

You can definitely see in the code what sort of stupid stuff I've had to do to work around some of the flaws with the inbuilt features. Why? Because I needed a (vis_contents) overlay that:
- can cover the entire width of the movable it's added to, whether it has a 32 px icon, 64, 96 etc, plus some to be safe.
- doesn't look misplaced around movables with an icon larger than 32 px.
- can be animated, so it doesn't rotate or scale (or rather, do so but in the opposite direction) with living mobs when resized, spinning or lying down or standing up, despite the KEEP_TOGETHER appearance flag, which ignores whether the overlay has the RESET_TRANSFORM flag.
- looks inset in the movable and doesn't spill into other movables of the same plane (that's what happened with normal inset overlays without the KEEP_TOGETHER flag)
- doesn't render above other movables covering the owner of the overlay.

If someone comes with a better and tested solution, go and tell I guess. How visuals are handled by this engine still eludes me at times, but I doubt planes would be a better solution to this.

The rest of the PR includes checks to make sure the overlay isn't added to movables with planes lower/higher than FLOOR_PLANE/ABOVE_ALL_MOB_PLANE, that are flying or being thrown, as well as certain types such as effects.

## Why It's Good For The Game
Adding some literal immersion to the game, since movables and mobs look like they're treading on water like Jesus right now. You can see a couple screenshots by scrolling down a little.

## Changelog

:cl:
image: walking on water will now immerse you in it... visually, at least.
/:cl:


vinylspiders:
Needs https://github.com/Skyrat-SS13/Skyrat-tg/pull/22606

vinylspiders (approved):

PR #22625, opened 2023-07-22
Hemophagus, The Hemophage Revamp, Part Four: Blood Depravity

GoldenAlpharex:
## About The Pull Request
This is the fourth part of the Revamp, this time focusing on the relationship between the tumor and the organs it corrupted.

What's changed?

<details>
<summary>The Hemophage species datum no longer handles any of the processes that are vital to Hemophages</summary>
  It's all handled in their organs now, making it a lot more customizable and far less likely to break, _especially_ given all of the recent SPECIES NUKING 2023 PRs from upstream.
</details>

<details>
<summary>Corrupted Organ behavior has been moved to components</summary>
  This means that all of the behavior that came from a Hemophage's tongue, liver and stomach are now no longer tied to just those very specific subtypes. Now any tongue can get corrupted and have the behavior you'd expect from a Hemophage's corrupted tongue. They also change visually, for the corrupted sprites, when available, or just a black color filter over them, in case there's no sprite available (looking at you, lungs).

Lungs can also be corrupted, and become "atrophied", however that change is just for flavor and doesn't do anything, because realistically, nobody that needs to breathe should have a Hemophage's tumor in them.
</details>

<details>
<summary>The Pulsating Tumor now corrupts corruptable internal organs</summary>
  Over a period of 60 seconds, livers, lungs, stomachs and tongues will become corrupted when in the same body as a tumor. Taking out a tumor before it has been in the same body as those organs, or taking out the organs themselves, will reset that timer. Once that timer is done, the appropriate `organ_corruption` component gets applied to the newly-corrupted organs, effectively giving them all of the features of authentic Hemophage organs!

In short, what that means is that all Hemophage organs can be replaced, _except for the Pulsating Tumor, but more on that below._
</details>

<details>
<summary>The Pulsating Tumor can now receive a coronary bypass more than once</summary>
  To address the fact that it was possible to fuck up Hemophages in a way that would absolutely require them to interact with Xenobiology in order to get a functional tumor again, I have made it so that a Pulsating Tumor can now receive another coronary bypass after running through at least one `on_life()` call, meaning that so long as they get revived once between every time they die and their hearts degrade past the point of non-functioning, they can be coronary-bypassed again.

Think of it as the tumor undoing/healing the coronary bypass, basically.
</details>

<details>
<summary>Corrupted organs now decay slower when the tumor is out of the Hemophage's body</summary>
  Previously, it would take 10 seconds for all of the organs to die, and I felt like that was a bit too fast. Now, it's 100 seconds. It's still fast, but it should hopefully give you enough time to get fixed again, or get some new organs, before they all start failing.
</details>

<details>
<summary>Corrupted organs no longer start decaying after going through the SAD, or otherwise getting your prefs re-applied to your character</summary>
  That was a big oopsie, introduced by some changes in some organ behaviors in mirrors, but it's all gone now, thanks to nothing actually being handled by the species datum anymore!
</details>

This took me far longer than I expected it to take, but I feel like I'm coming out of it with a far better understanding of elements, components and signals.

Hemophages deserved some more love, and were getting increasingly buggier and difficult to deal with, lately. This is one of the ways I've addressed that issue.

I also promised that I'd add organ corruption eventually, because I thought it would make sense, and it does, so here it is!

## Proof of Testing

<details>
  Honestly, there's a lot of stuff that I'd have to show, and it doesn't show itself very well with just images. I didn't feel like taking a video either, so have a few pictures proving some of the things worked.

image
image

</details>

## Changelog

:cl: GoldenAlpharex
add: A Hemophage's pulsating tumor will now corrupt nearby internal organs, hijacking them for its own purposes... Meaning that a Hemophage's organs (outside of the tumor itself) can be replaced by any other organs of the same kind, and after a minute, get corrupted and become functional Hemophage organs again.
qol: Corrupted organs now decay 10 times slower when no tumor is present in their body, taking around 100 seconds instead of 10 to fully decay. That means you finally have time to deal with it before the organs are fully dead, but still not that much.
qol: A Hemophage's pulsating tumor can now be coronary-bypassed more than once, granted that it spent at least some time in an alive person's body between the two bypasses! Yay for regeneration!
fix: Hemophage's corrupted organs no longer start rapidly decaying after going through the SAD, getting your preferences re-applied or otherwise even getting ahealed.
refactor: Hemophage code has been severely refactored, moving all of the organ-specific behavior to the actual organs, meaning that the species itself doesn't handle Hemophage-specific bodily functions anymore. Yay for customizability!
code: Hemophage code has been better spread out across multiple files, to be easier to sift through and find exactly what you want.
/:cl:


modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,113 @@
...
+		return
+
+	if(victim.stat == DEAD)
+		hemophage.balloon_alert(hemophage, "needs a living victim!")
+		return
+
+	if(!victim.blood_volume || (victim.dna && ((HAS_TRAIT(victim, TRAIT_NOBLOOD)) || victim.dna.species.exotic_blood)))
+		hemophage.balloon_alert(hemophage, "[victim] doesn't have blood!")
+		return
+
```
vinylspiders, line 73: You could put all of these checks into a separate `is_valid_target(hemophage, victim)` proc and it'd make this action a lot easier to follow.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,113 @@
...
+		hemophage.balloon_alert(hemophage, "stopped feeding")
+		return
+
+	var/drained_blood = min(victim.blood_volume, HEMOPHAGE_DRAIN_AMOUNT, blood_volume_difference)
+	// if you drained from a human w/ a client, congrats
+	var/drained_multiplier = (is_target_human_with_client ? BLOOD_DRAIN_MULTIPLIER_CKEY : 1)
+
+	victim.blood_volume = clamp(victim.blood_volume - drained_blood, 0, BLOOD_VOLUME_MAXIMUM)
+	hemophage.blood_volume = clamp(hemophage.blood_volume + (drained_blood * drained_multiplier), 0, BLOOD_VOLUME_MAXIMUM)
+
```
vinylspiders, line 123: Same here, perhaps an `attempt_feed(hemophage, victim)` that returns the volume of blood that was drained, or 0/ null if it failed in some way.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/hemophage_tumor.dm
```diff
@@ -0,0 +1,149 @@
...
+
+	bloodloss_rate *= is_dormant ? 1 / DORMANT_BLOODLOSS_MULTIPLIER : DORMANT_BLOODLOSS_MULTIPLIER
+
+	if(is_dormant)
+		owner.add_movespeed_modifier(/datum/movespeed_modifier/hemophage_dormant_state)
+
+	else
+		owner.remove_movespeed_modifier(/datum/movespeed_modifier/hemophage_dormant_state)
+
+	update_appearance()
```
vinylspiders:
```suggestion

	update_appearance()
	
	if(isnull(owner))
		return
	
	if(is_dormant)
		owner.add_movespeed_modifier(/datum/movespeed_modifier/hemophage_dormant_state)

	else
		owner.remove_movespeed_modifier(/datum/movespeed_modifier/hemophage_dormant_state)
```

`owner` can be null if the organ was removed already. The movespeed modifier in that case should still get handled fine since you do it in the `Remove()` proc as well, which has the mob's ref intact. But you'll want to guard against that here or you'll get a runtime.

vinylspiders (commented):
That's all that sticks out to me. Looks good otherwise!

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
GoldenAlpharex: Yeah, I meant to do that and forgot, I'll do that.

YakumoChen:
> To address the fact that it was possible to fuck up Hemophages in a way that would absolutely require them to interact with Xenobiology in order to get a functional tumor again

Actually the organ storage fridge slowly heals anything inside them when upgraded, to digress.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,141 @@
...
+#define HEMOPHAGE_DRAIN_AMOUNT 50
+/// The multiplier for blood received by Hemophages out of humans with ckeys.
+#define BLOOD_DRAIN_MULTIPLIER_CKEY 1.5
+
+
+/datum/component/organ_corruption/tongue
+	corruptable_organ_type = /obj/item/organ/internal/tongue
+	corrupted_icon_state = "tongue"
+	/// The item action given to the tongue once it was corrupted.
+	var/tongue_action_type = /datum/action/cooldown/hemophage/drain_victim
```
vinylspiders, line 11:
```suggestion
	var/datum/action/cooldown/hemophage/drain_victim/tongue_action
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,141 @@
...
+	. = ..()
+
+	if(!.)
+		return
+
+	var/obj/item/organ/internal/tongue/corrupted_tongue = corruption_target
+	corrupted_tongue.liked_foodtypes = BLOODY
+	corrupted_tongue.disliked_foodtypes = NONE
+
+	var/datum/action/tongue_action = corruption_target.add_item_action(tongue_action_type)
```
vinylspiders, line 24:
```suggestion
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,141 @@
...
+
+	if(corruption_target.owner)
+		tongue_action.Grant(corruption_target.owner)
```
vinylspiders, line 27:
```suggestion
	if(isnull(tongue_action))
		tongue_action = new
	if(corruption_target.owner)
		tongue_action.Grant(corruption_target.owner)
```

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
vinylspiders: Since this is attached to an organ, which exists in nullspace, this proc won't actually be able to add the action to the mob. The proc is written with the assumption that the item exists in the mob's loc.

You have two options: put the action type in `/obj/item/organ/internal/tongue/corrupted_tongue`'s `actions_types` var. That will cause it to automatically get added in `organ`'s  `on_insert` or `on_remove` and you will not have to worry about it.

Or option 2, if you want to keep it tied to the component (which you probably would, for the whole augmented/custom tongues stuff). Manually Grant/Remove the action. Which you are already doing half of here.  You just need to make sure it gets removed as well because that is not going to happen on its own due to the reasons stated above.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
vinylspiders: If you go with option 2 you'll need to handle removing the action too, so keep a ref.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
vinylspiders: Option 2

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
```diff
@@ -0,0 +1,141 @@
...
+
```
vinylspiders, line 28:
```suggestion

/datum/component/organ_corruption/tongue/on_tumor_removed(mob/living/carbon/tumorless)
	if(tumorless)
		tumorless.Remove(tongue_action)
	QDEL_NULL(tongue_action)
	return ..()

```

vinylspiders: Option 2

vinylspiders (changes requested):

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/hemophage/corrupted_tongue.dm
GoldenAlpharex: > Since this is attached to an organ, which exists in nullspace, this proc won't actually be able to add the action to the mob. The proc is written with the assumption that the item exists in the mob's loc.

That's actually why I'm granting the action manually. It goes through the `actions` automatically during `on_insert()` to grant them to the mob, and actually removes them during `on_remove()`, specifically for organs.

This is actually 100% functional, and works exactly as intended.

vinylspiders (approved):

GoldenAlpharex:
Been tested for a little while now, let's go.

PR #22642, opened 2023-07-23
sec uniform stuff - chest coverage on adjust, desc tweak, formal uniform thing

Hatterhat:
## About The Pull Request
- adjusting the sec uniform and variants now keeps the chest covered so as to prevent wacky interactions from a false assumption (adjustment doesn't always mean rolled down in this case)
- desc tweak: the nt belt buckle is now a lopland belt buckle because they're not NT or something like that
- formal uniform set to have no alt sprites so you dont accidentally get rid of its unique sprite

fashion officer potential or something also not having your torso exposed on altclick

## Proof of Testing

<details>

image
there's no way to see limb coverage so just trust me on that one  
</details>

## Changelog

:cl:
fix: The security uniform no longer exposes your chest when all you did was roll the sleeves up.
fix: The security formal uniform no longer inherits altsprites from the base uniform, since it didn't have any.
spellcheck: The security uniform's description now mentions a Lopland belt buckle, and not a Nanotrasen belt buckle.
/:cl:

### DIFF
--- a/modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
+++ b/modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
@@ -524,9 +524,11 @@
 */
 //Officer
 /obj/item/clothing/under/rank/security/officer
+	desc = "A tactical security uniform for officers, complete with a Lopland belt buckle."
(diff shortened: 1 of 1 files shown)


GoldenAlpharex (approved):
Looks good, cheers!

PR #22647, opened 2023-07-23
[MODULAR] Off-Duty check proc

softcerv:
## About The Pull Request
Introduces the `check_if_off_duty` proc to living mobs, allowing to check if the mob is currently on or off duty. This is only implemented through a debug item right now.

While this isn't added to anything normally obtainable in-game, this could be very helpful for future projects along with being useful for admins. 

## Proof of Testing

<details>

image

image

Checking as assistant
image  

</details>

## Changelog

:cl:
code: adds in a proc on mobs to check if they are on or off-duty
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/jobs/off_duty_check.dm
+++ b/modular_skyrat/master_files/code/modules/jobs/off_duty_check.dm
@@ -1,0 +1,51 @@
+/**
+ * Checks is the parent mob is currently off duty. Returns `FALSE` if the mob isn't clocked out.
+ *
+ * Arguements
+ * * `whitelisted_deparments` - This list contains the paths of `/datum/job_department`s we want to check if the parent mob is part of.
+ * If present, we will only check if the mob is off-duty if they are a member of one of the departments in the list
+ * Otherwise, if we have nothing in the list, we will check all departments
+ * * `blacklisted_jobs` - This is a list that contains paths of the `/datum/job`s that we will always return `TRUE` for
+ * This typically applies for jobs like assistant where clock status probably wouldn't really matter.
+*/
+
+/mob/living/proc/check_if_off_duty(list/whitelisted_deparments = list(), list/blacklisted_jobs = list(/datum/job/assistant))
+	if(!mind?.assigned_role)
+		return TRUE //If someone lacks a job or mind, we probably don't need to worry about their clock status.
+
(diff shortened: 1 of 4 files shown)


lessthnthree (approved):

modular_skyrat/master_files/code/modules/jobs/off_duty_check.dm
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Arguements
+ * * `whitelisted_deparments` - This list contains the paths of `/datum/job_department`s we want to check if the parent mob is part of.
+ * If present, we will only check if the mob is off-duty if they are a member of one of the departments in the list
+ * Otherwise, if we have nothing in the list, we will check all departments
+ * * `blacklisted_jobs` - This is a list that contains paths of the `/datum/job`s that we will always return `TRUE` for
+ * This typically applies for jobs like assistant where clock status probably wouldn't really matter.
+*/
+
+/mob/living/proc/check_if_off_duty(list/whitelisted_deparments = list(), list/blacklisted_jobs = list(/datum/job/assistant))
```
GoldenAlpharex: Any reason why this has to be on the `/living` level? Like, wouldn't we be better off having it on the `/human` level? I can't really see a usecase where we want to check if any random mob is actually off-duty.

GoldenAlpharex (changes requested):

modular_skyrat/modules/time_clock/code/console_tgui.dm
```diff
@@ -101,13 +101,22 @@
 			if(off_duty_check())
```
GoldenAlpharex, line 101: I'm curious, why didn't you actually use this new proc you added for this?

GoldenAlpharex (changes requested):

modular_skyrat/modules/time_clock/code/console_tgui.dm
softcerv: There might be an instance where someone has an ID that is off-duty but their mind is still on-duty.

GoldenAlpharex: I see.

GoldenAlpharex (approved):
Let's give it a go, then!

PR #4684, opened 2021-04-06
[NONMODULAR]Shocks & Zaps, Episode 2

Vizoplez:
## About The Pull Request
Somewhat WIP and may change. Changes shocks to be much more, or less lethal depending on your siemens coefficient, and what gear you are wearing.

## Why It's Good For The Game
Shocks wont even crit you at this current point in time, even with a cranked up SM and playing IPC (The race w/ the highest siemens coefficient, last I checked).

This actively discourages people from making cranked-up SM builds, and generally bypassing it to the grid, since they themselves might fall victim to the various effects this, and other future PRs will attempt to integrate, such as heart attacks (now in code); and potential 'flashing' of limbs (disintegration) 

## Changelog
:cl:
tweak: Oh god oh fuck. Electrical shocks are now much more lethal. Wear proper protection before tampering with equipment.
/:cl:

PR #22662, opened 2023-07-24
Revert powernet electrocute damage to tg values

Cyprex:
## About The Pull Request
What even was this code. The first if() usually returns negative numbers, the last if and the else are dead code. The second branch can return negative numbers too.
I just reverted this to tg.

The reason the negative numbers weren't noticed is because code grabs cell damage(presumably for machines) and calculates damage based on that and powernet damage and then takes the higher value. Cell damage defaults to 0.

## Proof of Testing

none, this is a quick webedit
<details>
  
</details>

## Changelog

:cl:
balance: Electric shocks from cables deal less damage.
/:cl:

### DIFF
--- a/code/modules/power/powernet.dm
+++ b/code/modules/power/powernet.dm
@@ -95,13 +95,7 @@
 	newavail = 0
 
 /datum/powernet/proc/get_electrocute_damage()
-	//SKYRAT EDIT BEGIN: Declamps burn damage w/ new effects. Original code: return clamp(20 + round(avail/25000), 20, 195) + rand(-5,5)
-	if(avail >= 1000 && avail <= 15000)
-		return round(avail/45000) + rand(-30,2) //Slight buff. Weaker electrical currents are somewhat mitigated.
-	if(avail >= 15000)
-		return round(avail/15000) + rand(-5,10) //You're dead.
-	if(avail >= 20000)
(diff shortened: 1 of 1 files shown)


Cyprex:
Oh- I'm unsure if having a commit change it from tg and then another commit that reverts it later can cause merge conflicts. If it can a //Skyrat edit comment might be useful?

GoldenAlpharex:
> Oh- I'm unsure if having a commit change it from tg and then another commit that reverts it later can cause merge conflicts. If it can a //Skyrat edit comment might be useful?

No, that wouldn't cause merge conflicts, since at the end the code is the same, and that's what matters for diffs.

vinylspiders:
Looks like the changes were from here https://github.com/Skyrat-SS13/Skyrat-tg/pull/4684. They didn't need to be nonmodular in the first place though.

>The reason the negative numbers weren't noticed is because code grabs cell damage(presumably for machines) and calculates damage based on that and powernet damage and then takes the higher value. Cell damage defaults to 0.

That would have been pretty funny if the shocks could _heal_ you.

lessthnthree (approved):

Cyprex:
> They didn't need to be nonmodular in the first place though.

True but honestly given the way this is written that's the least of my worries.

PR #22714, opened 2023-07-27
Correct powernet electrocution damage [NO GBP]

lessthnthree:
## About The Pull Request

Corrects an unintentional powernet nerf in https://github.com/Skyrat-SS13/Skyrat-tg/pull/22662 by adding Skyrat's adjustment to human maxHealth/health.

Also converts that to a define while we're at it.

## Proof of Testing

<details>

dreamseeker_k9e7nZaDpb

</details>

## Changelog

:cl: LT3
fix: Fixed unintentional nerf to electrocution damage when reverting to TG
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/power/powernet.dm
+++ b/modular_skyrat/master_files/code/modules/power/powernet.dm
@@ -1,0 +1,6 @@
+/datum/powernet/get_electrocute_damage()
+	if(avail >= 1000)
+		var/damage = clamp(20 + round(avail/25000), 20, 195) + rand(-5,5)
+		return damage * HUMAN_HEALTH_MODIFIER
+	else
+		return 0
--- a/code/__DEFINES/~skyrat_defines/mobs.dm
+++ b/code/__DEFINES/~skyrat_defines/mobs.dm
@@ -2,6 +2,8 @@
 
 #define PULL_OVERSIZED_SLOWDOWN 2
 
+#define HUMAN_HEALTH_MODIFIER 1.35
+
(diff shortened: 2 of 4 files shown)


modular_skyrat/master_files/code/modules/power/powernet.dm
```diff
@@ -0,0 +1,6 @@
+/datum/powernet/get_electrocute_damage()
+	if(avail >= 1000)
+		var/damage = clamp(20 + round(avail/25000), 20, 195) + rand(-5,5)
+		return damage * HUMAN_HEALTH_MODIFIER
+	else
+		return 0
```
GoldenAlpharex, line 6: I don't think it was an oversight, it directly stated in the changelog that the amount of damage received was lowered.

GoldenAlpharex (commented):
The rest is fine, but I don't know if we really have to make electrocution ramp up in damage. I'd need to see some actual computed numbers to see what kind of effect this has.

lessthnthree:
> The rest is fine, but I don't know if we really have to make electrocution ramp up in damage. I'd need to see some actual computed numbers to see what kind of effect this has.

I was originally going to make it closer to the old PR, but this is simply a 35% addition to match the extra 35 health (so you don't get two 'free' electrocutions before you're even seeing TG levels of impact. Defeats the point of gloves at that point.

600kW 44 -> 59
1.2mW 48 -> 65
1.8mW 92 -> 124

GoldenAlpharex (approved):
Alright, we can always revert this if we think that it's too intense.

PR #22745, opened 2023-07-28
[MIRROR] Add leash component to pAIs that keeps them within range instead of directly teleporting them back, increases default range to max range [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77030
--------------------

## About The Pull Request

Tries to keep pAIs in range of their owner by moving them closer when the owner moves, rather than jarringly teleporting every time the owner gets out of range. Does this by calculating the closest path a nearby tile and forcefully moving you there. Still a bit janky at times but is better than teleporting directly onto the owner 100% of the time I feel. Also prevents you from moving out of range, rather than forcefully teleporting you back.

Increases the default pAI range to the maximum (9 tiles)

## Why It's Good For The Game

New leashing makes being a leashed pAI significantly less jarring and obvious. Ideally we would also have a visible max range too.

Default pAI range was pretty small in my testing and I think it's not unreasonable to think a lot of people won't bother changing it. That they are leashed at all is the important part.

## Changelog

:cl:
qol: pAIs now try to stay within range of their owner, and teleport back only when necessary
qol: Default max pAI range has been changed to the maximum range you can choose (9 tiles)
/:cl:


vinylspiders:
Wait for the fix upstream

vinylspiders (approved):

PR #22768, opened 2023-07-29
[MIRROR] Sapient bots can play voice lines [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/76920
--------------------
## About The Pull Request

Part of the "throwing stuff at the wall and seeing what happens" series of changes.
We were discussing player bots and there was some small support for "what if bots couldn't talk except to other bots and primarily communicated via emoting and playing their prerecorded lines". 
I coded this but I'll be honest, my heart wasn't in it, especially because we have _already_ released sapient bots who can talk so I'd have to justify taking it away in a better way than "I want to see what would happen", so I'm not doing that (feel free to revert the last commit in this PR if you want to carry that cross yourself).

But I'm keeping the button that plays the prerecorded voice messages.
No these lists _aren't_ alphabetised, that's because I wanted to vaguely order them in order of "this is something the bot regularly says" to "this is something the bot rarely says (because it is rude or mean)". This priority is possibly the inverse of how players will want to use the button, but perhaps they should try being politer.

Additionally this PR ensures that if you put a pAI in a bot then it will be able to speak any languages the pAI could.

## Why It's Good For The Game

Saying "I should have been a plastic surgeon" is iconic medbot behaviour and players should be able to do it.
We don't currently have bot TTS voices which directly sound like the bots themselves to apply, and it might be mildly concerning if we did because I believe at least one of the voices is the IRL voice of a community member with some filters added?

## Changelog

:cl:
add: Player-controlled bots can now play the prerecorded lines associated with that bot.
add: Placing a pAI into a bot temporarily teaches that bot all the langauges the pAI could speak.
/:cl:


GoldenAlpharex (approved):
Neat!

PR #22789, opened 2023-07-31
Combat Shotgun is now equippable on back and Outfit slots

TheSS13Melon:
Allows the Combat shotgun to be equipped on the back.

Makes it more consistent with the riot shotgun.

## Proof of Testing

<details>
 image
</details>

## Changelog

:cl:
qol: Allows the Combat shotgun to be equipable on back slots and armor slots.
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/guns/guns.dm
+++ b/modular_skyrat/modules/sec_haul/code/guns/guns.dm
@@ -838,6 +838,19 @@
 	ammo_type = /obj/item/ammo_casing/shotgun/rubbershot
 	max_ammo = 8
 
+/obj/item/gun/ballistic/shotgun/automatic/combat
+	name = "\improper Peacekeeper combat shotgun"
+	desc = "A semi-automatic Nanotrasen Peacekeeper shotgun with tactical furnishing and heavier internals meant for sustained fire. Lacks a threaded barrel."
+	icon = 'modular_skyrat/modules/aesthetics/guns/icons/guns.dmi'
+	worn_icon = 'modular_skyrat/modules/aesthetics/guns/icons/guns_back.dmi'
+	lefthand_file = 'modular_skyrat/modules/aesthetics/guns/icons/guns_lefthand.dmi'
+	righthand_file = 'modular_skyrat/modules/aesthetics/guns/icons/guns_righthand.dmi'
+	inhand_icon_state = "shotgun_combat"
+	inhand_x_dimension = 32
+	inhand_y_dimension = 32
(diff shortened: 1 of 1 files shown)


YakumoChen:
The combat shotgun is a direct upgrade to the riot shotgun and probably one of the most powerful weapons in the game. How is this good for the game?

modular_skyrat/modules/sec_haul/code/guns/guns.dm
```diff
@@ -838,6 +838,19 @@
...
+	desc = "A semi-automatic Nanotrasen Peacekeeper shotgun with tactical furnishing and heavier internals meant for sustained fire. Lacks a threaded barrel."
+	icon = 'modular_skyrat/modules/aesthetics/guns/icons/guns.dmi'
+	worn_icon = 'modular_skyrat/modules/aesthetics/guns/icons/guns_back.dmi'
+	lefthand_file = 'modular_skyrat/modules/aesthetics/guns/icons/guns_lefthand.dmi'
+	righthand_file = 'modular_skyrat/modules/aesthetics/guns/icons/guns_righthand.dmi'
+	inhand_icon_state = "shotgun_combat"
+	inhand_x_dimension = 32
+	inhand_y_dimension = 32
+	w_class = WEIGHT_CLASS_BULKY
+	slot_flags = ITEM_SLOT_BACK | ITEM_SLOT_BELT | ITEM_SLOT_OCLOTHING
```
GoldenAlpharex, line 852: This isn't the proper way to override very specific variables, in this case, underneath the typepath, you'd just want to have `w_class` and `slot_flags`, but honestly I wouldn't make changes to the slot flags because they already fit on backs even without these changes.

GoldenAlpharex (changes requested):
I'm honestly not sure how useful this would be, I already did some tests in-game to see if the shotgun could already be worn on the back (and it could), and I don't think it should be worn on the belt, and otherwise I'm not certain that it's needed to make it so the combat shotgun can fit in suit storage, all in all. It's intentionally a bulkier shotgun (`WEIGHT_CLASS_HUGE` vs `WEIGHT_CLASS_BULKY`), so I'm not certain about the reason behind this whole PR as a whole, and the body of the PR hasn't really been convincing me of the upsides of this.

burgerenergy:
Shouldn't this type of thing be upstream, too? I don't see how it's Skyrat centric

GoldenAlpharex:
I honestly don't think this is required or really something that we should pursue, sorry.

PR #22829, opened 2023-08-02
Horny

Useroth:
image
image
image

image
image
image

:cl:
add: Added two variants of impala style horns, a smaller and bigger version
/:cl:


GoldenAlpharex (approved):
Looks fine to me now that the taller horns are at a more respectable size.

PR #22865, opened 2023-08-03
[MIRROR] Improved PDA Direct Messenger [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/75820
--------------------
## About The Pull Request

Fixes #76708, Closes #76729 (sorry Zephyr)

This PR expands the Direct Messenger UI, adding a chat screen for each available messenger that you can find, and moving message sending over to TGUI. 

This chat screen includes a message log that displays messages sent by you as well as messages received from the recipient. This gets rid of the previous chat log, which just had all messages thrown together that you received or have sent, in one big list. 

Furthermore, all messaging is now done inside the UI. This kills all TGUI popups you would ever need to send messages forever (except for quick replies). Use the input bar on the bottom, press Enter or the Send button, and it sends your message. Spam mode is now done in the UI too, via a text field you can find in the contacts list.

Additionally, because I have a habit of blowing things massively out of scope, I've also completely refactored how messages and chat logs are stored in the PDA messenger. I plan on using this in a PR that merges the chat client with the messenger, sometime in the future. Sorry this took so long.

Stuff left to do before I open this PR for review:
- [x] Add "recent messages"
- [x] Add "unread messages"
- [x] Add message drafts
- [x] Make photo sending not shit
- [x] Implement the edge cases for automated and rigged messages
- [x] Make sure shit isn't fucked
- [x] Profit

<details>
  <summary>Screenshots</summary>
  
dreamseeker_HIrEfrap5X
qIOWhVld4l
dreamseeker_LIqwi05i4O
BIYxNVjGL7
dreamseeker_IWdoSsUinC
L9DxzFHDEF
rAuDgqBLNE
dreamseeker_voui6S8MUf
image

</details>

## Why It's Good For The Game

The UI has largely stayed the same since modular tablets were added a year ago. Even better, direct messaging has been the same since PDAs were first added *more than a decade ago*. Imagine that.

Now we finally actually (!) make use of those brand new features that we got from the TGUI switch in this regard.
## Changelog
:cl: distributivgesetz
add: Updated Direct Messenger to v6.5.3. Now including brand new individual chat rooms, proper image attachments and a revolutionary message input field!
add: Added a "Reset Imprint" option to the PDA painter.
refactor: Refactored PDA imprinting code just a bit.
fix: PDAs should now properly respond to rigged messages.
/:cl:

PR #22867, opened 2023-08-03
Resolves conflicts in #22865

strandsofivy:
This just resolves left-over conflicts from the mirror of my upstream PR #22865 

Main things were the curator PDA messenger mute removal not applying correctly, and the `main.css` file being sorted and therefore messing up a lot of lines. Also the message that appears when you write PDA messages has been added back again so that feature does not get accidentally sent to the shadow realm.


vinylspiders:
Thanks!

PR #22874, opened 2023-08-04
[MIRROR] Lightgeist AI [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77287
--------------------
## About The Pull Request

The lightgeist AI controller was marked as "these aren't intended to exist outside of player control" but this is actually untrue.
Lightgiests can appear under AI control either from a rare vent clog event or in the "patch of eden" lavaland ruin.

As they heal things by "attacking" them, I made it so that they will target wounded creatures to "attack" them, though they will only heal them quite slowly, significantly less efficiently or effectively than a medibot (for example).
While making this change I also added a couple more parameters to their "healing hands" component, now they can no longer repair beepsky or heal cyborg limbs. It's fleshies only.

Lightgeists will attempt to heal _anything_ they can see which is both injured and has healable damage, which is notable because while this makes the "Patch of Eden" ruin a nice place of respite for a wounded miner you should be careful to let them finish up before you leave.
If they follow you out, they will attempt to heal any of the fauna that you are attacking. Worse still, most Lavaland mobs are not signatories of the Geneva convention and have no compunctions against killing field medics.

The majority of listed file changes in this PR is that I made some attempt at splitting our massive list of blackboard keys across several files, in order to cause myself a headache based on which of my (or other people's) open basic mob AI PRs gets merged first.
Also I fixed a bug where the goliath attack forecast would runtime if it killed a mob which qdels itself on death (guess how I found that out).

## Why It's Good For The Game

Adds a bit more character to a lavaland area and rare event.

## Changelog

:cl:
add: Lightgeists under AI control will selflessly heal any wounded creature that they see.
balance: Lightgeists can no longer repair non-organic tissue.
/:cl:


GoldenAlpharex (approved):
Nice.

PR #22924, opened 2023-08-07
[MIRROR] Base Female sprite tweaks [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77407
--------------------
## About The Pull Request

ASS STUFF HAS BEEN REMOVED BUT I STILL HATE IT

This PR tones down the proportions of the female base sprites, as currently they have about SIX extra pixels on the ass and a random pixel missing from the neck, which breaks some hairstyles & makes the neck look quite stupid.
It also adds a couple pixels to the male one because theirs was so stupidly SMALL it looked like they had no tailbone (still does, kind of).

Here is the current sprite 
image
& new sprite (only neck pixel removed)
image

## Why It's Good For The Game

Fixes some hairs

image

## Changelog

:cl:
image: fixes weird inconsistency on the neck and butt of the female base sprite
/:cl:


GoldenAlpharex (approved):
Should be fine, we can fix our modular heads as we see fit in the future if it becomes a problem down here.

PR #22963, opened 2023-08-09
[MIRROR] Anomaly gear: Polymorphic Field Inverter [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77383
--------------------
## About The Pull Request

https://www.youtube.com/watch?v=5bGrqJkpbLw

Adds a new device you can research called the "Polymorphic Field Inverter".
It requires Advanced Biotechnology and Anomalous Research to unlock.

Completing that research lets you print out a fancy belt. Using it for 5 seconds on a mob (dead or alive) which is organic in nature and isn't some kind of human or megafauna (monkeys are fine) will store a copy of its DNA. Scanning another creature will replace the original stored DNA. Storing DNA does absolutely nothing unless you have also inserted an active Bioscrambler core.

If the item is activated with a Bioscrambler core then equipping it grants you the ability to transform into the same kind of creature as the one you scanned (and back). As in spells such as Wildshape you share health with your transformed form and if you die while transformed then you die, so be careful when turning into something like a cockroach which has one hit point.

It has an associated experiment which makes the research free, which is to scan two humans with (non-prosthetic) organs that they didn't have when the round started. You can accomplish this through surgery, DNA infusion, or a bioscrambler anomaly. You can do it on some monkeys if nobody is cooperating with you.
I copied a teeny bit of this code from #73969 although I didn't actually end up needing most of it.

These are absolutely codersprites so if anyone wanted to make cooler ones, let me know.

## Why It's Good For The Game

Adds some fun tech linked to an anomaly core which has literally no in-game purpose.
There's probably some neat sandbox interactions you can get out of this.

## Changelog

:cl:
add: Adds a new technology which utilises Bioscrambler anomaly cores to allow you to transform into animals.
/:cl:


GoldenAlpharex (approved):
This is funny.

PR #23115, opened 2023-08-16
makes all suit-slot items e.g. jackets able to fit guns in their suit storage

Hatterhat:
## About The Pull Request
uuh someone remind me to get a new screenshot of a bomber jacket holding a shotgun or something i'll get back to you on that
(note: doesn't work for chaplain stuff. if this is blocking the pr's merge i guess i can finagle it)

does shooting people/open-carry as deterrence count as roleplay

## Proof of Testing
its right there

## Changelog

:cl:
balance: Every oversuit slot item, e.g. jackets/winter coats/etc. is now capable of holding a gun.
/:cl:


modular_skyrat/master_files/code/modules/clothing/suits/_suits.dm
```diff
@@ -2,6 +2,12 @@
 	/// Does this object get cropped when worn by a taur on their suit or uniform slot?
 	var/gets_cropped_on_taurs = TRUE
 
+/obj/item/clothing/suit/Initialize(mapload)
+	. = ..()
+	if(islist(allowed))
+		allowed += list(/obj/item/gun,
+						/obj/item/storage/belt/holster,)
```
Paxilmaniac: holy tabs batman

modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
```diff
@@ -787,6 +787,35 @@
 		/obj/item/gun/ballistic/rifle/boltaction, //fits if you make it an obrez
 		))
 
+/obj/item/storage/belt/holster/chameleon
```
Paxilmaniac: did you know that syndicate holsters in suit slots probably can break really bad due to the fact that not every belt has a belt_mirror sprite, which suit slot items use. I'm pretty sure I blocked syndie holsters from fitting in normal armor vests for this reason.

modular_skyrat/master_files/code/modules/clothing/suits/_suits.dm
```diff
@@ -2,6 +2,12 @@
 	/// Does this object get cropped when worn by a taur on their suit or uniform slot?
 	var/gets_cropped_on_taurs = TRUE
 
+/obj/item/clothing/suit/Initialize(mapload)
+	. = ..()
+	if(islist(allowed))
+		allowed += list(/obj/item/gun,
```
YakumoChen: This needs to be a list similar to what fits on security armour at minimum, item/gun is far too broad

YakumoChen:
While I largely disagree with this PR in general (we have a belt slot for open carry regardless of clothing) I think this should be limited to handguns at most, I can't see civilian jackets being standard to holster a shotgun or longarm.

kittysmooch:
let me hold a sword on that bad boy >:)

modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
GoldenAlpharex: I think that's a very valid point to bring up and consider.

modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
GoldenAlpharex, line 1: I feel like these changes should be made in a separate PR.

modular_skyrat/master_files/code/modules/clothing/suits/_suits.dm
GoldenAlpharex: I'm going to have to agree with Chen here, this is far too broad.

GoldenAlpharex (changes requested):
I'm overall not entirely convinced about this change, really.

Furthermore, I feel like holsters fitting in every single suit storage is a bit much as well, I think that it's a fine balance as-is to have to go for more armor-oriented suits in order to have the ability to wear holsters on suit storages instead of the belt slot.

Hatterhat:
lack of interest from maintainers so let's not

PR #23173, opened 2023-08-18
[NO GBP] How I found the differences between pick() and pick_weight() || Makes CIN surplus crates spawn with their other items again, rather than a thousand rifles

Paxilmaniac:
## About The Pull Request

image

I assure you that every one of these crates coming with ten rifles and nothing else is not a good thing.

## Proof of Testing

image

## Changelog

:cl:
fix: CIN military surplus crates now come with more than just rifle cases, as they should be
/:cl:


GoldenAlpharex (approved):
Can relate, I've done something similar in the past.

PR #23221, opened 2023-08-20
[MIRROR] Separate init_possible_values() from icon generation [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77660
--------------------

## About The Pull Request
`init_possible_values()` now only ever returns a list of values instead of both values and icons. The responsibility of icon generation has moved from this proc to a new `icon_for(value)` proc, which returns the icon/typepath of an atom. A runtime is thrown if any value does not have an icon.

This boosts production initialization times by 3+ seconds, but is primarily done to allow for further optimizations of this process. As an example, it is a mapping of value -> icon so that in the code that executes this we can have more fine grained control over tick overrun.

As a bonus, fixes #67092.


vinylspiders:
this one needs some work to get it functional

vinylspiders:
I got it functional, but still needs a bit of work: for posterity: still needs to have some of the modular icon gen refactored to take advantage of this PR's improvements, and to fix that linter a repath of the `body_markings` pref is required.

vinylspiders (approved):

PR #23277, opened 2023-08-23
MCR Disabler Barrel

projectkepler-ru:
## About The Pull Request

Adds the MCR disabler barrel using the grandfathered code for gandalf disabler beam that was still left in there
What this does is make the MCR an oversized disabler that you can't 1 hand
On one hand, it's a big disabler that you only need to hit 1 less shot to down someone, on the other, you are trading away a huge utility intentionally for this thing, with the right attachment it's a better disabler  and with a bayonet, you are ready to make yourself the bravest officer known to Lopland. 

You can research this thing from RnD enhanced MCR upgrades to get it unlocked at your local thingmabob

I don't plan to add it to cargo import since at that point you've way better options and this would just be a waste for that. 
Well imagine this, it's an oversized disabler, it can't kill people unless you attach a bayonet to it, and non-dead people are able to roleplay, dead people can't.  41 stamina damage vs 30 and I'd say it's pretty damn worth it

## Proof of Testing

<details>
  
image
image
image
image

</details>

## Changelog

:cl:
add: Added the MCR Disabler Barrel alongside it being printable in tech lathe
image: MCR disabler sprite, looks abit ugly though.
/:cl:


nikothedude:
Wasnt the entire reason this thoiing was removed was because it gave people an excuse to carry a egun around that they could just take the disabler off?

burgerenergy:
> Wasnt the entire reason this thoiing was removed was because it gave people an excuse to carry a egun around that they could just take the disabler off?

That's still true of the vanilla E-guns that are both readily obtainable from a mixture of cargo/crafting and, in somecases, just around though?

YakumoChen:
Making it a fully better disabler will just lead to officers using one any chance they get (eg, higher than green alert). 

Right now officers don't raid the armoury too often even when they technically can (non-lethal shotguns, CMGs on blue), they only really arm up for lethal threats.

With this I can see officers switching to MCRs at the drop of a hat because its not only superior to their disabler, but can be made lethal on demand.

Not to mention its easy to make them a powerhouse shitting disabler shots with upgrades + self charge.

burgerenergy:
MCR upgrades don't actually do anything for damage, just heat (which is fake) and charge.

The MCR is also a long rifle and cannot fit in the backpack, let alone the belt like the disabler can, so I don't see it as an issue personally.

In this case it also isn't in cargo, so it depends on what is usually the last research done (as it's very expensive and very questionable in worth.)

projectkepler-ru:
I did thought about making it buyable from cargo but seeing that it would be much better to just buy the proper disabler(which you can dual wield) at that point I didn't see a reason to let that be an option. The whole point of the MCR is that you can swap out the parts and turn it into a different gun and I feel like this fit that role just fine. And as for the parts? well I couldn't see how it's different from just using the advanced energy gun or a normal disabler at that point. I can see this being used more often because it's non lethal and security get to shoot twice as often with it.I didn't understand the other concerns

KathrinBailey:
drafted on request

YakumoChen:
> MCR upgrades don't actually do anything for damage, just heat (which is fake) and charge.
> 
> The MCR is also a long rifle and cannot fit in the backpack, let alone the belt like the disabler can, so I don't see it as an issue personally.
> 
> In this case it also isn't in cargo, so it depends on what is usually the last research done (as it's very expensive and very questionable in worth.)

They fit on armour suit slots and on belt slot, which is plenty enough for most uses, especially given security is most likely to have it.

> which you can dual wield

I don't see security dual-wielding disablers... ever, really. Its not really worth the accuracy loss

> The whole point of the MCR is that you can swap out the parts and turn it into a different gun and I feel like this fit that role just fine. And as for the parts? well I couldn't see how it's different from just using the advanced energy gun or a normal disabler at that point. I can see this being used more often because it's non lethal and security get to shoot twice as often with it.I didn't understand the other concerns

This is the problem, energy guns are not widely available so security does not get a lot of excuses to use one. MCRs are in the armoury round start, and you've mentioned the disabler damage is buffed, so security will always have a reason to use it over their normal disabler if given the chance, and it can be upgraded for more shot capacity.

nikothedude:
I do have to ask, though, what is different about MCRs that makes them fine to carry around at all times (if security is even allowed, or would be allowed, to do this)?

Just like you can't get a egun roundstart, you probably wouldn't be allowed to carry around a MCR that can be quickly made lethal, and if people WERE, thats not exactly the fault of giving it the disabler mode, for aformentioned reason of "we dont let people carry around eguns for no reason"

EDIT: do we not have eguns/laser guns in armory? Doesnt change my point that we wouldn't let people carry eguns around, but I swear we had some

YakumoChen:
> 

Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.

Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'

projectkepler-ru:
> > 
> 
> Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> 
> Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'

True but removing the disabler barrel would drain the cell

nikothedude:
> > 
> 
> Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> 
> Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'

I mean, non-lethal could easily mean "it can only be nonlethal", right?
SOunds more like a policy issue to me that we're letting people carry around lethal weaponry and saying "well it CAN be nonlethal"

YakumoChen:
> > > 
> > 
> > 
> > Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> > Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'
> 
> True but removing the disabler barrel would drain the cell

No? Removing the cell is the only thing that drains the cell, does it not

YakumoChen:
> > > 
> > 
> > 
> > Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> > Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'
> 
> I mean, non-lethal could easily mean "it can only be nonlethal", right? SOunds more like a policy issue to me that we're letting people carry around lethal weaponry and saying "well it CAN be nonlethal"

You are missing the point also where its also  literally a better disabler than the disabler gun

projectkepler-ru:
> > > > 
> > > 
> > > 
> > > Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> > > Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'
> > 
> > 
> > True but removing the disabler barrel would drain the cell
> 
> No? Removing the cell is the only thing that drains the cell, does it not


Sorry I didn't explain. that it's what this PR would make it do

burgerenergy:
> > > > 
> > > 
> > > 
> > > Blue alert lets you carry any 'non-lethal longarm' which means you're just a mouse click away from being Lethal if you want. We don't have a lot of desirable options in the armoury right now that let you do that.
> > > Again, this provides a buffed disabler, so there's incentive to carry these at any alert above green, instead of just 'when lethals are auth'd'
> > 
> > 
> > I mean, non-lethal could easily mean "it can only be nonlethal", right? SOunds more like a policy issue to me that we're letting people carry around lethal weaponry and saying "well it CAN be nonlethal"
> 
> You are missing the point also where its also literally a better disabler than the disabler gun

It is also a.) significantly larger b.) better by **a** shot (and ammo if/when cell upgraded)

Zenitheevee:
> Wasnt the entire reason this thoiing was removed was because it gave people an excuse to carry a egun around that they could just take the disabler off?

*deep breath*
It is also worth remembering that the biggest changes to MCR's happened not that long afterwards. You cant hot-swap batteries anymore and carry a large bag of extras anymore. While it is 100% worth keeping in mind that its still lethal on demand, It isn't going to boast the biggest issues MCR's had earlier.

Will people *probably* switch to MCR's just to be gamey? Yeah. Probably. Maybe. 
But balance wise... A roundstart MCR has a 12 shot battery. The disabler has a 20 shot battery. 
An MCR Disabler bolt in code does 41 damage. A normal disabler bolt does 30.
Disabler has a flat 600 damage per full charge boast, A bit over the MCR's starting 492... But the first upgrade already bumps it up to a nice 612, A fully upgraded MCR is at the level of 840(Correction: 984 Assuming nothing affects capacity besides recharge)... 

However, You'll likely still have to remove the barrel to make it lethal, meaning if you want a different barrel you'll have to pocket it for later.
Meanwhile.. I can just put two disablers in my satchel, Have a dedicated lethal gun without having to futz with it in my suit slot, And have the only difference being i have to shoot more to do the same damage.

PR #23303, opened 2023-08-24
[MODULAR] Adds the scryer to the loadout attempt 2

W4rd3nn:
## About The Pull Request
This adds the new modlink scryer to the loadout

It adds no more mechanical functionality over a PDA message. It really shouldn't be locked behind advanced modsuit research. This lets people get the silly lil necklace in the loadout.

## Proof of Testing

<details>

image

</details>

## Changelog

:cl:Motho
add: MODLink Scryer necklaces are now available in the loadout.
/:cl:

### DIFF
--- a/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_neck.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_neck.dm
@@ -265,6 +265,10 @@ GLOBAL_LIST_INIT(loadout_necks, generate_loadout_items(/datum/loadout_item/neck)
 	name = "Maid Neck Cover"
 	item_path = /obj/item/clothing/neck/maid
 
+/datum/loadout_item/neck/link_scryer
+	name = "MODlink scryer"
+	item_path = /obj/item/clothing/neck/link_scryer
+
 /*
 *	DONATOR
 */
--- a/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_neck.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/loadout_datum_neck.dm
@@ -266,7 +266,7 @@ GLOBAL_LIST_INIT(loadout_necks, generate_loadout_items(/datum/loadout_item/neck)
 	item_path = /obj/item/clothing/neck/maid
 
 /datum/loadout_item/neck/link_scryer
-	name = "MODlink scryer"
+	name = "MODlink Scryer"
 	item_path = /obj/item/clothing/neck/link_scryer
 
 /*


W4rd3nn:
OKAY!!! It's fixed. I have no idea what the issue was

lessthnthree (approved):

modular_skyrat/modules/loadouts/loadout_items/loadout_datum_neck.dm
```diff
@@ -265,6 +265,10 @@ GLOBAL_LIST_INIT(loadout_necks, generate_loadout_items(/datum/loadout_item/neck)
 	name = "Maid Neck Cover"
 	item_path = /obj/item/clothing/neck/maid
 
+/datum/loadout_item/neck/link_scryer
+	name = "MODlink scryer"
```
GoldenAlpharex:
```suggestion
	name = "MODlink Scryer"
```

GoldenAlpharex (changes requested):
My only issue with this is that the MODlink system is currently independent of telecommunications, so it works literally all the time, no matter what happens in the round.

I'd appreciate if we had it depend on telecomms like other telecommunications-related forms of communication, before having it available at round-start for everyone, because it's kind of a bit too strong otherwise, imo.

Think of it as an incentive to fix an issue for a feature to be allowed through. :)

W4rd3nn:
I mean, I can kinda see it but also station-bounced radios which are printable from the autolathe at the beginning of the round defeat any telecomms tampering.

Regardless, making it dependent on telecomms is out of my skillset, so if that's a dealbreaker, then I'll just wait for someone else to do it.

Nerev4r:
> My only issue with this is that the MODlink system is currently independent of telecommunications, so it works literally all the time, no matter what happens in the round.
> 
> I'd appreciate if we had it depend on telecomms like other telecommunications-related forms of communication, before having it available at round-start for everyone, because it's kind of a bit too strong otherwise, imo.
> 
> Think of it as an incentive to fix an issue for a feature to be allowed through. :)

Not to be a balancejak, but we already had that in the form of the NIF Hivemind, no? I think it's neat if space explorers w/ this feature for instance can phone home to their partner at the bar or w/e.

Paxilmaniac:
> 

This is literally the entire purpose of the modlink system. There are so so SO many ways to communicate directly with one other person or with a small group of other people without tcomms that this does not make sense to me as an argument. Comms being down prevents MASS communication of information for example screaming HELP NUKIES DORMS over common radio. You can't do that with a modlink. You're calling one guy who also happens to have one, and that one guy can only call one other guy at a time. And, again, as juniper said, if communication with literally anyone was a problem then surely NIF hiveminds must cease to exist too. Surely we must remove the emergency toolboxes on the interlink shuttle because they have radios that work regardless of tcomms in them.

CandleJaxx:
yeah, linking this to telecomms seems a bit of a stretch esp. when, as stated, one autolathe can immediately allow you access to common, which everyone can hear due to the intercoms

YakumoChen:
> This is literally the entire purpose of the modlink system. There are so so SO many ways to communicate directly with one other person or with a small group of other people without tcomms

Counterpoint: the link is cross-Z and requires no advance setup beyond owning the modlink in loadout

Most options I can think of that bypass tcomms are:
- very public (bounced radios)
- same z only (bounced radios)
- require mutual foreknowledge and setup (custom freq bounced radios)
- require intentionally acquiring/buying the equipment beforehand (all)
- User knows where recipient is (holocalls)

Even the hivemind you have to at least remember to buy in the round, modlink is a comms-free PDA you can click a loadout button and forget you have it until you need it

It's also hands free and won't be interrupted in combat or something

YakumoChen:
Ultimately this begs the question of "why is this needed in loadout if everyone who should get one also has a PDA?"

YakumoChen:
Also this will bypass ghost cafe protection which will make me mad because I keep removing radios from ghosts

W4rd3nn:
It's worth mentioning the scryers arent technically usable from roundstart. They still need a battery and I believe they also need their frequency set. This just means that people dont have to wait the 40-60 minute time period before advanced modsuit modules are researched

YakumoChen:
> It's worth mentioning the scryers arent technically usable from roundstart. They still need a battery and I believe they also need their frequency set. This just means that people dont have to wait the 40-60 minute time period before advanced modsuit modules are researched

https://github.com/tgstation/tgstation/pull/77957

It is usable from roundstart, and power cells are widely available everywhere (all flashlights contain one, for example)

I don't think a technicality refutes any of the points here.

It *does* add more mechanical functionality than a PDA message, so the Why isn't accurate.

CandleJaxx:
> Also this will bypass ghost cafe protection which will make me mad because I keep removing radios from ghosts

this feels more like a policy issue, if people are actively trying to find ways to contact the station in the ghost cafe, they should be looking at an ahelp

> > It's worth mentioning the scryers arent technically usable from roundstart. They still need a battery and I believe they also need their frequency set. This just means that people dont have to wait the 40-60 minute time period before advanced modsuit modules are researched
> 
> [tgstation/tgstation#77957](https://github.com/tgstation/tgstation/pull/77957)
> 
> It is usable from roundstart, and power cells are widely available everywhere (all flashlights contain one, for example)
> 
> I don't think a technicality refutes any of the points here.
> 
> It _does_ add more mechanical functionality than a PDA message, so the Why isn't accurate.

as for this, it seems to me that this adds the same level of difficulty as acquiring a station-bound radio to communicate does, since you either have to print + charge a battery, or find one in maint

YakumoChen:
> this feels more like a policy issue, if people are actively trying to find ways to contact the station in the ghost cafe, they should be looking at an ahelp

I do not want ghosts to have the opportunity and I do not want ghosts showing up on comms directories (like PDA or modlink lists)

> as for this, it seems to me that this adds the same level of difficulty as acquiring a station-bound radio to communicate does, since you either have to print + charge a battery, or find one in maint

A station bounced radio does not provide private, untappable, cross-Z, one-way-setup (both do not have to mutually agree on a frequency via other method of communication beforehand), telecomms-ignoring, discreet communication

GoldenAlpharex (approved):
Considering you still have to add a cell and set the frequency manually, I'm not as opposed to this as I was initially.

I'm hoping this won't become an issue with ghost roles (including the cafe), as we'll have to put in additional restrictions if that's the case.

PR #23409, opened 2023-08-30
Makes it harder to INITIAL D as snail

YakumoChen:
## About The Pull Request

Ticks snail shells with IMMUTABLE SLOW, so red slime potions don't work. Stable red slime cores still work, but that's not solvable with the current code. Blame @GoldenAlpharex for removing it from the species level.

Watching snails do laps around the hallway and make huge slippy trails was funny a few times, but lets not.

## Proof of Testing

image

## Changelog

:cl:
fix: Made it less trivial for snailpeople to crawl around at meth speeds.
/:cl:


GoldenAlpharex (approved):
Thank you for the fix, I thought they had it already.

PR #23454, opened 2023-09-01
MCR damage buff

burgerenergy:
## About The Pull Request

A while ago TG gave lasers a minor buff (#21088) but this was not folded over into our modular MCRs are the time, which use bespoke projectile code for whatever reason. This updates the projectiles, adding five across the board.

One caveat to my rendition includes heavy lasers, which the original buff did not. Heavy lasers were double, hence +10. As far as I can tell, the only thing on Skyrat that realistically uses heavy lasers are turrets and exos, neither of which would feel hampered by the MCR doing more damage then them. If I _didn't_ do it, then the lancer kit would feel even worse to use and remain incredibly underutilized and underpowered for the drawbacks.

Of note, this does mean a vanilla MCR would do (slightly) more damage then a single CMG lethal ballistic round, although that shoots twice.

MCRs should be doing the same damage as regular lasers and not be strictly inferior.

## Proof of Testing

It's a text edit...

## Changelog

:cl:
balance: Buffed MCR damage in line with an upstream generic laser buff
/:cl:

### DIFF
--- a/modular_skyrat/modules/microfusion/code/projectiles.dm
+++ b/modular_skyrat/modules/microfusion/code/projectiles.dm
@@ -16,7 +16,7 @@
 /obj/projectile/beam/laser/microfusion
 	name = "microfusion laser"
 	icon = 'modular_skyrat/modules/microfusion/icons/projectiles.dmi'
-	damage = 20
+	damage = 25
 
 /obj/projectile/beam/microfusion_disabler
 	name = "microfusion disabler laser"
@@ -36,7 +36,7 @@
 /obj/projectile/beam/laser/microfusion/superheated
 	name = "superheated microfusion laser"
 	icon_state = "laser_greyscale"
-	damage = 15 //Trading damage for fire stacks
+	damage = 20 //Trading damage for fire stacks
 	color = LIGHT_COLOR_FIRE
 	light_color = LIGHT_COLOR_FIRE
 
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
you could, additionally, just remove the damage var from the lasers in general, which would make them default to the TG laser damage, so we don't have to update this again later

burgerenergy:
That involves more coding rigamorale then I am comfortable with, so I will say out of scope unless someone esle opts to do it.

That said, unless you like inherited the damage value and made it a math equation you'd still have to do it for the unique ones, like the repeater. Again, not really within my talent set, I'm a simple girl. I edit number.

RatFromTheJungle:
> That involves more coding rigamorale then I am comfortable with, so I will say out of scope unless someone esle opts to do it.
> 
> That said, unless you like inherited the damage value and made it a math equation you'd still have to do it for the unique ones, like the repeater. Again, not really within my talent set, I'm a simple girl. I edit number.

you would actually just, remove your number define, as in, remove the "damage = XX" part, which would make it automatically use the base laser-projectiles damage

burgerenergy:
Ah, I see. But that would still just be only for the base vanilla (and clown version I guess) MCR, all the other ones would have to be specifically defined, anyway. The (vanilla) heavy laser is derived off the root, too. I could trim the two lines if a code maintainer would like, but it's insignificant bloat versus someone inept like ✨me✨ being able to accurately read the page and contrast the tools if they want to code dive.

(I also think it helps catch this sort of thing so long as we have to specifically define the alternative fire modes.)

lessthnthree (approved):

PR #23499, opened 2023-09-03
[MIRROR] replaces surgical and coroner duffels with surgery trays and medical carts, resprites razor [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77975
--------------------
## About The Pull Request
replaces all instances of surgical duffels with surgery trays, and all coroner duffels with morgue surgical trays.
they contain about the same items, with surgical trays/carts also having bone gel and tape, since their list of holdable items is much more limited. 
the surgery tray is a diagetic storage unit that displays any surgery tool it's holding in a small, almost world-state esque form on top of it. it can be carried around by dragging it on to yourself, but it will slow you down like an unzipped duffel would. it can also be deployed by activating it, and pulled around, but it will roll noisily in the process.
currently, all tool tiers are supported - from alien, to advanced, to cruel, to normal.
image
here you can see just a few of the possible combinations of tools - default, default morgue, advanced, and alien. but any combination of these tools should work together somewhat cleanly, as you can see here:
image
also adds a medical razor variant, because otherwise they wouldn't fit with the look of the other tools on the tray, before and after here:
image
## Why It's Good For The Game
having constantly visible tool storage like this means you don't have to worry about the one tool you need from the bag being stolen - if it is, you can see it before an operation starts, and plan around that! it also gives a little flexibility to mappers - if they'd like a more mobile cart for their medbay, or if they'd like the somewhat stationary tray. it also plain looks cool, and isn't quite as clunky as a duffelbag would be.
## Changelog
:cl:
add: adds medical carts and surgery trays
image: gives the surgery razor a unique sprite
/:cl:


lessthnthree (approved):

ORCACommander:
not sure why but the update script missed a bunch of maps

lessthnthree:
okay I'll look

PR #23528, opened 2023-09-05
millstone wrench, screwdriver, and building + stone stuff

jjpark-kb:
## About The Pull Request
title: millstone wrench, screwdriver, and building + stone stuff
the millstone can now be wrenched-- which anchors and unanchors it.
the millstone can now be screwdriver'd-- which deconstructs it.
the millstone can now be created using six cut stones.
you can find rough stones from any mineral wall (the mining kind).
you can cut the rough stones into the processed version (which is required for certain crafting).

currently, once the millstone is gone it is gone-- no replacing it; this aims to fix that.
additionally, the millstone may be in a location others disagree with, which with the wrench, it makes it possible to move it.
there was almost no way to easily get stone, unless you found an ore vein-- not very common; a 1/25 chance to get stone from a mineral way makes it not super spammy, but obtainable.
there was a logic error with cutting the rough stones-- it had an or rather than an and. that logic error prevented any cutting/processing of those rough stones, so this PR fixes that.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
qol: you can now construct, deconstruct, and anchor/unanchor the millstone
add: rough stones will now occasionally drop from mineral walls (the mining kind)
fix: you can now cut/process rough stones
/:cl:


Jamiemundy:
I like this a lot and wish for this to be added. I've personally anticipated the ability to get stone bricks so we could build with them too..

modular_skyrat/modules/primitive_cooking_additions/code/millstone.dm
```diff
@@ -88,6 +88,21 @@
...
+		attacking_item.play_tool_sound(src)
+		anchored = !anchored
+		balloon_alert(user, "[src] [anchored ? "anchored" : "unanchored"]")
+		return TRUE
+
+	if(attacking_item.tool_behaviour == TOOL_SCREWDRIVER)
+		attacking_item.play_tool_sound(src)
+
+		for(var/i in 1 to 6)
+			new /obj/item/stack/sheet/mineral/stone(get_turf(src))
```
vinylspiders:
```suggestion
			var/obj/item/stack/sheet/mineral/stone = new (get_turf(src))
			transfer_fingerprints_to(stone)
```

Just a small detail, if the mill had prints on it before so will the rock shards produced by dismantling it.

modular_skyrat/modules/primitive_cooking_additions/code/millstone.dm
```diff
@@ -88,6 +88,21 @@
...
+
+		Destroy()
```
vinylspiders:
```suggestion
		qdel(src)
```

vinylspiders (commented):
Looking at this as a whole this would be nice in the future to refactor this to use wrench_act() and screwdriver_act() instead of doing that behavior in `attackby()`.

Just making a note, not asking you to do that now (although if you do, I'll add the refactor label so you get more gbp)! Looks good to me other than the comments below.

vinylspiders (approved):

PR #23578, opened 2023-09-08
radical station centralization - prevents ds2/interdyne (lavabase)/tarkon from always spawning

Hatterhat:
## About The Pull Request
see title. takes away the always_spawn flag from ds2/lavabase/icebase/tarkon

no idea if this is del or balance so

radical station centralization

to be more serious i just personally think no ghostroles should be a guaranteed/forced spawn, ever. ds2 and ice/lavabase act as mini-stations with little to no relevance to the actually ongoing round, tarkon has its own little gameplay loop but after you rebuild the place there's not much else to do. i don't think this is a great philosophy to have for space ruin/ghostrole design

## Proof of Testing

<details>
i hit ctrl shift b but if you want i can take a screenshot later just tell me to do that  
</details>

## Changelog

:cl:
del: Thanks to the power of Legalese, Interdyne's continued presence on Lavaland has been made somewhat less permanent due to threats of legal action by Nanotrasen; this means that it doesn't always spawn anymore.
del: The Syndicate has reconsidered the continued deployment of DS2 in frontier space, and has resorted to flipping a coin on whether to deploy it or not, making it not always spawn.
del: Port Tarkon exploded. Or did it? We're not sure, it kinda fades in and out of reality when we're not looking. Maybe it's lost in another-another asteroid belt or something. Anyways, it's no longer guaranteed to appear.
/:cl:

### DIFF
--- a/modular_skyrat/modules/mapping/code/icemoon.dm
+++ b/modular_skyrat/modules/mapping/code/icemoon.dm
@@ -10,7 +10,6 @@
 	suffix = "icemoon_underground_syndicate_base1_skyrat.dmm"
 	allow_duplicates = FALSE
 	never_spawn_with = list(/datum/map_template/ruin/lavaland/skyrat/syndicate_base)
-	always_place = TRUE
 
 /datum/map_template/ruin/icemoon/underground/skyrat/mining_site_below
 	name = "Mining Site Underground"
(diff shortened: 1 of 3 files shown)


burgerenergy:
Ah, yes. Let's take these groupies with thriving communities and ensure that they only get to sometimes play their characters, as a prank. For reasons, including them always existing is bad ghostrole gameplay. What even?

GoldenAlpharex:
I don't like that reason. They're not creating that many issues, and if the players are having fun there? That's what matters to me.

PR #23580, opened 2023-09-08
Interdyne QoL map things

burgerenergy:
## About The Pull Request

Interdyne hasn't gotten touched up in a bit, so based on feedback from the current playerbase similarly to the DS-2 touch up I've gone ahead and swept through somethings. No drastic changes, nor any changes to the spawners themselves.

This is not the greater big remappening that is in the cards, just a pick me up.
<details>

- Fixed duplicate miner dorm names. (Icebox has four, why...)
- Add doorlocks to ALL dorms, although the miners still have a window. Weirdos.
- Lavaland's Dyne has been updated to have doublebeds and random sheets, too.
- Common request and practice was to shift the machines along the wall in RnD.
- Adds a public shower, praise be. This does make the bathroom a little awkwardly spacious, however the floor plan is what it is.
- Adds a roughshod brig section to intern miscreants, unfortunately removing one of the catwalks in the neck. Legitimately tragic.
- Made the couch in the library lounge southfacing, as it inherently looks better for three+ people.
- Added requested vending machines that otherwise required a trip to DS-2. (Or going to the station, but that's quite a trip for some smokes!)
- Fixes that kitchen vendor not being marked as off-station.
- Standardizes medical between the two versions. Additionally, adds a bloodchest, medipen supply, and a portable tank stand so the medical facility can actually do surgery on anesthetic now.
- Rearranges the kitchen and botany area.
- Standardizes the vault. Primarily it puts down the ammo bench instead of needing to be built on Lavaland.
- Adds a fax machine, although not in the Syndie loop so the station can interact with it. It would have gone in the DO's office, but there wasn't much space and the DO has no real reason to go in there after spawning, so visibility was a concern.
- In general a couple of rearrangements here and there.

</details>

Unrelatedly holy shit mappers have it easy these days

Quality of life improvements are good for the soul. Also Interdyne was getting really stinky with a lack of shower.

## Proof of Testing

image

</details>

(Not pictured as I forgot to save it, but I did test both versions.)
## Changelog

:cl:
qol: Both versions of Interdyne received some map touch ups. Standouts include; a fax machine, dorm locks, a make shift brig area, a revamped bathroom, botany, and kitchen, as well as backporting Icemoon improvements to the Lavaland version.
/:cl:


Zergspower:
Just a quick passing glance:
<details>
<summary>Icedyne:</summary>

image
Camera at the top has a bit of a gap

image
This button does nothing - maybe put the trash can next to the table?

The Fax seems a little out of place in the middle of the hall, on Icedyne it could go in the observation room or maybe in the RnD  room for both.
</details>

<details>
<summary>Lavadyne:</summary>

image
Same deal with the button not doing anything - maybe put the trash can next to the table?

This may or may not have been you but all the lightswitches except for the research ones are missing?
</details>

burgerenergy:
i hate buttons.

I cleaned up the lightswitches some, someone added it to Icedyne and never backported it to Lava like, well, a lot of things. Both are made up of maybe five areas so it isn't really useful, but might as well. 

I also shifted the trash and plant around, why not. Not sure why the plant didn't spawn for you, does on my tests.

Zergspower:
the plant being there in your screenshots make sense why the trashcan was there lol

Jolly-66 (dismissed):
I would like your CL to display some of the more important changes that would be player facing.

<details>
<summary> Review </summary>

> Common request and practice was to shift the machines along the wall in RnD.

I really don't like this. If people are groaning that stuff are in the way and they need to move things every round, its an IC issue in my eyes and not a mapping problem. Revert that change.

Use a random vendor helper here.
image

Two things here:
1. Subtype the vendor
2. Rearrange this corner. The vendor, while accessible in the corner, looks ugly.
image

Use a different colored trim here, the color clashing doesn't look good to me.
image

Another two things here:
1. Move this crate.
2. Rearrange the stuff on the table with the box. It looks goofy with the box on top of everything.
image

> Adds a fax machine, although not in the Syndie loop so the station can interact with it. It would have gone in the DO's office, but there wasn't much space and the DO has no real reason to go in there after spawning, so visibility was a concern.

I don't know about the fax machines.  Let me contact policy to make sure this would be an "okay" change to have. As it stands from a mapping PoV, I personally don't care, but I don't want to step on policy toes.

</details>

_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
```diff
@@ -1978,6 +2016,12 @@
 	dir = 1
 	},
 /obj/machinery/light/directional/south,
+/obj/structure/table,
+/obj/machinery/fax{
+	fax_name = "Interdyne";
```
RatFromTheJungle: interdyne always calls themselves 'Interdyne Pharmaceuticals" when talking to non-syndies, so the fax should read the same

burgerenergy:
image
image
image

I quite liked the contrast of the terracotta, but sure. I also apparently forgot to save when adding the kitchen machine wall line up to Lavaland, so that got added - you saw it already with the Icemoon, no surprises there. The RnD room can't be entirely reverted as to fit the fax machine, but the set up was shifted up one from original. Changelog also updated.

vinylspiders:
So this has been sitting a while, and I've gone through and confirmed that all of the requested changes have been made.

vinylspiders (approved):
I think it is time.

PR #23601, opened 2023-09-09
[MIRROR] [TEST-MERGE FIRST] Wound refactor number two: Full synthetic support [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/78124
--------------------
## About The Pull Request

Heavily refactors wounds AGAIN.

The primary thing this PR does is completely change how wounds are generated and added - while the normal "hit a guy til they bleed" stuff works about the same, asking for a specific type of wound, say, like how vending machines try to apply a compound fracture sometimes, isnt going to work if we ever get any non-organic wounds, which the previous refactor allowed.

With this PR, however...
* You can now ask for a specific type of wound via get_corresponding_wound_type(), which takes wound types, a limb, wound series, etc. and will try to give you a wound fitting those specifications - but also, a wound that can be applied to a limb.
* This will allow for a vending machine to apply a compound fracture to a human, but a collapsed superstructure (assuming my synth wounds go in) to a robot

There are now new "series types" and "wound specific types" that allow us to designate what series are "mainline" and randomly generatable, and what are "alternate types" and must be generated manually - you can see the documentation in wounds.dm. 

The behavior of limping and interaction delays has been abstracted to datum/wound from bone wounds to allow just, general ease of development

Pregen data now allows for series-specific wound penalties. Example: You could set a burn wound's series wound penalty to 40, which would make wound progression in the burn category easier - but it would not make it any easier to get a slashing wound. As it stands wound penalties are for wounds globally

Scar files are now picked in a "priority" list, where the wound checks to see if the limb has a biostate before moving down in said list. Example: Wounds will check for flesh first, if it finds it - it will use the flesh scar list. Failing that, they then check bone - if it uses that, it will use the bone scar list. This allows for significantly more modular scars that dont even need much proc editing when a new wound type is added

Misc changes: most initial() usage has been replaced by singleton datums, wound_type is now wound_types and thus wounds can accept multiple wound types, wounds can now accept multiple tool behaviors for treatment, wounds now have a picking weight so we can make certain wounds rarer flatly, 

This PR also allows for wounds to override lower severity wounds on generation, allowing eswords to jump to avulsions - but in spirit of refactoring, this has been disabled by default (see pregen data's competition_mode var).
## Why It's Good For The Game

Code quality is good!

Also, all the changes above allow wounds to be a MUCH more modular system, which is one of its biggest problems right now - everything is kinda hardcoded and static, making creative work within wounds harder to do.

## Changelog
:cl:
refactor: Refactored wounds yet again
fix: Wounds are now picked from the most severe down again, meaning eswords can jump to avulsions
fix: Scar descs are now properly applied
/:cl:


nikothedude:
DO NOT MERGE THIS TIL IM DONE WITH IT

code/__DEFINES/wounds.dm
```diff
@@ -53,49 +71,221 @@
...
+// Have to put it here so I can use it in the global list of wound series
+/// See muscle.dm
+#define WOUND_SERIES_MUSCLE_DAMAGE "skyrat_wound_series_muscle_damage" // We use a super high number as realistically speaking TG will never increment to this amount of wound series
+// SKYRAT EDIT ADDITION END
+
+/// A assoc list of (wound typepath -> wound_pregen_data instance). Every wound should have a pregen data.
+GLOBAL_LIST_INIT_TYPED(all_wound_pregen_data, /datum/wound_pregen_data, generate_wound_static_data())
+
+/// Constructs [GLOB.all_wound_pregen_data] by iterating through a typecache of pregen data, ignoring abstract types, and instantiating the rest.
+/proc/generate_wound_static_data()
```
GoldenAlpharex, line 137: Procs in the `__DEFINES` folder, shake my smh my head.

GoldenAlpharex (approved):
Should be ready for a test-merge.

PR #23607, opened 2023-09-10
[non modular] disables TG "hold up" for the foreseeable future

RatFromTheJungle:
## About The Pull Request
The main issues being:
- Doesn't make logs
- Doesn't really align with CI policy
- and the one use outside of this, is niche at best (hostage situations)

## Changelog

:cl:
del: Guns, collectively, can no longer right-click holdup people
/:cl:


GoldenAlpharex (approved):
I've never really liked the lack of a visual component to them, and considering they don't leave logs, I don't really see a point in having it. Once they'll have logs, and some visuals maybe, then we can reconsider.

PR #23628, opened 2023-09-11
[SEMI-MODULAR] Gives robotics bonesetters, non-medibot health analyzers, and some gauze

nikothedude:
## About The Pull Request

Title.

Adds 2 bonesetters, 4 stacks of gauze, and 2 "surplus" analyzers that cant be used for medibot assemblies to the robodrobe.

As it stands, robo already has to deal with wounds (dislocations) and as such might need something to tell them how to handle them (bonesetters). A whole suite of synth wounds is also going to come from upstream, making this a downright necessity.

Robotics will be taken by surprise by a bunch of new wounds that NOBODY has experienced before. Gauze is universally used and can be a quick treatment option while they figure it out, and analyzers can tell them straight up how to do it.

The analyzers are blacklisted from being used in medibots because I dont want robotics wasting all their analyzers on bots when they really are going to need them for wounds + I dont want to give robo 4 extra medbots at roundstart.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: 2 Bonesetters, 4 stacks of gauze, 2 health analyzers that cant be used for medibots, all in the robodrobe
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/construction.dm
+++ b/code/modules/mob/living/simple_animal/bot/construction.dm
@@ -273,6 +273,10 @@
 	switch(build_step)
 		if(ASSEMBLY_FIRST_STEP)
 			if(istype(W, /obj/item/healthanalyzer))
+				var/obj/item/healthanalyzer/analyzer = W // SKYRAT EDIT ADDITION BEGIN -- EXTRA ROBOTICS HEALTH ANALYZERS
+				if (!analyzer.can_be_used_in_medibot())
+					user?.balloon_alert(user, "no attachment ports!")
+					return // SKYRAT EDIT ADDITION END
 				if(!user.temporarilyRemoveItemFromInventory(W))
 					return
 				healthanalyzer = W.type
--- a/modular_skyrat/modules/medical/code/health_analyzer.dm
+++ b/modular_skyrat/modules/medical/code/health_analyzer.dm
@@ -1,0 +1,11 @@
+/// If TRUE, this analyzer can be used for medibot construction. If FALSE, it cannot. Returns TRUE by default.
+/obj/item/healthanalyzer/proc/can_be_used_in_medibot()
+	return TRUE
+
+/obj/item/healthanalyzer/no_medibot
+	name = "surplus health analyzer"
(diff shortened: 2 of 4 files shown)


GoldenAlpharex:
Why 4 health analyzers? Seems a bit overkill, especially if it's just to spit in their face when they try to use them to make medibots and just can't.

trefoils:
Science can print health analyzers from their lathe. You can print bonesetters from the autolathe. both of these are fairly early research too. This feels a tad unneeded? besides maybe the gauze. Why not given science the ability to print advanced health analyzers instead for the new cyborg wounds

nikothedude:
> Science can print health analyzers from their lathe. You can print bonesetters from the autolathe. both of these are fairly early research too. This feels a tad unneeded? besides maybe the gauze. Why not given science the ability to print advanced health analyzers instead for the new cyborg wounds

As the author of synth wounds, it's in my opinion that jobs who are supposed to treat wounds should have the tools to do so right from the start. Bonesetters and health analyzers, even though available early, aren't immediately available. The autolathe isn't guaranteed to exist either.

nikothedude:
> Why 4 health analyzers? Seems a bit overkill, especially if it's just to spit in their face when they try to use them to make medibots and just can't.

Spitting in their face?? What a weird way to say it

If 4 is overkill I can lower it, I just kinda picked a random number that felt good

nikothedude:
> Science can print health analyzers from their lathe. You can print bonesetters from the autolathe. both of these are fairly early research too. This feels a tad unneeded? besides maybe the gauze. Why not given science the ability to print advanced health analyzers instead for the new cyborg wounds


Also, you're assuming robos KNOW this. It's quite possible a robo might not know they can print them - but it's very likely for a robo to be aware they can just buy one, as it's pretty common for people to check their room's vendor when they try a job.

nikothedude:
All around having a bonesetter and analyzer isnt REALLY important for the simple reason of they can just _print_ them, but in my opinion its best if those things are already in the room they are supposed to be used in.

Zergspower:
would it benefit to instead make a robotics kit like how the syndiebase has?

lessthnthree (approved):

PR #23634, opened 2023-09-11
Void Raptor cargo door buttons

lessthnthree:
## About The Pull Request

Fixes label for the unloading side, sets door IDs as numericals so you quit locking yourself out when trying to change it with a multitool.

## Proof of Testing

<details>

https://github.com/Skyrat-SS13/Skyrat-tg/assets/83487515/28135549-c14c-4f29-bbeb-1ac5bce55680

</details>

## Changelog

:cl: LT3
fix: Fixed Void Raptor cargo door labels and IDs
/:cl:

### DIFF
--- a/_maps/map_files/VoidRaptor/VoidRaptor.dmm
+++ b/_maps/map_files/VoidRaptor/VoidRaptor.dmm
@@ -27727,13 +27727,13 @@
 	},
 /obj/effect/turf_decal/bot,
 /obj/machinery/button/door/directional/east{
-	id = "cargounload";
+	id = 68;
 	layer = 4;
-	name = "Loading Doors";
+	name = "Unloading Doors";
 	pixel_y = 6
 	},
 /obj/machinery/button/door/directional/east{
-	id = "cargoload";
+	id = 69;
 	layer = 4;
 	name = "Loading Doors";
 	pixel_y = -6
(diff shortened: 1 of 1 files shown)


_maps/map_files/VoidRaptor/VoidRaptor.dmm
```diff
@@ -27727,13 +27727,13 @@
 	},
 /obj/effect/turf_decal/bot,
 /obj/machinery/button/door/directional/east{
-	id = "cargounload";
+	id = 68;
```
vinylspiders, line 27730: Not sure which is which, but why not something more descriptive like `"cargounload_north"` and `"cargounload_south"` instead? 

So you can tell more easily at a glance which door the button goes to.

vinylspiders (changes requested):

_maps/map_files/VoidRaptor/VoidRaptor.dmm
vinylspiders: though 69 gave me a chuckle

lessthnthree: The multitool can only set numeric IDs on blast door controllers, so the problem is you can't construct a button that operates the door if the door has a non-numeric value.

You can only change the ID on a blast door while it's open, so if it's closed there's no method to get it operational.

vinylspiders: Ah good to know!

vinylspiders (approved):

PR #23647, opened 2023-09-12
Removes Changeling Horror-Form's Reduced Click Delay

Literallynotpickles:
## About The Pull Request
Guts the next move modifier (the reduced click delay; faster attack speed) for changeling's horror-form. **Bringing this upstream _(to here)_ per requests.**

this is a very bandaid change to changeling horror-form, but next move modifier (the reduced click delay); on top of everything else that this form offers, ergo;

- innate brute passive regen (against ballistic-only sec btw lol.)
- nearly the hp of an elite fauna
- 40 damage + wounding
- ventcrawl (was honestly tempted to remove this too)

~~...gets incredibly irksome in a server where we're not allowed to even metaknowledge changelings existing. I personally don't think this is going to make that specifically go away; but (hopefully) it can help alleviate it in-game until better code strolls along - assuming that metaprotections are staying.~~ Rant a little irrelevant since different server culture/gameplay.

Generally speaking about next move modifier - it isn't a good thing to have on something _this_ high-damaging when most means of it in-game are far more nichely limited in obtaining; like DNA Vault & Wishgranter, FOTN (Melee only, no smashing stuff...)

## Proof of Testing
One line removal of a next move mod. Probably fine.

## Changelog

:cl:
balance: Changeling Horror-Form no longer has reduced click-delay.
/:cl:

### DIFF
--- a/modular_skyrat/modules/horrorform/code/true_changeling.dm
+++ b/modular_skyrat/modules/horrorform/code/true_changeling.dm
@@ -32,7 +32,6 @@
 	attack_verb_continuous = "rips into"
 	attack_verb_simple = "rip into"
 	attack_sound = 'sound/effects/blobattack.ogg'
-	next_move_modifier = 0.5 //Faster attacks
 	butcher_results = list(/obj/item/food/meat/slab/human = 15) //It's a pretty big dude. Actually killing one is a feat.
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
i've half the mind to say horrorform should just be axed, because it's like, untennable, but it wouldn't be my call

lessthnthree (approved):

PR #23741, opened 2023-09-16
Makes flashlights not need cells to operate anymore, by consequences removes that one feature where you can change flashlight light power that nobody uses ever.

Paxilmaniac:
## About The Pull Request

Flashlights don't need power cells to work anymore, and as a consequence of this, that one feature of flashlights that you probably didnt know existed where you can make them stronger/weaker at the cost of more or less power consumption also has to go.

There's no real actual reason I can see for wanting the most basic of all things, a light, to require a cell to work that I can find or think of? There's so many exceptions to the cell requirement as well through various random items that most people just go out of their way to not use flashlights and use like a lamp or something of the sort. Of all the things to have a cell requirement, basic flashlights ain't one of them that needs to have it.

## Proof of Testing

image
image

## Changelog

:cl:
del: The ability to change a flashlight's intensity has been removed due to the fact they no longer have cells to draw more or less power from in those modes.
qol: Flashlights no longer pointlessly require power cells to function
/:cl:


burgerenergy:
A thought; could we keep high intensity mode but make it a on-cooldown sort of thing? Like, you ctrl+click it to turn it into HYPER MODE but it only lasts a minute~ before clicking back down for a minute or something.

Paxilmaniac:
> A thought; could we keep high intensity mode but make it a on-cooldown sort of thing? Like, you ctrl+click it to turn it into HYPER MODE but it only lasts a minute~ before clicking back down for a minute or something.

sounds like something that would be fun upstream you could try it there

vinylspiders:
"why was that sound change in there btw?"

It was to prevent the clicky sound that normally happens when you turn a flashlight off from playing when the battery dies.

Anyhow, I'm not sure we have to fully remove this feature. 

What if you instead made it so that the cells do not `process()`/are effectively unlimited charge, but leave the ability to remove the batteries? E.g. simply check if the cell is present once when turning on.

That way you can still do things like scavenge cells from flashlights but without the annoying aspects of having batteries run out on you. ~~And perhaps do something sneaky like deliberately remove the batteries from someone's flashlight.~~

HoodMilk:
This should also fix the bug where you can put MCR cells into flashlights and have an infinitely usable EMP device.

vinylspiders (changes requested):
Remaking my comment as a request: If you are doing this, make sure `modular_skyrat/master_files/sound/effects/flashlight.ogg` is deleted as it's not being used for anything else. Or alternatively add that portion of the code back.

vinylspiders (approved):

PR #23787, opened 2023-09-18
[Modular] New Scar Removal Medicine

sqnztb:
## About The Pull Request

New chem that'll clear temporal scars. Carpotoxin used to do it but it was removed with tg's #59382 with the 'we have a clear scar slots' button justification. That's cool but we don't have the clear scar button in character preferences. Could bring it back, or could add a bit of flavor to it and reintroduce the effect to a new chem. I chose the latter.

Scars are cool sometimes. Having to disable them, join, cryo, and then reenable them just to clear them up for a time is not. 

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: New medicine to clear up scars, made from synthflesh, ethanol, and miner's salve. Be sure to apply it via patches!
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/reagents/chemistry/reagents/other_reagents.dm
+++ b/modular_skyrat/modules/customization/modules/reagents/chemistry/reagents/other_reagents.dm
@@ -71,6 +71,7 @@
 		to_chat(M, span_notice("[pick("I feel oddly calm.", "I feel relaxed.", "Mew?")]"))
 	..()
 
+#define DERMAGEN_SCAR_FIX_AMOUNT 10
 /datum/reagent/medicine/dermagen
 	name = "Dermagen"
 	description = "Heals scars formed by past physical trauma when applied. Minimum 10u needed. Touch application only."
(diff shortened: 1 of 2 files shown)


lessthnthree (approved):

modular_skyrat/modules/customization/modules/reagents/chemistry/reagents/other_reagents.dm
```diff
@@ -70,3 +70,30 @@
...
+
+/datum/reagent/medicine/dermagen/expose_mob(mob/living/exposed_mob, methods=TOUCH, reac_volume, show_message = TRUE)
+	. = ..()
+	if(!iscarbon(exposed_mob))
+		return
+	var/mob/living/carbon/scarred = exposed_mob
+	if(scarred.stat == DEAD)
+		show_message = 0
+	if(!(methods & (PATCH|TOUCH|VAPOR)))
+		return
```
vinylspiders:
```suggestion
	if(!(methods & (PATCH|TOUCH|VAPOR)))
		return
	var/mob/living/carbon/scarred = exposed_mob
	if(scarred.stat == DEAD)
		show_message = FALSE
```
Just flipping these around so the early returns are first. And then in byond we use `FALSE` instead of `0` when using booleans (they are exactly the same under the hood though, it's just for readability).

vinylspiders (dismissed):
Looks good to me other than that!

modular_skyrat/modules/customization/modules/reagents/chemistry/reagents/other_reagents.dm
sqnztb: Since I was lazy as hell and this was a synthflesh copy/paste, synthflesh is the only spot in the codebase that still uses `show_message = 0` instead of `FALSE`. Kind of amusing.

lessthnthree (approved):

PR #23805, opened 2023-09-19
[Modular] Electric welder no longer gets a free cell on print

desertmote:
## About The Pull Request

Brings the electric welder in-balance with the other welders, which also print empty.

Consistency is good. Less messing around with dumping low tier cells just to replace them with high tier cells.

## Proof of Testing

<details>
  
dreamseeker_qGHfIpuoFh

</details>

## Changelog

:cl:
balance: Electric welders are printed without cell
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/cargo/exports/tools.dm
+++ b/modular_skyrat/master_files/code/modules/cargo/exports/tools.dm
@@ -1,2 +1,2 @@
 /datum/export/weldingtool/experimental
-	export_types = list(/obj/item/weldingtool/electric)
+	export_types = list(/obj/item/weldingtool/electric/no_cell)
--- a/modular_skyrat/modules/electric_welder/code/electric_welder.dm
+++ b/modular_skyrat/modules/electric_welder/code/electric_welder.dm
@@ -18,6 +18,11 @@
 	. = ..()
 	AddComponent(/datum/component/cell, cell_override, CALLBACK(src, PROC_REF(switched_off)))
 
+/obj/item/weldingtool/electric/no_cell/Initialize(mapload)
+	. = ..()
+	AddComponent(/datum/component/cell, start_with_cell = FALSE)
+	cut_overlays()
+
 /obj/item/weldingtool/electric/attack_self(mob/user, modifiers)
 	. = ..()
 	if(!powered)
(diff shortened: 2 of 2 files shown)


softcerv:
I don't think it is reasonable to compare this to other welders considering that this is meant to replace the experimental welder. 

Additionally, having the welder spawn without the cell adds unnecessary tedium.

desertmote:
~Its not tedious, you're just used to the luxury of it being the exclusive item ingame that prints with a cell included, not to mention it being fully charged, and T2.~

As well as real engineers swap out the cell on print anyway, for a T4 cell. Effectively making printing the electric welder LESS tedious than before.

softcerv:
> Its not tedious, you're just used to the luxury of it being the exclusive item ingame that prints with a cell included, not to mention it being fully charged, and T2.

Yes it is, the item that this replaces don't need anything additional to work. T2 cells aren't something that are extraordinarily expensive to print, nor do they have a very high capacity, I don't really see this as a real issue.

desertmote:
Sorry, refer to the second part of my initial comment about how this creates less tedium.

softcerv:
> As well as real engineers swap out the cell on print anyway, for a T4 cell. Effectively making printing the electric welder LESS tedious than before.


That's not true though, not everyone is going to go through the effort of printing out a new cell. Especially if there is something like a blob where a mass amount of welders need made. 

Why does the supermatter crystal have piping already made for it, when real engineers replace it anyways?

HoodMilk:
What if we just brought back  experimental welders?

YakumoChen:
We're removing flashlight batteries anyway, leaving electric welders as the last item to use the battery system. It's time to go back to experimental welders and remove the whole module.

ghost:
Tedium or not, this change is sound. Engineers swap the cell out immediately anyways, and have spare cells around their department. **As an engineer main,** I don't see an issue with this PR.

~~Plus, with Blob disabled, the kind of lrp "spam print 100 welders and leave them on the desk" doesnt exist any more so yay~~

Namwty:
> **As an engineer main,** I don't see an issue with this PR.

I would rather not have science desk duty more deadass than it is, people hate doing it, so why would we make it even worse for those who do it. Instead of printing welder and handing over, you will have to get a battery and charge it. Fun fact, you need to build multi cell charger, which also means you need to build circuitry printer, doesn't sound too bad, if not for overall science gameplay loop, which does not need to be more tedious.

Ya see, engineers won't be the ones who would need to print and charge EVEN MORE cells. Science already has some brainmelting mechanics, build every common machine in the game just to deconstruct em, build 12 damn toilets made out of three different materials, build 7 uranium walls and 6 plasma walls then scan em. Top that with xenobiology/ordenance and fact that you have to rush all of this without even ability to take a break, otherwise you fall behind in research points, which then causes people including **engineers** to go ham about "How science is useless and doesn't have *insert research/item here* at *insert time*". Even going to mention if you are solo scientist and have general expectancy of needing to upgrade every department. 

**Fact is, people will go to science for stuff, not to engineering.**

Tbf just replace em with experimental welders.

lessthnthree:
Superseded by https://github.com/Skyrat-SS13/Skyrat-tg/pull/23832

PR #23843, opened 2023-09-21
Removes some old code that wasn't removed when it should have been from a missed mirror

GoldenAlpharex:
## About The Pull Request
Double proc definitions for the `update_fire_overlay()` proc on carbon, I'm curious if it's one of the reasons behind the fire overlay that's just not going away. I couldn't test that it would fix that one issue, as it's not something that's easily reproduceable.

Less code that could cause bugs == more better

## Changelog

:cl: GoldenAlpharex
code: Removed some old fire overlay code, which might or might not fix some fire overlay issues. We'll have to see.
/:cl:


vinylspiders:
Nice catch, that's code's been squatting there for a long while it looks like

vinylspiders (approved):

PR #23871, opened 2023-09-23
mold mob fixes

tf-4:
## About The Pull Request

Fixes a few issues with mould mobs. It was a combination of my misunderstanding of how basic mob AI attacks work and forgetting about a component.

It also deletes the mobs on death because there tend to be loads of them and it's annoying to clear up.

Also removes the centaur mutagen splash on death as it did nothing but make a massive mess which is not fun.

## Changelog
:cl:
fix: Fixed some weirdness with mould mob attacks.
qol: Mold mobs delete themselves on death.
/:cl:

### DIFF
--- a/modular_skyrat/modules/mold/code/mold_mobs.dm
+++ b/modular_skyrat/modules/mold/code/mold_mobs.dm
@@ -70,7 +70,7 @@
 	SSvis_overlays.add_vis_overlay(src, icon, OIL_SHAMBLER_OVERLAY, layer, plane, dir, alpha)
 	SSvis_overlays.add_vis_overlay(src, icon, OIL_SHAMBLER_OVERLAY, OIL_SHAMBLER_OVERLAY_LAYER, EMISSIVE_PLANE, dir, alpha)
 
-/mob/living/basic/mold/diseased_rat/melee_attack(atom/target, list/modifiers, ignore_cooldown)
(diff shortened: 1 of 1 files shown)


RatFromTheJungle:
u should make the mobs explode and drop guns like pinatas

modular_skyrat/modules/mold/code/mold_mobs.dm
```diff
@@ -69,6 +70,18 @@
...
+	. = ..()
+	if(!isliving(target))
+		return
+
+	var/mob/living/ignite_target = target
+	if(prob(ignite_chance))
+		ignite_target.adjust_fire_stacks(additional_fire_stacks)
+
+	if(ignite_target.fire_stacks)
+		ignite_target.ignite_mob()
```
vinylspiders:
```suggestion
	ignite_target.ignite_mob()
```
`ignite_mob()` checks for this right away so you should be good with just calling the proc

https://github.com/Skyrat-SS13/Skyrat-tg/blob/f3e8e852f80ce9e6f33773abdd532bcbb7114887/code/modules/mob/living/living.dm#L1594-L1596

vinylspiders (changes requested):
Looks good to me though otherwise, assuming this is all good to go now

vinylspiders (approved):

PR #23885, opened 2023-09-24
Randomized exotic seeds crate

yorii:
## About The Pull Request

Randomizes all the seeds in the exotic seeds crate.

Makes the crate at least 99% less boring.

## Proof of Testing

<details>
<summary>Screenshot</summary>

image

</details>

## Changelog

:cl:
add: Added random seed contents to the exotic seeds crate.
/:cl:

### DIFF
--- a/code/modules/cargo/packs/organic.dm
+++ b/code/modules/cargo/packs/organic.dm
@@ -56,6 +56,8 @@
 		var/item = pick(contains)
 		new item(C)
 
+// SKYRAT EDIT START
+/*
 /datum/supply_pack/organic/exoticseeds
 	name = "Exotic Seeds Crate"
 	desc = "Any entrepreneuring botanist's dream. Contains twelve different seeds, \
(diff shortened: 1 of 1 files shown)


Nerev4r:
Just one tiny problem with it. The exotic seeds crate is necessary as a reliable source to get Seedlings, which aren't available anywhere else.

yorii:
> Just one tiny problem with it. The exotic seeds crate is necessary as a reliable source to get Seedlings, which aren't available anywhere else.

Ah I see, that's an oversight from my part, I actually pulled this PR out of a different fork of the codebase.
I will try and get a better sense of what loot table works best for Skyrat proper and adjust the items and weight values accordingly.

If anyone is familiar with any more requirements that otherwise would break the game please do inform me :)

HoodMilk:
Also exotic seed crates are the only way to get shrubs and (maybe?) rainbow flowers.

vinylspiders:
You should probably take this upstream, or create a modular version of this that coexists alongside the current one in light of the comments (like maybe a randomized version that is discounted because you don't know what you are getting?).

You could also just alternatively modularly overwrite the list (making it weighted) and override the `fill()` proc in `master_files` like you did here, but without the Skyrat edits. I just am not a big fan of the whole Skyrat edit commenting out entire blocks thing when it can be avoided is all.

yorii:
Closing this in favor of upstream PR: https://github.com/tgstation/tgstation/pull/78565

PR #23938, opened 2023-09-26
[MIRROR] Basic blob mobs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/78520
--------------------
## About The Pull Request

I remembered today that blob code is ass, especially blob spores.
There's still a lot to improve but I cleaned up _some_ of it by converting these mobs.
Now they use a newer framework and more signal handling as compared to circular references.

I _expect_ the behaviour here to largely be the same as it was or similar. I haven't added anything fancy or new.

This is a reasonably big PR but at least all of the files are small?
Everything here touched every other thing enough that it didnt make sense to split up sorry.

Other things I did in code:
- Experimented with replacing the `mob/blob` subtype with a component. Don't know if this is genius or stupid.
- AI subtree which just walks somewhere. We've used this behaviour a lot but never given it its own subtree.
- Blob Spores and Zombies are two different mobs now instead of being one mob which just changes every single one of its properties.
- Made a few living defence procs call super, because the only thing super does was send a signal and we weren't doing that for no reason. Also added a couple extra signals for intercepts we did not have.

## Changelog

:cl:
fix: Blob spores will respond to rallies more reliably (it won't runtime every time they try and pathfind).
fix: Blobbernaut pain animation overlays should align with the direction the mob is facing instead of always facing South
refactor: Blob spores, zombies, and blobbernauts now all use the basic mob framework. They should work the same, but please report any issues.
/:cl:


vinylspiders:
Needs bitrunning

vinylspiders (approved):

PR #24002, opened 2023-09-28
[MIRROR] Reworks transformation sting to be temporarily in living mobs, forever in dead mobs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/78502
--------------------
## About The Pull Request

- Reworks transformation sting. 
   - Transformation sting is now temporary, lasting 8 minutes (number not final) in humans. 
      - If used on a monkey, it lasts forever instead. 
      - While the target mob is dead or in stasis, the duration will not progress, making it functionally infinite until revived and taken off stasis, where it will resume its timer where it left off.
   - Chemical cost reduced to 33
   - DNA cost reduced to 2

- Removes TRAIT_NO_TRANSFORMATION_STING, instead just checks for TRAIT_NO_DNA_COPY
   - These were essentially the same traits, so I just combined the two

- Organizes some trait lists alphabetically

- Adds TRAIT_STASIS, to allow for reacting to mobs entering and exiting stasis via COMSIGS
   - Everything that checks IS_IN_STASIS now checks HAS_TRAIT TRAIT_STASIS, which is probably more performant, so that's a bonus. 

## Why It's Good For The Game

A lot of people don't like the current iteration of Transformation Sting, me included

Right now it's only use is for a meme - you make the entire station into felinids until you get lynched, and that's it. 

It's not really a healthy ability for ling's current kit, so this pr attempts to soft rework it to make it a bit more in line with how ling should be acting - turning it into a source of short term confusion between people, or using it on a body to cover your tracks. 

This accomplish it two fold - One, it is now cheap enough to use twice in rapid succession, allowing for quick on-the-spot "BE CONFUSED" situations while you abscond. Two, as mentioned in the last paragraph, you can poke a body of someone you murder to obfuscate the crime scene and maybe help out in taking over someone's identity. 

## Changelog

:cl: Melbert
balance: Transformation sting now lasts 8 minutes, down from permanent. However, the effect is paused for dead and stasis mobs, making it permanent SO LONG AS they stay dead or in stasis. The effect is also permanent if used on a monkey. 
balance: Transformation sting now costs 33 chemicals, down from 50. 
balance: Transformation sting now costs 2 dna points, down from 3. 
fix: Transformation sting works on monkeys again.
refactor: Refactored a bit of human randomization. 
/:cl:

Edit: This includes a refactor of the way that default species bodyparts are stored and rendered, because there was a lot of needless memory usage before and because the randomization of certain parts was not as controllable previously (causing screenshot tests to fail).


GoldenAlpharex (approved):
Should be good to go, great job @vinylspiders!

GoldenAlpharex:
Only failures were due to some rustg mishaps, and not related to the PR. Let's gooooo

PR #24018, opened 2023-09-29
TTS voice muffling on the half mask respirator can be toggled on or off

Paxilmaniac:
## About The Pull Request

See title

There are a lot of characters that use the half mask respirator specifically because its NOT the gas mask and having their voice be constantly muffled by a fashion choice is very stupid I think.

## Proof of Testing

I can't exactly test TTS things on local now can I.

## Changelog

:cl:
qol: The half mask respirator can have its TTS voice muffling properties toggled with control click
/:cl:


vinylspiders (approved):
👍

PR #24043, opened 2023-09-30
[MIRROR] Heretic Rebalance -  Main Knowledge gives free Side points, reduced rune drawing time, Codex Cicatrix is easier and roundstart knowledge [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/77939
--------------------
## About The Pull Request

Heretic Rebalance

Researching the Main Knowledge paths that unlock Side Paths will grant one Side Point that can be used only on those side paths. You can still spend normal knowledge points on them if you wish.

Rune drawing time has been reduced from 30->20 seconds. Codex drawing time has been reduced from 15->8.

Codex Cicatrix is now a roundstart knowledge, works as an amber focus when held in-hand and opened, and has had its recipe changed to: 1 of any non-standard pen (literally anything that isn't the base pen), any book, and either animal hide OR a corpse, any kind.

Added support for using a list inside ritual requirements and a special 'snowflake check' rituals can utilize.

The first non-path knowledge, the Mansus Hand Mark, has had its cost reduced from 2->1 points.
## Why It's Good For The Game

Heretic is an extremely top-heavy antagonist that is EXCEEDINGLY weak in the early-game, free get out of jail card aside, and gets utterly overwhelmed with options extremely quickly once hopping past the mid-game hurdle. You're completely starved for knowledge points to the.. point that you feel like you can't blow any on side paths, even though these are often essential as well! 

Once you hop the hurdle - 40 minutes in, usually - you suddenly find yourself blazing through path after path obtaining a ridiculous amount of points that you don't know what to do with and unlocking spell after spell all at once rather than slowly, meaning it becomes extremely hard to keep focus and actually use these things - attempting to keep 5 new spells in mind during combat will just cause you to stretch yourself thin and die.

This PR is meant to address this by giving Heretics a lot more leeway on the early game so they don't feel nearly as stifled as they are right now, which will also, ideally, help them unlock singular abilities earlier on so they can practice with them. **None of these changes are straight buffs to combat ability, they simply relax the extremely tight restrictions early-game heretics have.**

> Researching the Main Knowledge paths that unlock Side Paths will grant one Side Point that can be used only on those side paths. You can still spend normal knowledge points on them if you wish.

This idea in specific is supported by @ MrMelbert - which I'm using as a request to not instaclose this PR due to the slightly lower than neutral GBP I have. He supports this specific bit at least and doesn't outright disagree with me on the rest, as far as I'm aware from my last conversation with him on the subject a while back. Besides, I have a refactor on mirrors open!

Meta aside, this bit is meant to combat how claustrophobic early heretic feels by allowing the guy to pick one of two side paths for free* after unlocking that path's equivalent of Ash Jaunt, so they have *something* to use besides their sword and stink hand at the start.

> Rune drawing time has been reduced from 30->20 seconds. Codex drawing time has been reduced from 15->8.

The absurdly long drawing time is just extremely arduous and frustrating. It's way too slow but just not enough to let you zone out, and it has a ridiculously noticeable animation the whole way through which guarantees you being found out if you're found out midway through. It makes heretic even SLOWER early on as they need to set up one or several bases which takes forever.

> Codex Cicatrix is now a roundstart knowledge, works as an amber focus when held in-hand and opened, and has had its recipe changed to: 1 of any non-standard pen (literally anything that isn't the base pen), any book, and either animal hide OR a corpse, any kind.

Quite the buff, but I think all of these are good changes. Codex Cicatrix being a roundstart change lets one decide - do they want to rush the book to get multiplied influence strength, or do they want to get all the influences ASAP? It also lets them set up bases quicker at, again, the cost of having a blatantly evil item inside the bag.

The Codex recipe was annoyingly complicated, requiring you to run all around the station in yet another of Heretic's many fetch quests just for something that used to be part of the default kit, especially with how annoying getting animal hide can be. Now it's a lot easier to handle. 

The Amber Focus bit is just as a panic option for heretics if they get caught without one, at least they can use the book, and wielding a book in combat is awesome.

> The first non-path knowledge, the Mansus Hand Mark, has had its cost reduced from 2->1 points.

I don't think there ever really was a need to gate this behind two points, it's the very first unlockable research and it needs you to get two influences for some reason. The reasoning is probably 'heretic setup time so they dont start at full strength', which is solid, but is already well handled enough by their other systems and restrictions as mentioned in the rest of the PR.

Heretics have been on a constant trend of just.. not doing anything. Getting caught early and flopping or never managing to do things of note. They need changes, not because They're UnderPowered And The Competitive Balance Is Skewed, but because it's simply not that fun to *be* a heretic with these issues, and not that fun to FIGHT a heretic that merelay has a stink hand, a blade, two pieces of lint in their pocket, and a dead rat.

I also think the Ritual of Knowledge should 1. not cost points, and 2. be split across the 33% and 66% mark and give 2 points, but that's for another PR.
## Changelog
:cl:
add: Heretic Rebalance
balance: Researching the Main Knowledge paths that unlock Side Paths will grant one Side Point that can be used only on those side paths. You can still spend normal knowledge points on them if you wish.
balance: Rune drawing time has been reduced from 30->20 seconds. Codex drawing time has been reduced from 15->8.
balance: Codex Cicatrix is now a roundstart knowledge, works as an amber focus when held in-hand and opened, and has had its recipe changed to: 1 of any non-standard pen (literally anything that isn't the base pen), any book, and either animal hide OR a corpse, any kind.
code: Added support for using a list inside ritual requirements and a special 'snowflake check' rituals can utilize.
balance: The first non-path knowledge, the Mansus Hand Mark, has had its cost reduced from 2->1 points.
/:cl:


code/modules/antagonists/heretic/heretic_antag.dm
```diff
@@ -33,7 +33,13 @@
 	/// The path our heretic has chosen. Mostly used for flavor.
 	var/heretic_path = PATH_START
 	/// A sum of how many knowledge points this heretic CURRENTLY has. Used to research.
+<<<<<<< HEAD
 	var/knowledge_points = 2 //SKYRAT EDIT - ORIGINAL 1
+=======
+	var/knowledge_points = 1
+	/// How many side path points the heretic has. He gains one of these per main path that splits into two sidepaths. These can be used in place of knowledge points for side paths only.
+	var/side_path_points = 0
+>>>>>>> 2ab4fa3a923 (Heretic Rebalance -  Main Knowledge gives free Side points, reduced rune drawing time, Codex Cicatrix is easier and roundstart knowledge (#77939))
```
softcerv:
```suggestion
	var/knowledge_points = 2 //SKYRAT EDIT - ORIGINAL 1
	/// How many side path points the heretic has. He gains one of these per main path that splits into two sidepaths. These can be used in place of knowledge points for side paths only.
	var/side_path_points = 0
```

softcerv (approved):

PR #24225, opened 2023-10-09
fixes the bayonet x offset for the rengo rifle

Paxilmaniac:
## About The Pull Request

the bayonet on the qm's supergun was floating off the front of the barrel, that's not good i think

see above

## Proof of Testing

do i really gotta its like the most simple thing, you can see what it looks like here 
image

## Changelog

:cl:
fix: The quartermaster's Rengo rifle will no longer have floating bayonets
/:cl:


vinylspiders:
I wish these had an overlay for the inhand icons as well. Looks good though

vinylspiders (approved):

PR #24288, opened 2023-10-12
[MODULAR] You can now rename and describe any loadout item, not just plushies.

Iamgoofball:
## About The Pull Request

You can now rename and describe any loadout item, not just plushies.

Lets players customize their clothing and stuff more. Fun flavor!

## Proof of Testing

https://github.com/Skyrat-SS13/Skyrat-tg/assets/4081722/94d50a00-8352-47d2-94b4-411df58cb5bb

## Changelog

:cl:
qol: You can now rename and describe any loadout item, not just plushies.
/:cl:

### DIFF
--- a/modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
@@ -38,8 +38,8 @@ GLOBAL_LIST_EMPTY(all_loadout_datums)
 /datum/loadout_item
 	/// Displayed name of the loadout item.
 	var/name
-	/// Whether this item can be renamed.
-	var/can_be_named = FALSE
+	/// Whether this item can be renamed and described.
+	var/can_be_named = TRUE
 	/// The category of the loadout item.
 	var/category
 	/// The actual item path of the loadout item.
@@ -114,10 +114,13 @@ GLOBAL_LIST_EMPTY(all_loadout_datums)
 			else
 				stack_trace("[type] on_equip_item(): Could not locate backpack item (path: [item_path]) in [equipper]'s contents to set greyscaling!")
 
-	if(can_be_named && !visuals_only && (INFO_NAMED in our_loadout[item_path]))
+	if(can_be_named && !visuals_only && ((INFO_NAMED in our_loadout[item_path]) || INFO_DESCRIBED in our_loadout[item_path]))
 		var/obj/item/equipped_item = locate(item_path) in equipper.get_all_gear()
 		if(equipped_item)
-			equipped_item.name = our_loadout[item_path][INFO_NAMED]
+			if(INFO_NAMED in our_loadout[item_path])
+				equipped_item.name = our_loadout[item_path][INFO_NAMED]
+			if(INFO_DESCRIBED in our_loadout[item_path])
+				equipped_item.desc = our_loadout[item_path][INFO_DESCRIBED]
 		else
 			stack_trace("[type] on_equip_item(): Could not locate item (path: [item_path]) in [equipper]'s contents to set name!")
 
--- a/modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
+++ b/modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
@@ -114,7 +114,7 @@ GLOBAL_LIST_EMPTY(all_loadout_datums)
 			else
 				stack_trace("[type] on_equip_item(): Could not locate backpack item (path: [item_path]) in [equipper]'s contents to set greyscaling!")
 
-	if(can_be_named && !visuals_only && ((INFO_NAMED in our_loadout[item_path]) || INFO_DESCRIBED in our_loadout[item_path]))
+	if(can_be_named && !visuals_only)
 		var/obj/item/equipped_item = locate(item_path) in equipper.get_all_gear()
 		if(equipped_item)
 			if(INFO_NAMED in our_loadout[item_path])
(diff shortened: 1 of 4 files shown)


nikothedude:
so this is how skyrat dies.......................................................................................................................................................................................................................................................................................................

modular_skyrat/modules/loadouts/loadout_ui/loadout_manager.dm
```diff
@@ -208,10 +208,14 @@
...
 /datum/loadout_manager/proc/set_item_name(datum/loadout_item/item)
 	var/current_name = ""
+	var/current_desc = ""
 	if(INFO_NAMED in owner.prefs.loadout_list[item.item_path])
 		current_name = owner.prefs.loadout_list[item.item_path][INFO_NAMED]
+	if(INFO_DESCRIBED in owner.prefs.loadout_list[item.item_path])
+		current_desc = owner.prefs.loadout_list[item.item_path][INFO_DESCRIBED]
 
 	var/input_name = stripped_input(owner, "What name do you want to give [item.name]? Leave blank to clear.", "[item.name] name", current_name, MAX_NAME_LEN)
+	var/input_desc = stripped_input(owner, "What description do you want to give [item.name]? [MAX_MESSAGE_LEN] character max, leave blank to clear.", "[item.name] description", current_desc, MAX_MESSAGE_LEN)
```
GoldenAlpharex: TGUI stuff please.

Also, max input for the description of items should be 256, we don't need whole flavor texts in there.

modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
```diff
@@ -38,8 +38,8 @@ GLOBAL_LIST_EMPTY(all_loadout_datums)
 /datum/loadout_item
 	/// Displayed name of the loadout item.
 	var/name
-	/// Whether this item can be renamed.
-	var/can_be_named = FALSE
+	/// Whether this item can be renamed and described.
+	var/can_be_named = TRUE
```
GoldenAlpharex, line 42: Do we have anything with this set to FALSE? _Should_ we have anything that sets it to FALSE? Anything we should be concerned about for this?

GoldenAlpharex (dismissed):

modular_skyrat/modules/loadouts/loadout_ui/loadout_manager.dm
Iamgoofball: 256 is a bit low but u da boss, most donator item descriptions wouldn't fit in 256 characters fwiw

modular_skyrat/modules/loadouts/loadout_items/_loadout_datum.dm
Iamgoofball: only thing that used this var was plushies before

modular_skyrat/modules/loadouts/loadout_ui/loadout_manager.dm
Nerev4r: 512 maybe?

vinylspiders (commented):
I very much like the option of giving people the ability to add flavor to their loadout items, however I think this would be better done by applying the loadout prefs to `examine_more()` or something along those lines personally.

Reasons for this: It can become less clear what the item actually is. Descriptions could get quite long, and when you are trying to examine something for game-critical info it may become a hindrance.

Other thoughts: Obviously there is potential for abuse here: e.g. people just renaming things to stupid stuff like shown in your own video...another reason I'd prefer to keep the original descriptions intact (and possibly the names, even--could have the custom names be in parenthesis appended to the original ones).

I'm not the one who is going to be dealing with any of this in-game however, so that is going to fall on the rest of the staff to decide ultimately.

Nerev4r:
> I very much like the option of giving people the ability to add flavor to their loadout items, however I think this would be better done by applying the loadout prefs to `examine_more()` or something along those lines personally.
> 
> Reasons for this: It can become less clear what the item actually is. Descriptions could get quite long, and when you are trying to examine something for game-critical info it may become a hindrance.
> 
> Other thoughts: Obviously there is potential for abuse here: e.g. people just renaming things to stupid stuff like shown in your own video...another reason I'd prefer to keep the original descriptions intact (and possibly the names, even--could have the custom names be in parenthesis appended to the original ones).
> 
> I'm not the one who is going to be dealing with any of this in-game however, so that is going to fall on the rest of the staff to decide ultimately.

People renaming stupid stuff should probably just get honeypotted and banned. This feature has far too much good-faith potential (even just simple shit like 'worn-out hoodie' instead of 'hoodie') for us to be concerned about people that we can just. Tell them to knock it off. It would come off weird if we had, like. 'half-mask respirator (pattern m-4 rebreather)' instead of just having it be fully renamed.

Also no real game-critical stuff is going to be in the loadout.

vinylspiders:
> People renaming stupid stuff should probably just get honeypotted and banned. This feature has far too much good-faith potential (even just simple shit like 'worn-out hoodie' instead of 'hoodie') for us to be concerned about people that we can just. 

Fair enough, I still have concerns nevertheless. When used as intended it is a great thing to have though.

Plum-but-gayer:
this rules SO HARD the amount of flavor and in-game lore rep this will bring is HUGE!! +1
I think bad faith potential is pretty low, if someone tries to get cute with it and names their shirt "cryptographic sequencer" or w/e we can just bonk them, and they can already do this with plushies so I can't imagine any exceptional in game benefit from that. Also in general knowing the exact model of non-armor item a person is wearing isn't super important.

Namwty:
> People renaming stupid stuff should probably just get honeypotted and banned.

This
Generally saying this is no different to flavor text ( arguably just extension to it ), the whole "people would use this in bad faith" can be applied to flavor text and general character creation, just have it reinforced by same rules.

Regardless, this is good as some items have odd naming despite being able to be recolored. So you end up with "Black *item*" that is actually orange.

I think having interesting names with cool descriptions would look better than "Recolorable/Greyscale  cloak/suit/suitskirt". Mayhaps even reduce word bloat for those who don't like it ( example recolorable/greyscale or items like "Buttondown Shirt With Slacks" )

lessthnthree (approved):

PR #24339, opened 2023-10-14
[MIRROR] Miscellaneous Robotic Limb Fixes [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/78905
--------------------
## About The Pull Request

In PR #76817, the damage reduction for robotic limbs was changed from being flat to being multiplicative in nature. This wasn't supposed to have secondary mechanical effects on robot limbs, but as always a few slipped by. This PR just cleans up a few of these unintended mechanical changes.

First of all, the change made robotic limbs not fully immune to space damage. They remained immune to damage from pressure but not damage from temperature. Now people with only robotic limbs are immune to the temperature damage and the moodlets caused by low or high temperature. I have tested it, and it works for people who have all robotic limbs, people who have some robotic limbs and some missing, and even the extreme case of a nugget with H.A.R.S. who has nothing _but_ a robotic chest.

Due to technical limitations (`apply_damage` doesn't have a `required_bodytype` argument), people with some robotic limbs and some organic will still take temperature damage on their robotic limbs.

Second of all, androids are once again immune to crit damage. Normally, crit damage is just constant oxyloss, but this obviously doesn't work for species that don't breath. Instead, those species take a small amount of brute damage. Prior to the damage reduction change, this meant that Androids were effectively immune to crit damage. Now they're immune once again.

Finally, you can now cancel surgery on robotic limbs. I don't think this bug was actually caused by the above PR, but I decided to fix it while I was here.

Closes #77033.
## Why It's Good For The Game

The damage reduction change was explicitly not intended to make robotic limbs not-spaceproof, so the first change is by definition a bugfix. As far as I know, crit damage being brute for nobreath species is a much more recent change than the existence of androids, so I think that also qualifies as a bugfix.
## Changelog
:cl:
fix: Having all augmented limbs will make you properly spaceproof once again.
fix: Androids are immune to crit damage again.
fix: Surgery on robotic limbs can be canceled.
/:cl:


nikothedude:
will make synths spaceproof and temp-proof

Nerev4r:
user above me is a snitch


anyway idk if we can separate out temp-proof for them but idrc about synths being spaceproof tbh

nikothedude:
actually its worth mentioning, synths are already pressure-proof, just not temp-proof

code/modules/surgery/bodyparts/robot_bodyparts.dm
```diff
@@ -235,8 +235,46 @@
...
+	SIGNAL_HANDLER
+
+	var/all_robotic = TRUE
+	for(var/obj/item/bodypart/part in owner.bodyparts)
+		all_robotic = all_robotic && IS_ROBOTIC_LIMB(part)
+
+	if(all_robotic)
+		owner.add_traits(list(
+			TRAIT_RESISTCOLD,
+			TRAIT_RESISTHEAT,
```
vinylspiders, line 267:
```suggestion
			/* SKYRAT EDIT REMOVAL BEGIN - Synths are not immune to temperature
			TRAIT_RESISTCOLD,
			TRAIT_RESISTHEAT,
			SKYRAT EDIT REMOVAL END */
```

code/modules/surgery/bodyparts/robot_bodyparts.dm
```diff
@@ -235,8 +235,46 @@
...
+			TRAIT_RESISTLOWPRESSURE,
+			TRAIT_RESISTHIGHPRESSURE,
+			), AUGMENTATION_TRAIT)
+	else
+		owner.remove_traits(list(
+			TRAIT_RESISTCOLD,
+			TRAIT_RESISTHEAT,
```
vinylspiders, line 276:
```suggestion
```suggestion
			/* SKYRAT EDIT REMOVAL BEGIN - Synths are not immune to temperature
			TRAIT_RESISTCOLD,
			TRAIT_RESISTHEAT,
			SKYRAT EDIT REMOVAL END */
```

vinylspiders (commented):
We're just going to disable the temperature part for now.

lessthnthree (approved):

PR #24359, opened 2023-10-16
ant and worm farms

jjpark-kb:
## About The Pull Request
adds two new "ash" farming methods: ant and worm farms.
Ants farms slowly produce ores (with a weighted list), whereas worms can reproduce and produce fertilizer if given food.

This content works into other content for "ash" gameplay and adds more flavor for those who want to start a worm/ant farm.

## Proof of Testing

<details>
  
image
image
image

</details>

## Changelog

:cl:
add: added the worm/ant farm
balance: ash farming can be worm fertilized (same as regen core), but now only does one harvest
image: added the worm/ant farm, worm fertilizer
/:cl:


modular_skyrat/modules/ashwalkers/code/buildings/antfarm.dm
```diff
@@ -0,0 +1,64 @@
...
+	anchored = TRUE
+	/// If the farm is occupied by ants
+	var/has_ants = FALSE
+	/// the chance for the farm to get ants
+	var/ant_chance = 0
+	/// the list of ore-y stuff that ants can drag up from deep within their nest
+	var/list/ore_list = list(
+		/obj/item/stack/ore/iron = 20,
+		/obj/item/stack/ore/glass = 20,
+		/obj/item/stack/sheet/mineral/plasma = 14,
```
Nerev4r, line 16: Why?

jjpark-kb: the ores and stuff block their way from digging down, so they throw it out of their nest.

Nerev4r: As in, why is this something that Ashwalkers need?

jjpark-kb: its a slow method of gathering passive materials-- which means people can focus on other aspects of the game; also, there are a ton of ways to get materials now which are far easier than this.

lessthnthree (approved):

Nerev4r:
I don't think an entirely passive no-resources-needed source of plasma, silver, and gold that can be made out of deadass just sand probably just outside the airlock of the mining outpost en masse is a good idea or a necessary one. Especially with the stock market as an extant quandary, and especially with ArcMining looming over the horizon to do this properly. This should be taken back to the drawing board a bit and more properly integrated with the game like the bluespace miner is.

chemistrymain2:
Absolutely not, no.

jjpark-kb:
> I don't think an entirely passive no-resources-needed source of plasma, silver, and gold that can be made out of deadass just sand probably just outside the airlock of the mining outpost en masse is a good idea or a necessary one. Especially with the stock market as an extant quandary, and especially with ArcMining looming over the horizon to do this properly. This should be taken back to the drawing board a bit and more properly integrated with the game like the bluespace miner is.

They do require a resource to start, beyond the sand to build them. It's also weighted to give more iron or sand and way less silver or gold.

The stock market really made most, if not all resource generation useless. It is so easy to produce money and then buy resources-- and in far more quantities than the ant nest can do. This is a flavorful way for ash walkers to just get some ore resources passively, it can barely replace what normal mining can do for them; even if it did replace it, that wouldn't be the end of the world.

> Absolutely not, no.

Not a really helpful comment

modular_skyrat/modules/ashwalkers/code/buildings/antfarm.dm
```diff
@@ -0,0 +1,67 @@
...
+		/obj/item/stack/sheet/mineral/gold = 3,
+	)
+	// The cooldown between each worm "breeding"
+	COOLDOWN_DECLARE(ant_timer)
+
+/obj/structure/antfarm/Initialize(mapload)
+	. = ..()
+	if(!istype(get_turf(src), /turf/open/misc/asteroid/basalt))
+		qdel(src)
+		return
```
vinylspiders:
```suggestion
		return INITIALIZE_HINT_QDEL
```
When you want something to qdel in initialize you should return the hint and let the atom controller do it for you rather than just calling it yourself.

vinylspiders (changes requested):

modular_skyrat/master_files/code/game/objects/items/stacks/sheets/sheet_types.dm
```diff
@@ -129,3 +130,13 @@ GLOBAL_LIST_INIT(skyrat_snow_recipes, list(
 /obj/item/stack/sheet/mineral/snow/get_main_recipes()
 	. = ..()
 	. += GLOB.skyrat_snow_recipes
+
+// Sand
+
+GLOBAL_LIST_INIT(skyrat_sand_recipes, list(
+	new/datum/stack_recipe("ant farm", /obj/structure/antfarm, 20, time = 2 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS),
```
vinylspiders, line 137: As far as balance goes, it's probably fine--weight can be tweaked later if people think it's too OP. 

Only thing I would say is that you should probably add a range check for other anthills so you can't stack them infinitely in a certain radius. Anthills are like icebergs and there would be tunnels extending past where the entrance is.

I'm not sure that stack_recipes support such things (I know the crafting system does though). You might have to just add an `locate(/obj/structure/antfarm) in_view(radius, src)` check in Initialize() and refund the materials manually if it finds another anthill nearby.

vinylspiders (changes requested):

Nerev4r:
Two is a pretty comically low range for a low-grade bluespace miner.

jjpark-kb:
> Two is a pretty comically low range for a low-grade bluespace miner.

When it had no range and now one cannot be placed in a 5x5 square near another one-- I would say the issue was addressed.

vinylspiders (approved):

PR #24362, opened 2023-10-16
Sindano Inhand Sprite that were previously missing

projectkepler-ru:
## About The Pull Request
sindano inhand sprite that were previously missing

without this it has a jarring yellow dot when you look to the north because we're using the c20r as a base inhand icon 

## Proof of Testing
image
image
image
image
image
image

<details>
  Inhand icon

</details>

## Changelog

:cl:
image: gave the sindano  proper inhand icon for both variant
/:cl:


projectkepler-ru:
Worried it may not look very good rn, waiting on feedback

modular_skyrat/modules/modular_weapons/code/company_and_or_faction_based/carwo_defense_systems/submachinegun.dm
```diff
@@ -7,7 +7,9 @@
 	icon = 'modular_skyrat/modules/modular_weapons/icons/obj/company_and_or_faction_based/carwo_defense_systems/guns32x.dmi'
 	icon_state = "sindano"
 
-	inhand_icon_state = "c20r"
+	lefthand_file = 'modular_skyrat/modules/modular_weapons/icons/mob/company_and_or_faction_based/carwo_defense_systems/guns_lefthand.dmi'
+	righthand_file = 'modular_skyrat/modules/modular_weapons/icons/mob/company_and_or_faction_based/carwo_defense_systems/guns_righthand.dmi'
+	inhand_icon_state = "cindano"
```
vinylspiders: How come the `icon_state`s are spelled with a 'c'?

vinylspiders (changes requested):
Thanks for doing this--love to see inhand sprites. The colors look good. But there are a couple things which I think you picked up on in your worries:

1) There is a gap here and here between the mob and the gun with the East/West sprites, and it's also pushed up too far into the mob so it looks like part of the gun is just sawed straight off. The borders of the hidden bits need to conform to the mob's shape.

image

image

2) North/South is too lateral, it doesn't quite look right with how these guns are meant to be held.

image

3) The back of it is a bit long and skinny, it kinda looks like it's floating on certain backdrops.

image


The overall size and the way it's being held makes it read as more like a shotgun I think?

I think if you just sort of recolored the c20 sprite a bit and got rid of the yellow it'd look less 'off' and more like a SMG. 

dreamseeker_jrs7pKlLvc

That's what I would do, anyway.

modular_skyrat/modules/modular_weapons/code/company_and_or_faction_based/carwo_defense_systems/submachinegun.dm
projectkepler-ru: I'll get change that, must've been my brain damage

projectkepler-ru:
@vinylspiders  Willl make a patch now and see whats wrong
it looked fine on my end- might just be im using a much more squished monitor

projectkepler-ru:
The other reason why I wanted this inhand sprites was because the gun needs two hand to fire vs the c20r one handedness so recoloring the c20r wont do it for me

Paxilmaniac:
> The other reason why I wanted this inhand sprites was because the gun needs two hand to fire vs the c20r one handedness so recoloring the c20r wont do it for me

This is not correct, the reason i made it use the one hand c20 sprite to begin with was to show that it did not need two hands.

projectkepler-ru:
> > The other reason why I wanted this inhand sprites was because the gun needs two hand to fire vs the c20r one handedness so recoloring the c20r wont do it for me
> 
> This is not correct, the reason i made it use the one hand c20 sprite to begin with was to show that it did not need two hands.

weird? let me check again

vinylspiders (approved):

vinylspiders:
Pinged @GoldenAlpharex to take a final look, looks good to me though

GoldenAlpharex (approved):
Good work on this, and thank you @vinylspiders for the good sprite reviews too, very thorough!

PR #24392, opened 2023-10-17
[MODULAR] Fixes a nif related runtime with update_theme

vinylspiders:
## About The Pull Request

Race condition. In the parent `/datum/nifsoft/New()` it calls `update_theme()` which expects the `linked_soulcatcher` var to be set, but it doesn't get set until the end of `/datum/nifsoft/soulcatcher/New()`. So it runtimes.

This fixes that.

Also fixes a bonus runtime:

qALCzRpWAE

Less runtiming is good

## Proof of Testing

<details>
  
cgpHWBfjqy

</details>

## Changelog

Nothing really significantly player facing I don't think?


softcerv (approved):
This looks fine from a code perspective,
Thank you for catching this.

PR #24430, opened 2023-10-19
[MIRROR] A comprehensive refactor / cleanup of `bullet_hit` and `on_hit` to cut out a single bad species / mob proc [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79024
--------------------
## About The Pull Request

- Refactored `bullet_act`. Adds `should_call_parent` and refactors associated children to support that. 
   - Fixes silicons sparking off when hit by disabler fire. 
   - Desnowflakes firing range target integrity and cleans up its bullet-hole code a bit. 
   - Cleans up changeling tentacle code a fair bit and fixes it not taking off throw mode if you fail to catch something.
   - The Sleeping Carp deflection is now signalized 
   - Nightmare projectile dodging is now signalized and sourced from the Nightmare's brain rather than species
   - Refactored how cardboard cutouts get knocked over to be less snowflaked / use integrity
- Also adds projectile `on_hit` `should_call_parent` and cleans up a bit of that, particularly their arguments. 
   - On hit arguments were passed wrong this entire time, it's a good thing nothing relied on that. 

## Why It's Good For The Game

This is cringe. 

https://github.com/tgstation/tgstation/blob/1863eb2cd82e7cee4fdfff37b42d3fd0c7edd797/code/modules/mob/living/carbon/human/_species.dm#L1430-L1442

Bullets should overall act more consistent across mob types and objects. 

## Changelog

:cl: Melbert
fix: Silicons don't spark when shot by disablers
fix: Changelings who fail to catch something with a tencacle will have throw mode disabled automatically
fix: Fixes occasions where you can reflect with Sleeping Carp when you shouldn't be able to
fix: Fixes some projectiles causing like 20x less eye blur than they should be
refactor: Refactored bullet-mob interactions
refactor: Nightmare "shadow dodge" projectile ability is now sourced from their brain
/:cl:


vinylspiders (dismissed):

vinylspiders:
Actually this one might need a looking at because we do have modular overrides of these procs that might need some updating

vinylspiders (approved):

PR #24449, opened 2023-10-20
arrows can be converted

jjpark-kb:
## About The Pull Request
adds the ability to transform arrows to different types by attacking them with the conversion item (bronze tile to bronze arrow, bone to bone arrow, and sinew to ash arrow).

I also removed the forged arrow subtype-- after some well needed qol PRs, they became unused and not necessary.

THIS DOES NOT REMOVE THE RECIPES-- there may be cases where you need to be able to craft them so I kept the recipes.

Rather than being stuck in the crafting menu, more item interactions like this keeps "immersion."

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: you can change arrows into different types by hitting them with the conversion item (bronze tile for bronze, bone for bone, and sinew for ash)
del: removes the forged arrow subtype
/:cl:

### DIFF
--- a/modular_skyrat/modules/reagent_forging/code/forge_weapons.dm
+++ b/modular_skyrat/modules/reagent_forging/code/forge_weapons.dm
@@ -257,9 +257,11 @@
 	if(!stack_item.use(1))
 		return
 
-	new spawned_item(get_turf(src))
+	var/obj/item/ammo_casing/arrow/converted_arrow = new spawned_item(get_turf(src))
+	transfer_fingerprints_to(converted_arrow)
+	remove_item_from_storage(user)
+	user.put_in_hands(converted_arrow)
 	qdel(src)
-	return
 
 #define INCREASE_BLOCK_CHANGE 2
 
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/reagent_forging/code/forge_weapons.dm
```diff
@@ -239,22 +239,27 @@
...
+	var/obj/item/stack/stack_item = attacking_item
+	if(!stack_item.use(1))
 		return
-	return ..()
 
-/obj/projectile/bullet/arrow/forged
-	projectile_type = /obj/item/ammo_casing/arrow/forged
+	new spawned_item(get_turf(src))
+	qdel(src)
+	return
```
vinylspiders:
```suggestion
	var/obj/item/ammo_casing/arrow/converted_arrow = new spawned_item(get_turf(src))
	transfer_fingerprints_to(converted_arrow)
	remove_item_from_storage(user)
	user.put_in_hands(converted_arrow)
	qdel(src)
```
Just some small details/QoL. These procs are safer to use when doing this sort of thing. If you don't want the finished product to be put into the user's hands feel free to exclude that bit.

vinylspiders (changes requested):
Looks good to me otherwise!

vinylspiders (approved):

PR #24474, opened 2023-10-21
[MODULAR] Taur Beautification: What If They Were Cool Looking Edition (& Some Extra Stuff)

Nerev4r:
## About The Pull Request
Hi.
I saw the current Drider sprite the other day and nearly popped a blood vessel. So in search of a solution, I have ported some taur options (some as additions, some as replacements) from our distant cousins Virgo and Chompstation.

image
Starting off, the new Drider sprite. This has three (count 'em) color zones; the body, the stripes on the legs, and the stripe on the back. I did my best to replicate some kind of weirdly patterned black widow here.

image
This here is the new Naga sprite. Like the Drider one, this replaces the old one, but you will find that the shading is a lot more subtle.

image
Striped variant.

image
Rattle variant.
No, at this time, I cannot make a striped rattle variant.

For comparison's sake, 
image
Here's the old naga sprite.

image
And the old drider sprite. It's not much better.

image
Lastly, this particularly wet-looking tentacle sprite. This one is an alt, the old one will still be there.

I love it when sprites look nice.

## Proof of Testing
<details>
  Check right above here.
</details>

## Changelog
:cl:
image: The Naga taur option has been replaced with a new sprite, with two Naga variants.
image: A new Drider option, Tarantula, has been added; as well as a new Tentacle option.
image: A new plug tail option, as well as a new set of wings.
/:cl:


Nerev4r:
On behest of one of my allies, I have added an alternate plug tail. 
image

W4rd3nn:
> On behest of one of my allies, I have added an alternate plug tail. 

why's she got her hooha out

Nerev4r:
image
Okay, we're done here. Considering this ready, no further additions will be done.
One new wing option.

Namwty:
FuXIIKrWYAAY-88
Mfw I have to tell two people that characters they made not even week ago have to be scrapped or revamped.


My concern is that people who use OLD variants might get stomped by this change. There are too many variables that could happen, ya got **size scale** and all sorts of **clothes/bodyparts/markings** that may have looked good on old sprite, but might just not look good with new one or even something simple as people who used them like the old one regardless how improved the new one looks. So not sure why replacing exactly instead of adding as extra... **like we didn't remove old akula cause folk I guess used them AND STILL USE, regardless of how worse they **objectively** look compared to new akulas...**

Like for example concern of new drider model looking odd with clothes that HAVE legs/pants sprite. Old drider had front "legs", while new one doesn't, so it will just look even more odd, and the "no pants" options like sweater... well those are just limited.


EDIT: gets_cropped_on_taurs isn't something that all taurs have ( old drider doesn't for example ) and objectively isn't even good... half of clothes become unusable or simply bland, example Kimono/Yukata that main appeal is the lower part looking like robes. I mean again, it's like whatever nor I care to start a debate on it, it's easier for them to just delete the char and make something else.

Nerev4r:
> FuXIIKrWYAAY-88 Mfw I have to tell two people that characters they made not even week ago have to be scrapped or revamped.
> 
> My concern is that people who use OLD variants might get stomped by this change. There are too many variables that could happen, ya got **size scale** and all sorts of **clothes/bodyparts/markings** that may have looked good on old sprite, but might just not look good with new one or even something simple as people who used them like the old one regardless how improved the new one looks. So not sure why replacing exactly instead of adding as extra... **like we didn't remove old akula cause folk I guess used them AND STILL USE, regardless of how worse they objectively look compared to new akulas...**
> 
> Like for example concern of new drider model looking odd with clothes that HAVE legs/pants sprite. Old drider had front "legs", while new one doesn't, so it will just look even more odd, and the "no pants" options like sweater... well those are just limited.

Should be fine given most things have gets_cropped_on_taurs. If it's truly necessary I'll add the old drider sprite as an alt.

Nerev4r:
Upon further introspection, I have instead added the new Drider sprite as 'Tarantula' while retaining the old one.

GoldenAlpharex (approved):
Eh, sure, why not. The wings feel a bit off to me, but they're not really out of line for the rest of our wings, so I don't really have much against it.

PR #24512, opened 2023-10-22
removes reagent requirements from medicells, makes the oppressive force cell a slime extract thing

Hatterhat:
## About The Pull Request
- removes reagent requirements from medicells
- oppressive force relocator cell thing is now made with a bluespace slime extract like the body teleporter

i just did this because i saw that there was a pr that was failing unit tests b/c of this

## Changelog

:cl:
balance: Medicells no longer require reagents to print because reagent costs for lathe designs got depreciated. Rejoice.
balance: Oppressive force relocation cell is made like the body teleporter cell (with a bluespace slime extract or something).
/:cl:

### DIFF
--- a/modular_skyrat/modules/cellguns/code/medigun_research.dm
+++ b/modular_skyrat/modules/cellguns/code/medigun_research.dm
@@ -26,7 +26,6 @@
 	desc = "Gives cell-loaded mediguns improved brute damage healing functionality."
 	id = "brute2medicell"
 	materials = list(/datum/material/plastic = SHEET_MATERIAL_AMOUNT, /datum/material/glass = SHEET_MATERIAL_AMOUNT, /datum/material/plasma = HALF_SHEET_MATERIAL_AMOUNT)
-	reagents_list = list(/datum/reagent/medicine/c2/libital = 10)
 	build_path = /obj/item/weaponcell/medical/brute/tier_2
 
 /datum/design/medicell/burn2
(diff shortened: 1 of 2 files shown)


lessthnthree (approved):

lessthnthree:
Straight up removing the requirement with no balance has to be looked at.

vinylspiders:
I feel like these should either be moved into the pill press/chem factory options or put into the crafting menu with the reagent requirements intact.

The whole point of these is the reagents within, and it doesn't make any sense to remove them. Kinda cheapens the point of chemistry.

chemistrymain2:
> I feel like these should either be moved into the pill press/chem factory options or put into the crafting menu with the reagent requirements intact.
> 
> The whole point of these is the reagents within, and it doesn't make any sense to remove them. Kinda cheapens the point of chemistry.

TG has a PR merged  that removes the code for reagent support in protolathes.

vinylspiders:
>TG has a PR merged that removes the code for reagent support in protolathes.

I'm aware. Not a fan of it.

Or as a compromise I guess we could just give them a little extra -something- to build. Just feels too easy now.

HoodMilk:
Just make the lathes print an empty medicell that you craft into the other cells with chemicals (would also have the benefit of actually showing you what chems you need too)

lessthnthree:
That could work, an empty cell and you put libital in it to make the brute cell, etc.

lessthnthree:
> Just make the lathes print an empty medicell that you craft into the other cells with chemicals (would also have the benefit of actually showing you what chems you need too)

Let's make the medicell a subtype of hypovial, then it can be printed in the chemmaster with the chemicals inside.

lessthnthree:
Making a cell using reagents can be done in a different PR. To keep things moving I'm okay with removing the reagents, however will be raising the material cost of the cell a bit.

PR #24520, opened 2023-10-23
nerfs the new engineering machine's prybar, toolspeed also: removes the RCD from it

RatFromTheJungle:
## About The Pull Request

the tools reference being slower, in exchange for being three in ones, or roundstart jaws (looking at you prybar), but they weren't actually slower at all

## Proof of Testing

<details>
  
its just a numbers change

</details>

## Changelog

:cl:
balance: made the colony fabricator's tools actually slower, removed the RCD per request
/:cl:

### DIFF
--- a/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
+++ b/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
@@ -26,7 +26,7 @@
 	hitsound = 'sound/items/drill_hit.ogg'
 	usesound = 'sound/items/drill_use.ogg'
 	w_class = WEIGHT_CLASS_SMALL
-	toolspeed = 1
+	toolspeed = 1.2
 	random_color = FALSE
 	greyscale_config = null
 	greyscale_config_belt = null
@@ -95,7 +95,7 @@
 		survive <b>forcing doors open</b>."
 	icon = 'modular_skyrat/modules/colony_fabricator/icons/tools.dmi'
 	icon_state = "prybar"
-	toolspeed = 1
+	toolspeed = 1.3
 	force_opens = TRUE
 	custom_materials = list(
 		/datum/material/iron = SHEET_MATERIAL_AMOUNT * 1.75,
--- a/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
+++ b/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
@@ -26,7 +26,7 @@
 	hitsound = 'sound/items/drill_hit.ogg'
 	usesound = 'sound/items/drill_use.ogg'
 	w_class = WEIGHT_CLASS_SMALL
-	toolspeed = 1.2
+	toolspeed = 1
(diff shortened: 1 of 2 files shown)


Paxilmaniac:
> the tools reference being slower, in exchange for being three in ones

Erm, they only say they're slower than power tools.

Hatterhat:
> > the tools reference being slower, in exchange for being three in ones
> 
> Erm, they only say they're slower than power tools.

if you want to nitpick you could probably argue that the desc should be more clear about this but i think the mild tedium of having to juggle a radial for the combi driver is enough

modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
```diff
@@ -26,7 +26,7 @@
 	hitsound = 'sound/items/drill_hit.ogg'
 	usesound = 'sound/items/drill_use.ogg'
 	w_class = WEIGHT_CLASS_SMALL
-	toolspeed = 1
+	toolspeed = 1.2
```
ghost: Just to make sure, this one's description explicitly says "Compared to other power drills these tend to be \
		<b>not much quicker than unpowered tools</b>." So it SHOULD still be quicker than unpowered ones.

Paxilmaniac: Just leave it at 1 toolspeed. You don't even gain much from it being three in one compared to power tools, considering that you still need to carry a separate crowbar.

RatFromTheJungle: if they're 1 toolspeed, they should be normal sized like the other powertools; or powertools should be made small sized, because essentially-power-tools but only one is normalsized (crowbar) is a bit off pilled with the standard currently set, when you can do stuff like (image), so I decided it'd be better to just slow them down a bit so they aren't always questionably a storage upgrade to powertools for non-engineers (minus the prybar)
image

Paxilmaniac: Make the driver normal sized in that case, I thought I did already. The convenience of multiple in one NOTHING if you're staring at action bars longer than just having the tools split.

chemistrymain2: > if they're 1 toolspeed, they should be normal sized like the other powertools; or powertools should be made small sized, because essentially-power-tools but only one is normalsized (crowbar) is a bit off pilled with the standard currently set, when you can do stuff like (image), so I decided it'd be better to just slow them down a bit so they aren't always questionably a storage upgrade to powertools for non-engineers (minus the prybar) image

Powertools should probably be made small sized, yes.

YakumoChen: They aren't because it would completely invalidate the tool belt as a belt, you either choose to have a belt and faster tools, a box with slower tools, or half your backpack space to have fast tools

lessthnthree (approved):

PR #24543, opened 2023-10-24
issue fixes and qol for ant, worm, and ash farming

jjpark-kb:
## About The Pull Request
you can now feed worm and ant farms from a plant bag
plant bags can now directly harvest from ash farms
worms will reproduce when having 2 (was bugged and required 3 due to logic)
fertilizer will not be used if it doesn't upgrade something in the plant

there are a few things that didn't make sense-- worms can't reproduce when having 2? it was a logic error: using > instead of >=. there are also a few qol things that should have been added way earlier

## Proof of Testing

<details>
  
image
image

</details>

## Changelog

:cl:
qol: you can now directly feed worm/ant farms from plant bags
qol: you can now directly harvest ash farms with plant bags
fix: worm farms can reproduce with 2 now (instead of the 3 due to a logic error)
fix: fixed fertilizer being deleted even if it didn't upgrade an ash farm
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
@@ -116,6 +116,7 @@
 		balloon_alert(user, "feeding the worms")
 		for(var/obj/item/food/selected_food in attacking_item.contents)
 			if(!do_after(user, 1 SECONDS, src))
+				in_use = FALSE
 				return
 
 			qdel(selected_food)
--- a/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
@@ -32,7 +32,7 @@
 
 	COOLDOWN_START(src, worm_timer, 1 MINUTES)
 
-	if(current_worm > 2 && current_worm < max_worm)
+	if(current_worm >= 2 && current_worm < max_worm)
 		current_worm++
 
 	if(current_food > 0 && current_worm > 1)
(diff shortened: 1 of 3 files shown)


xXPawnStarrXx:
Worms were intended to be made for tribals, right? Would it be fair to maybe make them only craftable by tribals or do you intend for crew to use them as they have been?

jjpark-kb:
> Worms were intended to be made for tribals, right? Would it be fair to maybe make them only craftable by tribals or do you intend for crew to use them as they have been?

I try for most of my content to be open to whomever wants to use it— worms were just intended for those who use ash farming, of which I know some station people do ash farming

vinylspiders (approved):

modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
```diff
@@ -107,6 +107,23 @@
...
+	if(istype(attacking_item, /obj/item/storage/bag/plants))
+		if(in_use)
+			balloon_alert(user, "currently in use")
+			return
+		in_use = TRUE
+
+		balloon_alert(user, "feeding the worms")
+		for(var/obj/item/food/selected_food in attacking_item.contents)
+			if(!do_after(user, 1 SECONDS, src))
+				return
```
vinylspiders:
```suggestion
			if(!do_after(user, 1 SECONDS, src))
				in_use = FALSE
				return
```
Think you'll want to set this bool here, or else it might get stuck in the `in_use` state.

vinylspiders (changes requested):

vinylspiders (approved):

PR #24548, opened 2023-10-24
Marks the flatpacked Fabricator and RTG imports as 'restricted'

ghost:
## About The Pull Request
It really makes no sense for the crowbar to be restricted, but not the thing that can print infinite crowbars (plus a lot more)...
And the improperly shielded RTG just doesn't make sense to be all-access either. Giving untrained and unauthorized crew access to radiation spewers?

If the code was in place, I would prefer to make them restricted to only Engineering so that other command staff don't just buy them to kit their department with budget-JoL and tools, but import code runs off guncargo code which only had guns in mind rather than the broad spectrum of departmental accesses...

Really it just doesn't make sense that the prybar is restricted but not the thing that prints them. Seems like an oversight.
And the RTG just doesn't make sense to not be restricted because, again, it's not shielded. Radioactive contaminants shouldn't be all-access.

## Proof of Testing
It's really that shrimple

image

## Changelog

:cl:
balance: the flatpacked Fabricator and RTG are now restricted-access imports
/:cl:

### DIFF
--- a/modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
+++ b/modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
@@ -29,6 +29,7 @@
 /datum/armament_entry/company_import/akh_frontier/deployables_fab/rapid_construction_fabricator
 	item_type = /obj/item/flatpacked_machine
 	cost = CARGO_CRATE_VALUE * 6
+	restricted = TRUE
 
 // Various smaller appliances than the deployable machines below
 
@@ -80,3 +81,4 @@
 
 /datum/armament_entry/company_import/akh_frontier/deployables/rtg
 	item_type = /obj/item/flatpacked_machine/rtg
+	restricted = TRUE


RatFromTheJungle:
image

HoodMilk:
To be fair anyone with lathe access can print out uranium and start flooring the station with it. But the fabricator board makes sense.

SpaceLoveSs13:
The entire point of the PR was to allow anyone to start a functional system from scratch anywhere when needed in dire situations. Unnecessarily restricting the access of the fundamental item in the block is not required.

SomeRandomOwl:
> The entire point of the PR was to allow anyone to start a functional system from scratch anywhere when needed in dire situations. Unnecessarily restricting the access of the fundamental item in the block is not required.

If something the lathe can make is restricted from normal purchase, then the lathe itself should also be restricted so people cant work around the restriction

ghost:
> If something the lathe can make is restricted from normal purchase, then the lathe itself should also be restricted so people cant work around the restriction

This pretty much summed up the reasoning behind the PR yeah thanks 4 being better with words

lessthnthree (approved):

PR #24555, opened 2023-10-24
Pollution doesn't spawn when the subsystem is offline + only player-crafted bonfires produce smoke

GoldenAlpharex:
## About The Pull Request
It annoyed me that pollution would keep spawning even after the subsystem was turned off, so it won't do that anymore.

I also made it so only player-made bonfires would produce smoke, because otherwise it's a pain every time we have custom maps, or every time /tg/ adds a new thing with a bonfire in an enclosed area, which ends up smoking entire rooms/z-levels for zero gain.

Smoke will now only happen when it's meant to happen.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl: GoldenAlpharex
fix: Pollution (smoke et al.) will no longer be added to turfs when the subsystem is offline. Basically, you can now turn off pollution properly.
qol: Only player-made bonfires will produce smoke.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -326,7 +326,7 @@ GLOBAL_LIST_INIT(wood_recipes, list ( \
 	new/datum/stack_recipe("loom", /obj/structure/loom, 10, time = 1.5 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS), \
 	new/datum/stack_recipe("mortar", /obj/item/reagent_containers/cup/mortar, 3, check_density = FALSE, category = CAT_CHEMISTRY), \
 	new/datum/stack_recipe("firebrand", /obj/item/match/firebrand, 2, time = 10 SECONDS, check_density = FALSE, category = CAT_TOOLS), \
-	new/datum/stack_recipe("bonfire", /obj/structure/bonfire, 10, time = 6 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS), \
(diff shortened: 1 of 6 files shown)


code/game/objects/items/stacks/sheets/sheet_types.dm
```diff
@@ -326,7 +326,7 @@ GLOBAL_LIST_INIT(wood_recipes, list ( \
 	new/datum/stack_recipe("loom", /obj/structure/loom, 10, time = 1.5 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS), \
 	new/datum/stack_recipe("mortar", /obj/item/reagent_containers/cup/mortar, 3, check_density = FALSE, category = CAT_CHEMISTRY), \
 	new/datum/stack_recipe("firebrand", /obj/item/match/firebrand, 2, time = 10 SECONDS, check_density = FALSE, category = CAT_TOOLS), \
-	new/datum/stack_recipe("bonfire", /obj/structure/bonfire, 10, time = 6 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS), \
+	new/datum/stack_recipe("bonfire", /obj/structure/bonfire/player_made, 10, time = 6 SECONDS, one_per_turf = TRUE, on_solid_ground = TRUE, category = CAT_TOOLS), /* SKYRAT EDIT - Pollution - ORIGINAL: /obj/structure/bonfire */ \
```
vinylspiders, line 329: There really should be an object flag that gets sets in `/obj/item/stack/proc/make_item` to mark player-crafted atoms as such. Something you could add upstream if you wanted.

vinylspiders (approved):

PR #24597, opened 2023-10-26
laser carbines for everyone* - disabler/laser carbines now for sale by microstar

Hatterhat:
## About The Pull Request
- adds disabler SMGs and laser carbines to microstar at the same price points of their semi-auto variants
- uncomments the laser gun/egun crates

- disabler SMGs and laser carbines makes them more available without having to jump through hoops (buying a crate and cracking it or something idk)
- at first i thought that having laser guns and eguns not available as dep-orderable was silly when carbines and SMGs were, so i uncommented them. then i thought that maybe guncargo existing means that this doesn't need to happen. then i remembered that you can buy MCRs and the sindano SMGs through it so i thought "yeah let's just go back on that"

## Changelog

:cl:
balance: Disabler SMGs and laser carbines (the full-auto sidegrades to the disabler and laser gun, respectively) are now available for purchase through Microstar.
balance: Cargo now stocks laser gun and energy gun crates again, which means that they can be ordered through security's dep-order console.
/:cl:

### DIFF
--- a/modular_skyrat/modules/company_imports/code/armament_datums/microstar_energy.dm
+++ b/modular_skyrat/modules/company_imports/code/armament_datums/microstar_energy.dm
@@ -11,9 +11,9 @@
 	item_type = /obj/item/gun/energy/disabler
 	cost = PAYCHECK_CREW * 5
 
-/datum/armament_entry/company_import/microstar/basic_energy_long_weapons/disabler_smg
+/datum/armament_entry/company_import/microstar/basic_energy_weapons/disabler_smg
 	item_type = /obj/item/gun/energy/disabler/smg
-	cost = PAYCHECK_CREW * 5
+	cost = PAYCHECK_CREW * 7 // slightly more expensive due to ease of use/full auto
 
 /datum/armament_entry/company_import/microstar/basic_energy_weapons/mini_egun
 	item_type = /obj/item/gun/energy/e_gun/mini
@@ -32,7 +32,7 @@
 
 /datum/armament_entry/company_import/microstar/basic_energy_long_weapons/laser_carbine
 	item_type = /obj/item/gun/energy/laser/carbine
-	cost = PAYCHECK_CREW * 5
+	cost = PAYCHECK_CREW * 7 // slightly more expensive due to ease of use/full auto
 
 /datum/armament_entry/company_import/microstar/basic_energy_long_weapons/egun
 	item_type = /obj/item/gun/energy/e_gun
--- a/modular_skyrat/modules/company_imports/code/armament_datums/microstar_energy.dm
+++ b/modular_skyrat/modules/company_imports/code/armament_datums/microstar_energy.dm
@@ -11,6 +11,10 @@
 	item_type = /obj/item/gun/energy/disabler
 	cost = PAYCHECK_CREW * 5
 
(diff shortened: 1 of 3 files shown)


Stalkeros2:
add tasers too pls? :pleading_face:

GoldenAlpharex (changes requested):
What good does this do to the game? I'm not really convinced by the PR body.

Plum-but-gayer:
I think adding carbines/smgs to microstar at like a slightly increased cost (+50/100 creds) might work a lil better. Historically I've been a big fan of bringing laser crates back but now that there's no stock limit of how much we can order from corporate orders there's do difference in the HoS buying 3 allstars with corpo vs buying 3 allstars with crates.

imo just add the two new guns with a small price bump, leave the crates. Keeps our gun buying mechanics consistent.

Iamgoofball:
@GoldenAlpharex its guncargo, the justification for guncargo is the same as it's always been

chemistrymain2:
> What good does this do to the game? I'm not really convinced by the PR body.

These are guns conventionally available in crates and instead they should be available through guncargo for the same reason we allow high-power ballistics to be purchasable through guncargo.

The laser carbines and disabler smgs are simply tg guns instead of our custom ballistics, but they're balanced (if not weaker than) our current available guns.

Hatterhat:
> What good does this do to the game? I'm not really convinced by the PR body.

As it stands, the laser carbines and disabler SMGs are available through crates - and nothing else. This is in contrast to the energy gun and laser gun, which are mainly available through the imports system (guncargo).

The uncommenting of the crates for the laser guns and eguns, and the addition of the disabler SMG and laser carbine to the imports system, unifies the two, so that they are both available through both sources; I'm open to killing off the crates and including the carbine/SMG crates in that, and leaving them in the imports.

As above, they're intended to have the same overall damage output as their semi-auto counterparts; double battery, but half damage. I suppose I could also bump up the price as well, to reflect the increased ease of use autofire grants them.

Plum-but-gayer:
So what's the word on this? Is it an issue of bringing egun/laser crates back or is there something else that's an issue?

projectkepler-ru:
> So what's the word on this? Is it an issue of bringing egun/laser crates back or is there something else that's an issue?

golden is busy me think

modular_skyrat/modules/company_imports/code/armament_datums/microstar_energy.dm
```diff
@@ -11,6 +11,10 @@
 	item_type = /obj/item/gun/energy/disabler
 	cost = PAYCHECK_CREW * 5
 
+/datum/armament_entry/company_import/microstar/basic_energy_long_weapons/disabler_smg
+	item_type = /obj/item/gun/energy/disabler/smg
+	cost = PAYCHECK_CREW * 5
```
GoldenAlpharex: I feel like they could probably be just a tad bit more expensive than the non-automatic variant, as I suspect they're likely stronger than said non-automatic variants. Maybe just a `PAYCHECK_CREW * 6` could be enough, for now at least.

GoldenAlpharex (changes requested):

GoldenAlpharex (approved):
Looks good, let's try it out!

PR #24651, opened 2023-10-29
Soupier soups

xXPawnStarrXx:
## About The Pull Request
I have made a modular addition that replaces the output of almost all soups in the game, removing the dozens of excess reagents, meaning that you only get pure reagents when making soups. Some have been adjusted due to their difficulty to produce between 30-60u from a single batch. (Chicken curry, for example)

This is modular because after discussion with /TG/ they want it to remain that way on their end, for some strange reason.

Soups were in a strange place, because a lot of our players, especially more OCD ones, (like myself) found it incredibly disturbing to see soups that had a large number of unknown reagents or get slapped in the face by a 30 line box of garbage ingredients.

## Proof of Testing

<details>
From this after cooking:

image

To something like this:
image
image

  
</details>

## Changelog

:cl:
qol: made soup actually produce just soup, instead of an ungodly reagent slurry.
code: changed some code around soup, making it produce cleaner output.
/:cl:

### DIFF
--- a/modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
+++ b/modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
@@ -1,0 +1,414 @@
+/datum/reagent/consumable/nutriment/soup
+	nutriment_factor = 20 // Upped to counter the loss of random ingredients.
+	//Some soups will create between 30-60u to counter the loss of additional reagents.
+
+/datum/chemical_reaction/food/soup
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+// Meatball Soup
+/datum/chemical_reaction/food/soup/meatballsoup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/meatball_soup = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/vegetable_soup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/vegetable_soup = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/nettlesoup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/nettle = 30
+	)
+	ingredient_reagent_multiplier = 0.0 // No more acid!
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/wingfangchu
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/wingfangchu = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/hotchili
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/hotchili = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/coldchili
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/coldchili = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/clownchili
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/clownchili = 60
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/chili_sin_carne
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/chili_sin_carne = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/tomatosoup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/tomato = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/eyeballsoup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/eyeball = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/misosoup
+	results = list(
+		/datum/reagent/consumable/nutriment/soup/miso = 30
+	)
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
+
+/datum/chemical_reaction/food/soup/bloodsoup
+	results = list(
+		// Blood tomatos will give us like 30u blood, so just add in the 10 from the recipe
+		/datum/reagent/blood = 10
+	)
+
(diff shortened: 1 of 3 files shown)


chemistrymain2:
I honestly prefer the nonpure reagents, both because it gave you more soup per pot as a result of dilution and because it also meant that if the chef made you like poison plants to put in your soup (or they were cross contaminated somehow) you'd fucking die.

xXPawnStarrXx:
> I honestly prefer the nonpure reagents, both because it gave you more soup per pot as a result of dilution and because it also meant that if the chef made you like poison plants to put in your soup (or they were cross contaminated somehow) you'd fucking die.

I'm already working on finding out how to get the soups to produce more based on quality of reagents, I just received several people who like the idea of removing it so I wanted to get this out first. 
With the new allergy trait and the already existing medical one, If a chef cooks only soup, it just straight up removes the ability for you to use the kitchen and I wanted to remove that because most of our players are here for roleplay, it's not fun to roleplay being in ghost chat for anywhere up to 30 minutes(depending on medical's skill and station crewing) because you got hungry.

Iamgoofball:
this is stupid because it means that if i make poison tomatoes, the poison is not present in the soup

if your ingredients have garbage in them it should be represented in the soup, this is the "strange reason" the OP is referring to and doesn't say explicitly because it's reducing the capability of the roleplay sandbox

xXPawnStarrXx:
> this is stupid because it means that if i make poison tomatoes, the poison is not present in the soup
> 
> if your ingredients have garbage in them it should be represented in the soup, this is the "strange reason" the OP is referring to and doesn't say explicitly because it's reducing the capability of the roleplay sandbox

That is absolutely fine; On /TG/, unless policy changed in the last few months, you were not allowed to commit widespread poisonings. Even with an OPFOR they must be targeted or small scale, so the only way this is happening is sheer incompetance or maliciousness in skirting the rules using that as a cover. It's also why I made this modular, so it only affects Skyrat. If you want to poison people with food, then you can on /TG/.

EDIT; If you still want to poison people, there's nothing stopping you literally adding the reagents to the bowl alongside the soup. it's literally a reagent container and would display the same way. Still allowing smallscale poisoning.

chemistrymain2:
> > this is stupid because it means that if i make poison tomatoes, the poison is not present in the soup
> > if your ingredients have garbage in them it should be represented in the soup, this is the "strange reason" the OP is referring to and doesn't say explicitly because it's reducing the capability of the roleplay sandbox
> 
> That is absolutely fine; On /TG/, unless policy changed in the last few months, you were not allowed to commit widespread poisonings. Even with an OPFOR they must be targeted or small scale, so the only way this is happening is sheer incompetance or maliciousness in skirting the rules using that as a cover. It's also why I made this modular, so it only affects Skyrat. If you want to poison people with food, then you can on /TG/.
> 
> EDIT; If you still want to poison people, there's nothing stopping you literally adding the reagents to the bowl alongside the soup. it's literally a reagent container and would display the same way. Still allowing smallscale poisoning.

Accidental incompetence killing the crew is fine, and mass poisonings are acceptable: Policy seems to have changed in that regard, because there's been more than a few times a chef or someone else has poisoned all of the chef's food and got everyone who ate it sick _and it was ruled acceptable._

So either admins have fucked up repeatedly, or it's not something that's policy blacklisted by OPFOR or antagonism.

xXPawnStarrXx:
> So either admins have fucked up repeatedly, or it's not something that's policy blacklisted by OPFOR or antagonism.

Back when I was staff; There was a very strict rule. No permitting mass death/casualties. (No spreading viruses, no delams, no random bombings, no mass murdering, no mass poisoning.) So unless they've decided to waive that in the last few months, I feel it's just not being enforced. It's not very good for roleplay being wordlessly murdered or poisoned.

softcerv:
Is this going to require new recipes to be manually added in when new soups are added upstream?

modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
softcerv, line 1: I feel like you could use inheritance to trim down a lot of this code.

modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
```diff
@@ -0,0 +1,414 @@
+/datum/reagent/consumable/nutriment/soup
+	nutriment_factor = 20 // Upped to counter the loss of random ingredients.
+	//Some soups will create between 30-60u to counter the loss of additional reagents.
+
+/datum/chemical_reaction/food/soup
+	ingredient_reagent_multiplier = 0.0
+	percentage_of_nutriment_converted = 1.0
```
vinylspiders, line 7:
```suggestion
```
How come we are blanket-applying this to all soups? Below you seem to be tweaking it at an individual soup level which seems preferable to me.

That said personally, I think it would be best to work with the folks who made the soup system upstream on this to come to a better solution if all you want is cleaner soups. I'm sure they'll have some good ideas to point you in the right direction (see https://github.com/tgstation/tgstation/pull/74205).

The whole PR would be better suited to upstream frankly but I won't press on that if you really want to keep it here. And you did claim TG said they weren't interested in this change so there is that.

Also--for what it's worth I think that poisoned ingredients _should_ make poison soup. Maybe you can devise some method of condensing the examine text's ingredient list down to the soup itself, and only be able to see the full reagents through some other method? Because it does look a whole lot neater that way I'll admit, and feels more immersive (how does one even 'see' exact measurements of nutriment in a pot of soup??).

vinylspiders (changes requested):

modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
xXPawnStarrXx: I applied it repeatedly here because when I adjusted the parent soup, it did not apply to all, even if they didn't have individual modifiers. It was the only way I could ensure they all followed the same pattern.

xXPawnStarrXx:
> Is this going to require new recipes to be manually added in when new soups are added upstream?

Sadly yes.

modular_skyrat/modules/soupiersoups/code/soup_mixtures.dm
vinylspiders: > I applied it repeatedly here because when I adjusted the parent soup, it did not apply to all, even if they didn't have individual modifiers. It was the only way I could ensure they all followed the same pattern.

Ah I see what you mean. Those particular soups have their own vars defines so you have to redefine them...

I think we should workshop this one a bit more. While I understand the intent of this and think it's going in the right direction, I feel like having the reagents from ingredients always get cooked off is completely counter to the intent of the system. There are better solutions, I think.

One thing you could maybe do instead of altering the recipes themselves, is add a new type of soup pot (or add a toggle on the existing one) that doesn't transfer reagents from the ingredients. 

https://github.com/Skyrat-SS13/Skyrat-tg/blob/65d21d4d71bc961e566eeee20aaeeb810a27d0b4/code/modules/food_and_drinks/recipes/soup_mixtures.dm#L182

You can stop it from happening here whenever that pot/pot setting is involved. Etc. That way we if people want to make 'pure' soups more easily they can use the special pot or setting, without removing a feature from the people who like the current system (as well as not being a nightmare to maintain).

xXPawnStarrXx: Yeah, I'm looking into how the base reaction works and instead seeing if I can make the subtype consume the usual offenders in the soup.

softcerv:
> > Is this going to require new recipes to be manually added in when new soups are added upstream?
> 
> Sadly yes.

While I understand the meaning behind this PR, I don't think that it justifies adding a additional thing to consider when dealing with mirror PRs. Although it's ultimately up to the other maintainers to decide if this is acceptable.

W4rd3nn:
missed an opportunity to call this souper soups

xXPawnStarrXx:
I intend to come back at this from another more modular direction sometime.

PR #24654, opened 2023-10-30
[MIRROR] Basic Pirate NPCs [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79284
--------------------

## About The Pull Request

Converts hostile pirate NPCs to basic mobs - specifically, a subtype of trooper. As their behavior is not meaningfully distinct from other troopers, this conversion mostly just sticks them on the existing AI behavior while keeping the rest the same.

Pirates do have one new thing going for them, though, to differentiate them from other troopers. They use the new **plundering attacks** component, which means that every time they land a melee attack, they steal money from the bank account of whoever they hit. This requires the target to be wearing an ID with a linked bank account, so it's not the hardest thing in the world to hide your money from them - but it's still something to be wary of! If killed, any mob with this component will drop everything they've stolen in a convenient holochip.
## Why It's Good For The Game

Takes down 5 more simplemobs, and (I think) converts the last remaining trooper-type enemy to be a basic trooper. (It's possible there's more I've forgotten that could use the same AI, though.)

The money-stealing behavior is mostly good because I think it's funny, but it also makes the pirates something a little distinct from "yet another mob that runs at you and punches you until you die". They still do that, but now there's a little twist! This can be placed on other mobs too, if we want to make any other sorts of thieves or brigands.
## Changelog
:cl:
refactor: Pirate NPCs now use the basic mob framework. They'll be a little smarter in combat, and if you're wearing your ID they'll siphon your bank account with every melee attack! Beware! Please report any bugs.
/:cl:


Compu18AI:
this
is epic

vinylspiders (approved):

PR #24791, opened 2023-11-04
[modular] you can place torches on walls

jjpark-kb:
## About The Pull Request
title-- allows you to place torches on walls

some maps have torches on walls-- and if you ever for some reason took a torch, good luck placing it back!

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: you can place torches on walls (closed turfs) now
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/turfs/closed_turfs.dm
+++ b/modular_skyrat/modules/ashwalkers/code/turfs/closed_turfs.dm
@@ -6,15 +6,15 @@
 
 // Try to place a torch on the wall such that we can only see it from one side
 /turf/closed/proc/place_torch(obj/item/flashlight/flare/torch/torch_to_place, mob/user, params)
-	if(user.dropItemToGround(torch_to_place, silent = FALSE))
+	if(user.transferItemToLoc(torch_to_place, user.drop_location(), silent = FALSE))
 		var/found_adjacent_turf = get_open_turf_in_dir(src, get_dir(src, user.loc))
 		var/list/modifiers = params2list(params)
 
 		// Center the icon where the user clicked.
 		if(LAZYACCESS(modifiers, ICON_X) && LAZYACCESS(modifiers, ICON_Y))
 			//Clamp it so that the icon never moves more than 16 pixels in either direction (thus leaving the wall turf)
-			torch_to_place.pixel_x = clamp(text2num(LAZYACCESS(modifiers, ICON_X)) - 16, -(world.icon_size/4), world.icon_size/4)
-			torch_to_place.pixel_y = clamp(text2num(LAZYACCESS(modifiers, ICON_Y)) - 16, -(world.icon_size/4), world.icon_size/4)
+			torch_to_place.pixel_x = clamp(text2num(LAZYACCESS(modifiers, ICON_X)) - 16, -(world.icon_size/3), world.icon_size/3)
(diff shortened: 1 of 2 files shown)


modular_skyrat/modules/ashwalkers/code/turfs/closed_turfs.dm
```diff
@@ -0,0 +1,7 @@
+/turf/closed/attackby(obj/item/attacking_item, mob/user, params)
+	if(istype(attacking_item, /obj/item/flashlight/flare/torch))
+		user.dropItemToGround(attacking_item)
+		attacking_item.forceMove(src)
+		return
```
vinylspiders:
```suggestion
		return place_torch(attacking_item, user, params)

	return ..()

// Try to place a torch on the wall such that we can only see it from one side
/turf/closed/proc/place_torch(obj/item/flashlight/flare/torch/torch_to_place, mob/user, params)
	if(user.dropItemToGround(torch_to_place, silent = FALSE))
		var/found_adjacent_turf = get_open_turf_in_dir(src, get_dir(src, user.loc))
		var/list/modifiers = params2list(params)

		// Center the icon where the user clicked.
		if(LAZYACCESS(modifiers, ICON_X) && LAZYACCESS(modifiers, ICON_Y))
			//Clamp it so that the icon never moves more than 16 pixels in either direction (thus leaving the wall turf)
			torch_to_place.pixel_x = clamp(text2num(LAZYACCESS(modifiers, ICON_X)) - 16, -(world.icon_size/4), world.icon_size/4)
			torch_to_place.pixel_y = clamp(text2num(LAZYACCESS(modifiers, ICON_Y)) - 16, -(world.icon_size/4), world.icon_size/4)

		// Try to put the torch in the adjacent turf relative to the user. This way it's not visible from the other side of the wall
		if(found_adjacent_turf)
			torch_to_place.forceMove(found_adjacent_turf)
		else
			torch_to_place.forceMove(src) // no open turfs for some reason
			return TRUE

		// The item itself is in the adjacent turf, so we need to shift the icon one tile over to put it in the wall
		switch(get_dir(found_adjacent_turf, src))
			if(NORTH)
				torch_to_place.pixel_y += world.icon_size
			if(SOUTH)
				torch_to_place.pixel_y -= world.icon_size
			if(EAST)
				torch_to_place.pixel_x += world.icon_size
			if(NORTHEAST)
				torch_to_place.pixel_y += world.icon_size
				torch_to_place.pixel_x += world.icon_size
			if(SOUTHEAST)
				torch_to_place.pixel_y -= world.icon_size
				torch_to_place.pixel_x += world.icon_size
			if(WEST)
				torch_to_place.pixel_x -= world.icon_size
			if(NORTHWEST)
				torch_to_place.pixel_y += world.icon_size
				torch_to_place.pixel_x -= world.icon_size
			if(SOUTHWEST)
				torch_to_place.pixel_y -= world.icon_size
				torch_to_place.pixel_x -= world.icon_size
		return TRUE
```
Changes here:

1) Sets the torch's loc to that adjacent tile so it doesn't show up from the other side. I don't think there's any other way to do this.
2) Torch goes where the cursor is like placing stuff on tables. You could get rid of this part if you'd prefer it to always be centered (though you'd have to temporarily set `item_flags` to `NO_PIXEL_RANDOM_DROP` before calling `dropItemtoGround()` as well if you go that route).

vinylspiders (changes requested):

vinylspiders (approved):

PR #24800, opened 2023-11-05
Snowfarming for Demis.

xXPawnStarrXx:
## About The Pull Request

It always struck me as strange that there was a small amount of basalt around the demi-camp on icebox because it allowed them to plant. Making them significantly more restricted than their ashy counterparts, so to counter that I have on request added the component to snow tiles, allowing you to plant seeds into the soil.

Brings a level of equality to ashwalkers and icebox demi-humans. Meaning they aren't limited on farming space.
## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: Added component to snowtiles, allowing seeds to be planted in them.
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/buildings/ash_farming.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/ash_farming.dm
@@ -274,3 +274,7 @@
 /turf/open/misc/asteroid/basalt/Initialize(mapload)
 	. = ..()
 	AddComponent(/datum/component/simple_farm)
+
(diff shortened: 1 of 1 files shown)


Paxilmaniac:
There is so much of the dirt they can plant in around their new camp that i really don’t think this is needed.

xXPawnStarrXx:
> There is so much of the dirt they can plant in around their new camp that i really don’t think this is needed.

That is true, but they have less if they choose to use the anthills.

jjpark-kb:
rather than planting directly in the snow (and by that, I mean putting the component on initialize), how about you put the component on snow on the getDug() proc?

PR #24809, opened 2023-11-06
[MIRROR] Golems cannot turn into Plasmamen [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79249
--------------------
## About The Pull Request

Fixes #79155 
Also cleans up the code in plasma river "lava" a bit

The plasma river code was a little crusty and confusing and made some interesting decisions.
To start with it would do _two_ burn damage per tick to people entering it.
Then _15 toxic and 25 more burn damage_ to specifically humans?
This seems... unintended? So now it does the same amount of damage equally to anyone regardless of if they are human or not.

In terms of the "turning dead people into dead plasmamen" behaviour we will now:
- Not attempt to transform any limbs which cannot be dismembered, with the assumption that those aren't meant to be replaceable.
- Not attempt to transform anyone with the undismemberable trait, with the assumption that their body is supposed to be immutable.
- Not attempt to transform anyone with the new "don't transform me, plasma river" trait which replaces an explicit species check for being an android or plasmaman (already).

I also simply made golems immune to the plasma river _anyway_, because people expect them to be able to wade into lava and the fact that this is "cold lava" probably shouldn't impede that.

I also made golems unable to become husks because... none of the three occasions where this could happen really apply and also if they get husked they turn into gross-looking humans instead of golems until fixed, which is weird.

Finally I did some more fucking around with husking because it's visually weird.
I added a new flag to bodyparts (applied to robotic and skeletal parts) which means that they will never be replaced with a husk icon. It might be preferable to instead make custom husk sprites for these limbs? But half of them are on species which cannot be husked in the first place so this would only appear when they are transplanted onto a species which can.

## Changelog

:cl:
fix: The plasma river is about as deadly for animals as it is for humans.
fix: Golems can now wade in the plasma river unscathed.
fix: Undismemberable limbs will no longer be dismembered by the plasma river.
balance: Golems and plasmamen cannot become husked.
image: Robotic and Skeletal parts will remain distinct while the rest of the body is husked.
/:cl:


vinylspiders:
Needs transformation sting which I still have to work on a bit

vinylspiders (approved):

PR #24852, opened 2023-11-08
Fixes flavor texts sometimes bluescreening

### LINKED ISSUE #24847: Inspecting Akula's in Particular results in a bluescreen window
Round ID: 9856

if you shift click an Akula and then hit the "look closer", you will be hit with a bluescreen, other species/characters that are not akula's are fine and work as intended

to reproduce this step simply "inspect" an akula closer and it will occur

Majkl-J:
## About The Pull Request

Fixes https://github.com/Skyrat-SS13/Skyrat-tg/issues/24847

For some arcane reason each time I try to examine an akula, the flavor text screen bluescreens with the error of "replace" not being a method of an element.

So to solve this the species lore now gets merged into one string upon getting parsed

Reading flavor text good, yes?

## Proof of Testing

No idea what to put here? Me examining an akula? I'll just put it here when requested.

I did also test if the link highlight still works, and it does, so functionally there should be no change.

<details>
  
</details>

## Changelog

:cl:
fix: Flavor text should no longer bluescreen
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/mob/living/examine_tgui.dm
+++ b/modular_skyrat/master_files/code/modules/mob/living/examine_tgui.dm
@@ -82,7 +82,7 @@
 		obscured = (holder_human.wear_mask && (holder_human.wear_mask.flags_inv & HIDEFACE)) || (holder_human.head && (holder_human.head.flags_inv & HIDEFACE))
 		custom_species = obscured ? "Obscured" : holder_human.dna.species.lore_protected ? holder_human.dna.species.name : holder_human.dna.features["custom_species"]
 		flavor_text = obscured ? "Obscured" :  holder_human.dna.features["flavor_text"]
-		custom_species_lore = obscured ? "Obscured" : holder_human.dna.species.lore_protected ? holder_human.dna.species.get_species_lore() : holder_human.dna.features["custom_species_lore"]
+		custom_species_lore = obscured ? "Obscured" : holder_human.dna.species.lore_protected ? holder_human.dna.species.get_species_lore().Join("\n") : holder_human.dna.features["custom_species_lore"]
 		ooc_notes += holder_human.dna.features["ooc_notes"]
 		if(!obscured)
 			headshot += holder_human.dna.features["headshot"]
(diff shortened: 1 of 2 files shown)


vinylspiders (changes requested):
Appreciate the attempt to fix, but this is just masking the issue. It is caused by Akula being the only race that has `lore_protected` set to `TRUE`, which leads to this code here passing the results of `get_species_lore()` to the function you've added the try/catch to.

https://github.com/Skyrat-SS13/Skyrat-tg/blob/e568fd216a40ad3f31c0428cb5e0bf96e4651e7c/modular_skyrat/master_files/code/modules/mob/living/examine_tgui.dm#L85

The function `formatURLs(arg)` expects a string, but it's getting a list from `get_species_lore()`. You need to convert that list to a string before sending it to tgui.

Please remove your try/catch (we want it to give us an error in cases like this so we can actually find the bug) and instead alter the linked line to be the following:

`custom_species_lore = obscured ? "Obscured" : holder_human.dna.species.lore_protected ? holder_human.dna.species.get_species_lore().Join("") : holder_human.dna.features["custom_species_lore"]`

vinylspiders:
https://github.com/Skyrat-SS13/Skyrat-tg/issues/24847

Majkl-J:
Testing requesting changes, will press the review button again once that is over

vinylspiders:
Nice work, thanks! 👍 

I am almost certain this should fix it, the only thing I'm unsure of is how the formatting will look. Feel free to tweak that if you want, but as long as it doesn't bluescreen I'm ok with it as is (assuming it doesn't look awful).

vinylspiders (approved):

vinylspiders:
(by formatting I mean, the Akula's `get_species_lore()` proc--it looks like its paragraphs can be broken into multiple list entries and neatened up a bit. But that is something I would consider beyond the scope of this PR and if you feel like doing that you can open a separate one for it to get more gbp!)

Majkl-J:
Now confirmed to work on my end. I might tweak the formatting a bit later separetely, simple newlines will do the work for now.

Also sorry for re-requesting approval, didnt notice it was already given and wanted to make sure it works before I asked for it

vinylspiders:
> Now confirmed to work on my end. I might tweak the formatting a bit later separetely, simple newlines will do the work for now.
> 
> Also sorry for re-requesting approval, didnt notice it was already given and wanted to make sure it works before I asked for it

Don't worry about it--I still get email notifications either way. 

The newlines (if you're referring to the "\n" arg in Join) won't actually get applied to anything unless you move each part you want to get the newline into separate list entries. You can see how that's done in the `get_species_lore()` for plasmamen. Just something to keep in mind when/if you end up doing that.

vinylspiders (approved):

PR #24888, opened 2023-11-09
Removes illegal tech from legally bought suppressors' deconstructor value

### LINKED ISSUE #24779: Suppressor from cargo has illegal tech value
## Round ID: 
9832/don't matter really

## Reproduction:
Go to cargo, open company imports and go to Vitezstvi, spend 100 credits for a suppressor and you got illegal tech researched.

It's like makarov from contraband locker situation, but even easier to get. Illegal tech can already be obtained trough methods like xenoarch ( which became easier ), antags and stray pods, so making it easy as this to obtain is odd and most likely just something that was overlooked.
Think suppressor should be tweaked not to have illegal tech value.
tech

vinylspiders:
## About The Pull Request

Closes https://github.com/Skyrat-SS13/Skyrat-tg/issues/24779

What it says on the tin. Illegal tech was easily attainable this way. 

Created a new subtype, `/obj/item/suppressor/standard` which is considered legal, and that is now what is purchasable through cargo.

Uplink/opfor suppressors use the original type and can be deconstructed as before to attain illegal tech.

Also rearranged some modular overrides in techweb/all_nodes just in case upstream changes something.

Fix an oversight

## Proof of Testing

<details>
<summary>Two kinds, one legal one not.</summary>
  
image

image

</details>

## Changelog

:cl:
fix: legally purchased suppressors no longer give illegal tech when deconstructed
/:cl:

### DIFF
--- a/modular_skyrat/modules/assault_operatives/code/armaments/armament_utility.dm
+++ b/modular_skyrat/modules/assault_operatives/code/armaments/armament_utility.dm
@@ -3,7 +3,7 @@
 	category_item_limit = 8
 
 /datum/armament_entry/assault_operatives/utility/suppressor
-	item_type = /obj/item/suppressor
+	item_type = /obj/item/suppressor/standard
 	cost = 1
 
 /datum/armament_entry/assault_operatives/utility/thermals
--- a/modular_skyrat/modules/assault_operatives/code/armaments/armament_utility.dm
+++ b/modular_skyrat/modules/assault_operatives/code/armaments/armament_utility.dm
@@ -3,7 +3,7 @@
 	category_item_limit = 8
 
 /datum/armament_entry/assault_operatives/utility/suppressor
-	item_type = /obj/item/suppressor/standard
+	item_type = /obj/item/suppressor
 	cost = 1
 
 /datum/armament_entry/assault_operatives/utility/thermals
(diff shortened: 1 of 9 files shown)


Plum-but-gayer:
throws my hands into the air and cries on the floor i was about to push a fix for this 😧
anyway ty for fixing stuff

vinylspiders:
> throws my hands into the air and cries on the floor i was about to push a fix for this 😧 anyway ty for fixing stuff

Sorry!

modular_skyrat/modules/assault_operatives/code/armaments/armament_utility.dm
```diff
@@ -3,7 +3,7 @@
 	category_item_limit = 8
 
 /datum/armament_entry/assault_operatives/utility/suppressor
-	item_type = /obj/item/suppressor
+	item_type = /obj/item/suppressor/standard
```
GoldenAlpharex: I feel like they can have the illegal one, considering they're from the Syndicate, but I'm not overly worried about it either.

vinylspiders: sure

GoldenAlpharex (approved):

PR #24913, opened 2023-11-10
[MODULAR] Anesthetic mask will give an error message when placed via mapping

vinylspiders:
## About The Pull Request

These should cause CI to fail when placed via mapping. Now, they will! They will also self-delete if accidentally spawned via adminnery.

Also fixed a lot of bugs with the code. They were storing a hard ref in a weakref...fixed it so they actually use a real weakref, now, instead of trying to pretend the hardref was a weakref.

Fixes bugs, makes anesthetic machine work better. Players picking up NODROP items is bad.

## Proof of Testing

<details>
<summary>Still works</summary>

dreamseeker_bSqpM4IvAb

dreamseeker_6uoxjJpISJ

</details>

<details>

<summary>The new runtimes</summary>

image

</details>

## Changelog

:cl:
refactor: fixed hard dels associated with anesthetic machine, & the mask item will now cause ci to fail when placed via mapping
fix: you no longer have to violently beat the anesthetic machine to death with the anesthetic tank to load it
fix: fixed a bug that would cause the anesthetic machine to become unusable when a mask gets destroyed (e.g., in a fire or explosion)
/:cl:

### DIFF
--- a/_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
+++ b/_maps/RandomRuins/IceRuins/skyrat/icemoon_underground_syndicate_base1_skyrat.dmm
@@ -2401,10 +2401,6 @@
 	pixel_x = -5;
 	pixel_y = 5
 	},
-/obj/item/clothing/mask/breath/anesthetic{
-	pixel_x = 8;
-	pixel_y = 5
-	},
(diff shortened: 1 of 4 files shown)


Zergspower:
that works!

jjpark-kb (approved):

PR #24915, opened 2023-11-10
dug snow farming, plants and ants & MORE

jjpark-kb:
## About The Pull Request
allows the ability to farm on snow that has been dug, both plants and ants
this does change it for basalt as well-- so it will have to be dug first before farming plants or ants

closes #24800

tribals should stick togetha~
it makes more sense that you need to prepare the tiles a bit at least, both plants and ants

## Proof of Testing

<details>
  
image
image

</details>

## Changelog

:cl:
add: you can farm plants and ants on dug snow
balance: you must now dig the basalt to farm plants and ants
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/buildings/antfarm.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/antfarm.dm
@@ -28,8 +28,8 @@
 /obj/structure/antfarm/Initialize(mapload)
 	. = ..()
 	var/turf/src_turf = get_turf(src)
-	if(!istype(get_turf(src), /turf/open/misc/asteroid/basalt))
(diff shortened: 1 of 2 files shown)


PawnAtWork:
Sand might become an issue for iceboys; Maybe make it so that the hills produce a little more sand or instead get some when digging up snow?

jjpark-kb:
> Sand might become an issue for iceboys; Maybe make it so that the hills produce a little more sand or instead get some when digging up snow?

You can get sand from breaking the rocks on icebox. I don't think making sand appear from digging snow would be a good change, at least in this PR.

vinylspiders (approved):
I really wish these sprites had icon smoothing

PR #24932, opened 2023-11-12
some changes for the russian derelict station

jjpark-kb:
## About The Pull Request
the russian derelict will no longer use the old window spawners, and instead will use full tile spawners.
I replaced the windows where the "singularity" would have been with reinforced plasma windows, because that is what it should have been.
the derelict now has a small sm shard, as well as a grounding rod.
the derelict has a bluespace miner board and some bluespace crystals nearby.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: added a supermatter shard to the russian derelict station
add: added a bluespace miner board and bluespace crystals to the russian derelict station
add: the russian derelict station now uses fulltile window spawners instead of the directional window spawners
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/skyrat/russian_derelict_skyrat.dmm
+++ b/_maps/RandomRuins/SpaceRuins/skyrat/russian_derelict_skyrat.dmm
@@ -1,0 +1,17208 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"ab" = (
+/obj/machinery/atmospherics/pipe/smart/manifold4w/supply/visible,
+/obj/effect/mapping_helpers/burnt_floor,
+/obj/effect/decal/cleanable/dirt,
+/turf/open/floor/plating/airless,
+/area/ruin/space/ks13/engineering/atmos)
+"ac" = (
+/obj/structure/lattice,
+/obj/structure/door_assembly/door_assembly_public,
+/turf/template_noop,
+/area/space/nearstation)
+"ae" = (
+/obj/structure/door_assembly/door_assembly_mai,
+/turf/open/floor/plating/airless,
+/area/ruin/space/ks13/service/jani)
+"ag" = (
+/obj/structure/table_frame,
+/obj/effect/decal/cleanable/blood/gibs/old,
+/obj/effect/decal/cleanable/dirt,
+/turf/open/floor/iron/dark,
+/area/ruin/space/ks13/medical/morgue)
+"ah" = (
+/obj/effect/spawner/structure/window/reinforced,
+/turf/open/floor/plating/airless,
+/area/ruin/space/ks13/service/cafe)
+"aj" = (
+/obj/effect/decal/cleanable/glass,
+/obj/effect/mapping_helpers/broken_floor,
+/obj/effect/decal/cleanable/dirt,
+/turf/open/floor/plating/airless,
+/area/ruin/space/ks13/hallway/aft)
+"ak" = (
+/obj/structure/cable,
+/turf/open/floor/plating,
+/area/ruin/space/ks13/engineering/sb_bow_solars_control)
+"al" = (
+/obj/machinery/gravity_generator/main,
+/turf/open/floor/iron/airless,
+/area/ruin/space/ks13/engineering/grav_gen)
+"an" = (
+/obj/structure/chair{
+	dir = 1
+	},
+/obj/structure/window/spawner/directional/south,
+/obj/effect/decal/cleanable/dirt,
+/turf/open/floor/iron,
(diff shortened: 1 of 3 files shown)


Plejek:
**Big drone saves the day**

KiriKaz:
BIG DROOOONE! Hell yeah!

Paxilmaniac:
> the russian derelict will no longer use the old window spawners, and instead will use full tile spawners.

If you're doing stuff like this you really ought to take it upstream

jjpark-kb:
> > the russian derelict will no longer use the old window spawners, and instead will use full tile spawners.
> 
> If you're doing stuff like this you really ought to take it upstream

my concern is that, even if I do make this upstream, there are still things I will need to modify the map for skyrat-exclusive content-- so I believe this is fine.

ghost:
Seconding to take most of this upstream, though AFAIK they're intentionally the old method - it's an old station with old construction methods.
They could really use a power source though.

The BS Miner is the only Skyrat addition that would be added, and can just be added later or individually. Perhaps just have the board spawn in the same place as the drones do to save on mapping conflicts.

Jolly-66 (commented):
Since I was the one to last touch this upstream (to my knowledge),

The directional window spawners are intentional and go against the dated theme of this ruin.

I also don't understand the addition of miners and bluespace in general, there *shouldn't* be anything in the tech fab that requires blue space, if anything, they should be blaclisted per a change upstream ages ago (unless that was reverted here).

Lastly, the SM shard. This is just "my" issue I have with anything in space in general: Someone is just going to come and take it, regardless of the rules. People play by "me first, you second" mentality (this obviously doesn't apply to everyone).

I'm not on board with these changes, especially the window changes. This is an **old** ruin its supposed to be dated, old, and look the same. You should also try to push these upstream, but I have a strong feeling the latter of the two won't be accepted.

Jolly-66:
> Seconding to take most of this upstream, though AFAIK they're intentionally the old method - it's an old station with old construction methods. They could really use a power source though.

I doubt /tg/ will take a SM shard cause they try to keep KS13 and Charlie different in their approaches of "what are you doing", i.e Charlie is the carbon "learn to play the game" while drones are "learn to play engineering" per-se. That's just my take. If power is a critical issue, I say expand and add a new solar array somewhere, OR give them another crate. I do agree they need more power, but a SM shard is not the way to go (They have a turbine but who the hell actually knows fusion).

KiriKaz:
directional windows are a fine pick abt the commit, but there _are_ doubled up spawners that do need a little touching up around the map, albeit unrelated to the current changes

jjpark-kb:
> The directional window spawners are intentional and go against the dated theme of this ruin.

Yea, I had surmised that the window spawners were intentional-- then the question is whether I should just create a new ruin for drones that do not have this intention?

>I also don't understand the addition of miners and bluespace in general, there shouldn't be anything in the tech fab that requires blue space, if anything, they should be blaclisted per a change upstream ages ago (unless that was reverted here).

The drones have no way of leaving the station. So obtaining materials to keep the station looking nice would be nice. I understand that they can self-cannibalize part of the station, but that just makes it almost impossible to actually "restore" the station, which is one of the drones' tasks. The bluespace is added because bluespace miners require it.

>Lastly, the SM shard. This is just "my" issue I have with anything in space in general: Someone is just going to come and take it, regardless of the rules. People play by "me first, you second" mentality (this obviously doesn't apply to everyone).

It really is unfortunate that certain community members have this mentality, though I'm not sure if this should stop us from adding content to ghost roles? Also, the supermatter shard is orderable on the cargo request console.

>I doubt /tg/ will take a SM shard cause they try to keep KS13 and Charlie different in their approaches of "what are you doing", i.e Charlie is the carbon "learn to play the game" while drones are "learn to play engineering" per-se

That makes sense for /tg/; however, I would argue that learning how to setup a supermatter crystal shard engine from scratch is very much in line with "learning to play engineering."

> If power is a critical issue, I say expand and add a new solar array somewhere, OR give them another crate. I do agree they need more power, but a SM shard is not the way to go (They have a turbine but who the hell actually knows fusion).

I mean yea, I could add another solar crate so they could learn to do that from scratch. On the topic of turbine-- it is so very limited because they do not have gas miners, so I'm going to ignore the turbine as a valid source of power.

>I'm not on board with these changes, especially the window changes. This is an old ruin its supposed to be dated, old, and look the same. You should also try to push these upstream, but I have a strong feeling the latter of the two won't be accepted.

So I pose the question that I asked earlier-- do you think I should just create a new ruin for drones to do these things?

jjpark-kb:
Alright, so I do agree that we should leave KS13 alone and let it have its own lore and such, and that I create a separate map. It is best to keep the aesthetics and choices that KS13 has for KS13, and any of the design choices I wanted to make would be better suited for an entirely different map.

PR #24933, opened 2023-11-12
Removes the fishing minigame

jjpark-kb:
## About The Pull Request
title

the server sometimes experiences some REALLY bad stuff that makes playing the fishing minigame nigh impossible.

## Proof of Testing

<details>
  
image
image

</details>

## Changelog

:cl:
del: removed the fishing minigame
del: removed the trash from the water basin fishing source
/:cl:

### DIFF
--- a/code/modules/fishing/fishing_minigame.dm
+++ b/code/modules/fishing/fishing_minigame.dm
@@ -257,7 +257,7 @@
 		send_alert("miss!")
 		start_baiting_phase()
 	else if(phase == BITING_PHASE)
-		start_minigame_phase()
+		complete(TRUE) //SKYRAT EDIT: Remove Fishing Minigame ORIGINAL: start_minigame_phase()
 	return COMSIG_MOB_CANCEL_CLICKON
 
 /// Challenge interrupted by something external
--- a/modular_skyrat/modules/reagent_forging/code/water_basin.dm
+++ b/modular_skyrat/modules/reagent_forging/code/water_basin.dm
@@ -103,6 +103,4 @@
 		/obj/item/fish/gunner_jellyfish = 15,
 		/obj/item/fish/needlefish = 10,
 		/obj/item/fish/armorfish = 10,
-		/obj/effect/spawner/random/maintenance = 10,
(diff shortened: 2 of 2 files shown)


nikothedude:
never heard of this problem before, can someone confirm

also, wHy remove the trash

jjpark-kb:
> never heard of this problem before, can someone confirm

The server chugs at times, I don't see why you need confirmation of this.

> also, wHy remove the trash

It is broken and has been broken for sometime, it was going to be removed eventually.

nikothedude:
> > never heard of this problem before, can someone confirm
> 
> The server chugs at times, I don't see why you need confirmation of this.

no i mean confirmation taht server lag obliterates the miningame

nikothedude:
actually why am i asking for someone else to confirm their word is as fine as yours w/e

Nerev4r:
It's mostly been insulated from lag as far as I'm aware. @Ghommie Can you weigh in on this?

Paxilmaniac:
the trash removal change is fine but

I play on that dollar store college dorm internet connection and the only times the fishing minigame is so effected by lag that I couldnt use it, I'm also lagging hard enough that _nothing else in the game would have been playable in the first place_. I think using the server as an excuse for removing it is falsified and hiding why you actually want it gone.

Nerev4r:
Conjecture of intent isn't really necessary here. I'd remark myself, though, that fishing was pretty doable the other day (when I tried it for the first time in months) even with a really shitty rod setup. I'll check w/ the author in devcode.

jjpark-kb:
discussion in dev-code; something else will be done instead

Ghommie:
> It's mostly been insulated from lag as far as I'm aware. @Ghommie Can you weigh in on this?

The fishing minigame doesn't run on a TGUI window anymore. It has its own subsystem, and severe server lag will impact it just like any other subsystem, if not worse: it's supposed to fire every decisecond, and the delays to the subsystem firing can easily be much larger than that. 

In the end, you can do what you want, I don't mind really. Even if the minigame were removed in favor of a do-after, the problem of latency spikes won't make the experience much more pleasant anyway.

PR #25035, opened 2023-11-16
[tm only] disable gutlunch breeding

jjpark-kb:
This is to be TM'd ONLY!

Disables Gutlunches from breeding-- sorry ashies.


Zergspower:
:squints:

vinylspiders:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/25066 This is now fixed

PR #25063, opened 2023-11-17
The Bird Is The Word: Avian Traits quirk for tweeting, chirping, and shrieking

Ephemeralis:
## About The Pull Request

Adds a simple traits quirk for no cost that allows avian demihumans and other bird-like characters to speak as befitting their chosen identity.

To prevent weirdness with picking both canine & avian traits on the same character (and it being uncertain which one would apply at roundstart), I've also made all racial speech traits mutually exclusive with one another.

I've also added two trait defines for canines and avians to bring them in line with the assignment feline aspect performs, and two helper macros to make checking for their presence easier for future contributors.

Lets people with bird-like characters lean into their identity more, and finally frees them from being third-class citizens in terms of conversational flavour.

## Proof of Testing

<details>

YkARP5PuFJ

G479c5C5sb

</details>

## Changelog

:cl: yooriss
add: The 'Avian Traits' quirk is now available for all those of a bird-like persuasion (be they extraplanar visitors or otherwise) to peep, squawk, murmur and shriek their lungs out.
change: You can now pick only one racial speech quirk (feline, canine, avian) at a time.
add: Macros for canine and avian archetypes are now available for future use, making more species-specific interactions with things (similar to felinids and laser pointers) easier to make.
/:cl:

### DIFF
--- a/code/__DEFINES/~skyrat_defines/traits.dm
+++ b/code/__DEFINES/~skyrat_defines/traits.dm
@@ -51,6 +51,12 @@
 // felinid traits
 #define TRAIT_FELINE "feline_aspect"
 
+// canine traits
+#define TRAIT_CANINE "canine_aspect"
+
+// avian traits
+#define TRAIT_AVIAN "avian_aspect"
+
 // chameleon mutation
 #define TRAIT_CHAMELEON_SKIN "chameleon_skin"
 
--- a/code/__DEFINES/~skyrat_defines/traits.dm
+++ b/code/__DEFINES/~skyrat_defines/traits.dm
@@ -66,9 +66,6 @@
 /// The trait that determines if someone has the oversized quirk.
 #define TRAIT_OVERSIZED "trait_oversized"
 
-/// Caninid trait
-#define TRAIT_CANINE "trait_canine"
-
 /// Cargo Loader trait
 #define TRAIT_TRASHMAN "trait_trashman"
 
--- a/modular_skyrat/modules/organs/code/tongue.dm
+++ b/modular_skyrat/modules/organs/code/tongue.dm
@@ -42,7 +42,7 @@
 	signer.verb_yell = "shrieks"
 
 /obj/item/organ/internal/tongue/avian/Remove(mob/living/carbon/speaker, special = FALSE)
-	..()
+	. = ..()
 	speaker.verb_ask = initial(verb_ask)
 	speaker.verb_exclaim = initial(verb_exclaim)
 	speaker.verb_whisper = initial(verb_whisper)
--- a/modular_skyrat/modules/organs/code/tongue.dm
+++ b/modular_skyrat/modules/organs/code/tongue.dm
@@ -27,6 +27,28 @@
 	speaker.verb_sing = initial(verb_sing)
 	speaker.verb_yell = initial(verb_yell)
 
+/obj/item/organ/internal/tongue/avian
+	name = "avian tongue"
+	desc = "A short and stubby tongue that craves seeds."
+	say_mod = "chirps"
+	icon_state = "tongue"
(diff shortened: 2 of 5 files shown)


desertmote:
There are odds you will get the request for this PR to be modified in the same vain as this PR: https://github.com/Skyrat-SS13/Skyrat-tg/pull/13295.

PRs like this have been made before and denied because the addition of extra tongues was found to be bad implementation, good luck!

SomeRandomOwl:
Honkpocket is correct, it'd be best to implement this without changing peoples internal organs and modifying their tongue.

Ephemeralis:
This would break congruency with how canine traits are handled, and previous commit history indicates the current tongue-based solution is what was arrived at over pushing trait-linked overrides inside things like say procs. Which way is best to proceed?

Ephemeralis:
Testing with dual canine & avian traits quirks revealed some weirdness, which I've addressed by allowing selection of only one racial speech quirk at a time, which probably should've been the case to begin with.

Brought this up on #development-code and it seems that the tongue-based approach is fine as long as `copy_traits_from` is used, so we should be okay.

I've also linked in race traits for canines and avians and helper macros to allow for future interactions with these quirks to be done more easily. Chocolate causing toxin damage for canines, anyone?

code/__DEFINES/~skyrat_defines/traits.dm
```diff
@@ -51,6 +51,12 @@
 // felinid traits
 #define TRAIT_FELINE "feline_aspect"
 
+// canine traits
+#define TRAIT_CANINE "canine_aspect"
```
vinylspiders, line 55: You forgot to get rid of the old define, line 70

modular_skyrat/modules/organs/code/tongue.dm
```diff
@@ -27,6 +27,28 @@
...
+
+/obj/item/organ/internal/tongue/avian/Insert(mob/living/carbon/signer, special = FALSE, drop_if_replaced = TRUE)
+	. = ..()
+	signer.verb_ask = "peeps"
+	signer.verb_exclaim = "squawks"
+	signer.verb_whisper = "murmurs"
+	signer.verb_yell = "shrieks"
+
+/obj/item/organ/internal/tongue/avian/Remove(mob/living/carbon/speaker, special = FALSE)
+	..()
```
vinylspiders:
```suggestion
	. = ..()
```

modular_skyrat/modules/organs/code/tongue.dm
```diff
@@ -27,6 +27,28 @@
...
+	speaker.verb_ask = initial(verb_ask)
+	speaker.verb_exclaim = initial(verb_exclaim)
+	speaker.verb_whisper = initial(verb_whisper)
+	speaker.verb_sing = initial(verb_sing)
+	speaker.verb_yell = initial(verb_yell)
+
```
vinylspiders, line 51: I think it's fine to do it like this, although is there really no 'say_mod' equivalent for the other verbs? Feels like that should be added upstream (not asking you to, just thinking aloud).

I have no problems with this given there is no other way far as I can tell...The canine tongue does it the same way.

vinylspiders (changes requested):

vinylspiders:
> There are odds you will get the request for this PR to be modified in the same vain as this PR: #13295.
> 
> PRs like this have been made before and denied because the addition of extra tongues was found to be bad implementation, good luck!

The issue back then with the hemophage abilities being lost when choosing an augmented organ was fixed a few months back, so adding new tongues isn't as big a deal nowadays so long as they are making sure to call `copy_traits_from()` (which they are doing).

vinylspiders (approved):

PR #25131, opened 2023-11-19
[MISSED MIRROR] Puts all traits in the globalvars file + CI Testing (#79642)

vinylspiders:
## ORIGINAL PR: https://github.com/tgstation/tgstation/pull/79642

Fixes #76349

I didn't know that people needed to add any new traits to a global list so they can be easily read in View Variables, and was pretty shocked to find out many other people didn't know it was a thing. Let's make it a thing by testing it using a new CI Python Linter to check this. But oh no-

image

There were about 200+ missing traits. Alright, so let's do the following:

* Move trait defines to their own dedicated folder in the `_DEFINES` folder.
* Split up the traits mega-file into different files, for better organization. One for the macros, one for the sources, and a few for the "trait declarations"
* Run the linter a load of times and add everything to the globalvars file, removing anything that's no longer used and figuring out where the best categorization of it is. also minor code improvements. also rename all of the ones that look weird. also fix list indentations
* Also alphabetize the lists because it's easy
* Move everything to a new `traits_by_type` list, while keeping the admin one the way it is for the time being while we figure out a better way to show that list to admins.
* Profit

Mapping trait injectors will now work for any type of trait. You shouldn't add any trait via this injector though, but you're no longer limited to coders remembering to add it to that critical list you needed.

Lays the framework for a better view variables experience. This work is too lengthy to presently do, but hopefully we can get this done sooner rather than later. we will need a code-accessible way to view these traits for such a framework to be implemented, so let's just do that.

Future steps are to break down the mega-declarations file into a folder full of separate files by typepath, but that requires a lot of auditing. Does need to happen one day though, there's a lot of mob traits mingled with datum traits and auuugh we gotta do this later this PR is already massive.

there's probably ways to game this but this catches _my_ mistakes so good luck to everyone else (it should work for 99% of everyone)

Nothing applicable to players. However, to mappers, the mapping trait injector should always be able to add any kind of trait (which is rather good for the times when you need it).

---

CI is going to love this one...


GoldenAlpharex (approved):
Good catch!

PR #25154, opened 2023-11-20
[FIX] Chameleon Is Toggleable Again

Nerev4r:
## About The Pull Request
I have plans that I cannot share with you right now, because the haters will sabotage me.

This was working a long time ago, but got lost in a random code audit. This mutation is hardly usable without it.

## Proof of Testing
<details>
  
image

</details>

## Changelog
:cl:
fix: Chameleon mutation is toggleable again.
/:cl:

### DIFF
--- a/code/datums/mutations/chameleon.dm
+++ b/code/datums/mutations/chameleon.dm
@@ -8,7 +8,6 @@
 	text_lose_indication = "<span class='notice'>You feel oddly exposed.</span>"
 	instability = 25
 	power_coeff = 1
-	power_path = /datum/action/cooldown/spell/chameleon_skin_activate //SKYRAT EDIT
 
 /datum/mutation/human/chameleon/on_acquiring(mob/living/carbon/human/owner)
 	if(..())
--- a/code/datums/mutations/chameleon.dm
+++ b/code/datums/mutations/chameleon.dm
@@ -8,6 +8,7 @@
 	text_lose_indication = "<span class='notice'>You feel oddly exposed.</span>"
 	instability = 25
 	power_coeff = 1
+	power_path = /datum/action/cooldown/spell/chameleon_skin_activate //SKYRAT EDIT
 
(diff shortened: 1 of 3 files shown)


Namwty:
Frankly should be mainstream ( imo )

From skyrat point, it's good, cause you can deactivate it to have to RP immersion ( I love chameleon, it's niche and fun ), but god it's annoying to go invisible during convos

From balance point, it's still good, **as in most cases it will increase instability** ( and no, I don't think rerolling DNA constantly for it is exactly a concern )
Increased instability means it won't be stackable with other mutations + makes chromosomes slightly more useful if someone wants to work with those. ( Instability matters so you can't stack it with stuff powergamey stuff like fiery sweat )

code/datums/mutations/chameleon.dm
```diff
@@ -8,6 +8,7 @@
 	text_lose_indication = "<span class='notice'>You feel oddly exposed.</span>"
 	instability = 25
 	power_coeff = 1
+	power_path = /datum/action/cooldown/spell/chameleon_skin_activate //SKYRAT EDIT
```
vinylspiders: Would you mind putting this in master_files as an override? You already have the file made
```suggestion
```

modular_skyrat/master_files/code/datums/mutations/chameleon.dm
```diff
@@ -0,0 +1,24 @@
+/datum/action/cooldown/spell/chameleon_skin_activate
```
vinylspiders, line 5:
```suggestion
// toggleable chameleon skin
/datum/mutation/human/chameleon
	power_path = /datum/action/cooldown/spell/chameleon_skin_activate

/datum/action/cooldown/spell/chameleon_skin_activate
```

vinylspiders (changes requested):

vinylspiders (approved):

PR #25202, opened 2023-11-23
Removes Kilostation completely from the Automapper

StrangeWeirdKitten:
## About The Pull Request

Kilostation is no more, sadly. These templates are just codebase junk now and there's no reason to keep them around.

Nothing will be visible

## Proof of Testing

If it passes CI, it works. 

image

## Changelog

:cl: The Sharkening
refactor: Removes unused Kilostation Automapper templates
/:cl:


Jolly-66 (approved):
Someone is going to groan that they want Kilo back.

I don't care this is fine.

StrangeWeirdKitten:
> Someone is going to groan that they want Kilo back.
> 
> I don't care this is fine.

Bubberstation reintroduced kilo and we keep it updated on our end. If anyone from Skyrat wants to bring it back to Skyrat, they are more than welcome to and it would realistically be a very simple PR. I know it was removed from TG after birdshot was introduced. But that's for any Skyrat contributor to decide and for the maintainers to want Kilo back. But yeah, there's no need to have automapping templates for a map that doesn't exist anymore. (Plus you wouldn't need automapping templates in the first place now)

vinylspiders (approved):
It's also not hard to bring things back if need be, we have all the commit history.

PR #25250, opened 2023-11-25
[MIRROR] Slime type datum, and cleans up slime related code [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79852
--------------------
## About The Pull Request

I have decided to help out with basic mobbing slimes, and found out, that this code is ancient and crusty. So I have decided to refactor and tweak things bit but but, until the moment I will hit AI related stuff.

I might have gone a bit overboard, and performed a melange of fixes. I have done the following in order:

- Slime colour, core type and mutation list are now held by a single `slime_type datum`. This means the slime's core is no longer spawned using text operations on the slime's colour, and the slime's mutation is no longer a single large switch statement that returns an exactly four element length array. Icons are still based on the slime's colour define, but that is now more changeable.
- Autodocced all the slime procs and vars I could find, while also renamed them to be more descriptive. Also made Booleans actual booleans. Also added code to confused code segments. For example, did you know that slimes would forgive all its stacks of Feeling Very Attack Right Now, if it has exactly one level of Discipline? I didn't, and I almost turned the 1 in the check into a TRUE before I had a second look.
- Added defines to the slime hunger checks
- Cleaned up a lot of single letter vars in all code that referenced the now changed variables.
- Large amount of if-chain pyramids have been turned into early returns. More readable.
- Made the xenobiology camera shortcuts and actions always do the same thing, though the action buttons still loop through all of the conents. In addition, after seeing a comment lamenting about varied load orders, I have made autolinking the monkey recycler is now done in LateInitialize, which didn't exist when it was made.
- I have included an UpdatePath that changes the colours of premapped slimes, but I did not do anything for the other values.
- I noticed slimes were unable to attack pacifist humans, due to trait checking the human instead of the attacking slime. This has been fixed.
- The slime scanner has an examine block, and it now describes your slimes' "mutation attempt chance", instead of like, performing math in case the slime's current colour is one of the mutations, or hiding it if it has none. This value is needed to breed rainbow slimes.

This has been an extremely large scrubbing. I am willing to try rip it up into smaller chunks, even if most of it is just single letter vars and if pyramids.

## Why It's Good For The Game

If code is more readable and expandable, it will aid in further refactors.

## Changelog

:cl:
refactor: Slimes's colour, core type and mutation list is now held in a slime type datum
code: Slime's variables have been documented, and renamed a bit to add clarity. Please report bugs that might stem from renaming.
fix: Slimes are not longer prevented from attacking pacifist humans.
qol: Slime scans now display the actual amount of genetic instability, instead of hiding it if a slime doesn't mutate further, or tweaking it if it might mutate back into itself. This will make it easier to parse which slime to breed further to get a rainbow slime.
/:cl:


code/game/objects/items/devices/scanners/slime_scanner.dm
```diff
@@ -13,41 +13,44 @@
 	throw_range = 7
 	custom_materials = list(/datum/material/iron=SMALL_MATERIAL_AMOUNT*0.30, /datum/material/glass=SMALL_MATERIAL_AMOUNT * 0.20)
 
-/obj/item/slime_scanner/attack(mob/living/M, mob/living/user)
+/obj/item/slime_scanner/attack(mob/living/living_mob, mob/living/user)
 	if(user.stat || !user.can_read(src)) //SKYRAT EDIT: Blind People Can Analyze Again
```
vinylspiders:
```suggestion
	if(user.stat || !user.can_read(src)) //SKYRAT EDIT CHANGE - Blind People Can Analyze Again - ORIGINAL : if(user.stat || !user.can_read(src) || user.is_blind())
```

vinylspiders (dismissed):

SomeRandomOwl (approved):

PR #25257, opened 2023-11-26
[MIRROR] Bumps compile to 515 [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79134
--------------------

## About The Pull Request

LSP supports it, let's GOOOOOO
I've removed the 515 tests since they're stable, alongside the libcall wrapper. left the rustgcall wrapper cause yaknow memes
Just removed all the 515 and 514 particular define wrappers. gaming

## Changelog
:cl:
server: Minimum compile version has been bumped to 515. clients still support 514 but we're gonna start using 515 restricted features for serverside now.
/:cl:

PR #25274, opened 2023-11-27
[MIRROR] Micros bucketJoin with operator"" [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79949
--------------------

## About The Pull Request

We make timers a lot. Making a unique string for each of them wastes time (string churn).
It also means we have to do an extra ref() in the bucketJoin proc.

If we instead throw all the shit we care about in a list and just read off it later, we get pretty decent savings ((0.013 | 0.022) -> (0.009, 0.012)) (It was way worse when ref() was hyper expensive) It's not much, but since timers are hot I think it's worthwhile.

It also lets us add further debug information, if we want it.
Could optimize this further if we had less stuff in the list, depends on what we want displayed as it was on insertion and what we want displayed as it was at moment of print.

Also also this is 100% the reason I did 515 in the first place and I need to be free

## Why It's Good For The Game

Uhhhhhhhh more flexability in timer readouts? Cost I was worried about is mostly gone cause ref() got better I think


GoldenAlpharex:
This is going to have to wait until https://github.com/Skyrat-SS13/Skyrat-tg/pull/25257 is merged.

vinylspiders (approved):

PR #25295, opened 2023-11-28
[MODULAR] Fixed hypospray wait times

misterghast:
Fixed hypospray wait times so they actually used the constants other than "WAIT_INJECT"  (and their counterparts based on variant). "WAIT_SPRAY", "DELUXE_WAIT_SPRAY" "SELF_SPRAY", and "DELUXE_SELF_SPRAY" are now used for hypos' spray functions, which not only makes the CMO's hypospray actually have an instant use option like it should, but also makes them more tweakable overall.  "SELF_INJECT" is now also used for a hypo's inject function when used on yourself.

## About The Pull Request

Added a new "selected_wait_time" variable in the afterattack method that is dynamically defined based on the hypospray's current use type and then used for the length of the doafter timer.

## Why It's Good For The Game

I like hyposprays and just want them to be good. Also it annoys me.

## Changelog

:cl: Ghasterino
fix: Made the hypospray actually use the "WAIT_" values for the respective mode it's using.
fix: Made the CMO hypospray actually instant, instead of seeming instant in code but not actually instant in game.
/:cl:


modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
```diff
@@ -218,7 +218,20 @@
...
+	if(mode == HYPO_INJECT)
+		if(target == user)
+			selected_wait_time = inject_self
+		else
+			selected_wait_time = inject_wait
+	else
+		if(target == user)
+			selected_wait_time = spray_self
+		else
+			selected_wait_time = spray_wait
```
vinylspiders:
```suggestion
	var/selected_wait_time
	if(target == user)
		selected_wait_time = (mode == HYPO_INJECT) ? inject_self : spray_self
	else
		selected_wait_time = (mode == HYPO_INJECT) ? inject_wait : spray_wait
```
Just looks a little nicer, I think

vinylspiders (changes requested):

modular_skyrat/modules/hyposprays/code/hyposprays_II.dm
misterghast: I genuinely didn’t know that kind of syntactic sugar was in DM. Thanks for the improvement!

vinylspiders (approved):

Issue #22980, opened 2023-08-10
Synth eyes are not in the eye area.


trefoils:
Do eye surgery on a synth go to remove the eyes oh the eyes are not in the eye area.. they are in the head slot, but to insert new eyes you need to put them in through targeting the eyes. This needs to be fixed and streamlined again. On non synth races you can remove the eyes by targeting the eyes for surgery. Having to do two surgeries to replace synths eyes is needless and clearly not intended something got borked somewhere.

PR #25308, opened 2023-11-28
The synth brain surgery now targets the chest and only applies for synths

Majkl-J:
## About The Pull Request

It is a synth positronic brain surgery. The synth brain is defined with `zone = BODY_ZONE_CHEST` yet the surgery targets the head. It also doesn't even check for only synths, as the name might imply.

Well now it does

The surgery target should be on the place where the organ is, to avoid confusion between one and the other.

## Proof of Testing

<details>
 Here it is appearing on a synth
image

Here it is not appearing on john doe
image

</details>

## Changelog

:cl:
qol: Synth brain surgery now actually targets the synth brain location
/:cl:


vinylspiders:
There's also https://github.com/Skyrat-SS13/Skyrat-tg/issues/22980 that you could fix while you're at it, if you'd like (you can open another pr to get fix gbp for both)

vinylspiders (approved):

PR #25332, opened 2023-11-29
[MIRROR] Removes context from tgui [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/80003
--------------------

## About The Pull Request
*Most context
This is a precursor PR for moving to React. Newer React (>16) does not have the `context` parameter, so it needs removed. I want to make converting to React look slightly less intimidating, so this is just short-circuiting context here. 

This shouldn't be considered a long term solution.
## Why It's Good For The Game
Moving to react is good actually
## Changelog
N/A hopefully no noticeable change


SomeRandomOwl:
#25094

tgui/packages/tgui/interfaces/NtosCrewManifest.jsx
```diff
@@ -3,14 +3,14 @@ import { useBackend } from '../backend';
 import { Button, Section, Table } from '../components';
 import { NtosWindow } from '../layouts';
 
+export const NtosCrewManifest = (props) => {
```
vinylspiders: This got accidentally moved up I think? It used to be below the skyrat edit

vinylspiders (changes requested):

vinylspiders (approved):

PR #25371, opened 2023-12-01
[MIRROR] Remove /datum/game_mode, we SSdynamic now [again] [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79965
--------------------
I don't remember what was hard about this last time it took me like 20 minutes this time so I'm scared.

Removes dynamic simulations, only I have used them and it's a lot more complicated now with this. I plan on making Dynamic simulations a part of moth.fans anyway


vinylspiders:
This should come after the contractor pr

PR #25394, opened 2023-12-02
[NO GBP] i accidentally made the twitch overdose not kill you through tox that hard oops

Hatterhat:
## About The Pull Request
uuuh when i was doing some stuff with the previous PR i forgot to change this number back because i was trying to make it properly return MOB_UPDATE_HEALTH (and then gave up) please help

i think twitch ODs should probably be a little better at killing you through tox damage

## Changelog
none because it just builds off the last PR (oops)


Paxilmaniac:
What is this guy doing.

vinylspiders (approved):

PR #25407, opened 2023-12-03
[MIRROR] Falling down a z-level while standing up can (occasionally) break your legs. Felinids will now always land on their feet (for better or for worse).  [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/80054
--------------------
## About The Pull Request

- Falling down a z-level while standing up will apply the damage directly to your legs. 
   - This damage has the chance of rolling wounds, so falling z-levels -> broken legs. 
   - If you are lying down when you fall, the damage is spread through your body like normal. 

- Felinids who fall down a z-level while standing will always land on their feet. 
   - Instead of being floored for 5 seconds / levels fallen, you will receive a slowdown for 2 seconds / levels fallen. 
   - However, while they have cat reflexes, they do not have cat legs. Landing on your feet will deal additional damage. 
      - For falling 1 z-level, this is ~18 damage (instead of ~12).
      - For 2 z-levels, this is ~50 damage (instead of ~32). 

- Cats (basic mobs) are capable of falling down z-levels without sustaining damage. 

image

## Why It's Good For The Game

Add some more depth to falling down. (No pun intended)

Rather than dropping down a level and walking it off after getting up, you might dislocate your leg. 

Also adds some more species interactions. Felines might be able to give people the slip easier when dropping down into the undertram, thanks to their cat-ness. 

## Changelog

:cl: Melbert
add: Falling down a z-level while standing now damages your legs rather than your entire body. This also means falling down multiple z-s may rarely break your legs. 
add: Felinids now land on their feet if they fall down a z-level while standing. This replaces the knockdown with a short slowdown, and also has the trade-off of causing more leg damage. 
add: Cats can now fall z-levels without sustaining damage. 
/:cl:


vinylspiders (approved):

comrade-dagmar:
tg taking away the one good thing felinids have, not happy

Syalea:
Ah yes, lets take the one good thing for felinids away, They can't even lick their wounds to stop bleeding ( no one knew how to do it). Now they receive damage if they fall down a level. why don't we buff synthetic people or podpersons even more?

vinylspiders:
> Ah yes, lets take the one good thing for felinids away, They can't even lick their wounds to stop bleeding ( no one knew how to do it). Now they receive damage if they fall down a level. why don't we buff synthetic people or podpersons even more?

Please check the code itself or test in game before you get upset!! While the TG description says otherwise, felinids do in fact still take NO DAMAGE from falling down z-levels.

I made sure to port over the no-fall-damage for tajaran and felinids that we already had in our overrides to the new code.

MrMelbert:
lol

comrade-dagmar:
ah much better, forgive me- I am just a serial changelog reader. Im very thankful for both your work, I jumped the gun and read the changelog text too literally from the original. <3 My apologies...

PR #25408, opened 2023-12-03
Clean soups

xXPawnStarrXx:
## About The Pull Request
Ya boi's back at it again with his anti-slurry-soup propaganda. **_This time, It's modular._**

The idea of this PR is to make all of the items that would be emulsified into the soups, create extra soup. So the bigger and better quality of plants that you use, the more soup you get! Currently due to my limited skill coding, I'm not sure how to get it to select only specific items so it will convert everything to soup.

EDIT: Adjusted the blood soup's reaction to require saltwater, instead of normal water to prevent it infinitely looping. (Saline seems more reasonable for making extra blood anyway.)

EDIT 2: Thanks to the incredible work of @vinylspiders this PR is now fully functional and entirely optional to use in game, If you want your slurry. You can have it, or you can toggle it off with ALT+RMB.

image

To those who will have an issue with this and the lack of ability to poison, bowls are reagent containers, it's not hard to just add a little bit to the bowl when no-one's looking.

Clean soups for chefs, so they can serve soups without the paranoia of more "Unknown reagent" than soup in a bowl.

## Proof of Testing

<details>
  I've made soups go from this:
  
image

  To this:
  
image

What can be achieved with incredible quality reagents.
image

New toggleable options!

image

</details>

## Changelog

:cl:
qol: made soups clean, for chefs.
/:cl:


Stalkeros2:
mmm soup :)

xXPawnStarrXx:
I have tested every soup I can find the recipe for. So far, it's resulted in the outcome that I have had to edit one recipe, that being blood soup (which shouldn't even be a recipe, grinding the damn tomatoes results in more blood than the actual recipe, putting that in a bowl has the exact same outcome...)

code/modules/food_and_drinks/recipes/soup_mixtures.dm
```diff
@@ -710,7 +710,7 @@
 
 /datum/chemical_reaction/food/soup/bloodsoup
 	required_reagents = list(
-		/datum/reagent/water = 10,
+		/datum/reagent/water/salt = 10, // SKYRAT EDIT
```
vinylspiders:
```suggestion
		/datum/reagent/water/salt = 10, // SKYRAT EDIT CHANGE - ORIGINAL: /datum/reagent/water = 10,
```
Would be nice to also put in the comment the reason behind it

modular_skyrat/modules/supersoups/code/soup_mixtures.dm
```diff
@@ -0,0 +1,10 @@
+/datum/chemical_reaction/food/soup/clean_up(datum/reagents/holder, datum/equilibrium/reaction, react_vol)
+    . = ..()
+    var/souptype = results[1]
```
vinylspiders:
```suggestion
    if(!length(results))
        return
    var/souptype = results[1]
```
I don't know if there will ever be a case of a soup with no results, but you should guard against it regardless.

vinylspiders (changes requested):
This is better than before, but instead of making this an automatic reaction I would like this to be something that you have to manually trigger.

code/modules/food_and_drinks/recipes/soup_mixtures.dm
xXPawnStarrXx: Oh, I forgot to comment. It causes an infinitely looping reaction that never ends, I'm not entirely sure what caused it.

xXPawnStarrXx:
> This is better than before, but instead of making this an automatic reaction I would like this to be something that you have to manually trigger.

I have plans to come back at this later. I wanted to get the groundwork functional and available to those who'd actually like to cook soups first, I don't understand much BYOND code and it'd likely be a while before I worked out how to do so on a trigger as well as adding a conversion list, instead of converting ALL to soup.

vinylspiders:
> > This is better than before, but instead of making this an automatic reaction I would like this to be something that you have to manually trigger.
> 
> I have plans to come back at this later. I wanted to get the groundwork functional and available to those who'd actually like to cook soups first, I don't understand much BYOND code and it'd likely be a while before I worked out how to do so on a trigger as well as adding a conversion list, instead of converting ALL to soup.

I understand, though my only issue with that is this is that it makes the old soup features essentially unavailable to people who might have wanted to use that. It goes from forcing the reactions to go one way->forcing them to go another way.

Ideally you give the players full control. I think something as simple as a toggle on the soup pot itself could be fine for this.

xXPawnStarrXx:
> I understand, though my only issue with that is this is that it makes the old soup features essentially unavailable to people who might have wanted to use that. It goes from forcing the reactions to go one way->forcing them to go another way.
> 
> Ideally you give the players full control. I think something as simple as a toggle on the soup pot itself could be fine for this.

While I agree with this, soups don't see much use, if any at all right now. I'm not sure how I'd even start with adding a toggle with my current coding ability. I just wanted such a neat mechanic actually be used, rather than sat effectively decaying.

modular_skyrat/modules/supersoups/code/soup_mixtures.dm
```diff
@@ -0,0 +1,12 @@
+/datum/chemical_reaction/food/soup/clean_up(datum/reagents/holder, datum/equilibrium/reaction, react_vol)
+    . = ..()
+    if(!length(results))
```
vinylspiders:
```suggestion
	if(!length(results))
		return
	var/obj/item/reagent_containers/cup/soup_pot/our_pot = holder.my_atom
	if(!istype(our_pot) || !our_pot.emulsify_reagents)
		return
```
Something like this, then just add a modularly overidden `var/emulsify_reagents` to the soup pot atom. And you can toggle it by alt clicking or however you want.

Should not take long to add, if you need any more help implementing the toggle let me know. It will be easier if you make the override file for the soup pot obj so I can directly suggest code to that.

vinylspiders (changes requested):

modular_skyrat/modules/supersoups/code/soup_mixtures.dm
xXPawnStarrXx: Could I make "var/emulsify_reagents" a list? Because I've written a list but haven't worked out how to impliment it. I'm tired as hell right now, so I'm not going to be touching much until the morning.

vinylspiders: I would keep that as a boolean, to actually let you toggle whether you want it to let the cleaning happen _at all_.

From there if you want an additional list of soups that can/can't be cleaned (or whatever you are trying to do there) yeah sure that'd work. You'd want to make that a var/static/list or a global though so you don't have redundant copies.

xXPawnStarrXx: Yeah. I'm probably going to need a lot of help with this later... I'm more than inept at BYOND code. I'm just taking what I've seen elsewhere and jerry-rigging it together into something cohesive.

vinylspiders: No worries--just start with simply making a master_files override for the `/obj/item/reagent_containers/cup/soup_pot` and add that new var, and I can help out a bit.

xXPawnStarrXx: I would also have to find out a key combination suitable. Since altclick won't work.
I don't even know how to assign a key to things.

vinylspiders: Is alt already being used by something? There's also the less-used `alt_click_secondary()` (alt-right-click). You could make it an action button, or make it toggled by attacking it with an item...lot of possibilities.

xXPawnStarrXx: Yeah, alt click let's you set volume transfer.

xXPawnStarrXx: > Is alt already being used by something? There's also the less-used `alt_click_secondary()` (alt-right-click). You could make it an action button, or make it toggled by attacking it with an item...lot of possibilities.

I'm currently working on that fix as I couldn't sleep because it's irking me. How does one apply something like this. If it's easier for you, my discord is the same as my username here

MrMelbert:
> Ideally you give the players full control. I think something as simple as a toggle on the soup pot itself could be fine for this.

Give chefs a wooden soup spoon that lets them meld the ingredients together by clicking on the pot (after a do_after)

xXPawnStarrXx:
These changes have broken everything and I'm getting errors from places I've not even touched. I'm going to sleep before I get stressed and just erase the whole project.

vinylspiders:
> These changes have broken everything and I'm getting errors from places I've not even touched. I'm going to sleep before I get stressed and just erase the whole project.

You had the wrong kind of spacing in your file. I have fixed it and written a baseline that you can work from should you want to alter it to use a spoon to mix rather than the alt-right-click.

xXPawnStarrXx:
> > These changes have broken everything and I'm getting errors from places I've not even touched. I'm going to sleep before I get stressed and just erase the whole project.
> 
> You had the wrong kind of spacing in your file. I have fixed it and written a baseline that you can work from should you want to alter it to use a spoon to mix rather than the alt-right-click.

Alt+right-click works for me, means we keep it all just to the pot.

vinylspiders (approved):
Well, I think it can be improved (I would prefer it to instantly do the 'cleaning' reaction when stirred or toggled, which right now it does not, not until you add reagents anyway) but if you are not up to that task this is fine in its current state.

PR #25418, opened 2023-12-03
Balance fans love her | Adds one-hit shield bracers to both cargo and the loadout

Paxilmaniac:
## About The Pull Request

image

I've been floating the idea of shield generators for a while now, there was supposed to be a belt that goes with this that's a different type of shield, but the code for it sucks and didn't work, so we get just the gloves for now.

The Milano shielded bracers, as they are called, are gloves that give you a one-hit personal shields that recharge after a minute.

The one thing I hear a lot people be afraid of when it comes to fighting is the classic "1 hit and you have probably lost" stuff we have around, like pies, bolas, what have you. The best solution I came up with for this is simply an item that lets you alleviate your one-hit fears but blocking the first two hits. People can be less afraid to initiate combat if they know they won't instantly die or be stunned before they get the chance to react.

And its because of that, that I made it not only available through cargo, but also able to be spawned with in the loadout in your gloves slot. Not being afraid of instantly losing combat is something everyone should have access to if they want.

## Proof of Testing

<details>

https://github.com/Skyrat-SS13/Skyrat-tg/assets/82386923/c6f9dbe2-160d-4754-a35d-d8b30c445cea
image
image
  
</details>

## Changelog

:cl:
add: A new pair of gloves has been added to imports under microstar, or under the gloves section of the loadout. These gloves are shield bracers, and will give you a two-hit personal shield against any attack.
image: Sprites for the new shield bracers done by meee :333
/:cl:


modular_skyrat/modules/shield_belts/code/cooler_shielded_component.dm
```diff
@@ -0,0 +1,63 @@
...
+	var/obj/effect/temp_visual/shield_pulse_effect = new /obj/effect/temp_visual/shield_pulse(get_turf(wearer))
+	shield_pulse_effect.icon = shield_icon_file
+	shield_pulse_effect.icon_state = shield_overlay_icon_state
+	shield_pulse_effect.color = shield_overlay_color
+	shield_pulse_effect.alpha = shield_overlay_alpha
+
+/obj/effect/temp_visual/shield_pulse
+	duration = 0.5 SECONDS
+
+/obj/effect/temp_visual/decoy/twitch_afterimage/Initialize(mapload)
```
Hatterhat: wrong path looks like

Paxilmaniac: its so over

ChickyRooster:
Pax out here making the best PRs. I wish I learned this power.

Speaking of, I was talking with someone about something like this as a MODsuit module, that might be an idea too, instead of gloves? It would be a toggle that consumes power.

However, local balance fan here;

The half a minute cooldown concerns me, as this could mean someone using this could attack a security officer with it, tank two stun baton hits, and rush away for 30 seconds before going in for another attack, maybe up the cooldown to a minute? Especially with this being available in the loadout I can imagine it becoming a MUST HAVE item, tanking 2 full hits of any damage is incredibly strong, ESPECIALLY every 30 seconds.

Paxilmaniac:
> tank two stun baton hits, and rush away for 30 seconds before going in for another attack

Every secoff has a disabler if someone is hit and running you and you can't disabler enough to stop them getting away, then you weren't going to win regardless of if they had a shield or not.

SpringSkipper:
This is an awful idea.
You've balanced your machine guns/rapid-fire disablers on the idea of having the same approximate DPS as the slow-firing versions. Now you're throwing **your own balance choices** out the window since rapid-fire weapons would be vastly preferable for breaking shields.
This issue is dramatically more pronounced with weapons like bolt-action rifles or melee attacks (which have on average twice the cooldown of ranged weapons), removing the elements of skill in using these weapons and landing hits in exchange for more spray-and-pray.

one-tap stuns and knockdowns are frustrating and poor game design with the way this server handles combat, I sympathize with that. But I'd much sooner see the particular offending items nerfed than to slap this bandaid over top everything in a way that fixes the issue by leaving the majority of combat methods in the game shattered and broken.

Plum-but-gayer:
this rules put mantis in loadout next

Namwty:
Will be real, cooldown should be moderately higher ( it's too short, wayyy to short ) + EMP effect should be a thing.

But only reason why this isn't tooooo bad in terms of balance, cause gripper gloves ( and other tackling gloves ) exist, which are essentially insta win any battle, cause you can't exactly "fail" a tackle now, unlike past. Then you have other gloves which grant you stuff like shock immunity and armor ( bracers etc ). Can't also use this in modsuit, which is fan favorite.

So for sec mains for example it's this ( which is just... 1/2 bullet protection against 24 round smg ) vs insta win tackling gloves.... which arguably is hard choice.

**Now I will be real, I'm not a fan of it being in a loadout, this should be a cargo item that is buyable at reasonable price ( something that's like armor )**

**For example, riot shields aren't buyable exactly to public, but these are basically better version of riot shields.... which don't require any sec access/"legal breaking" or any money spent.**

Imo shields shouldn't be inferior to this, but dunno.

Also making this a "everyone should select this" type of choice, is bad... cause well, people who won't take it for cosmetic reasons will beat themselves up and feel like shit. So mayhaps cargo is better option.


But at the same time, I'm not too sure if this is even good thing to add, blocking is really strong ( See me blocking cannonball with it ( happened with hallucination armor, altought not sure if it's same, code wise ) )


But this all is my opinion, as I generally think even grippling gloves will trump this to hell... like not even close. People sleep on those gloves, but yeah... I would be carefull by adding too many of these defensive items ( especially considering we got new armor + CIN is planned a buff.... might be too much too fast )

EDIT: But yeah, also maybe it's best to fight "1 hit combat" by dealing with core of issue, not by adding these items ( even if it's sick in idea )
But at the same time, I wouldn't mind this, cause combat is really... odd right now, as in you can win first within seconds with barely any sweat.... that honestly feels soo rough that you just get urge to "fail" and let that nightmare walk free for extra time.

EDIT2: Maybe add that trait which acts as if you have "chunky fingers/big hands" to it, like insulated gloves have, just an idea to make it less powerfull.

SpringSkipper:
> EDIT: But yeah, also maybe it's best to fight "1 hit combat" by dealing with core of issue, not by adding these items ( even if it's sick in idea ) But at the same time, I wouldn't mind this, cause combat is really... odd right now, as in you can win first within seconds with barely any sweat.... that honestly feels soo rough that you just get urge to "fail" and let that nightmare walk free for extra time.

I mean, also worth noting that this is a nerf to all antags (unless they also spawn with them), detectives (everyone is wearing the same gloves now), and one-armed characters.

Since overly-easy stuns and knockdowns are the actual problem, it might be better to expand on stamina mechanics? Maybe the 'resist' button lets you literally resist out of stun/knockdown with a stamina cost, or reduces stun/knockdown time based on how high your stamina bar is?
"_(Name) rolls to their feet with a surge of adrenaline!_" or something along those lines.

nikothedude:
this should NOT be in the loadout for the same reason guns arent in, knives arent in, armor isnt in, unless youve actually conferred with someone higher up and determined why yes we want this to be a feature EVERYONE ALWAYS has access to, you should NOT be giving it to people in the loadout

HoodMilk:
Does this work with all damage, or just damage from other players? Being able to use these fighting megafauna or facetanking bombs is kinda wacky.

SpringSkipper:
> Does this work with all damage, or just damage from other players? Being able to use these fighting megafauna or facetanking bombs is kinda wacky.

As best I understand it works against all _attacks_, so it wouldn't protect from bombs or lava but would protect against fauna and make megafauna fights dramatically easier.

Stalkeros2:
what do you mean 'her'
does that mean you're actually british
:(

chemistrymain2:
I do dislike this one on the basis that it makes rapidfire weapons better than slowfire weapons.

burgerenergy:
While I dislike this on principle for the stated reasons, this shouldn't be in the loadout regardless. The item being _guaranteed_ to be on someone's person (if they choose to) is vastly different from them playing the minor effort tax to get a weapon from cargo, or by whatever other means, before they do mischief. It encourages people who plan, or at least prepare, as opposed to just it being a mechanically pro move to always have outside of engineering insulated gloves.

Ephemeralis:
Balancejak concerns about loadout are a little overblown. Loadout causes someone to start with these things at roundstart. The alternative is making them a cargo purchase, which at CREW*3 is what, like... 150 credits? Most roles start with more credits in their ID than this. The difference in the functional time-to-acquisition of these gloves is maybe five or ten minutes at best. Really not a huge deal.

If people want to meta loadout fixtures for maximum robustness, they're only depriving themselves of interactions with the round and other players. The answer to "why can't I if the option is there" is *roleplay*.

Increased resilience in combat produces longer and more interesting altercations in rounds, which is clearly what Paxil has been going for between this and the armorcargo PRs. I think this'd definitely be worth a testmerge to get the 7 days of "oo shiny" out the way and see how things settle.

StrangeWeirdKitten:
Why not just put it inside the techweb?

ReturnToZender:
I decently enjoy the _idea_ of the changes, however I feel there is a key system that is being overlooked;
The idea of a singular, free hit compared to normal TG is clearly the intention of Skyrat changes like adding 135 health, or making stun batons require 3 (4? On Skyrat? Forgive me, I haven't played much Skyrat) hits instead of 2. Longer combat and more opportunities to escape lead to more engaging interaction, if all the corpses pile up in the first five minutes then no one has fun. That said, I feel the additional health and strikes are more than sufficient, although I personally would hope that they would be more complete and nullify things like wounds on the first strike.

With that said, this is redundant, and while I enjoy the extra sprites, I personally feel like the implementation of the "one free hit" concept would be better off being done on the mob rather than through a loadout option of a magic shield.
I feel like you want to add things that are "cool" or "neat" and the fact that there are other, dynamic systems in the game get in the way of that, by making you have to consider how a system like this will affect everyone (who, as mentioned above can purchase this) and when people disagree with your careless changes, you call them "balancejaks" and similar. I don't exactly know what this is supposed to mean, but it makes me sad that you'd refer to people giving valid criticism to your idea in such a manner.

I agree with the intent behind your changes, though I feel the existing system is sufficient, but I feel the way you handle criticism and attempt to solve these issues is rather poor, and could certainly be done better in the future. Kindness is few and far between on SS13.

Love,
    Moffington

burgerenergy:
> Balancejak concerns about loadout are a little overblown. Loadout causes someone to start with these things at roundstart. The alternative is making them a cargo purchase, which at CREW*3 is what, like... 150 credits? Most roles start with more credits in their ID than this. The difference in the functional time-to-acquisition of these gloves is maybe five or ten minutes at best. Really not a huge deal.
> 
> If people want to meta loadout fixtures for maximum robustness, they're only depriving themselves of interactions with the round and other players. The answer to "why can't I if the option is there" is _roleplay_.
> 
> Increased resilience in combat produces longer and more interesting altercations in rounds, which is clearly what Paxil has been going for between this and the armorcargo PRs. I think this'd definitely be worth a testmerge to get the 7 days of "oo shiny" out the way and see how things settle.

Someone having to go to cargo instead of it being guaranteed isn't much of a hurdle, but it's a hurdle that separates between "I will **always** have this" versus "hm maybe I'll do some fighting, let's go get the no downsides gloves." The latter means you might be taken by surprise, and that there is some form of intent.

There is no roleplay in waddling to cargo, making an order, and picking up the parcel bag off the floor. Nothing is happening there. There just is intent and a minor tax to go and _do_ it if you want it, which matters when you're doing round set up for non-assistant jobs before falling into other stuff.


Loadout aside, though, I am also against this entirely now, as it dumpsters melee weapons even more. Melee weapons are already painfully weak in Skyrat with how ubiquitous firearms are, and most of the weapons you, Paxil, added all have magazines 7+ rounds even at the smallest, so they don't really care about this. Whereas anyone trying to bonk someone with a baton, or a e-sword, or whatever have you has to get in several dud hits while they're fighting, dud hits that have a realistic chance of getting regenerated by kiting given everyone runs around at more or less parity speed.  And then this also dumpsters slower, powerful weapons like the MCR lance, any of the snipers, shotguns, etc. Combat doesn't need to be slow. At the end of the day spacemans is WASD and M1 in a broom closet sized screen space, with people always being able to be revived and healing taking less then thirty seconds. It doesn't need to be drawn out by and item that will make any fights extremely lopsided if one side doesn't have it, while also being more effective then the actual shields that take hand space and value by way of being guaranteed.

SpringSkipper:
> Loadout aside, though, I am also against this entirely now, as it dumpsters melee weapons even more.

They didn't mention it in the pull request details, but they edited for balance by reducing it to only one shield charge and also giving the gloves chunky fingers, which means you can't use most guns with them.

PR #25460, opened 2023-12-06
[Non-Modular] Allows the use of All Nighter and Heavy Sleeper combined

desertmote:
## About The Pull Request

Removes the quirk incompatibilities of the all nighter and heavy sleeper quirk.

I couldn't find out why these two were incompatible together. Being able to sleep for a longer time while benefiting from it isn't unfair or detrimental to anything, it just means you're clicking the sleep verb less often.

## Proof of Testing

<details>
  FIFTY SECONDS? 😲

Code_EY7YT85wJi

</details>

## Changelog

:cl:
qol: You can now pick both Heavy Sleeper and All Nighter quirks together so you can enjoy sleeping just a bit longer.
/:cl:


vinylspiders:
I made them incompatible because All Nighter gives the same exact trait as Heavy Sleeper (`TRAIT_HEAVY_SLEEPER`) but with the added features (undereye bags etc).

They do not stack and are not multiplicative. So you are just getting free quirk points by putting them together. It's gonna stay blacklisted for that reason.

vinylspiders:
If you think they should be separated and allowed together in some form you'd want to make a unique trait for heavy sleeper + add trait checks for that wherever `TRAIT_HEAVY_SLEEPER` exists. But until someone does there's no reason to have them together.

PR #25461, opened 2023-12-06
Void Raptor pharmacy tweaks

Lutowski:
## About The Pull Request

This PR adjusts the pharmacy on Void Raptor to have more walking space, and more comfortable to do chemistry in overall. Slightly adjusted round-start placement of chem lab tables and rack, with chem vendor moved there for security. Racks in chem storage replaced with shelves because bottles on shelves feels more aesthetically appropriate. Glass shutters on desks are external instead of internal. Buttons close shutters seperately now. 

South part of pharmacy was stretched out by one tile into the hallway. Queue space is one tile wide instead of two now, and the bench and plants facing the hallway remains same. Hallway width remains same!

The table placement clogging up the entrance made the pharmacy cramped and stifling to move around in, especially on high-pop when you have doctors running in and out constantly which is overstimulation hell. The queue facing east desk had too much space than deserved, so one tile width is absorbed. The hopeful intent is that with a spacious pharmacy, it can be treated as a proper roleplay space to teach and hang out in.

South part of pharmacy was stretched out to make this happen because the queue space facing east desk is hardly used, and if it was, it's usually by one person. It hasn't been practical live, and it was a little more room the pharmacy could've had.

Chem vendor moved to lab because that's obviously the chemist's vendor, while railings and fake stair were added to uplift the aesthetic.

## Proof of Testing

pharm1
image
pharm2
## Changelog

:cl:
add: Void Raptor pharmacy and lab is slightly more spacious and prettier.
qol: Chem vendor moved from pharmacy to chemistry lab on Void Raptor.
/:cl:


Thlumyn:
I love the aesthetic of it! Fake stairs are such a nice look there. 

The blue trim that was in the hall sort of wraps to the side to nowhere now - maybe that needs to be moved up one tile?
image

Another comment - it looks like when the disposal was moved there is still this section here leading to nowhere: 
image

Looks like the pharmacy junction goes to that section too (I deleted the pipes on top of it for clarity), might want to move that junction to line up with the new disposal chute
image

Thlumyn:
Might also want to move these fire shutters up one so that they're in between walls 
image

Lutowski:
> Might also want to move these fire shutters up one so that they're in between walls

Hi, thank you for the feedback. I fixed just about everything you pointed out and pushed the newest commit. While I was at it, I moved the chem storage shelves further in because having them out in the middle was weird. I also added two tube racks because it's an underused equipment and what's really a chem storage without one?

I also moved the ChemMasters to be next to the dispenser instead of across so it's less mouse travel.

storage

I moved the pharmacy junction pipe to this red circle, tested the new disposal bin and it sends trash to cargo as it should. Is this spot fine? If not, I'd love suggestions since I'm not that super experienced with disposal pipes.
junctionpipe

Thlumyn:
Don't forget to put the mail sorting mapping helper for chemistry on that junction so that it knows which packages to sort (if you are using the destination tagger in cargo) 

image

Lutowski:
> Don't forget to put the mail sorting mapping helper for chemistry on that junction so that it knows which packages to sort (if you are using the destination tagger in cargo)

Oh gosh I nearly forgot. Thanks for pointing it out! I added, tested and it sorts mail like it should. PR is ready for review & merge.

vvx

Gandalf2k15 (approved):

PR #25469, opened 2023-12-06
Nerfs Blueshift a little

desertmote:
## About The Pull Request

Blueshift had some map inconsistencies that lead to a needlessly tougher time to "fuck shit up", compared to main maps.

Experienced gamers who expected to be able to tackle antagonistic schemes in classic ways shouldn't be left dissapointed.

## Proof of Testing

The vault no longer has two high-sec vault doors, like the other maps its first door is an unsecured yet access-restricted white glass door.
dreamseeker_NuIWqmoBgp

The AI upload has been widened, it no longer bottlenecks you inbetween two turret tiles. Why is this needed? A classic tactic to play on the turrets are non-smart foam metal. This doesn't work if all that's allowed for space is a 1 tile gap. Not many satellites or uploads have 1 tile turret gaps, and I'd like to prevent any cases of this if I can.
dreamseeker_GF3MxLactB

The armoury gets a ladder once again.
dreamseeker_wtARyakZKn

It used to be like this, plus it cuts out on the necessity of a jetpack once again. This is a good weakspot, and prevents security from "just patching up the hole" with floortiles, making any breaking attempts require an RCD to break the tiles once again.
dreamseeker_IgPH3D47Qv

The prison gets a soda vendor! Why is this important? Well all prisons should have one, there's only one or two prisons (can't tell if its down to one now that I added it to BS) because they contain welding fuel and oxygen. These are break-out tools necessary for antagonists!
dreamseeker_ogOCV48Kje

## Changelog
:cl:
balance: Blueshift's vault, armoury, AI upload and prison are a little more antag friendly!
/:cl:


W4rd3nn:
Hate all deez

jk these are good

Plum-but-gayer:
yayyy antag balance

RatFromTheJungle:
the vault and armory shouldn't have ladders to places you can't elsewise get too, that equally lead to isolated places, especially not a place bordering somewhere fairly unrelated in maints (specifically the vault ladder)

Gandalf2k15 (approved):
Good changes.

PR #25485, opened 2023-12-07
[MISSED MIRROR] Icon Autoslicing (#79659)

vinylspiders:
## ORIGINAL PR: https://github.com/tgstation/tgstation/pull/79659

Ok so you know all the dmis we have that are made to work with the smoothing system? carpets, walls, etc.

The proper way to edit those is to convert them into a png with 5 "states' it in (one for 0 connections, one for horizontal, one for vertical, one for all cardinals and one for all directions) and then modify THAT, then run it through [the cutter
tool.](https://github.com/tgstation/icon-cutter)

But none ever does that, because we explain it fucking nowhere. So instead, let's keep all those "base" files in the repo, alongside the configs they work with, and "cut" the pngs into dmis as a part of the build process.

I wrote a guide for how to interact with this system as a spriter, you can find it
[HERE](https://github.com/LemonInTheDark/tgstation/blob/slice-the-sky/icons/Cutter.md).

[Adds a icon cutter build
task](https://github.com/tgstation/tgstation/commit/52143d2e96498de92421d516e0dd3f23936f88d8)

This relies on action ninja's hypnagogic (find more [here](https://github.com/actioninja/hypnagogic)), a rust based icon cutter.
It operates inline with the file structure, searching the codebase for templates and resource files and compiling them down to dmis.

It can do way more then just bitmask stuff, but that is what we are using it for rn.

Hope is to prevent for eternity the "I'm just gonna edit each of these 255 icon states that's how this carpet was made right?" meme, and allow more expansive use of smoothing in future

[Adds a lint that ensures config files work
right](https://github.com/tgstation/tgstation/commit/21eeab9cf831c5fdac5a9b366478a9dab285c20c)

Checks to ensure they have a paired png and dmi, and also avoids issues with uncompiled changes by double checking that nothing happens before/after a cutter run

[Pulls all non smoothed states out of structures into bespoke dmis](https://github.com/tgstation/tgstation/commit/a730e0cb47fc0a622fe265bccc296cec8d3a8fea)

This is required because the cutter cannot output named icon states, only the actual cut icon

[Does something similar to
walls](https://github.com/tgstation/tgstation/commit/40780e9481103c8ee9e16538d1c2d0cdc124eeb9)

Moves reinforced walls decon stuff from their icon to a var on the type and a set of states in the reinforced_states dmi

Moves falsewalls into their own dmi, this involved some changes to gamecode to ensure falsewalls knew which dmi to use and what key. Makes falsewalls display as such in editor rather then just walls

Moves smoothrock's gibonite overlays into their own file for similar reasons

[Same thing different day
(Floors)](https://github.com/tgstation/tgstation/commit/9a3da3b69705278f39af109ac5ce86d27c2479a1)

Pulls bespoke floor icon states into their own file, splits up neon carpets into multiple files to make cutting possible

[Actually adds the cut templates and their matching png files](https://github.com/tgstation/tgstation/commit/1bd8920dc90d1ee1b934b6dadc39f2331854f5fa)

Not much to report here, outside of I changed the prefix for bamboo walls to bamboo_wall so it works with false_walls

image

None should have to manually edit cut dmis. Ever.
Also this makes adding a new smoothed thing trivial, don't even need to know what tool you're using to do it. V good v good. Sets us up nicely for wallening's well, wall of sprites.

Some structural decisions, we are essentially committing build artifacts here. That's the best way of handling it because otherwise mappers could need to run build.bat before opening a map, and that is stupid!

:cl:
refactor: (Almost) all smoothed icons can now be edited in their pre cut forms
/:cl:


vinylspiders:
Wait for https://github.com/tgstation/tgstation/pull/80175 before merging, there is a bug with false walls.

SomeRandomOwl (approved):

PR #25486, opened 2023-12-07
[Modular] Hash is no longer recursive

SomeRandomOwl:
## About The Pull Request

This fixes a bug with hash which caused it to infinity react when it was in lava, as when hash bricks are in lava it reacts and becomes dabs. But fun fact, its a chemical reaction because of the THC in the hash bricks, and guess what also has THC in it? Dabs, which resulted in the newly created DABS which has even more THC in it than hash to react and create MORE dabs, which then ALSO reacts. Which lead to a infinite recursion of dabs creating itself and duplicating over and over infinity over a very very short time frame

Credit to @Iamgoofball for helping identify this bug

No more server crashes

## Proof of Testing

<details>

dreamseeker_lRZXFND0pb
  
</details>

## Changelog

:cl: SomeRandomOwl and Iamgoofball
fix: Hash Bricks and Dabs are no longer recursive 
/:cl:

### DIFF
--- a/modular_skyrat/modules/morenarcotics/code/thc.dm
+++ b/modular_skyrat/modules/morenarcotics/code/thc.dm
@@ -46,6 +46,10 @@
 	ph = 6
 	taste_description = "skunk"
 
+/datum/reagent/drug/thc/concentrated
+	name = "Concentrated THC"
+	description = "TCH in pure concentrated form"
+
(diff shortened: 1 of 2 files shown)


GoldenAlpharex (approved):
Looks good, thanks for the fix!

vinylspiders (approved):

PR #25506, opened 2023-12-08
LOOC'ing while in crit..? Or while unconscious..? You got it!

dis-integrates-the-integration-tests:
## About The Pull Request

This PR _should_ allow you to LOOC while in soft-crit/hard-crit/unconscious. Yeah.

Imagine this: 
There's a full skeleton pirate crew, and there's a HoS

The HoS has been put into softcrit and is about to be brought onto the ship, and be unrevivable for a very long time probably, like it usually happens

The HoS does not want to be practically RR'd, since they play Skyrat like once a week, and that's fully okay.

The HoS cannot LOOC while in crit, and now has only 4 real options:
- Ahelp the pirate crew
- Salt on Discord about how they got forgotten
- Whisper something LRP to try and tell them
- Salt in DChat about how "BRO I'VE BEEN DEAD FOR 2 HOURS HOW DID THEY NOT NOTICE"

If LOOC was enabled while in crit though.. 
- LOOC "pls no rr"
and that'd most likely be the end of it. Generally, I feel like there are many more benefits to enabling crit LOOC than negatives. By enabling it, it also allows people to consent to certain things while in crit, such as Kidnapping, if the contractor happened to accidentally bomb someone to death, or to "tell your knifeplay enthusiast Yes - Dom surgeon which specific organs you really hope are missing when you wake up from the haze of masochistic pain." - Friendly Player

Additionally, I feel like the positive effects of such change will be **way** more apparent once the Opt-In system takes place.

## Proof of Testing (UPDATED!!!!)

<details>
  
 
image

image
image
image
image
image
image

</details>

## Changelog

:cl: KannaLisvern
qol: You can now LOOC while in crit/unconscious. 
/:cl:


Namwty:
While I'm not exactly fan of "LOOC: pls no RR", cause most of cases the person... well basically asks for it ( kinda like being chased by validhunter who you already sent 3 times to medbay... yet they still come back", but that's just me OVERREACTING ( maybe ), but still... like LOOC shouldn't be a shield that people can use to avoid consequences.... regardless this will have more positives than negatives.

Reality is simple, this would reduce frustrations, and prevent IC - OOC salt bleed ( or make it easier to deal with for staff members ).

Also finally which should be still mentioned, we do have... non-con ERP related stuff... so, mayhaps it's good to have crit/related stuff LOOC for just consent reasons, if folk are into that department.

SomeRandomOwl:
>like LOOC shouldn't be a shield that people can use to avoid consequences.... regardless this will have more positives than negatives.

If people attempt to do that the other party can ahelp that technically, as people shoulden't be trying to avoid consequences of their own actions.

> Reality is simple, this would reduce frustrations, and prevent IC - OOC salt bleed ( or make it easier to deal with for staff members ).

Same goes for that, if people use it to salt we can deal with it accordingly as well staff wise.

-------------------

All in all on this as a whole, i can see the benefits but the downsides above make it iffy? Looking at this from a purely staff perspective i can see ways people might abuse this, flat out. Some people tend to get heated when the action is pumping and people are in the moment when there is combat. I get it, but i see it often happen where it leads to salt.

But there are also often times where this would be useful, For example: Im crit on a table and the doctor tending me seems kinda new lost or confused? I could use it then to help guide them because i have a strong understanding of medical OOCLY, Which in my eyes would be an alright use case.

ToasterStrudes:
I will say, a better example where this would be good to have is something like.. with the new synth wounds, some roboticists don't know they're a thing, or perhaps don't know how to do the steps to fix them or how to find out how. It would be good to be able to do something like "hey! By the way you need to do X to fix this" so on or so forth. It's a good way to coach newer/inexperienced players of how to do things should they need it.

desertmote:
Hi! This edit is insufficient. In the way you've made it, it allows people to LOOC while dead! Even ghosting and then returning to your body will still allow you to LOOC.
This is a really bad idea, it will be a salt fountain. Please, for the sake of all the good medbay players, change the code to correspond to the title and body of the PR.

dis-integrates-the-integration-tests:
> Hi! This edit is insufficient. In the way you've made it, it allows people to LOOC while dead! Even ghosting and then returning to your body will still allow you to LOOC. This is a really bad idea, it will be a salt fountain. Please, for the sake of all the good medbay players, change the code to correspond to the title and body of the PR.

.... **OH FUCK.**
Shit, okay, this is cool, time to fix
Thanks for pointing it out!

dis-integrates-the-integration-tests:
Ok, _should_ not allow you to LOOC while dead anymore
I can't actually test it, because admin rank allows you to LOOC anytime anywhere as anything, but looking at the code **I'm 98% sure** it should work.

SpaceLoveSs13:
> Ok, _should_ not allow you to LOOC while dead anymore I can't actually test it, because admin rank allows you to LOOC anytime anywhere as anything, but looking at the code **I'm 98% sure** it should work.

You can always deadmin in your test server to test things out as a normal player

Floofies:
Your CI is failing but the failure is completely unrelated to your PR. Looks good to me

dis-integrates-the-integration-tests:
> You can always deadmin in your test server to test things out as a normal player

gonna try this rn

dis-integrates-the-integration-tests:
Updated proof of testing screenies

GoldenAlpharex (approved):
I'm okay with giving this a try, I think I'll want to test-merge it for a few days first though, just to see how people act with it.

My main concern is people just using it to salt when they get put into crit or whatever, and that's one of the reasons why it couldn't be used while in crit or dead, but I'm willing to see how it goes, as I do agree that it would be quite handy in many cases.

GoldenAlpharex:
Haven't heard of any issues coming from this, so I'll just go ahead and merge it.

PR #25512, opened 2023-12-09
Suppressorsweep 2023

Paxilmaniac:
## About The Pull Request

Two things:

Suppressors no longer change the w_class of guns.

The lanca and miecz can have suppressors attached now.

Suppressors don't provide too much of an advantage at all, and yet increase the size of any gun they are put on by an entire w_class, which while fine enough for pistols and whatnot, means that rifles with suppressors can't be stored literally anywhere except on your back, maybe. A pretty big tradeoff for something you don't gain much from, don't you say?

The lanca and miecz not being suppressable before was a simple matter of me being too lazy to make suppressor sprites for them before, but now they have them, woe.

## Proof of Testing

image

image

image

## Changelog

:cl:
qol: The lanca and miecz can be suppressed.
qol: Suppressors no longer increase the weight class of a gun they're attached to.
image: New sprites for lanca and miecz suppressors done by me.
/:cl:


Plum-but-gayer:
shhhhmaxxing

SomeRandomOwl:
You say you fail to compile, are you on the beta branch of byond (515) we pushed a update that requires you to be on version 515 in order to compile

Paxilmaniac:
> You say you fail to compile, are you on the beta branch of byond (515) we pushed a update that requires you to be on version 515 in order to compile

I was on the beta, I just had to pull from master and it seems fixed now

modular_skyrat/modules/modular_weapons/code/suppressor_size_change_override.dm
```diff
@@ -0,0 +1,8 @@
+/obj/item/gun/ballistic/install_suppressor(obj/item/suppressor/added_suppressor)
+	suppressed = added_suppressor
+	update_appearance()
```
vinylspiders:
```suggestion
	. = ..()
	w_class -= added_suppressor.w_class // we do not want suppressors to change the w_class of ballistic weapons
```
Might as well just piggyback on the parent proc in case they add anything upstream there. You can get rid of the added weight class from calling parent by doing the inverse

modular_skyrat/modules/modular_weapons/code/suppressor_size_change_override.dm
```diff
@@ -0,0 +1,8 @@
...
+
+/obj/item/gun/ballistic/clear_suppressor()
+	if(!can_unsuppress)
+		return
+	return ..()
```
vinylspiders, line 15:
```suggestion
	if(isitem(suppressed))
		var/obj/item/removed_suppressor = suppressed
		w_class += removed_suppressor.w_class // so we don't end up with a lower w_class than what we started with
	return ..()
```
This is kind of screwy but calling the parent is going to subtract the w_class so we need to add it here in order to preserve the no change behavior.

modular_skyrat/modules/modular_weapons/code/suppressor_size_change_override.dm
```diff
@@ -0,0 +1,8 @@
+/obj/item/gun/ballistic/install_suppressor(obj/item/suppressor/added_suppressor)
```
vinylspiders, line 2:
```suggestion
// suppressors do not change w_class
/obj/item/gun/ballistic/install_suppressor(obj/item/suppressor/added_suppressor)
```

modular_skyrat/modules/modular_weapons/code/suppressor_size_change_override.dm
```diff
@@ -0,0 +1,8 @@
...
+	suppressed = added_suppressor
+	update_appearance()
+
+/obj/item/gun/ballistic/clear_suppressor()
```
vinylspiders, line 8:
```suggestion
// suppressors do not change w_class
/obj/item/gun/ballistic/clear_suppressor()
```

vinylspiders (changes requested):
Seems fine to me otherwise

vinylspiders (approved):

PR #25516, opened 2023-12-09
[MIRROR] Attack chain refactoring: Broadening `tool_act` into `item_interact`, moving some item interactions to... `atom/item_interact` / `item/interact_with_atom` [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79968
--------------------
## About The Pull Request

Implements half of this (with some minor changes): 

image

The ultimate goal of this is to split our attack chain in two: 
- One for non-combat item interactions
  - Health analyzer scanning
  - using tools on stuff
  - surgery
  - Niche other interactions
- One for combat attacking
  - Item hit thing, item deal damage. 
  - Special effects on attack would go here.  

This PR begins this by broadining tool act into item interact. 

Item interact is a catch-all proc ran at the beginning of attack chain, before `pre_attack` and such, that handles the first part of the chain. 

This allows us to easily catch item interaction and cancel the attack part of the chain by using deliberate bitflag return values, rather than `TRUE` / `FALSE`*. 

*Because right now, `TRUE` = `cancel attack`, no matter what, which is unclear to people. 

Instead of moving as much as possible to the new proc in this PR, I started by doing some easy, obvious things. More things can be moved in the future, or technically they don't even need to move in a lot of cases.  

## Changelog

:cl: Melbert
refactor: Refactored some methods of items interacting with other objects or mobs, such as surgery and health analzyers. Report if anything seems wrong
/:cl:


vinylspiders:
Includes https://github.com/Skyrat-SS13/Skyrat-tg/pull/25517

vinylspiders (approved):

PR #25568, opened 2023-12-12
Fixes reinforced walls reverting to the TG icon files when updating

vinylspiders:
## About The Pull Request

These should really be set to a var upstream instead of hardcoded but oh well.

Bugfix

## Changelog

:cl:
fix: fixes reinforced walls getting reverted back to the tg icon
/:cl:


GoldenAlpharex (approved):
Thanks for the quick fix!

PR #25572, opened 2023-12-12
[MIRROR] Sign up for Cargorilla from the lobby [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/79776
--------------------
## About The Pull Request

If the station rolls the "Cargo Gorilla" trait, a button will now be visible on the lobby.
Clicking on this button before the round has started will add you to a list of participants, one of whom will be selected to become a gorilla when the round begins.
If nobody signs up (because they're really boring I guess) the job will instead appear on the latejoin menu.
Once someone has become the gorilla the button will disappear.

dreamseeker_ntP3OayAuV

While implementing this I noticed that an inverted check means we were never populating the "GLOB.cargo_sloth" field which means the station trait wasn't even working.

BEHIND THE SCENES
This also adds a generic "job station trait" which can be expanded in the future.
Future developers can extend this to add other "rare jobs" with relative ease.
By default I have made it so all subtypes of this trait are mutually exclusive, only one can roll at a time.

This also means that I have converted "cargo gorilla" into a job, which applies most of the code previously located in the mob's typepath or in the station trait.
The fact that it is a job means that **admins** can enable any number of gorillas to be present on the latejoin menu (but not the roundstart one, as it is not possible to add Cargo Gorilla to your occupation preferences) if they so desire.
The random beurocratic station trait, event, and traitor item (and the job console) are not able to add gorilla slots.

Because I changed "Cargo Gorilla" to a job it now no longer exists on the map until a player gains the role, and there wasn't a non-hacky way to copy the name of this round's cargo sloth. Instead I just added a small cargo gorilla name list.

## Why It's Good For The Game

Makes the presence of a fun trait more visible to players.
Means that people who aren't observing get a chance to be a monkey.
This is a framework several other people have wanted to exist for their own features.

## Changelog

:cl: Jacquerel and Fikou
qol: If the station rolls the "Cargo Gorilla" station trait. you will be able to sign up for the role from the game lobby.
qol: If nobody signs up to be the Cargo Gorilla then you can select it from the Late Join menu and arrive on the arrival shuttle.
fix: The Cargo Gorilla will actually spawn.
/:cl:


code/__DEFINES/jobs.dm
```diff
@@ -208,7 +181,16 @@
...
+#define JOB_DISPLAY_ORDER_SECURITY_MEDIC 39 //SKYRAT EDIT ADDITON
+#define JOB_DISPLAY_ORDER_CORRECTIONS_OFFICER 40 //SKYRAT EDIT ADDITON
+#define JOB_DISPLAY_ORDER_NANOTRASEN_CONSULTANT 41 //SKYRAT EDIT ADDITON
+#define JOB_DISPLAY_ORDER_BLUESHIELD 42 //SKYRAT EDIT ADDITON
+#define JOB_DISPLAY_ORDER_ORDERLY 43 //SKYRAT EDIT ADDITION
+#define JOB_DISPLAY_ORDER_SCIENCE_GUARD 44 //SKYRAT EDIT ADDITION
+#define JOB_DISPLAY_ORDER_BOUNCER 45 //SKYRAT EDIT ADDITION
+#define JOB_DISPLAY_ORDER_ENGINEER_GUARD 46 //SKYRAT EDIT ADDITION
+#define JOB_DISPLAY_ORDER_CUSTOMS_AGENT 47 //SKYRAT EDIT ADDITION
+#define JOB_DISPLAY_ORDER_EXP_CORPS 48 //SKYRAT EDIT ADDITON
```
Fikou: why dont you guys just like put 2 newlines between these and the tg ones and move these to like 101, 102, 103 etc and never have to deal with it again

vinylspiders: good point

PR #25585, opened 2023-12-13
[MODULAR] Clothing fixes #1 | Billions must holster

Ephemeralis:
## About The Pull Request

This PR does these simple things:

- Right-clicking almost any suit slot item with the shoulder holsters item in hand will attempt to fit the holster to that item, allowing it to be worn in its suit storage slot afterwards.
- It makes chaplain equipment renamable with a pen similar to forged equipment and other toys.

The holster fix is going here instead of upstream because I don't particularly want to argue with tg balancejaks about gun storage issues.

Holsters are cool. Customize your tactical drip with a personalized holster and promptly suffocate to death when the internals-storing slot you took for granted for years is occupied with your gun. Visibly and outwardly appear to be packing heat to anyone that looks at you with your badass holster.

Chaplains can now customize their vendor equipment (that can actually hold bibles) to better suit their given bit. This one has annoyed me in particular on a few occasions.

## Proof of Testing

<details>

dreamseeker_fR9MkcJPLD

dreamseeker_vrXo9p4mnW

https://github.com/Skyrat-SS13/Skyrat-tg/assets/966289/2ab95a50-52f5-4169-8caa-d92a6277fb0c

</details>

## Changelog

:cl: yooriss
add: Right clicking with shoulder holsters on any suit slot item will now attempt to fit the holster to that item, allowing it to be worn in its suit storage slot afterwards. Holsters can also be renamed and customized with a pen, too!
add: The Chaplain's vendor equipment can now be renamed and customized with a pen to better suit their themes/religion. Never go without your trusty bible again!
/:cl:


Samarai:
this is the best PR ever 

the only issue i can say is that it sorta invalidates the purpose of special holsters like the thermal pistol holster/energy holster/detective holsters (ignoring that detective holster can store ammo boxes iirc) having the ability to be placed on armor, while (in theory) leather shoulder holsters are meant to be able to be put on anything that _isn't_ armored (since armored items tend to use the armor vest list of allowed items, which purposefully excludes the basic leather shoulder holster), essentially letting you trade off armor for drip and slightly more storage for small weapons/ammo. if it would somehow be possible to make it so suit items that have armor are exempt from the initialize bit that causes them to gain the ability to attach the holster, i think that'd be the ideal solution, but really that's such a tiny balance nitpick that ultimately doesn't cause any significant issues, it just makes the harder-to-get holsters less desirable.

Ephemeralis:
> this is the best PR ever
> 
> the only issue i can say is that it sorta invalidates the purpose of special holsters like the thermal pistol holster/energy holster/detective holsters (ignoring that detective holster can store ammo boxes iirc) having the ability to be placed on armor, while (in theory) leather shoulder holsters are meant to be able to be put on anything that _isn't_ armored (since armored items tend to use the armor vest list of allowed items, which purposefully excludes the basic leather shoulder holster), essentially letting you trade off armor for drip and slightly more storage for small weapons/ammo. if it would somehow be possible to make it so suit items that have armor are exempt from the initialize bit that causes them to gain the ability to attach the holster, i think that'd be the ideal solution, but really that's such a tiny balance nitpick that ultimately doesn't cause any significant issues, it just makes the harder-to-get holsters less desirable.

The `Initialize` approach was causing unit test failures (and honestly kind of wasn't a very good approach to begin with), so I've rejigged the whole thing to use a pre_attack_secondary action that adds the holster type to an individual suit storage item's allowed list after a 5 second channel.

This should hopefully fix the failures and also potentially allow us to far more easily extend what can receive a holster at a later point if we want to go down that path.

desertmote:
This isn't a fix. This is an addition if not both an addition and a balance change. Please use the changelog appropriately.

The five second do_after is not necessary. There doesn't have to be a do_after condition, let alone a very long one. The balance concern only comes down to whether or not the belt can be worn, not how long it takes to put on, as that's just needless tedium.

Ephemeralis:
> The five second do_after is not necessary. There doesn't have to be a do_after condition, let alone a very long one. The balance concern only comes down to whether or not the belt can be worn, not how long it takes to put on, as that's just needless tedium.

I've reduced it to 1.5s since I think at least *some* delay makes it feel like you're actually doing something to the holster versus the entire feature being a clear code workaround.

PR #25604, opened 2023-12-13
Use preloaded MODlink scryers in loadout & exosuit fabs

Ephemeralis:
## About The Pull Request

Adjusts loadout and exosuit fabs to use the "loaded" MODlink scryer object, allowing them to actually be functional at roundstart/when printed with minimal setup.

The base type that was previously given via loadout had no frequency set, and the only way you could get it set was to hope robotics had a MOD crate lying around with pre-loaded scryers that have the NT frequency prebaked, then pestering them to copy the frequency across via multitool for you. 

Very annoying and almost certainly not intended. No more!

This makes MODlink scryers an actually usable loadout choice, and a viable alternative to holocalling that doesn't consume all of your bodily fluids like the NIF version does.

## Proof of Testing

<details>
  
dreamseeker_MGEMXyPdJQ

</details>

## Changelog

:cl: yooriss
fix: Loadout MODlink scryers now come with the NT frequency by default, allowing them to actually be used without requiring complicated setup. Try them out again and holocall all your friends!
/:cl:


SomeRandomOwl:
They where originally edited this way in loadout because cafe cant and shouldn't be able to call station. That should be fixed first and formost

Ephemeralis:
> They where originally edited this way in loadout because cafe cant and shouldn't be able to call station. That should be fixed first and formost

I've added a modular overload to address this:

dreamseeker_EtK73TqFKN

code/modules/research/designs/mechfabricator_designs.dm
```diff
@@ -1953,7 +1953,7 @@
 		/datum/material/glass = SMALL_MATERIAL_AMOUNT * 3,
 	)
 	construction_time = 5 SECONDS
-	build_path = /obj/item/clothing/neck/link_scryer
+	build_path = /obj/item/clothing/neck/link_scryer/loaded //SKYRAT EDIT
```
vinylspiders: Could you please make this varedit a master files override instead of a skyrat edit? With simple varedits like this it's always preferred to do it that way.

You will have to create the `modular_skyrat\master_files\code\modules\research\designsmechfabricator_designs.dm`

modular_skyrat/master_files/code/game/objects/items/devices/mod_link.dm
```diff
@@ -0,0 +1,13 @@
...
+	if(istype(user))
+		var/turf/user_turf = get_turf(user)
+		var/area/user_area = get_area(user_turf)
+
+		// if we're in a ghost-cafe, we can't call or be called.
+		// this will still work on the interlink (/area/centcom/interlink)
+		if(istype(user_area, /area/centcom/holding))
+			return FALSE
+
+	return .
```
vinylspiders:
```suggestion
```
No need for this, that will happen automatically

modular_skyrat/master_files/code/game/objects/items/devices/mod_link.dm
```diff
@@ -0,0 +1,13 @@
+/obj/item/clothing/neck/link_scryer/can_call()
+	. = ..()
+	var/mob/living/user = loc
+	if(istype(user))
+		var/turf/user_turf = get_turf(user)
+		var/area/user_area = get_area(user_turf)
```
vinylspiders:
```suggestion
		var/area/user_area = get_area(user)
```
That proc works on atoms as well as turfs so I think you can just directly pass user and it should work.

vinylspiders (changes requested):
Otherwise, looks good to me.

modular_skyrat/master_files/code/game/objects/items/devices/mod_link.dm
```diff
@@ -0,0 +1,13 @@
+/obj/item/clothing/neck/link_scryer/can_call()
+	. = ..()
+	var/mob/living/user = loc
+	if(istype(user))
```
vinylspiders: You could also make this an early return to reduce the amount of indentation in the code.

vinylspiders (changes requested):

code/modules/research/designs/mechfabricator_designs.dm
Ephemeralis: All done, and good to know!

modular_skyrat/master_files/code/game/objects/items/devices/mod_link.dm
Ephemeralis: not quite sure i follow here - you mean inverting the `istype(user)` check and just continuing on base indent underneath it? is this strictly a code format thing or

vinylspiders: Exactly--strictly code style to make it ever so slightly more readable.

modular_skyrat/master_files/code/game/objects/items/devices/mod_link.dm
```diff
@@ -0,0 +1,11 @@
...
+		var/area/user_area = get_area(user)
+
+		// if we're in a ghost-cafe, we can't call or be called.
+		// this will still work on the interlink (/area/centcom/interlink)
+		if(istype(user_area, /area/centcom/holding))
+			return FALSE
```
vinylspiders:
```suggestion
	if(!istype(user))
		return FALSE
		
	var/area/user_area = get_area(user)

	// if we're in a ghost-cafe, we can't call or be called.
	// this will still work on the interlink (/area/centcom/interlink)
	if(istype(user_area, /area/centcom/holding))
		return FALSE
```

vinylspiders: This is all I meant

vinylspiders (changes requested):

vinylspiders (approved):
Looking good!

lectronyx:
Oversight: This also messed up modlink scryers for DS-2 and Interdyne characters.

PR #25664, opened 2023-12-17
[MIRROR] Last Resort can be used while dead [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/80352
--------------------
## About The Pull Request

This PR allows Last Resort to be used while dead, in crit, or during your reviving stasis.
Additionally it increases the stun duration applied to bystanders from 2 to 4 seconds.
The confusion applied to bystanders has been increased from 3 to 12 seconds.

Additionally it fixes a runtime which was being triggered by Last Resort where a body part was being passed into the argument for `bodypart_remove` which expected to be passed a human, not a limb.

## Why It's Good For The Game

Last Resort is a rarely-used ability which makes you very vulnerable and requires you to find (or create) a corpse as a fairly fragile mob in order to return to the round (it's a last resort in other words). It also costs 20 chemicals, which means you need to have saved those to do this instead of Regenerative Coma or Fleshmend.

Because it was only usable while _alive_ (not even when in crit) it was also actually quite difficult to trigger because it requires two inputs (there's a confirmation box about whether you are sure you want to gib yourself) during a narrow window (where you know you are going to lose a fight, but haven't been knocked unconscious yet). Being usable while in crit or dead means it can now be activated during the relatively common scenario where you are dead, handcuffed, and being dragged towards a machine which will destroy your body without first needing to wait for regenerative coma to complete _and_ to regenerate an additional 20 chemicals.
This likely makes changelings potentially stickier and harder to permanently remove, to which I think: good, they're too _easy_ to deal with at the moment and seeing them get reborn due to careless corpse disposal is fun.

Additionally I have _slightly_ increased the duration of the stun and disorient applied to onlookers. Previously it applied a two second stun (and a 3 second confusion, 2 seconds of which was spent stunned), however the headslug only spawns after a one second delay. This makes its chance of escaping essentially nil, even if the corpse bursts directly on top of an air vent.
With a slightly longer stun there is some chance that it will escape, though there is still a reasonable chance that if there's more than one person around, they will manage to catch you before you crawl into a vent.

## Changelog

:cl:
balance: Last Resort can now be used while unconscious or dead.
balance: Last Resort stuns bystanders for slightly longer.
/:cl:


Gandalf2k15:
orgaN

PR #25707, opened 2023-12-18
[MIRROR] qm uses cargo tech icons [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/80411
--------------------
## About The Pull Request
qm now has the "x-in-training" intern title instead of "intern quartermaster", similar to the other heads of staff
qm's hud and id card icons are now the cargo tech's icon (but blue), following the other heads of staff which have blue versions of their dept (blue wrench, blue cross, blue... sec badge?? blue exclamation)

## Why It's Good For The Game
makes the qm more visually consistent with the other heads, makes it clearer that the qm is the head of the cargo department
the current icon is kinda hard to decipher too? you can tell the cargo tech's is a crate, but the qm's is just like a weird head dude (which the hop already has, making those ids hard to distinguish since theyre both blue)? or a stamp?

## Changelog
:cl:
fix: qm's intern id is now "quartermaster-in-training"
image: qm's id and hud icons now use the cargo tech icons (but blue) to be consistent with the other heads of staff
/:cl:


ghost:
We'll need to remove the QM icons from our modular hud file

vinylspiders (approved):

PR #25729, opened 2023-12-19
Revert "[TM-ONLY] Early mirror of fixing TGUI images not allowing clicks"

vinylspiders:
Reverts Skyrat-SS13/Skyrat-tg#25725

Whoops, did not mean to actually merge this. Revert whenever the actual mirror arrives so the diffs are correct.


GoldenAlpharex (approved):
I'll just wait until the PR from upstream is in before updating the server again, I doubt it'll take long before it comes around.

PR #25758, opened 2023-12-21
Map votes are no longer random

GoldenAlpharex:
## About The Pull Request
What it says on the tin. I feel like it being weighted random kinda defeats the purpose of having it be a vote, since we already get so many options that most maps will actually have very similar percentages of being picked, leading to very random outcomes, which feels pointless.

Now you'll get the map that you voted for.

## Proof of Testing

<details>
  I'll be honest, I tried to test it, and it just told me that there was only one map that was eligible and that it thus just set it to Void Raptor, and I didn't really feel like messing with configs on my test server to change that.
</details>

## Changelog

:cl: GoldenAlpharex
qol: Map votes are once again going to respect the majority vote, instead of being a weighted random vote.
/:cl:

### DIFF
--- a/code/datums/votes/map_vote.dm
+++ b/code/datums/votes/map_vote.dm
@@ -2,7 +2,7 @@
 	name = "Map"
 	message = "Vote for next round's map!"
 	count_method = VOTE_COUNT_METHOD_MULTI
-	winner_method = VOTE_WINNER_METHOD_WEIGHTED_RANDOM
+	winner_method = VOTE_WINNER_METHOD_SIMPLE // SKYRAT EDIT - Map votes shouldn't be using weighted random - ORIGINAL: winner_method = VOTE_WINNER_METHOD_WEIGHTED_RANDOM
 
 /datum/vote/map_vote/New()
 	. = ..()
--- a/code/datums/votes/map_vote.dm
+++ b/code/datums/votes/map_vote.dm
@@ -2,7 +2,7 @@
 	name = "Map"
 	message = "Vote for next round's map!"
 	count_method = VOTE_COUNT_METHOD_MULTI
-	winner_method = VOTE_WINNER_METHOD_SIMPLE // SKYRAT EDIT - Map votes shouldn't be using weighted random - ORIGINAL: winner_method = VOTE_WINNER_METHOD_WEIGHTED_RANDOM
+	winner_method = VOTE_WINNER_METHOD_WEIGHTED_RANDOM
 
 /datum/vote/map_vote/New()
 	. = ..()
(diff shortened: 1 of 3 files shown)


vinylspiders (approved):

Nerev4r:
random is preferable over ice > tram > ice > tram tbh

desertmote:
I'm not writing this comment to convince you to keep weighted random, but I do want to try and explain why it is a good thing. Weighted random allows maps to roll that otherwise would hardly have a shot at rolling during the timeslot.
For example during my late timeslot (11pm CET) I could guarantee I'd open Skyrat into blueshift. Or when I open Skyrat at my average timeslot (5pm CET) I could bet on either the station being Tram or Icebox.
With weighted random I've been opening Skyrat and seeing different maps. I've seen birdshot more often on my lowpop rounds, and more of non-blueshift on highpop, that in my opinion is a healthy and fun map rotation.
Sure, it takes some getting used to not to see majority vote win; but its not that important, is it? Its just a map vote, whats more important the integrity of a vote, or a map rotation that isn't player-weighted? (Which is a much scarier concept IMO)

Lutowski:
As a european player, I'd be lucky if I saw Blueshift even once a week. The only maps I EVER get to play is the bastard trifecta of Tram, Icebox and Meta. I'm really sick of these maps. I'd like to actually see Delta and Birdshot in my timezone for once, so please reconsider.

Bubberstation has us beat on map variety, every vote has lots of maps to choose from, even if they all have their own issues or charms. If we're sticking with majority vote again we should at least consider having more maps in rotation.

GoldenAlpharex:
I wouldn't really consider seeing Birdshot more often as a positive, as I don't really think that map is really well-suited for Skyrat to begin with.

In general, I simply disagree that we should have votes if the end result is nearly equal chances for every map to get picked. At that point, it might as well be fully random, and I'd rather not have that. So it's either that we get multi-choice map votes but the result is no longer random, or we get single choice map votes and the results could be random, so that the weights actually matter in the end result. I don't want votes where tne vote you cast basically doesn't matter.

burgerenergy:
> As a european player, I'd be lucky if I saw Blueshift even once a week. The only maps I EVER get to play is the bastard trifecta of Tram, Icebox and Meta. I'm really sick of these maps. I'd like to actually see Delta and Birdshot in my timezone for once, so please reconsider.
> 
> Bubberstation has us beat on map variety, every vote has lots of maps to choose from, even if they all have their own issues or charms. If we're sticking with majority vote again we should at least consider having more maps in rotation.

That's an issue more so with maps having player count requirements, which probably could go given we run three hour rounds anyway, not voting.

Gandalf2k15:
nyet

PR #25768, opened 2023-12-21
[MIRROR] Lizard spines no longer "float" on characters without lizard tails. [MDB IGNORE]

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/80456
--------------------

## About The Pull Request

closes: #26468 
closes: #26496
closes: #26516
closes: #26528
Also contains some nice stuff for our modular files that we didn't adjust back in earlier mirrors. - Useroth

Fixes #69658.

As the title says, this PR resolves the issue where lizard spines always have a "tail" portion even if there isn't actually a tail for them to be on. This has been done via two steps:

1. Spine icons have been separated into "spine" and "tailspine" icon states, for those on the back and those on the tail, respectively.
2. Tailspines now have their own, separate bodypart overlay, which is added by the tail if the owner has spines. This bodypart overlay appears if both a tail and spines are present, and is removed if this ceases to be true.

image

This image displays the results of this change. From left to right, it shows:

1. A lizard with both a tail and spines.
2. A lizard with spines but not a tail (specifically, one where the tail was removed).
3. A lizard with a tail but not spines (specifically, one where the spines were removed).
4. A lizard with the "short" tail style, which does not fit the shape of tail spines and thus doesn't have them.
5. A lizard with a felinid tail surgically attached, which also does not receiver tail spines.

Which tails will successfully render tail spines is now handled by a "spine_key" define in the tail sprite datum. Currently, such a key only exists for the standard lizard tail shape, but should anyone wish to sprite spines for other sorts of tails, it is now a simple matter to do so.

Relatedly, handling for tail spines has been moved to the base tail object, rather than lizard tails in particular. This functionally changes nothing, but may make life easier for downstreams that want to add spines to tails that are not explicitly _lizard_ tails.

Finally, this PR also fixes a bug introduced in #79687, which made bodypart overlays not update at the right time when external organs are inserted or removed. Specifically, bodyparts were redrawn _before_ the overlays were added or removed, meaning they would not update until something else forced a redraw. This is no longer the case.
## Why It's Good For The Game

"Floating" lizard spines are incredibly ugly visually, and a major obstacle to adding lizard tails that are not of the "standard" shape. In addition to cases where a lizard loses their tail mid-round, this also allows tailless or short-tailed lizards to actually use one of lizards' main customization options without looking like a total buffoon.

The extensibility, as stated before, may be nice for downstreams that wish to place spines on more tails, and may even be useful here if someone wishes to sprite spines for the "short" tail style.

Also makes the code nicer, as the tail/spine wagging interaction previously relied on a very ugly two-way link between the two organs that has been rendered unnecessary by the tail handling everything.

Finally, it's good to fix a bug that completely destroyed the usability of feature manipulation surgery.
## Changelog
:cl:
fix: Feature manipulation surgery will now properly update the patient's appearance.
refactor: The tail portion of lizard spines will no longer draw on people who do not have a tail.
/:cl:


Gandalf2k15:
organ pr

Majkl-J:
Important question is Does wagging still work as intended?

Gandalf2k15:
> Important question is does wagging still work as intended?

nope, so that needs fixed

PR #25892, opened 2023-12-29
Cleans up ~Skyrat_defines

StrangeWeirdKitten:
## About The Pull Request

Alright, so I dug through all of Skyrat's defines and I've trimmed them down. There are three types of defines.
- The reasonable define. It's used in multiple checks. It's a standardized define. It massively helps code readability. These are all fine.
- The "I'm a variable" define. You don't need to write defines that are used once and will realistically only be used once in the entire codebase. Bonus points when it's set to a number or a string which is only going to be used once, ever.
- The abandoned consistency project defines. These are old ruins of systems which appear be abandoned ideas. For example, armor, and items using power cells and the charge they use. Everything on /obj/item uses ``power_use_amount = 50`` with only two items deviating from that (the electric welders). We don't need an entire list of power usage defines when we don't use them at all.
- The litteral abandoned defines. Defines which are never used and are old relics of old, deleted code.

Nothing changes, but it might shave a second off initilization time and coders can maybe see some consistency of rhyme and reason again.

## Proof of Testing

image

## Changelog

:cl:
code: Cleans up unused or unneeded defines from Skyrat modular
/:cl:


Kapu1178:
compiler macros do not affect initialization time or compile times.

ss13's compile times are quite literally, 99% type node expansion.

Iamgoofball:
this PR directly downgrades the readability of the project :-1:

StrangeWeirdKitten:
These defines are completely unused, or completely unneeded, or do not follow any other consistency across anywhere else. You are free to point to a specific define for further discussion.

modular_skyrat/modules/apc_arcing/apc.dm
```diff
@@ -1,3 +1,10 @@
+/// Lower excess value for APC arcing, 5% chance to arc
+#define APC_ARC_LOWERLIMIT 2500000
+/// Moderate excess value for APC arcing, 10% chance to arc
+#define APC_ARC_MEDIUMLIMIT 5000000
+/// Upper excess value for for APC arcing, 15% chance to arc
+#define APC_ARC_UPPERLIMIT 7500000
```
YakumoChen: Can be further defined or code commented to portray what the number translates to in metric.

Iamgoofball:
Are you actually incapable of understanding the benefit of not having magic numbers floatimg around the codebase? Removing unused defines, sure, but quit removing actively in use defines, it reduces code readability for literally zero benefit.

PR #25898, opened 2023-12-29
[DNM] Raises the minimum players required for some events.

Useroth:
:cl:
config: Raised the minimum players required for some events.
/:cl:


Gandalf2k15 (changes requested):
wait dont science need these anomalies sometimes?

softcerv:
> wait dont science need these anomalies sometimes?

I think they could buy the cores from Cargo, but I could be wrong on that.

StrangeWeirdKitten:
The only thing anamolies give you is cores that don't need to be imploaded.

PR #25918, opened 2023-12-30
Modular Engines + The Return of Lord Singuloth(and more)

Gandalf2k15:
## About The Pull Request

This adds a few things to make engineering a little bit more dynamic:

- Modular Engines
- More Engine Types

### Modular Engines
The modular engine system uses my automapper system to choose a random engine type at the start of every shift to replace the default engine. The default engine has an equal chance to spawn however. So this will mean you will still get the supermatter engine sometimes.

The modular engine system will have the following engine types to pick from:
- Supermatter engine(unchanged)
- Singularity Engine
- Tesla Engine
- ~~PACMAN~~ Penitence Engine
- RBMK Engine
- GA37W EBWR Engine
- Antimatter Engine
- Quantum Phase Inversion Drive(TM)

For those of you that remember the old singularity I added, it has been renewed and tested, the systems that it once used are now gone so I've had to redo the engine. Here are the deets:
It uses hawking radiation instead of radiation(basically just a way for me to transmit energy from the singularity to the rad collectors). It functions in almost the exact same way as before(just without the radiation component).

The bugs it had before have been removed and it now functions correctly.

Calculations for Hawking Radiation:
Given:

ACCUMULATOR_STORED_OUTPUT = 0.04
stored_energy_per_collector = 1000 energy units
number_of_collectors = 6
Calculation Steps
Power Output for Each Collector:

power_output_per_collector = min(stored_energy, (stored_energy * ACCUMULATOR_STORED_OUTPUT) + joules_to_energy(2000))
We need to calculate joules_to_energy(2000) as before.
joules_to_energy(2000):

= 2000 * (1 SECONDS) / SSmachines.wait
= 2000 * 10 / 20
= 1000 energy units
For stored_energy_per_collector = 1000:

power_output_per_collector = min(1000, (1000 * 0.04) + 1000)
= min(1000, 40 + 1000)
= min(1000, 1040)
= 1000 energy units (since 1000 is less than 1040)
Total Power Output to the Powernet:

total_power_output = power_output_per_collector * number_of_collectors
= 1000 * 6
= 6000 energy units

The RBMK engine will be ported from TBA

The Antimatter Engine will be ported from TBA

The Penitence is made inhouse. Give it blood.

The GA37W will be added when I finish it. LOL.

The QPID will not be completed until the next solar cycle.

Each map has it's own set of engine templates which need to be fabricated. You can have an unlimited amount per map. Like I said, it just uses the automapper system.

Having more variety of engines to pick from for the engineering team is good as it just makes a very very important job a little bit more interesting.

## Proof of Testing

image

## Changelog

:cl:
add: Modular engine system, see:
add: LOTS OF ENGINES.
/:cl:


Erol509:
Would be nice if we had a item like syndicate beacon, which can at least work as a last hope of containing tesla, since tesla doesnt eat whole station after breaking out.

Erol509:
Tesla zaps by default dont generate any power, and if you add ZAP_GENERATE_POWER, tesla will make 200 MW with tesla coils

Zergspower:
that PACMAN is - an extremely low change i hope given how expensive it is to run and how little power it makes now?

StrangeWeirdKitten:
Pacmans make so little power and use a ton of fuel. It's not going to be a viable engine that's neither fun or engaging. They make 15 KW maximum per PACMAN and the plasma usage is not ignorable. You'd need to at least make it upgradable or increase the maximum output and lower the plasma usage to run a station for three hours.

Gandalf2k15:
> Pacmans make so little power and use a ton of fuel. It's not going to be a viable engine that's neither fun or engaging. They make 15 KW maximum per PACMAN and the plasma usage is not ignorable. You'd need to at least make it upgradable or increase the maximum output and lower the plasma usage to run a station for three hours.

I'll see about making a different engine then

ORCACommander:
out of curiosity, are the sprites for the singularity going to stay the same as TG or are we bringing back the ones that were ported over from i forget where but was stripped out in the rsc downsizing?

PR #26213, opened 2024-01-22
sell your power today!

jjpark-kb:
## About The Pull Request
allows you to sell your power for some money (in cargo)

we should really be rewarding engineering for producing a TON of power, and what better way than to give cargo money so they can spend it on the station-- not only that, but we have CC sending power to us from seemingly nowhere, this helps give more reason why and how they can. many other stations use the powerator to sell their excess power and we use that excess power to recharge our SMES when they are low (on lowpop)

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: added the powerator  (purchase-able and BEPIS-tech), a machine to sell power for cargo credit
/:cl:


Useroth:
Hmm... I think this could use a built-in variable tracking how much power was sold and how many credits were made via it.
Also I'm wondering about the general balance cause a properly set basic engine gives a reasonable surplus already.

Though, I think this is fine to testmerge.

Cyprex:
This is difficult to execute because power isn't balanced- some curve with diminishing returns(logarithmic?) might work well however. Definitely needs balancing though!

Cyprex:
Thinking about it providing diminishing returns based on how much power was already exported sounds good, price goes lower from a stable equilibrium as you export power(saturating the market)

PR #26214, opened 2024-01-23
hybridizing mining + more

jjpark-kb:
## About The Pull Request
brings back ores to the mineral walls (the previous mining) and makes mining level influence your manual processing of boulders

arc mining, imo, is great... but it can be of great torture to miners who just want to mine the other way. so rather than forcing the miners to use the machines, bringing back the previous mining allows them the choice to use the new mining or the previous mining.

also, added some skill influence to boulders-- the higher your mining level, the faster and less exhausted you get manually processing the boulders.

## Proof of Testing

Works on my local with no runtimes

## Changelog

:cl:
add: added back the default turf mining
add: boulders are now affected by your mining skill
balance: you get more mining xp from boulders
/:cl:


Cyprex:
Isn't finding ore supposed to be a sign of vents? I don't really think this is necessary, ultimately?

projectkepler-ru:
> Isn't finding ore supposed to be a sign of vents? I don't really think this is necessary, ultimately?

defending the mining machine on the initial run can be a chore, that and their placement can be very,,,,, awkward

PR #26336, opened 2024-02-02
Refactor of the lowpop power recharge system

Useroth:
## About The Pull Request

Previously the system, in some circumstances, could run in a patchy manner that left the station without power at times.
I've also broken up and simplified the logic a little more to be more predictable (ie the critically low SMES stuff, which might have left smes that are about to discharge uncharged).
Made it fire more often by default too.
Also made it not announce every recharge.


Gandalf2k15 (approved):
Lgtm

PR #26342, opened 2024-02-03
Changes to robotic repair surgeries.

xXPawnStarrXx:
## About The Pull Request

Previously, the robotic repair surgeries required both wire and a welder, which seems reasonable, but if organics don't need consumable resources to get healed, why should synthetics, that's not really fair. It now uses crowbars for brute or wirecutters for burn damage. (You can use the other in an emergency, but don't expect it to be quick.)

Running out of wire constantly just because one poor synthetic has gotten the overheat trauma and is around 1000 burn constantly... or watching a roboticist accidentally welderbomb because they forgot to turn it off in a panic when refilling.

## Proof of Testing

<details>

image

image

</details>

## Changelog

:cl:
qol: changed robotic repairs to use crowbar/wirecutters instead of welder/wire
/:cl:

### DIFF
--- a/modular_skyrat/modules/synths/code/surgery/robot_healing.dm
+++ b/modular_skyrat/modules/synths/code/surgery/robot_healing.dm
@@ -35,8 +35,8 @@
 		)
 
 /datum/surgery_step/robot_heal
-	name = "repair body (welder/cable)"
-	implements = list(TOOL_WELDER = 100, /obj/item/stack/cable_coil = 100)
+	name = "repair body (crowbar/wirecutters)"
+	implements = list(TOOL_CROWBAR = 100, TOOL_WIRECUTTER = 100)
 	repeatable = TRUE
 	time = 25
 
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/synths/code/surgery/robot_healing.dm
```diff
@@ -84,7 +87,7 @@
 
 /datum/surgery_step/robot_heal/initiate(mob/user, mob/living/carbon/target, target_zone, obj/item/tool, datum/surgery/surgery, try_to_fail = FALSE)
 	if(..())
-		while((heals_brute && target.getBruteLoss() && tool.tool_use_check(user, 1)) || (heals_burn && target.getFireLoss() && tool))
+		while((heals_brute && target.getBruteLoss() && tool/*.tool_use_check(user, 1)*/) || (heals_burn && target.getFireLoss() && tool))
```
jjpark-kb: This is in a modular file-- I don't see the need to comment it out.

jjpark-kb (changes requested):
Will merge once you've addressed my concern(?)

modular_skyrat/modules/synths/code/surgery/robot_healing.dm
xXPawnStarrXx: OH! I forgot to remove that part, I'll handle it now.

jjpark-kb (approved):

PR #26367, opened 2024-02-06
[MODULAR] Adds a new reagent meant for cooling synthetic burns: Dinitrogen Plasmide

nikothedude:
## About The Pull Request
Direct port of https://github.com/NovaSector/NovaSector/pull/445

Title.

Compared to hercuri, dinitrogen plasmide is a worse coolant, but it doesnt cool the bodytemp of the target, meaning its much more useful in situations where you really dont want that.

Added this to all synth treatment kits, and the robodrobe.

## How This Contributes To The Nova Sector Roleplay Experience

I realized a while ago that hercuri is not the best thing to use for synthetic burns. It often caused more problems than it solved, especially since we dont have many good ways to purge chems on a synth. Take a synth who took a lava bath, theyll be coming into robo with multiple critical burns and the rest severe. Youll want to use a good coolant for that, as not to destroy them, but hercuri will just build up in their systems, and when theyre revived, theyll immediately overdose and die again.

Plasmide, in this scenario, will simply slow them down a lot and give them a hefty movespeed malice for about a minute.

Keeping hercuri's ability to cool burns is important, IMO, for ghetto treatment, as hercuri is in burn kits and in medical. Plus, if that aformentioned synth was alive, the hercuri might be able to fight off the bodytemp caused by the burn wounds.

## Proof of Testing

<details>
  
image

</details>

## Changelog

:cl:
add: Dinitrogen Plasmide, a chem used to cool synthetic burns. Does not have the side effects of hercuri, making it safer to use.
/:cl:


nikothedude:
me when conflict

PR #26395, opened 2024-02-07
[MODULAR] Vore Part 1: Prefs + Smell/Taste

softcerv:
## About The Pull Request
This PR does two things:
First is, adding preferences for vore separate from the one that already exists (I'll explain this later.)

Second is adding character preferences for taste and smell along with verbs to lick and smell other adjacent humans with text set. If no preference is set, it will be impossible to taste/smell someone.

Basically, people may want to use the vore system for RP reasons while not being directly into vore, so it makes sense to have separate mechanical prefs for this.

Letting people set a smell/taste adds a neat little bit of optional flavor. 

## Proof of Testing

<details>
  
image
image
image
image

</details>

## Changelog

:cl:
add: Adds mechanical vore preferences
add: Adds taste/smell character preferences along with verbs to taste/smell others that have text set.
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
+++ b/modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
@@ -88,11 +88,11 @@
 /// Returns a list containing all of the humans adjacent to the user.
 /mob/living/proc/get_adjacent_humans()
 	var/list/hear_list = orange(1, src)
-	for(var/mob/living/carbon/human/hearer as anything in hear_list)
-		if(ishuman(hearer))
+	for(var/mob/living/carbon/human/nearby_human as anything in hear_list)
+		if(ishuman(nearby_human))
 			continue
 
-		hear_list -= hearer
+		hear_list -= nearby_human
 
 	return hear_list
 
--- a/modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
+++ b/modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
@@ -55,3 +55,45 @@
 		dropItemToGround(equipped_item, TRUE)
 
 	return TRUE
+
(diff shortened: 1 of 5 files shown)


modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
```diff
@@ -55,3 +55,57 @@
...
+
+	var/smell = target.get_taste_or_smell(TRUE)
+	if(!smell)
+		to_chat(src, span_warning("[target] doesn't seem to have a smell."))
+		return FALSE
+
+	to_chat(src, span_notice("[target] smells like [smell]."))
+
+/// Returns the smell/taste of the parent mob
+/mob/living/carbon/human/proc/get_taste_or_smell(smell = FALSE)
```
Useroth: I think, in the code, this is akin to calling a proc with a magical number.
In my opinion it should either be SMELL and TASTE defines as the argument, or this should be wrapped in two separate procs.

modular_skyrat/modules/modular_items/lewd_items/code/verbs.dm
```diff
@@ -55,3 +55,57 @@
...
+
+	var/string_to_return = preferences.read_preference(/datum/preference/text/taste)
+	if(smell)
+		string_to_return = preferences.read_preference(/datum/preference/text/smell)
+
+	return string_to_return
+
+
+/// Functions similiar to get hearers in view, but it only counts humans.
+/proc/get_human_hearers_in_view(view_radius, atom/source)
```
Useroth, line 102: Since it is not called within this PR, so I can't infer how often it will be called and performance impact due to it, it might be worth re-writing the proc based on get_hearers_in_view as opposed to removing non-humans them from the list.

Useroth (changes requested):

Useroth (approved):

PR #26440, opened 2024-02-10
[MIRROR] Refactors fire overlays once again to make it not get stuck so often

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/81367
--------------------
## About The Pull Request

Maybe finally fixes #77701 

A big reason why this kept happening is because fire uses standing overlays. But fire isn't managed by mobs anymore. Meaning in some situations, fire can cease to exist but the overlay can still be on the mob. So it gets stuck. 

So like, why use standing overlays anymore? We can just hook `update_overlays` signal.
Isn't that neat. 

## Changelog

:cl: Melbert
refactor: Fire effects get added to mobs in a different way now. Maybe it will get stuck less. Report any oddities. 
/:cl:


jjpark-kb:
this requires a lot more than I thought! Our xenos...

PR #26448, opened 2024-02-10
probability bsminer change

jjpark-kb:
## About The Pull Request
when you use your hand on a bluespace miner, you can now focus on specific products from it-- tripling the chance. the catch is that it will prevent all other mats from spawning.

bluespace miners are great at producing every low-tier mat, so doing this should allow people to get a bit more of what they want with a drawback attached.

## Proof of Testing

<details>
  
image
image
image

</details>

## Changelog

:cl:
add: you can now increase the probability of certain mats from bluespace miners at the cost of not producing the other mats
/:cl:


xXPawnStarrXx:
Oooh, I like that. I've always wanted it to be a filter thing, 'cause then you can focus on the rarer mats. It's not broken either, 'cause xenobiology slimes can just do that.

PR #26729, opened 2024-03-03
forging weapons with perfect hits, and mindless killer

jjpark-kb:
## About The Pull Request
There is a new component added to forging weapons (could be added to other weapons as well if we deem it fine) that increases the damage the weapon does to any mob that is mindless-- aka, not controlled by a player. The component can either add flat damage, a multiplier, or both at the same time (with the flat applying first). 

When forging, you will get something called "perfect hits" occasionally. These only affect weapons at the moment, but depending on the amount of perfect hits, you could increase the effectiveness (the force) of the weapons (up to 3 additional force). Now that may seem... trivial but I assure you that that extra 3 force really does add up.

I did lower the force of some of the items to make their unwielded force half of their wielded force so I could just the force multiplier on the two-handed component... but pair that up with the mindless killer component and you have a killer weapon! Take the sword for example: it has an unwielded force of 10, but that goes up to 20 for mindless mobs; it has a wielded force of 20, but that goes up to 40 for mindless mobs. That example also didn't factor in if there was a perfect amount of perfect hits-- 13 force (unwielded, mind) to 26 force (unwielded, mindless), and 26 force (wielded, mind) to 52 force (wielded, mindless)!

This change should help make forged weapons feel great against the fauna of the mining planet, whether it be lavaland or icemoon. Reality is, melee sucks greatly against fauna unless you are trying to get the trophies for the crusher/glaive. With these changes, hopefully melee can be more up to par with ranged-- you get greater damage at the risk of getting hit, or do lesser damage with less risk of getting hit.

## Proof of Testing

Tested locally.

## Changelog

:cl:
balance: forged sword, katana, spear, axe, and hammer had their force lowered (so wielding multiplier can double)
balance: bokken had its force very slightly increased (so wielding multiplier can halve)
fix: you can no longer get infinite amounts of experience for hammering forged items (the hits required check is earlier and with a return)
add: added perfect hits, which grant a great deal of smithing experience and can increase the damage of forged weapons (up to 3 additional force)
add: added mindless killer component to all forged weapons-- deal twice the damage to any mob that has no mind (no mind means not player controller)
/:cl:


Gandalf2k15 (approved):

modular_skyrat/modules/reagent_forging/code/mindless_killer.dm
```diff
@@ -0,0 +1,47 @@
...
+		return COMPONENT_INCOMPATIBLE
+
+	obj_parent = parent
+
+	mindless_force = mindless_force_override
+	mindless_multiplier = mindless_multiplier_override
+
+	RegisterSignal(parent, COMSIG_ATOM_EXAMINE, PROC_REF(examine))
+	RegisterSignal(parent, COMSIG_ITEM_PRE_ATTACK, PROC_REF(do_force))
+	RegisterSignal(parent, COMSIG_ITEM_AFTERATTACK, PROC_REF(undo_force))
```
Maia-J, line 20: Butchering interrupts and cancels the attack chain in the on_attack point. These bonuses are applied in pre_attack, and removed in post_attack. The chain goes `pre_attack -> on_attack -> attack -> post_attack` Meaning you can stack these bonuses by butchering and cancelling the action repeatedly.

PR #26818, opened 2024-03-09
[MIRROR] Mini-Nukie Update: The Bulldog is a 2-shot burst fire gun edition (and lone ops get some spare ammo for their bulldog)

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/81715
--------------------

## About The Pull Request

The bulldog shotgun, available to ops, now has a burst fire of 2 rounds. This means it launches a fairly rapid volley of bullets per click.

Currently, the lone operative spawns with a single bulldog shotgun. Instead of just the shotgun, the operative now comes with some spare magazines for their gun.

## Why It's Good For The Game

The bulldog still seems to be really unpopular, and still feels like it isn't quite doing as much as other options for operatives. Even for the cheap price, it just doesn't pull its weight. It lacks staying power or range to face several opponents at once. You can certainly buy ammo to make it better at engaging at range, but this sometimes begs the question as to why you bothered with the bulldog over something else. 

It is primarily only really good at killing one person at a time in close quarters, but the person you would be using that power against is usually a more armored opponent like the Captain or Head of Security. Lots of other nuclear operative weapons simply do that much better than the bulldog.

If the bulldog is meant to be mulching single targets one at a time, it should absolutely be doing that reliably. Because people really like shotguns. So why not unload two buckshot straight into someone's cranium with a single pull of the trigger for a nice crunchy splat.

Lone operative got themselves a bulldog, but were cheated out of the free ammo they would have received if they had simply bought the bulldog from their uplink. Lone ops are not nearly as scary as an actual midround nuclear assault, primarily because their objective doesn't come with the benefit of an infiltrator or a bomb to deliver. They're a punishment mechanic, technically. But even as a punishment, they don't really do a terribly good job at punishing. Since the bulldog is focused on putting down the captain as fast as possible, then it stands to reason that the lone operative gets as much help accomplishing that as necessary.

## Changelog

:cl:
balance: Bulldog Shotguns now have a 2-round burst fire.
balance: Lone Operatives now come with some additional Bulldog Shotgun magazines.
/:cl:


code/modules/projectiles/guns/ballistic/shotgun.dm
```diff
@@ -148,10 +148,18 @@
...
 	fire_delay = 10 //Skyrat edit - Original: 0
 	pin = /obj/item/firing_pin/implant/pindicate
 	fire_sound = 'sound/weapons/gun/shotgun/shot_alt.ogg'
+=======
+	burst_size = 2
+	fire_delay = 1
+	pin = /obj/item/firing_pin/implant/pindicate
+	fire_sound = 'sound/weapons/gun/shotgun/shot_alt.ogg'
+	actions_types = list(/datum/action/item_action/toggle_firemode)
+>>>>>>> 91f698f3ea1 (Mini-Nukie Update: The Bulldog is a 2-shot burst fire gun edition (and lone ops get some spare ammo for their bulldog) (#81715))
```
softcerv:
```suggestion
	burst_size = 2
	fire_delay = 10 //Skyrat edit - Original: 1
	pin = /obj/item/firing_pin/implant/pindicate
	fire_sound = 'sound/weapons/gun/shotgun/shot_alt.ogg'
	actions_types = list(/datum/action/item_action/toggle_firemode)
```

softcerv (approved):

PR #26880, opened 2024-03-13
Two New Skills

jjpark-kb:
## About The Pull Request
adds two new skills: construction and primitive (survivalist). construction is gained through crafting and building, and primitive is from most content that is... primitive.

More skills with some benefits means that people should have to focus on that content to get the best benefits and people can't just hop onto a role and expect to be better than someone who has been playing the role since the start of the shift. It helps keep Mary Sues in line.

## Proof of Testing

Tested locally.

## Changelog

:cl:
add: added two new skills: construction and primitive
/:cl:


jjpark-kb:
github actions is rude!

PR #26961, opened 2024-03-20
[MIRROR] [No GBP] Vending examine message now properly shows the name of the vending refill required.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/82101
--------------------
## About The Pull Request

Vending refills were piping in the typepath of the refill as opposed to the name since their examine was referencing a type path. This instead pulls the `machine_name` and `name` to form the correct string on examine to confirm that the value is accurate.
image

## Why It's Good For The Game
Fixes #82095. That should probably not be shown in-game as is and now actually showcases the correct message on examine.

## Changelog
:cl:
fix: Vending machines now show the correct vending refill name when examine and not full.
/:cl:


code/modules/vending/_vending.dm
```diff
@@ -342,6 +342,27 @@ GLOBAL_LIST_EMPTY(vending_machines_to_restock)
...
+		. += span_notice("It should have a handfull of credits stored based on the missing items.")
+	else if (credits_contained > PAYCHECK_CREW)
+		. += span_notice("It should have at least a full paycheck worth of credits inside!")
+		/**
+		 * Intentionally leaving out a case for zero credits as it should be covered by the vending machine's stock being full,
+		 * or covered by first case if items were returned.
+		 */
+
+
+>>>>>>> 014ffc818be ([No GBP] Vending examine message now properly shows the name of the vending refill required. (#82101))
```
softcerv:
```suggestion
/obj/machinery/vending/examine(mob/user)
	. = ..()
	if(isnull(refill_canister))
		return // you can add the comment here instead
	if((total_loaded_stock() / total_max_stock()) < 1)
		. += span_notice("\The [src] can be restocked with [span_boldnotice("\a [initial(refill_canister.machine_name)] [initial(refill_canister.name)]")] with the panel open.")
	else
		. += span_notice("\The [src] is fully stocked.")
	if(credits_contained < CREDITS_DUMP_THRESHOLD && credits_contained > 0)
		. += span_notice("It should have a handfull of credits stored based on the missing items.")
	else if (credits_contained > PAYCHECK_CREW)
		. += span_notice("It should have at least a full paycheck worth of credits inside!")
		/**
		 * Intentionally leaving out a case for zero credits as it should be covered by the vending machine's stock being full,
		 * or covered by first case if items were returned.
		 */



```

softcerv (approved):

code/modules/vending/_vending.dm
```diff
@@ -361,6 +361,7 @@ GLOBAL_LIST_EMPTY(vending_machines_to_restock)
 		 */
 
 
+
```
softcerv:
```suggestion
```

PR #27268, opened 2024-04-11
Modular: Prioritize same z level machines for technet get_available_servers() function

tmyqlfpir:
## About The Pull Request

This is a improvement to the currently existing behavior of `get_available_servers()`. When a machine that uses tech net research (lathes, exomachines, etc), it will scan for an existing research server present across all z levels.

The limitation comes when there exists another server on a different z level, but is on the same planet as the station (ice moon map for example). If Interdyne was to have their own server for ice moon map, it will still use the station's research server, because the macro `CONNECT_TO_RND_SERVER_ROUNDSTART` will always return the first item in the list of servers (which is likely to always be NT's servers).

How this PR fixes the limitation is by first checking for any same z level research servers, and adding it to the start of the server list. This ensures `CONNECT_TO_RND_SERVER_ROUNDSTART` will favor research servers located in the same z level. If there are no servers located on the same z level, it will then fallback on the original code.

This improves the overall robustness needed to ensure planet based maps with Interdyne base ruin can have their own research servers, and require players to do research instead of relying on NT.
## Proof of Testing

I have ran this on my local server, and can verify that it does not break or change the same machines linked to NT's research server, or Interdyne's research server (if built).

Tramstation's lower level lathes (such as in engineering) will still use the research server located on the upper z level as expected.
<details>
  
The changes overall
Untitled

A machine being constructed on Tarkon, and returning the assigned server located on the same z level
tarkon

</details>

## Changelog
:cl:
fix: Prioritize same z level research servers when assigning a machine to a tech net
/:cl:


tmyqlfpir:
I've come to realize that this solution may not be the most elegant, and will cause maintainer issues for future merges. Unfortunately as it stands, any station that is based on a planet with lavalands z level, will assume the Interdyne base is part of the station, and thus force them to be unable to use their own R&D server.

For DS-2 and Tarkon as they are based in space, they can use their own R&D server without possible conflicting with the NT station.

Currently the least problematic solution is to keep the way things are, and allow Interdyne to syphon off of NT's tech net.

Edit: This has been resolved by future PR https://github.com/Skyrat-SS13/Skyrat-tg/pull/27512

PR #27299, opened 2024-04-13
[MIRROR] Fixes to zombie regen (and incidentally behavior of TRAIT_NODEATH)

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/82602
--------------------

## About The Pull Request

Zombie regen was being cancelled out by the brute damage you take while in crit with TRAIT_NOBREATH (which zombies have) and since they have TRAIT_NODEATH they would, at a certain point, end up nearly permacritted unless they were somehow killed (which TRAIT_NODEATH would prevent in almost all cases).

Zombies now have TRAIT_NOCRITDAMAGE so they don't take this damage, and carbon/death() and human/death() now take TRAIT_NODEATH into account, since there are hard calls to death() littered everywhere instead of going through the actual body damage dealing code. If you have NODEATH you'll still die if gibbed.
## Why It's Good For The Game

Fixes #82601 
Fixes TRAIT_NODEATH actually making it so you... nodeath...
## Changelog

Being in critical condition no longer damages zombies.
:cl:
fix: Being in critical condition no longer damages zombies.
code: TRAIT_NODEATH will actually stop you dying in (most) cases.
/:cl:


jjpark-kb:
CAREFUL, another PR is being done that reverts some of this-- so this needs to be merged first!

PR #27312, opened 2024-04-14
Removes decaying waste as a pollutant

lessthnthree:
## About The Pull Request

Removes decaying waste as a pollutant. It provides no value to the station, and remains even when spaces are cleaned/renovated. Given maints spawn with garbage, it's impossible to eliminate and constantly spreads.

## How This Contributes To The Skyrat Experience

No more "The unmistakable odour of decaying waste bombards your nostrils." in spaces.

## Changelog

:cl: LT3
del: The unmistakable odour of decaying waste will no longer bombard your nostrils
/:cl:


jjpark-kb:
I think rather than removing it, we need to find better ways to remove it IC.

PR #27473, opened 2024-04-25
Creates Round removal opt-in (bootlegged off of nova's antag opt-in.)

Swiftfeather:
## About The Pull Request

I'm Draft PRing this early so I can watch github CI's go through and see diffs as I change all the words to say round removal instead of antag.

I consider this an _alternative_ to opting in/out of being antagged, as the best way to opt out of being antagonized is to stay at the several places that exist to not get antagonized (interlink, ghost cafe)

This is a pref we staff and maintainers downstream on bubberstation want, I mentioned it and useroth told me I should feel free to PR it upstream so here I am.

## Changelog

:cl:
add: Added Round Removal preferences
/:cl:


Swiftfeather:
This is ready for you to make better @Majkl-J

modular_skyrat/modules/rr_opt_in/code/objective_item.dm
```diff
@@ -0,0 +1,20 @@
+/datum/objective_item
+	/// The opt in level all owners of the item must meet for this to be eligible as an objective target.
+	var/opt_in_level = RR_OPT_IN
```
Maia-J:
```suggestion
	var/opt_in_level = RR_OPT_OUT
```

modular_skyrat/modules/rr_opt_in/code/objective.dm
```diff
@@ -0,0 +1,69 @@
+/datum/objective
+	/// The default opt in level of this objective. Only targets with opt in above or at this will be considered for this objective.
+	var/default_opt_in_level = RR_OPT_IN
```
Maia-J:
```suggestion
/datum/objective
	/// The default opt in level of this objective. Only targets with opt in above or at this will be considered for this objective.
	var/default_opt_in_level = RR_OPT_OUT
```

Majkl-J (commented):
Judging from all the overrides, shouldn't these be opt out by default? Or the overrides done the other way around, alternatively

Swiftfeather:
It has been a few weeks since I posted this, but I do not play skyrat so I am asking for a review on its functionality. Policy matters aside, is everything working as intended?

ReturnToZender:
Don't agree with Command + Sec being opted-in by default, particularly the former, but I am a big fan of the idea as a whole. The default expectation, for _everyone,_ should always be that round-removal is not allowed save for in extreme circumstances. Still, that's more of a policy thing-- code looks good.

Also the target parts of the code should be separate from the RR parts. That is to say, opting out of round-removal should not prevent you from being a target, but it should be a separate preference, and _that_ can be forced on for sec/command

modular_skyrat/master_files/code/modules/mob/living/examine_tgui.dm
```diff
@@ -57,17 +57,27 @@
...
+			var/e_prefs_hypno = preferences.read_preference(/datum/preference/choiced/erp_status_hypno)
+			var/e_prefs_nc = preferences.read_preference(/datum/preference/choiced/erp_status_nc)
+			var/e_prefs_v = preferences.read_preference(/datum/preference/choiced/erp_status_v)
+			var/e_prefs_mechanical = preferences.read_preference(/datum/preference/choiced/erp_status_mechanics)
+			ooc_notes += "ERP: [e_prefs]\n"
+			ooc_notes += "Non-Con: [e_prefs_nc]\n"
+			ooc_notes += "Hypnosis: [e_prefs_hypno]\n"
+			ooc_notes += "Vore: [e_prefs_v]\n"
+			ooc_notes += "ERP Mechanics: [e_prefs_mechanical]\n"
+			ooc_notes += "\n"
```
softcerv, line 72: I don't see why it is necessary to add more indents here. The logic is the exact same. Please revert this.

modular_skyrat/modules/rr_opt_in/code/objective_item.dm
```diff
@@ -0,0 +1,20 @@
...
+	var/opt_in_level = RR_OPT_OUT
+
+/// Returns TRUE if we have no owners, or all owners's effective opt in level is above [opt_in_level]. FALSE otherwise.
+/datum/objective_item/proc/owner_opted_in()
+	if (!length(item_owner))
+		return TRUE
+	for (var/mob/living/player as anything in GLOB.player_list)
+		if ((player.mind?.assigned_role.title in item_owner) && player.stat != DEAD && !is_centcom_level(player.z)) // is an owner, copypasted from objective_items.dm owner_exists()
+			if (player.mind.get_effective_opt_in_level() < opt_in_level)
+				return FALSE
```
softcerv, line 12: Change the logic in this to utilize early `continue`s rather than nested `if` statements.

modular_skyrat/modules/rr_opt_in/code/mind.dm
```diff
@@ -0,0 +1,84 @@
...
+/// Refreshes our ideal/on spawn round removal opt in level by accessing preferences.
+/datum/mind/proc/update_opt_in()
+	var/datum/preferences/preference_instance = GLOB.preferences_datums[lowertext(key)]
+	if (!isnull(preference_instance))
+		ideal_opt_in_level = preference_instance.read_preference(/datum/preference/choiced/rr_opt_in_status)
+
+		for (var/antag_category in GLOB.rr_optin_forcing_on_spawn_antag_categories)
+			if (antag_category in preference_instance.be_special)
+				on_spawn_rr_opt_in_level = RR_OPT_LEVEL_ANTAG
+				break
```
softcerv, line 49: Please change this to use early returns.

modular_skyrat/modules/rr_opt_in/code/mind.dm
```diff
@@ -0,0 +1,84 @@
...
+/// Sends a bold message to our holder, telling them if their optin setting has been set to a minimum due to their antag preferences.
+/datum/mind/proc/send_rr_optin_reminder()
+	var/datum/preferences/preference_instance = GLOB.preferences_datums[lowertext(key)]
+	var/client/our_client = preference_instance?.parent // that moment when /mind doesnt have a ref to client :)
+	if (our_client)
+		var/rr_level = get_rr_opt_in_level()
+		if (rr_level <= RR_OPT_OUT)
+			return
+		var/stringified_level = GLOB.rr_opt_in_strings["[rr_level]"]
+		to_chat(our_client, span_boldnotice("Due to your antag preferences, your antag-optin status has been set to a minimum of [stringified_level]."))
```
softcerv, line 60: This too.

modular_skyrat/modules/rr_opt_in/code/mind.dm
```diff
@@ -0,0 +1,84 @@
...
+/datum/mind/proc/get_rr_opt_in_level()
+	if (on_spawn_rr_opt_in_level > RR_OPT_OUT)
+		return on_spawn_rr_opt_in_level
+
+	var/datum/preferences/preference_instance = GLOB.preferences_datums[lowertext(key)]
+	if (!isnull(preference_instance))
+		for (var/antag_category in GLOB.rr_optin_forcing_midround_antag_categories)
+			if (antag_category in preference_instance.be_special)
+				return RR_OPT_LEVEL_ANTAG
+	return RR_OPT_OUT
```
softcerv, line 80: Please change this to be an early return.

modular_skyrat/master_files/code/modules/mob/living/examine_tgui.dm
```diff
@@ -57,17 +57,27 @@
...
+
+	if(!CONFIG_GET(flag/disable_rr_opt_in_preferences)) //Round Removal opt in stuff
+		var/rr_prefs = holder.mind?.ideal_opt_in_level
+		var/effective_opt_in_level = holder.mind?.get_effective_opt_in_level()
+		if(isnull(rr_prefs))
+			rr_prefs = preferences.read_preference(/datum/preference/choiced/rr_opt_in_status)
```
softcerv, line 78: What does this part of the code do if preferences are null when it runs?

softcerv (changes requested):
There are some things I noticed

softcerv:
While I do think that it is good that people now have more of a choice over Round Removal, I don't like the fact that it forces both command and security to take the risk of being removed from a shift that typically lasts over 2 hours minimum. This is something that I think will stop a lot of people from playing command jobs and will possibly encourage negative behavior for those that do still play those roles. 

With all of that being said, I am not going to block or close this PR as this is primarily a policy issue. I still want my requested code changes to be addressed though.

R3dtail:
Yeah even as a long-term command player I ... don't see how forcing me to accept being told I can't play the round if an antag decides they want to round remove me is fun and engaging for me.  I aim to present myself as open to interactions as possible, and there's significant leeway for how antag interactions can go.  Forcing round removal isn't ideal and definitely would make me want to not play the roles I previously enjoyed most of all, because round remova ldoes not, in my eyes, add anything to roleplay.  It should definitely be something that someone should be able to opt in or out of  and most certainly shouldn't be forced.  Some people are more okay with their characters being RR'd  if an antag interaction goes south.  That's fine, this option is good for them.  Some characters aren't okay with themselves being round removed.  I know the go-to common response is "Just don't play command then" or "Make a different character" but on furry-based servers that's not a very realistic expectation.    Round removal's fine on servers whose rounds are on average an hour.  Not so much  on places where the minimum round length is over 2 hours.  That is a significant chunk of time in a day.

PR #27479, opened 2024-04-26
adds the new movement speed modifier to raptor cybernetics

aKromatopzia:
## About The Pull Request

Consistency. Also, do I have permission to just put these into the teshari mutantparts file so I can use the defines there? they're undef'd at the end so any numerical changes do unnoticed.

## Changelog
:cl:
add: teshari move speed change applies to raptor cybernetics
/:cl:


aKromatopzia:
I will never understand that bot

PR #27561, opened 2024-05-01
[MIRROR] Moves "sprite accessories" (e.g. Hair, Undergarments, Mutant Bits) from `GLOB` to a datasystem

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/82847
--------------------
This is just a revitalization of #80275.

## About The Pull Request

On the tin, basically demotes everything related to setting up and storing these bulky lists generated from reading `/datum/sprite_accessory` subtypes from living in a global space that will instead be in a compartmentalized subsystem for accesses. Also a lot of code modernization and micro-improvements (unquantifiable)

## Why It's Good For The Game

Same exact expected results, just accessed in a different way.

image

There's a few reasons why I want this to happen.
* The `GLOB` space is too clogged. There are at least a thousand variables on `GLOB`, and it's extremely painful to access stuff on production/local through view variables when you're debugging stuff like this. It's also painful when there is stuff that _should_ live on `GLOB` that you might want to see in VV/Debugger but are forced to either have to scroll a mile to find what you want or wait a long while for it to load. The less bulky lists we have of stored initialized datums, the better.

* `make_datum_reference_lists()` is a consequence of wack stuff like this where we're reliant on certain things being initialized in the `GLOB` portion of world initialization _before_ subsystems/static variables load - most of these datum lists in the aforementioned proc doesn't _really_ need to be ready to go before `world.New()` for example. We'll sadly have to abuse `PreInit()` for now, but it really is something that has to be ready to go due the critical dependence that stuff like Preferences has on it.

* We don't have to have the procs live in a global namespace either. Instead of passing in `GLOB.XList` or `DSstorage.XList` every single time, we can instead just move the proc setup on the subsystem and use `XList` in a more native fashion.

* It's easier to find what you need. To me, it's a lot nicer to ctrl+click the DS and go to the variables to find something I'm looking for instead of having to scavenge around for any footprint/trace of the global I want to look for. This is more trivial than the other two, but that's something I like to think about when I go to bed.

I also had to refactor a bit of the code to accommodate the limitations of the new DS system, but it should be a lot cleaner anyways.

## Changelog

Not relevant

---

Also nothing should have broken but it's a good thing we have screenshot unit tests to prove me wrong.


SpaceLoveSs13:
@Majkl-J You are a godsend.

PR #27620, opened 2024-05-07
Increases Sleeping Carp Cost and reduce contractor baton stamina

projectkepler-ru:
## About The Pull Request
This increases the cost of sleeping carp item, both on time progression and TC-wise.
Normally, in TG and in bubber

Scarp is fine

problem is, it's not here, the average skyrat player simply was not built to deal with Scarp.

This does not prevent it from being purchased, but more so discourage it

This also revert skyrat contractor baton back to 85 stamina damage instead of 115, which I thought was way too much
You wouldn't enjoy it if I bought Scarp and then go miner, go space ruining till I get a swat boot to become immune to slip, steal captain glove to be immune to shock, obtain drake armor set to become immune to fire

The fact is, Scarp is beatable, but the method to do so is not something most player will know, for now, put this on the backburner. 
## Proof of Testing
N/A
<details>
  
</details>

## Changelog
:cl:
balance: Sleeping Carp now cost 25 TC and has a progtot time of 140 minutes
balance: Contractor baton deals 85 stamina damage instead of 115, which was the skyrat buff
/:cl:


projectkepler-ru:
@Useroth 
The numbers need to be adjusted a fair bit, it need to account for the 3 hours round
that
Or
Pop Check.

mariadkgdmgdmi:
25 tc for scarp??? that shit doesn't cost even 17!

PR #28068, opened 2024-06-08
Fixes brain transfers

### LINKED ISSUE #28008: Brains lose soul
Been an issue for a while now but if you remove a players brain there soul isn't attached, so the new body their brain is wont actually let the player revive. As their soul is in their old body. so the new humonkey will show as dnr and their old brainless body will still be twitching with life

Majkl-J:
## About The Pull Request

fixes #28008

Fix bug good

## Proof of Testing
Typical forgot to screenshot moment

## Changelog

:cl:
fix: Brains get souls transferred again
/:cl:


tmyqlfpir:
Thank you!

PR #28103, opened 2024-06-10
[MIRROR]  The Syndicate MODs are now resistant to batons knockdown.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/83510
--------------------
## About The Pull Request

Adds a new Module preinstalled into the nukie suits, and purchasable in the uplink,  grants resistance to batons knockdown.

## Why It's Good For The Game

At the last coderbus meeting, i personally asked to Fikou and Mothblocks for more baton counters to be introduced, they seemed to be ok with the idea, so here we are.

Introduces an unremovable module which grants to the wearer resistance to baton knockdown, it comes preinstalled in all syndicate modsuits.

With the fix of "rest to prevent disarm", stun batons have gotten even more powerful than they once were, and while i do agree secret kinda abusable tech is bad for the health of the game, the end result is that the stun meta is even more prominent than ever, so it's a good time introduce a small counter to it.

From a thematic perspective, Syndie suits should at least make people feel afraid, which is hard to do when a single prod from a stun stick will make you drop like a sack of potatoes, lose your desword and get slashed to bits.

On the gameplay side, wearing one of these will make you a target for the entire station, if a traitor is willing to expose themselves, or in the case of nukies, forced to; they should probably have a small edge on what's the most common tool used to stop them.

Keep in mind that this only prevents the knockdown, it does nothing for  the stamina damage, you'll still get stamcritted in a couple of hits, nothing changes on that front.

This is merely to prevent a scenario where a single baton hit will immediately lose you a fight.

## Changelog

:cl:
add: adds the MOD shock-absorption module, into the game.
add: The MOD shock-absorption module into the the uplinks, costs 4 TC.
balance: Nukie modsuits come with the shock_absorption module preinstalled.
/:cl:


Erol509:
Perfect balance, definetly

SpaceLoveSs13:
Might need to modularly revert this

PR #28423, opened 2024-06-27
[MIRROR] Space Heaters no longer deplete their starting cell instantly, and scale appropriately

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/84360
--------------------

## About The Pull Request

The value for base heating energy now utilizes a multiplier of the define for cell charge.

Space heaters start with a larger cell due to the fact that their old cell was woefully underperforming with the new values.

Also makes the improvised chem heater also use the define and not a hardcoded value.

## Why It's Good For The Game

I tested the values, and while the cell definitely doesn't immediately deplete from the temperature adjustments anymore, it certainly still feels, at best, a lethargic adjustment. I suspect this may need a rework, but all I am interested in doing is getting it working right now.

## Changelog
:cl:
fix: Space heaters do not completely drain their starting cell while trying to change their room temperature by a few degrees.
/:cl:


SpaceLoveSs13:
Dependant on #28376

PR #28442, opened 2024-06-28
New Bonesetters and bloodfilters

xXPawnStarrXx:
## About The Pull Request

cMttSR8

With /TG/s upcoming changes to removing easy purging, I turned my attention to bloodfilters and bonesetters. Both rather... lackluster in their uses and space consuming. But fret not, I'm here to help! I'm not going to be nice and give you both in one, I'm adding THREE items; The Combitool, the alien bonesetter and the alien bloodfilter.

The combitool; This nifty tool is a combined bloodfilter and bonesetter, how does it work? Same as the other combo tools!
The alien bloodfilter; A faster bloodfilter, incase your patient is full of 1000u of absinthe (Yes, that's actually been something I've seen.)
The alien bonesetter; This is a spicy tool... It can realign hairline and compound fractures, at a cost of 20 and 40 damage to the limb respectively. Make sure to treat your patient between uses, you can easily kill them... that'd be sad, wouldn't it?... ~~unless it's the third time that same idiot walked infront of the damned tram~~

You can combine both tools into one storage slot, meaning you can carry another item or two. Full suite of alien tools, never hurts.

## Proof of Testing

<details>

image
image
Right leg was a hairline (20 damage), Right arm was a compound (40 damage).

image
image

</details>

## Changelog
:cl:
add: Added new medical tools
/:cl:


lessthnthree:
For QoL could you please make the blood filter not go 'ping!' every time it cycles?

It will be even worse with a rapid blood filter.

xXPawnStarrXx:
> For QoL could you please make the blood filter not go 'ping!' every time it cycles?
> It will be even worse with a rapid blood filter.

I think that's a base part of the blood filter surgery, I'll see what I can do to make it better.

PR #28933, opened 2024-07-19
[MIRROR] HUD traits now apply their corresponding hud automatically. Most clothing/item/etc sources of HUDs now only use traits

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/84984
--------------------

## About The Pull Request

Currently if you want to apply a HUD you usually add both its trait and the HUD itself. Only exceptions are things like simplemobs where you should avoid adding the hud trait since it adds security/med DB access and such, but there is no cases where you'd want to apply the trait and not apply the hud.

Requested by Melbert about a week ago.
image

## Why It's Good For The Game

This makes working with HUDs significantly easier, as you no longer have to bother with manually adding/removing them. Also potentially removes an edge case where if your hud could get removed while keeping the trait.

## Changelog
:cl:
refactor: HUD traits now apply their corresponding hud automatically
/:cl:


SpaceLoveSs13:
#28856

PR #29153, opened 2024-07-31
[MIRROR] Turfs in lua will no longer null out their reference on deletion.

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/85429
--------------------

## About The Pull Request
Turfs are static objects and you can safely hold reference to them because their reference does not change on deletion.

## Why It's Good For The Game
Makes lua scripts more stable since signals don't get unregistered on turf deletion.

## Changelog
:cl:
admin: Turfs in lua will no longer become invalid on deletion.
/:cl:


SpaceLoveSs13:
#29210

PR #29235, opened 2024-08-07
[Modular] Allows the SAD to apply client quirks too so when you get humonkied you don't need to beg admins for help

Odairu:
## About The Pull Request
does as title says
This is moreso for QoL for players and admins so you don't lose all your quirks when you end up needing to be humonkied then sadded

Yes its abusable, but no more so than the sad already is, and I added a bit more logging so some random dude can't just keep his name the same and shuffle quirks around to get them all like pokemon
## Proof of Testing
image
<details>
  
</details>

## Changelog
:cl:
qol: SAD now applies quirks
/:cl:


Majkl-J (approved):
Looks good to me + we used to have this on bubber it's just gone now for some reason

PR #29389, opened 2024-08-15
[MIRROR] Digitigrade clothing sprites (feat. GAGS and 0 sprite bloat) 

SkyratBot:
Original PR: https://github.com/tgstation/tgstation/pull/85406
--------------------
## About The Pull Request

Use GAGS to auto-generate digitgrade leg sprites based off of 1 basic template + color sampling.
Icons are cached relatively aggressively, like female sprite variations. So no need to worry about that. 

The result: All of these sprites with only **3** icons added (the base template, and the masks)
Obviously it's not perfect - some of the sprites (like the engineer's jumpsuit) lose some "luster", but it does the job if I do say so myself.

image

Only applied to undersuits, for now. 

## Why It's Good For The Game

- Allows Lizards to show off their handicap, rather than hiding it or necessitating they use skirts. 

- (Ideally) leads to an uptick in Digitigrade abuse (due to being easier to identify), which is also a net positive.

- Implemented without any common issues revolving alt bodytype clothing sprites (those being maintainability and sprite bloat)

## Changelog

:cl: Melbert
add: Auto-generated digitigrade clothing sprites for most jumpsuits
/:cl:


projectkepler-ru:
I fucked something up @Majkl-J

